PR #5, opened 2022-01-21
Added emoji suggest feature

Devorein:


mattcasey:
I tried it out.. can you fix the CSS so there's a white background, and remove the grey boxes around the emojis?

image

Devorein:
Yeah absolutely. Also, another thing I've noticed is that it doesn't add a scroll on the y-axis for overflow.

PR #8, opened 2022-01-23
Added basic support for grouped dropdown items

Devorein:


components/editor/BangleEditor.tsx
```diff
@@ -41,17 +42,21 @@ const picker = {
...
 
-const NUM_SUGGESTIONS = 3;
+const totalDropdownItems = dropdownGroups.reduce((current, dropdownGroup) => current + dropdownGroup.items.length, 0);
+const dropdownItems: HTMLDivElement[] = [];
 
 function placeSuggestion() {
   const suggestion = document.querySelector('#suggestion') as HTMLDivElement;
   suggestion.style.display = picker.open ? 'block' : 'none';
   const rect = document.getElementsByClassName('autocomplete')[0]?.getBoundingClientRect();
+  console.log({ rect });
```
mattcasey, line 52: i think you can remove this?

components/editor/BangleEditor.tsx
```diff
@@ -96,6 +102,39 @@ const options: Options = {
   ],
 };
 
+function suggestionItemClickHandler() {
+  console.log(123);
```
mattcasey, line 106: console.log

components/editor/BangleEditor.tsx
```diff
@@ -96,6 +102,39 @@ const options: Options = {
...
+  placeSuggestion();
+  if (!picker.range) return;
+  const tr = picker.view.state.tr
+    .deleteRange(picker.range.from, picker.range.to)
+    .insertText(`Clicked`);
+  picker.view.dispatch(tr);
+  picker.view.focus();
+}
+
+const suggestionComponent = <div id="suggestion" style={{ display: "none" }}>
```
mattcasey, line 119: please use single quotes.. maybe we should add a prehook for linting

components/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface DropdownGroup {
```
mattcasey, line 1: rather than a generic types.ts file, i think this would be better inside Dropdown.ts or DropdownGroup.ts? You could also import the json file from there instead of putting it in BangleEditor.tsx, which i'm worried is getting pretty large already

styles/index.css
```diff
@@ -18,16 +18,33 @@
...
 }
-#suggestion:hover > div:hover,
-#suggestion .selected {
+
+.suggestion-group-item.selected, .suggestion-group-item:hover {
   background-color: wheat;
   cursor: pointer;
 }
 
+.suggestion-group-items {
```
mattcasey, line 39: is there a reason you're using css classes vs styled components?

components/types.ts
Devorein: Yep, I agree. `BangleEditor.tsx` is getting too big.

styles/index.css
Devorein: Not really I just figured that since we are using CSS styles for the dropdowns, we might use that. But I will switch everything to styled components.

components/editor/BangleEditor.tsx
Devorein: Yeah absolutely

components/editor/BangleEditor.tsx
Devorein: Absolutely

PR #9, opened 2022-01-24
Initial integration with bangle.io

Devorein:


mattcasey (approved):
LGTM once the TS error is fixed

PR #19, opened 2022-01-27
Improve folder organization

mattcasey:
Don't want to spend too much time refactoring, but I think small cleanups like this as we go are good. 
- added 'adapters' folder based on the conversation in Discord
- moved style sheet into theme folder
- added SASS - it's supported natively by next.js once you install the module


README.md
```diff
@@ -33,9 +33,17 @@ We stand on the shoulders of giants:
...
+### Folder Structure
 
-#### Theming
+- `/pages`: pages rendered by Next.js
+- `/theme`: colors, typography, etc.
+- `/components`: React components
+- `/context`: React utils
+- `/hooks`: React utils
+- `/adapters`: utils to interact with external apis and other io layers
+- `/lib`: utils that don't fit anywhere else
```
mattcasey, line 44: Do you think having this in the README will be helpful to someone? I don't often see projects list out their folders, and half of these are standard-ish.

Devorein: Hmm, not really sure. As you said half of these are quite obvious what they store, so I would say it's a bit redundant.

Devorein (approved):

README.md
mattcasey: ok i'm gonna remove it, it's one less thing to maintain

PR #20, opened 2022-01-27
Dark mode support for floating menu component

Devorein:
Had to transfer a bit code from `@bangle.dev/react-menu` package and refactor/change it as per our need.


mattcasey (changes requested):
@Devorein can you tell me what it was that you couldn't override? Altho we are copying all the code from bangle.io over, i want to avoid doing the same for bangle.dev since we can install it via npm. if you need to patch that code, we can use `patch-package`. But also, there is a method called `renderMenuType` which might give you the control you need? See the "Customizing floating menu type" example here: https://bangle.dev/docs/examples/react-floating-menu/.

mattcasey (changes requested):
after talking, we agreed to just add a README with the current version number explaining that we wanted to update the tooltips, etc.

Devorein:
I've added a `readme.md` file under `components/editor/@bangle.dev` to mention why we decided to use our own implementation of the package for the specific version.

mattcasey (approved):
perfect, thanks!

PR #22, opened 2022-01-28
Store space, users, and pages in local storage

mattcasey:


components/common/page-layout/Sidebar.tsx
```diff
@@ -13,25 +13,15 @@ import ListItem from '@mui/material/ListItem';
...
-
-const pages = [
-  { id: 1, path: '', title: 'First Page' },
-  { id: 2, path: '', title: 'Second Page' }
-]
+import { isTruthy } from 'lib/types';
+import { useUser } from 'hooks/useUser';
+import { useSpace } from 'hooks/useSpace';
+import { useSpaces } from 'hooks/useSpaces';
+import { usePages } from 'hooks/usePages';
```
mattcasey, line 24: does this pattern make sense? I figure when we start using a db, this code won't have to change, and these hooks would just be calling a REST endpoint using SWR instead of localstorage

hooks/useSpace.ts
```diff
@@ -0,0 +1,22 @@
+import { useRouter } from 'next/router';
+import { Space } from 'models';
+import { useSpaces } from './useSpaces';
+
+export function useSpace () {
```
mattcasey, line 5: perhaps we should be passing in the space id here, instead of it grabbing it from some global state? i went with what's easiest for now, which is useSpace() calls useRouter().. we'll have to pass spaceId down in a number of places

components/common/page-layout/Sidebar.tsx
Devorein: Yes, I do follow the same pattern, but instead of keeping the API-related hooks under `hooks` folder, I keep it under either `api/mutations` or `api/queries` folder. But one thing I would suggest would be to make subdirectories inside hooks as well for example `hooks/adapters` and `hooks/api`, where all the above hooks would lie.

hooks/useSpace.ts
Devorein: Hmm, I think it would be better if we pass in the space id to the hook.

hooks/useUser.ts
```diff
@@ -3,6 +3,6 @@ import { activeUser } from 'seedData';
 import { useLocalStorage } from './useLocalStorage';
 
 export function useUser () {
-  return useLocalStorage<Contributor>(`space.profile`, activeUser);
+  return useLocalStorage<Contributor>(`charm.v1.space.profile`, activeUser);
```
Devorein, line 6: I think we should store the ls key prefix as a constant in a separate module preferably called `constants` so that we can change it easily in the future.

Devorein (changes requested):

components/common/page-layout/Sidebar.tsx
mattcasey: hmm yea, maybe `hooks/data` would be good? or `hooks/models`

hooks/useUser.ts
mattcasey: agreed!

Devorein (approved):

PR #23, opened 2022-01-28
Allow saving favorites

mattcasey:
image

Added a Favorite toggle to the top right of the page. I realized my hooks don't automatically share state between properties w/o Context or somehting else.. but in this case I was able to just pass the state down :)

### DIFF
--- a/components/common/page-layout/PageLayout.tsx
+++ b/components/common/page-layout/PageLayout.tsx
@@ -103,7 +103,7 @@ export function PageLayout ({ children }: { children: React.ReactNode }) {
   const [user, setUser] = useUser();
   const router = useRouter();
   const thisPage = pages.find(page => page.title === pageTitle);
-  const isFavorite = thisPage && user.favorites.some(({ pageId }) => pageId === thisPage.id);
+  const isFavorite = thisPage && user?.favorites.some(({ pageId }) => pageId === thisPage.id);
 
   const handleDrawerOpen = () => {
     setOpen(true);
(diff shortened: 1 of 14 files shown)


hooks/useLocalStorage.tsx
```diff
@@ -5,23 +5,28 @@ const PREFIX = 'charm.v1';
...
-  return null;
+  return defaultValue || null;
 }
 
 export function useLocalStorage<T = any> (_key: string, defaultValue?: T) {
 
   const key = getKey(_key);
 
   const [value, setValue] = useState<T>(() => {
-    return getStorageValue(PREFIX + key, defaultValue);
```
mattcasey, line 24: this was a bug.. we weren't actually retrieving from the right place!

components/common/page-layout/PageLayout.tsx
```diff
@@ -125,6 +142,12 @@ export function PageLayout({ children }: { children: React.ReactNode }) {
             <Typography noWrap component='div' sx={{ flexGrow: 1, fontWeight: 500 }}>
               {pageTitle}
             </Typography>
+            {/** favorite toggle */}
+            <Tooltip title={isFavorite ? 'Remove favorite' : 'Add to favorites'} arrow placement='bottom'>
+              <IconButton sx={{ ml: 1 }} onClick={toggleFavorite} color="inherit">
+                {isFavorite ? <FavoritedIcon color='secondary' /> : <NotFavoritedIcon color='secondary' />}
```
Devorein, line 152: Shouldn't we use a gold color for the FavouritedIcon? It might make it stand out a bit.

mattcasey: I think that could be too distracting, since everything else is black at the moment. Favorites are nice for the power users but we aren't trying to optimize on people creating favorites. Also following Notion's example, they have it just black.

mattcasey: FWIW I think the top-right corner is yet to be fully designed. If we end up with a few more icons that could use color, I could see us adding yellow here..

PR #29, opened 2022-01-28
Full support for json based content schema

Devorein:
The whole app might crash at first due to the page content being a markdown string rather than JSON schema. I would suggest removing the corresponding key from local storage completely.

### DIFF
--- a/models/Page.ts
+++ b/models/Page.ts
@@ -4,7 +4,7 @@ export const PATH_BLACKLIST = ['settings'];
 export interface Page {
   id: string;
   title: string;
-  content: string;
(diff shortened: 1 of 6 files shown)


models/Page.ts
```diff
@@ -19,4 +19,21 @@ export interface PagePermission {
   pageId: string;
   userId: string;
   level: 'full_access' | 'editor' | 'view_comment' | 'view';
+}
+
+export interface PageContent {
```
mattcasey, line 24: thanks for adding this here!

mattcasey (approved):

PR #30, opened 2022-01-28
Fixed theming codeblock, scrollbar and callout component

Devorein:


components/editor/BangleEditor.tsx
```diff
@@ -77,13 +79,34 @@ export default function Editor({ markdown }: { markdown: string }) {
       tablePlugins(),
       columnResizing,
       floatingMenuPlugin(),
+      NodeView.createPlugin({
+        name: "blockquote",
+        containerDOM: ["blockquote"],
+        contentDOM: ["span"]
+      }),
+      NodeView.createPlugin({
```
mattcasey, line 87: do we still need codeBlock.plugins() and  blockquote.plugins() above?

components/editor/BangleEditor.tsx
```diff
@@ -77,13 +79,34 @@ export default function Editor({ markdown }: { markdown: string }) {
...
   });
 
-  return <ReactBangleEditor state={state}>
+  return <ReactBangleEditor state={state} renderNodeViews={({ node, children }) => {
+    if (node.type.name === 'blockquote') {
+      return <BlockQuote>
+        {children}
+      </BlockQuote>;
+    } else if (node.type.name === "codeBlock") {
+      console.log(children)
```
mattcasey, line 102: console.log

mattcasey (approved):
approved with a question

components/editor/BangleEditor.tsx
Devorein: yes, it doesn't work otherwise, unfortunately.

components/editor/BangleEditor.tsx
Devorein: Nice catch

theme/styles.scss
```diff
@@ -167,3 +162,272 @@ div#__next > div {
   font-weight: bold;
   text-align: left;
 }
+
```
Devorein, line 165: I had to move the `style.css` from `@bangle.dev/core/style.css` as it overwrote our `blockquote` styling. End goal would be to completely gradually remove styling from this file and port them to styled components

mattcasey: agreed

PR #31, opened 2022-01-28
Allow creating and editing workspaces

mattcasey:


pages/[domain]/settings/account.tsx
```diff
@@ -9,10 +9,16 @@ import FieldLabel from 'components/settings/FieldLabel';
 import Avatar from 'components/settings/LargeAvatar';
 import { setTitle } from 'components/common/page-layout/PageTitle';
 import { useUser } from 'hooks/useUser';
+import * as yup from 'yup';
+import { yupResolver } from '@hookform/resolvers/yup';
 
-interface FormValues {
-  username: string;
-}
+const schema = yup.object({
```
Comment (mattcasey) [inline pages/[domain]/settings/account.tsx:15]: this is pretty cool... react-hook-form integrates with yup, which offers a lot of nice validation handling and error messages

hooks/useSpace.ts
```diff
@@ -13,10 +14,26 @@ export function useSpace () {
     throw new Error('Space not defined for domain: ' + domain);
   }
 
-  function setSpace (space: Space) {
-    const newSpaces = spaces.map(s => s.domain === space.domain ? space : s);
-    setSpaces(newSpaces);
+  function setSpace (_space: Space | null, silent?: boolean) {
```
mattcasey, line 17: i'm not particularly happy with this method, its talking to localStorage directly.. but it should be much simpler in the future. it is handling side-effects like updating the 'spaces' key in local storage so it doesn't trigger a redraw until we reload

Devorein:
Right now we allow spaces inside a domain name. I think this would pose a problem for us. What do you think about it?
image

pages/[domain]/settings/account.tsx
Comment (Devorein) [reply pages/[domain]/settings/account.tsx:15]: I agree. Yup is my go-to library both for server-side and client-side validation

hooks/useSpace.ts
Devorein: Agreed

pages/[domain]/settings/workspace.tsx
```diff
@@ -54,6 +57,13 @@ export default function WorkspaceSettings () {
     reset(values);
   }
 
+  function deleteWorkspace () {
```
Comment (Devorein) [inline pages/[domain]/settings/workspace.tsx:60]: Shouldn't we remove the data related to space and its pages from localstorage after deleting it?

Devorein (changes requested):

pages/[domain]/settings/workspace.tsx
Comment (mattcasey) [reply pages/[domain]/settings/workspace.tsx:60]: that is actually happening if you look at useSpace, it's inside of that method right now.

Devorein (approved):

PR #35, opened 2022-01-30
Fixed styling for emoji suggest palette

Devorein:
Updated the styles for the emoji-suggest component, work nicely with both light and dark mode.
We can now render emoji inside a callout, but it's quite difficult to trigger the emoji-suggest component on clicking the callout emoji. I have to try it out again.

image
image


components/common/page-layout/Sidebar.tsx
```diff
@@ -58,15 +57,15 @@ export default function Sidebar({ closeSidebar, favorites }: SidebarProps) {
   const [spaceFormOpen, setSpaceFormOpen] = useState(false);
   const favoritePages = favorites.map(fav => pages.find(page => page.id === fav.pageId)).filter(isTruthy);
 
-  function showSpaceForm () {
+  function showSpaceForm() {
     setSpaceFormOpen(true);
   }
 
-  function closeSpaceForm () {
+  function closeSpaceForm() {
```
mattcasey, line 64: is this intentional? most of the code includes a space between name and parenths... if we change this decision i want to do it all at once so we dont end up with both

components/editor/@bangle.dev/README.md
```diff
@@ -4,7 +4,11 @@ This folder was created to customize packages under the [@bangle.dev](https://gi
...
 
-1. `MenuGroup` and `Menu` components were copied over to met our design guidelines (using styled components rather than plain css)
+1. `MenuGroup` and `Menu` components were copied over to met our design guidelines (using styled components rather than plain css for custom/dynamic theming)
 2. `Icon` component was copied over to utilize MUI's own tooltip component rather than the one `react-menu` provides out of the box
 3. `MenuButtons` had to be copied as it relied on the `Icon` component. Since we added our own version of `Icon` we had to make use of our own custom `MenuButtons` component
 4. Some of the functionalities provided by `MenuButtons` does not suit our use cases
+
+## `react-emoji-suggest@0.28.4-alpha.2`
+
+1. `EmojiSuggest` component was copied over to met our design guidelines (using styled components rather than plain css for custom/dynamic theming)
```
mattcasey, line 14: thanks for keeping it updated

theme/styles.scss
```diff
@@ -8,26 +8,8 @@ div#__next > div {
...
-}
-
-.bangle-emoji-square:hover {
-  cursor: pointer;
-}
-
-.bangle-emoji-square.bangle-is-selected {
-  background: rgb(223, 223, 223);
+body button {
+  background: inherit;
```
mattcasey, line 12: this is a very global style, also affects Mui components... it would be good to have a comment here explaining why we need it?

mattcasey (approved):
approved but with comments

components/common/page-layout/Sidebar.tsx
Devorein: Not actually. I don't know how or why that changed. I don't remember changing anything from this module :(

components/editor/@bangle.dev/README.md
Devorein: No problem as a matter of fact I forgot to update it for the other packages.

theme/styles.scss
Devorein: Good catch. Let me examine this a bit.

components/common/page-layout/Sidebar.tsx
Devorein: Since it's a linting issue we decided to handle it in a separate PR.

PR #36, opened 2022-01-30
Support for strike and underline text formatting

Devorein:
Strike and underline formats are now supported
image

Updated code styling to resemble that of notion
image


theme/index.ts
```diff
@@ -119,6 +122,9 @@ export const createThemeLightSensitive = (mode: PaletteMode) =>
       sidebar: {
         avatarHighlight: mode === 'dark' ? 'rgba(255, 255, 255, .2)' : '#ccc',
         background: mode === 'dark' ? sidebarBackgroundColorDarkMode : sidebarBackgroundColor,
+      },
+      code: {
+        color: "#EB5757"
```
mattcasey, line 127: theres really no reason to add it to the palette here unless you're handling dark and light mode..

components/editor/BangleEditor.tsx
```diff
@@ -53,6 +55,16 @@ const specRegistry = new SpecRegistry([
   tableRow
 ]);
 
+const StyledReactBangleEditor = styled(ReactBangleEditor)`
+  code {
+    padding: ${({ theme }) => theme.spacing(0.5)} ${({ theme }) => theme.spacing(1)};
+    border-radius: ${({ theme }) => theme.spacing(0.5)};
+    background-color: ${({ theme }) => theme.palette.mode === "dark" ? sidebarBackgroundColorDarkMode : sidebarBackgroundColor};
```
mattcasey, line 62: I believe this line can just be `({ theme }) => theme.palette.sidebar.background`

mattcasey (changes requested):
i think this is mixing up a few ways to import colors, but i would recommend:
```
code: {
      background: mode === 'dark' ? darkBgColor : lightBgColor,
      foreground: mode === 'dark' ? darkColor : lightColor
}
 ```

theme/index.ts
Devorein: But then if we want to use this color at other places we would have to copy the same hex value, and more importantly what if we decide to change it in the future. I guess we can do a search and replace in VsCode but don't you think it's a bit better to keep it in a central place?

components/editor/BangleEditor.tsx
Devorein: This is actually handled in the #35 PR. I've attached two more properties to the `background` interface to handle a lighter and darker version of our base background color.

mattcasey: I like what you did in the other PR as well! that's what i was trying to get at, thanks

mattcasey (approved):

PR #38, opened 2022-01-30
Add linting

mattcasey:
This adds lint test to github CI and also a pre-commit hook via husky.
* I skipped `next lint` for now, adding eslint myself. Next lint has some nice things in it but i feel like it's more magic than we need. I also left out prettier because I like having the space before parenths of a function and couldn't find any way to override it :/


.github/workflows/test_and_deploy.yml
```diff
@@ -21,8 +21,10 @@ jobs:
           node-version: "16"
       - name: Install npm dependencies
         run: npm ci --no-audit --no-fund
-      - name: Build Next.js App
-        run: npm run build
+      - name: Lint
+        run: npm run lint:prod
+      # - name: Build Next.js App
```
Devorein, line 26: Why are we skipping the build step?

Devorein (changes requested):

.github/workflows/test_and_deploy.yml
mattcasey: ah, i had removed it since we don't run tests, and it gets run in the 'deploy' step.. but I forgot it is useful for making sure the build is correct even on PRs

Devorein (approved):

PR #46, opened 2022-02-01
Add ability to delete pages

mattcasey:
image


components/common/page-layout/Sidebar.tsx
```diff
@@ -134,6 +134,12 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
     }, 100);
   }
 
+  function deletePage (pageId: string) {
+    const newPages = pages.filter(p => p.id !== pageId);
+    router.push(newPages[0] ? `/${space.domain}/${newPages[0].path}` : `/${space.domain}`);
```
mattcasey, line 139: i don't love that we just send you to the first page, but this is OK for now i guess? not sure what would be better unless we do like Notion and let you view deleted pages.

Devorein (approved):

PR #47, opened 2022-02-01
Allow setting page title

mattcasey:
Only thing that doesn't work: the title isn't updated in the sidebar until you refresh the page. When we move to HTTP, I would use SWR to trigger a new http request for the pages in the sidebar instead.

### DIFF
--- a/theme/styles.scss
+++ b/theme/styles.scss
@@ -108,7 +108,7 @@ div#__next > div {
 }
 
 .ProseMirror {
-  padding: 5px 15px;
+  padding: 5px 0;
 }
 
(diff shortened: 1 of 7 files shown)


theme/styles.scss
```diff
@@ -108,7 +108,7 @@ div#__next > div {
 }
 
 .ProseMirror {
-  padding: 5px 15px;
+  padding: 5px 0;
```
mattcasey, line 111: removed padding on the sides so the content is flush with page title

Devorein (approved):

PR #48, opened 2022-02-01
Fixed callout emoji suggest container position

Devorein:


mattcasey:
it doesnt look like it's fixed yet, here's what i see:
image

mattcasey (changes requested):
see comment

mattcasey (approved):

PR #50, opened 2022-02-02
Remove suggestion mark after inserting code and callout block

Devorein:


components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
```diff
@@ -94,13 +94,20 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]>
...
       editorExecuteCommand: () => {
-        return (state, dispatch) => {
-          return insertNode(state, dispatch, state.schema.nodes.codeBlock.create(
-            { language: "Javascript" },
-            Fragment.fromArray([
-              state.schema.text("console.log('Hello World');")
-            ])
-          ))
+        return (state, dispatch, view) => {
+          rafCommandExec(view!, (state, dispatch) => {
```
mattcasey, line 98: what is this command? i guess it comes from bangle.dev/io?

mattcasey (approved):

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
Devorein: Yep exactly. Although I had to change it a bit.

PR #53, opened 2022-02-03
Button for adding icon and cover for a page

Devorein:
image
image
Clicking on Add Icon inserts a random emoji, just like notion, and clicking on Add Cover inserts our wide logo


Devorein:
Yeah, sure thing. I will work on that then

PR #54, opened 2022-02-03
Insert image block from palette dropdown

Devorein:
image
image
image
You can now insert an image block from the palette dropdown. Alignment is also supported. But at the moment we only support embedding a link to an image. Image resize, caption and drag & drop still are a few issues we need to solve.


components/editor/@bangle.dev/base-components/image.ts
```diff
@@ -0,0 +1,318 @@
+import type { RawPlugins, RawSpecs } from '@bangle.dev/core';
```
mattcasey, line 1: What are your thoughts on putting this outside of @bangle.dev, since it's a completely new plugin? Did you consider it? I would have thought it would be outside since it's completely our own code, but I can also see how it makes more sense to go with the other components.

components/editor/@bangle.dev/base-components/image.ts
```diff
@@ -0,0 +1,318 @@
...
+              drop(view, event) {
+                if (event.dataTransfer == null) {
+                  return false;
+                }
+                const files = getFileData(
+                  event.dataTransfer,
+                  acceptFileType,
+                  true,
+                );
+                console.log({files})
```
mattcasey, line 113: console.log

components/editor/Image.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  if (!node.attrs.src) {
+    return (
+      <PopperPopup popupContent={(
+        <Box>
+          <MultiTabs tabs={[
+            [
+              'Upload',
+              <div>
+                <Button variant='contained'>Choose an image</Button>
```
mattcasey, line 87: It looks like this should work but I noticed it isn't hooked up yet.. intentional?

components/editor/@bangle.dev/base-components/image.ts
Devorein: Thanks!!!

components/editor/Image.tsx
```diff
@@ -0,0 +1,166 @@
...
+                setAlign(alignLabel as string);
+              }}
+            >
+              {alignIcon}
+            </ListItem>
+          ))}
+        </Controls>
+        { /* eslint-disable-next-line */}
+        {<img width={500} contentEditable={false} draggable src={node.attrs.src} alt={node.attrs.alt} />}
+        <ImageCaption>
```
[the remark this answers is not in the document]
Devorein: I think that would be best. It's been quite difficult to make the caption work for some reason. I will take a look at it for a while and if I can't find a solution we can scrap it for now.

theme/index.ts
```diff
@@ -155,14 +155,22 @@ export const createThemeLightSensitive = (mode: PaletteMode) => createTheme({
     },
     MuiButton: {
       defaultProps: {
-        disableRipple: true
+        disableRipple: true,
+        variant: 'contained'
```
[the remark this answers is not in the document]
Devorein: Yeah I just saw you adding these two props to most of the buttons, and I think we can do the same for `ListItem` as well

PR #56, opened 2022-02-05
Notion patterns for page covers

Devorein:


mattcasey (changes requested):
- the popup that appears when I click "Change cover" doesn't go away when I click outside of it or on something else.
- the title and "Add icon" option should appear below the cover image. This is also getting in the way of the "change cover" button from appearing while hovering (because you're hovering over the title html element, not the cover)
- make gallery first, and use a random image from our gallery when adding a cover by default - let's get rid of the charmverse logo as default backdrop, it's a poor quality image anyway.

Devorein:
image

Looks far better now thanks to your feedback 😄 Completed all of them apart from the last one 🤔 I'm not sure I quite understand what you mean. We can bring it up in the meeting.

Devorein:
image
I think I understood it. This might be what you meant in the last point. Keeping the page cover controls aligned with the editor content.

mattcasey (approved):

mattcasey:
One more thing: if I have an icon set on the page, it gets in the way of mousehover and the 'change cover' button doesn't appear.

PR #57, opened 2022-02-07
Fixed page cover controls now showing on hover

Devorein:
Hovering over the page cover control works fine now 😄 
image

### DIFF
--- a/components/common/PopperPopup.tsx
+++ b/components/common/PopperPopup.tsx
@@ -5,7 +5,7 @@ import * as React from 'react';
 
 interface PopperPopupProps {
   popupContent: React.ReactNode
-  children: React.ReactNode
+  children?: React.ReactNode | null
 }
 
 export default function PopperPopup (props: PopperPopupProps) {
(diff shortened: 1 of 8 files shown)


mattcasey (changes requested):
- the emoji picker doesn't disappear after i select an emoji (but it should, unlike the cover photo where we wait til you click outside). it also doesn't close if i click outside it
- The emoji element should be added before the content section, or use 'position: absolute'. Right now, it slides down the page as content is added. Here's an example:
image

Devorein:
I actually tried doing the second one. It was a lot more difficult than just moving the components around. At first, I tried to swizzle `@bangle.dev/react/ReactEditor.tsx` component but there was an error that kept popping up over and over again. I tried fixing that but couldn't unfortunately. So ended up with the Icon below the page content and making every one of them absolute to place on our screen.

Devorein:
Regarding the emoji suggest not closing when clicking outside its from `@bangle.dev`, so we have the same behavior for callout block. But its definitely possible 🚀

PR #58, opened 2022-02-07
make font families consistent, and move some style files to root

mattcasey:
Also I removed a bunch of dead code from their sidebar that we don't need. I think it makes sense to get rid of anything we're not using, so long as it doesn't hinder our ability to compare and review upstream changes to the code we are using. it'll require less maintenance over time.


mattcasey:
going to pre-emptively merge, just an organization change and not much to review

PR #61, opened 2022-02-08
Image resize feature

Devorein:
Added two image resize handlers on both sides of the image. Hovering over the image triggers their visibility. Just drag them from left to right to see the magic happen 😉 
image
One small thing you might notice is the big black cursor on the left side. I'm not quite sure where or how that is being generated. If you want I can take a look at it.


components/editor/Image.tsx
```diff
@@ -7,9 +7,12 @@ import AlignHorizontalRightIcon from '@mui/icons-material/AlignHorizontalRight';
 import DeleteIcon from '@mui/icons-material/Delete';
 import ImageIcon from '@mui/icons-material/Image';
 import { Box, ListItem, Typography } from '@mui/material';
-import { HTMLAttributes, useState } from 'react';
+import React, { HTMLAttributes, useState } from 'react';
 import ImageSelector from './ImageSelector';
 
+const MAX_IMAGE_WIDTH = 750; const
```
mattcasey, line 13: this should be two lines

components/editor/Image.tsx
```diff
@@ -64,9 +67,119 @@ function EmptyImageContainer (props: HTMLAttributes<HTMLDivElement>) {
...
+  clientX: number,
+  position: 'right' | 'left'
+  isDragging: boolean
+  setIsDragging: React.Dispatch<React.SetStateAction<boolean>>
+}
+
+function ImageResizeHandle (
+  { isDragging, setIsDragging, position, imageWidth, setClientX, setImageWidth, clientX }: ImageResizeHandleProps
+) {
+  const sx: Partial<{left: number, right: number}> = {};
```
mattcasey, line 106: this is OK but i think it'd be better as a property of your component, that way the styles are in one place and u might have other css rules to change:
```
const StyledImageResizeHandle = styled(Box)<{ position: string }>`
  ${({ position }) => position === 'left' ? 'left: 15px' : 'right: 15px'};
`'
...
<StyledImageResizeHandle position='left' />



```

mattcasey:
approved, because this works in Chrome but it does not in Firefox so that is an issue still. I tested out react-dnd for dragging instead but couldn't really get it to work, although we're using it in a number of places :( I feel like it may be related to bangle.dev but didn't keep digging.

components/editor/Image.tsx
Devorein: I miss prettier 🤣

PR #62, opened 2022-02-09
Allow logging in via web3 wallet

mattcasey:
With this update, Metamask is now required to be able to access the demo content


theme/index.ts
```diff
@@ -151,13 +151,13 @@ export const createThemeLightSensitive = (mode: PaletteMode) => createTheme({
...
   },
   components: {
-    MuiTypography: {
-      styleOverrides: {
-        root: {
-          whiteSpace: 'nowrap'
-        }
-      }
-    },
+    // MuiTypography: {
```
mattcasey, line 154: @Devorein i commented this out otherwise the modals are messed up, but we spoke about moving the CSS rule to somewhere else in Discord. not sure if you have yet or if this is OK for now?

PR #63, opened 2022-02-09
Floating menu link added

Devorein:
image

### DIFF
--- a/components/editor/@bangle.dev/react-menu/LinkSubMenu.tsx
+++ b/components/editor/@bangle.dev/react-menu/LinkSubMenu.tsx
@@ -1,0 +1,132 @@
+import { link } from '@bangle.dev/base-components';
+import { EditorView } from '@bangle.dev/pm';
+import { useEditorViewContext } from '@bangle.dev/react';
+import styled from '@emotion/styled';
+import CancelIcon from '@mui/icons-material/Cancel';
+import CheckCircleIcon from '@mui/icons-material/CheckCircle';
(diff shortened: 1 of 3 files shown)


mattcasey (changes requested):
The implementation works well for me in Firefox, great work! Some UX requests:
- Save Icon: I completely missed the "Save" icon on the right. I think it deserves its own spot next to the text input. You could make it 50% opacity unless we detect changes to the input. And use the Save MUI icon - a check is normally used to show the save was successful.
- Remove Icon: Can we use the Trash Icon instead of "X" and the tooltip should say "Remove link" instead of "Clear" (I think X icon/ "Clear" is too close to the standard UI for closing a modal)?
- Most editors have a "Copy link" button where "Go to site" is. We use a library 'react-copy-to-clipboard' which makes it very easy to add.
- I wanted to see if I could find a more intuitive way to 'go to link'. I like how Notion lets you click the original link - and this is possible in bangle.dev by registering the link component with `link.spec({ openOnClick: true })`. But we need to make the floating menu appear on hover somehow? If that is not supported, can we just change the icon to "Launch" to signal it is sending you to another website.
- could we do some very minimal validation - even `/(http(s?)):\/\//i.test( userInput )`. A cool feature would be showing the favicon, page title and description from the page, but we'll save that for another day!

Updated icons:
image

mattcasey:
I noticed there's also a "LinkOff" icon from mui that could work instead of the trash can.

PR #66, opened 2022-02-09
Crypto price block

motechFR:
Provides a component for viewing the price of a crypto-currency.

image

Reusable fiat currency selector that can be imported across the app

image

Quote currency configurable in the UI, or in code

image

===
TODO: Integrate this into the Bangle editor


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
mattcasey, line 1: like! I think we could include rules for linting too as it seems to trip people up

components/common/RelativeTime.tsx
```diff
@@ -0,0 +1,74 @@
+import { useEffect, useState } from 'react';
+
+function convertToUnixSeconds (timestamp: number | Date): number {
```
mattcasey, line 3: Normally I'd say this file is a great candidate for a couple unit tests, but we don't have testing infra set up just yet :/

components/common/RelativeTime.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const now = Date.now();
+
+  const differenceInSeconds = timestampDifferenceInSeconds(now, timestamp);
+
+  const differenceInMinutes = Math.floor(differenceInSeconds / 60);
+
+  let label = '';
+  let timeoutToSet = 0;
+
+  let timeout: any;
```
mattcasey, line 32: I believe we'd want to memorize this so that it doesn't get reset on every refresh. It's definitely a higher level challenge in the React world of things though. @Devorein  mentioned he wanted to help with this and will send you a separate PR.

components/common/RelativeTime.tsx
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  // Refresh this every second
+  // eslint-disable-next-line prefer-const
+  timeout = setTimeout(() => {
+    // TODO - Improve this quick hack for forcing re-renders
+    refreshLabel(Math.random().toString());
+  }, timeoutToSet);
+
+  return (<span>{label}</span>);
```
mattcasey, line 73: I think you can just return `label` no need for an extra html tag? Another React tip: sometimes you need to group multiple React elements as one but don't want a full HTML element, you can use `<React.Fragment>{label}</React.Fragment>` or just `<>{label}</>` but again, not necessary here since it's just returning a string really.

components/common/RelativeTime.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return timestamp / 1000;
+  }
+  return timestamp;
+}
+
+function timestampDifferenceInSeconds (first: number | Date, second: number | Date): number {
+  return Math.floor(convertToUnixSeconds(first) - convertToUnixSeconds(second));
+}
+
+export function RelativeTime ({ timestamp }: {timestamp: number | Date}) {
```
mattcasey, line 19: Something to consider... It actually looks like you've written a hook! I don't think this needs to be a component since the output is really just a string. Or this component could just be importing your hook internally, sort of a 'utility component' if someone prefers that sort of API. cc @Devorein ?

components/common/form/InputSearchCrypto.tsx
```diff
@@ -0,0 +1,62 @@
+import { Autocomplete, Box, TextField } from '@mui/material';
+import Image from 'next/image';
+import { CryptoCurrencyList, CryptoCurrency } from '../../../models/Currency';
```
mattcasey, line 3: we have the basePath setup so you can skip the relative import and do `import ... from 'models/Currency'`

components/common/form/InputSearchCrypto.tsx
```diff
@@ -0,0 +1,62 @@
...
+      onChange(value as CryptoCurrency);
+    }
+  }
+
+  return (
+    <Autocomplete
+      onChange={(event, value) => {
+        emitValue(value as any);
+      }}
+      id='currency'
```
mattcasey, line 32: do you need an id here? is it possible this input might appear multiple times on the page?

components/common/form/InputSearchCrypto.tsx
```diff
@@ -0,0 +1,62 @@
...
+const cryptoLogoPaths: Record<CryptoCurrency, string> = {
+  ETH: '/cryptoLogos/ethereum-eth-logo.svg',
+  BTC: '/cryptoLogos/bitcoin-btc-logo.svg',
+  BCH: '/cryptoLogos/bitcoin-cash-bch-logo.svg',
+  ETC: '/cryptoLogos/ethereum-classic-etc-logo.svg',
+  LTC: '/cryptoLogos/litecoin-ltc-logo.svg',
+  XRP: '/cryptoLogos/ripple-xrp-logo.svg'
+};
+
+export function InputSearchCrypto ({ onChange }: {onChange: (value: CryptoCurrency) => any}) {
```
mattcasey, line 19: it's better to return `void` instead of `any` if you don't care about the response of onChange but let me know if I'm forgetting something.

components/common/form/InputSearchCrypto.tsx
```diff
@@ -0,0 +1,62 @@
...
+      autoHighlight
+      renderOption={(props, option) => (
+        <Box component='li' sx={{ '& > img': { mr: 2, flexShrink: 0 } }} {...props}>
+          <Image
+            loading='lazy'
+            width='20px'
+            height='20px'
+            src={cryptoLogoPaths[option as CryptoCurrency]}
+          />
+          {'    '}
```
mattcasey, line 44: there might be some MUI components to help with layout.. MenuList/ ListItemIcon https://mui.com/components/menus/

components/editor/BangleEditor.tsx
```diff
@@ -163,16 +166,20 @@ export default function BangleEditor (
...
         name: 'blockquote',
         containerDOM: ['blockquote'],
-        contentDOM: ['span']
+        contentDOM: ['div']
       }),
       NodeView.createPlugin({
         name: 'codeBlock',
         containerDOM: ['pre'],
-        contentDOM: ['span']
+        contentDOM: ['div']
```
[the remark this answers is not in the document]
Devorein: not any tbh. It works both ways. All of the content will be stored inside `contentDOM` tag. It's a bit more logical to think that a paragraph is nested inside a `div` than inside a `span`

PR #70, opened 2022-02-10
Show notistack error when uploading file of large size

Devorein:
Allow user to upload files only of size <=1Mb. Had to delete the `package-lock.json` due to lockfileversion incompatibility for some reason -_-
image


components/editor/ImageSelector.tsx
```diff
@@ -2,6 +2,7 @@ import { Button, TextField } from '@mui/material';
 import { Box } from '@mui/system';
 import MultiTabs from 'components/common/MultiTabs';
 import PopperPopup from 'components/common/PopperPopup';
+import { useSnackbar } from 'notistack';
```
mattcasey, line 5: Did you know MUI has a snackbar? https://mui.com/components/snackbars/. I'm not sure the pros/cons but generally i think it would be best to stick with what MUI offers since it can inherit our theme automatically. But let's talk about it in our meeting.
We actually need to unify some snackbar-type messages from focalboard and the wallet3 connects, so it would be good to decide on one.

pages/_app.tsx
```diff
@@ -288,7 +286,15 @@ function DataProviders ({ children }: { children: ReactNode }) {
         <DatabaseBlocksProvider>
           <PageTitleProvider>
             <UserProvider>
-              {children}
+              <SnackbarProvider
```
mattcasey, line 289: this component is called 'DataProviders' because it is just being used for localstorage hooks at the moment... i think u could just put this one in the main component.

mattcasey (changes requested):
move the snackbar context to the main component in _app.tsx

mattcasey (approved):

PR #71, opened 2022-02-10
Press enter to go to bottom of editor

Devorein:


mattcasey (changes requested):
Almost! It is inserting an extra line when I hit Enter.. it should just go to the first line of the editor
image

PR #73, opened 2022-02-10
Feat/editor tweaks

mattcasey:


components/editor/@bangle.dev/tooltip/suggest-tooltip.ts
```diff
@@ -61,7 +61,7 @@ function specFactory({
           {
             'data-bangle-name': markName,
             'data-suggest-trigger': mark.attrs.trigger,
-            'style': `color: ${markColor}`,
+            'style': markColor ? `color: ${markColor}` : '',
```
mattcasey, line 64: I am confident this is the right place to remove the blue color when using the autosuggest palette, but it doesn't seem to matter and I'm still seeing the blue style being added some how... (I tried deleting this entire line and it didnt matter)

PR #74, opened 2022-02-10
Implement an Error boundary

mattcasey:
I ran into an issue today in prod where the localStorage data for contributors was blowing up the page.. this resulted in an error inside the RouteGuard component and an infinite redirect to the login page. I think this should fix it so it at least dumps you to an error page


components/common/errors/ErrorBoundary.tsx
```diff
@@ -0,0 +1,36 @@
+import { Component, ErrorInfo, ReactNode } from 'react';
```
mattcasey, line 1: this whole class was just copied from https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/error_boundaries/

PR #76, opened 2022-02-11
set up db models in prisma

mattcasey:
Questions, maybe in a future iteration:
- should we just make "Page" model a special type of "Block" similar to "Board"? We might want to investigate turning bangle.dev contents into blocks as well in that case. It feels right, though, because I can't see getting rid of the block model being used in focalboard
- we want to let people have private pages that transcend workspaces (!) I guess it means spaceId is optional on pages.
- should the foreign key for a user be a uuid or their wallet address? Examples where we need a user id are: created/updated timestamps, page and space permissions, favorites, UI settings, etc. Wallet seems way more natural..


Devorein:
1. I do think it's best if we convert everything to block. Storing everything in a large JSON content would become infeasible real fast, focalboard using the block mode provides a stronger reason
2.  Yep if that's something we want to support it becomes unnecessary to store it
3. Like you said it feels natural to store their wallet address but it will become really difficult to keep track of the actual user when they decide to use multiple wallet addresses. Also, it's not something the end-user would really care about or even see unless they open up the networks tab in their browser console.

mattcasey:
For now after talking w/Alex and @Devorein , we decided:
1. Page, Space, Favorites, everything could just be blocks. But for now, we'll keep them separate. We don't know enough about focalboard's blocks just yet, and it shouldn't be too hard to migrate in the future. Maybe a transitional state would be to create a 'charmverse blocks' table, but I want to focus right now on just hooking up the database w/models we are familiar with.
2. so, spaceId is optional. if it doesn't exist, it will appear under "Private" category for the "createdBy" user id.
3. we'll use a uuid for now to keep things simple. we could change our mind int he future, but i feel like we should figure out what type of 'decentralized' storage we'll be using, it may come with a solution.

PR #77, opened 2022-02-11
Setup automatic linter fixing on save for tsx and ts files in VSCode

motechFR:
When saving .ts or .tsx files in VScode, ESLint will automatically apply all fixes to that file


Devorein (approved):

mattcasey:
much appreciated!

PR #85, opened 2022-02-15
Feat/models api

mattcasey:
TODO:
- [x] create space
- [x] update space
- [x] delete space
- [x] create profile
- [x] edit profile
- [x] get users
- [x] get workspaces
- [x] save workspaces
- [x] get pages
- [x] save pages
- [x] delete pages
- [x] get favorites
- [x] save favorites
- [x] delete favorites
- [x] get space permissions
- [x] save space permissions
- [x] get focalboard blocks
- [x] create focalboard blocks
- [x] delete focalboard blocks
- [x] update focalboard blocks
- [ ] save changes to page content
- [ ] get bounties
- [ ] save bounties
- [ ] delete bounties


.env
```diff
@@ -1 +1,2 @@
-DATABASE_URL=postgres://postgres:postgres@localhost:5432/charmverse
\ No newline at end of file
+DATABASE_URL=postgres://postgres:postgres@localhost:5432/charmverse
+AUTH_SECRET=4sjrrA4kJwa4jCcLNMXN6xvpGwa4MHn3
```
mattcasey, line 2: note that this is just for local dev! we will have a different value in production

charmClient.ts
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable class-methods-use-this */
```
mattcasey, line 1: i've been putting all the backend calls in here. It follows the same pattern in focalBoard, where they put all their calls is inside a file called `octoClient.ts`. Makes it easy to see everything at once..

components/common/CreateSpaceForm.tsx
```diff
@@ -1,21 +1,41 @@
...
+    .required('Name is required')
+});
+
+export type FormValues = yup.InferType<typeof schema>;
 
 interface Props {
-  onCancel: () => void;
-  onSubmit: (values: Space) => void;
+  onCancel?: () => void;
+  onSubmit: (values: Prisma.SpaceCreateInput) => void;
```
mattcasey, line 38: I'm leaking prisma models into the front-end, which won't work long-term... this is just time-saving for now (you can see how they let you do things like create a workspace, permissions, pages, etc all in one call to prisma.. nice!)

charmClient.ts
motechFR: Cool idea @mattcasey
I like having one place for definition and docs of API calls.
Does prisma help generate openAPI docs?

mattcasey: Nothing native, and that lib "next-crud" would have been able to generate swagger, however I decided not to use it because it seemed like it was going to take too much time to configure, and by default it adds a lot more endpoints than we needed. Someone asked this same question  on their [github](https://github.com/prisma/prisma/discussions/5757) and they referenced this: https://github.com/pantharshit00/prisma-docs-generator which could be useful

mattcasey: Also I want to point out that we might go "all in" on blocks.. if you look how Notion and Focalboard do it, everything including your workspace is just a type of block. This greatly simplifies the API layer, and then u just need to know about what types of blocks exist. It will just take time to unify our models with what focalboard's got

PR #89, opened 2022-02-16
Integration of BangleEditor and FocalBoard

Devorein:


mattcasey (approved):
Look good.. i was sharing with @motechFR a way to embed the page logic from Editor.tsx into BangleEditor.tsx which I think would help.. basically, just pass those components as children in. What do you think? Then you would'nt eed to pass in setPage or Page.

We can do that in a separate PR

PR #90, opened 2022-02-17
remember extensions when using Chrome debugger

mattcasey:
Currently the Chrome debugger starts with a new profile each time you run it, but it's a pain to reinstall the Metamask extension each time. One option is to set the user directory to "false" and it will use your personal profile, but with this update we save it inside `.vscode` so that it just remembers what you did the last time, and we only have to install Metamask once.
https://stackoverflow.com/questions/51725854/run-vscode-chrome-debugger-with-its-extensions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ tsconfig.tsbuildinfo
 
 # user data
 *.csv
+.vscode/vscode-chrome-debug-userdatadir/
 
 # debug
 npm-debug.log*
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ tsconfig.tsbuildinfo
 
 # user data
 *.csv
+.vscode/vscode-chrome-debug-userdatadir/
```
mattcasey, line 20: ignoring the director since there's a ton of folders that get generated and there may be personal data

Devorein (approved):

PR #91, opened 2022-02-17
Columnblock and ColumnLayout node spec and views

Devorein:
Tried my best but couldn't manage to get the resize feature working. We can probably create a new PR with the fix but for now, static 2 and 3 columns layout was what I was able to do


mattcasey (changes requested):
Can you remove the CSS styling around the borders of the columns? And this is already in prod right? I think it would be helpful for future devs to mention why this update is still needed, I'm not too sure actually.

Also, sort of interesting is if you embed an image or iframe you can resize the column by resizing the internal content :)
image

mattcasey (approved):
Approving because I don't think my request to fix the CSS needs to block this PR from being merged!

Devorein:
I can remove the CSS styles on this branch. Also I did notice that resizing partially works when we embed an image inside and resize it 😛

PR #93, opened 2022-02-17
Youtube link copy paste to embed iframe

Devorein:


components/common/AutoFocusedTextField.tsx
```diff
@@ -0,0 +1,14 @@
+import { TextField, TextFieldProps } from '@mui/material';
+import { useEffect, useRef } from 'react';
+
+export function AutoFocusedTextField (props: TextFieldProps) {
```
motechFR, line 4: @Devorein Looks good. I am working on our form tooling, so let's see how to bring in what you built here

Devorein: I actually ended up deleting it. Material UI already provides an auto-focused text-field and I completely forgot to use that one 😛

components/editor/Iframe.tsx
```diff
@@ -103,12 +122,12 @@ function EmptyIFrameContainer (props: HTMLAttributes<HTMLDivElement>) {
...
       }}
-      {...props}
+      {...rest}
     >
       <StyledEmptyIFrameContainer>
-        <VideoLibraryIcon fontSize='small' />
+        {type === 'embed' ? <PreviewIcon fontSize='small' /> : <VideoLibraryIcon fontSize='small' /> }
         <Typography>
-          Embed an iframe
+          {type === 'video' ? 'Insert a video' : 'Insert an embed'}
```
motechFR, line 130: Great UX detail

PR #102, opened 2022-02-21
Persist bounties

motechFR:
**Improvements**
- Save bounties to the database
- List bounties in the current workspace

@mattcasey Can you give me feedback on Prisma schema for Bounties?

Required before merge:
- Reimplement form using  React--form-hooks
- Handle editing behaviour
- Implement InputSearchCollaborator


charmClient.ts
```diff
@@ -221,14 +221,15 @@ class CharmClient {
   }
 
   listBounties (workspaceId?: string): Promise<Bounty []> {
-    return http.GET<Bounty[]>('/api/bounties');
+    const requestQuery = workspaceId ? `?workspaceId=${workspaceId}` : '';
+    return http.GET<Bounty[]>(`/api/bounties${requestQuery}`);
```
mattcasey, line 225: you don't need to build the query string yourself with this wrapper! just
`http.GET<Bounty[]>('/api/bounties', { spaceId })`

components/bounties_v2/BountyEditor.tsx
```diff
@@ -61,8 +65,18 @@ function bountyEditorFields (bounty?: IBounty): IInputField<IBounty> [] {
 
 export function BountyEditor ({ onSubmit, bounty, mode = 'create' }: IBountyEditorInput) {
 
+  const [space] = useCurrentSpace();
+  const [user] = useUser();
+
   async function submitted (value: IBounty) {
+    console.log('Submit captured', value);
```
mattcasey, line 72: dont forget to delete console logs ;)

components/bounties_v2/BountyEditor.tsx
```diff
@@ -61,8 +65,18 @@ function bountyEditorFields (bounty?: IBounty): IInputField<IBounty> [] {
...
+    console.log('Space', space);
     if (mode === 'create') {
+      delete (value as any).linkedTaskId;
+      value.workspaceId = space!.id;
+      value.createdBy = user!.id;
+      value.assignee = user!.id;
+      value.reviewer = user!.id;
```
mattcasey, line 79: i'm assuming these lines will change right? ie we don't need a default assignee, and it seems odd that it would be the person creating the bounty. but it makes sense that reviewer can be the person that created the bounty by default

components/bounties_v2/BountyList.tsx
```diff
@@ -4,18 +4,20 @@ import Loader from 'components/common/Loader';
...
 
+  const [space] = useCurrentSpace();
   const [bountyList, setBountyList] = useState([] as IBounty []);
   const refresh = useRef(true);
   const [displayBountyDialog, setDisplayBountyDialog] = useState(false);
 
   async function refreshBounties () {
     try {
-      const foundBounties = await BountyService.listBounties();
+      const foundBounties = await BountyService.listBounties(space?.id);
```
mattcasey, line 20: spaceId should probably be required to list bounties right? You could either check that space exists, or if you are certain it exists, I prefer we override it and use `space!.id` rather than make listBounties() take an optional parameter, which would never make sense imo

mattcasey: The thing about getCurrentSpace() is that space may not exist on the first page load, but it definitely exists by the time the user has a chance to interact with the page.. hence i think it's ok to use `!` sometimes to assert that it is defined

pages/api/bounties/index.ts
```diff
@@ -0,0 +1,36 @@
+
+import { NextApiRequest, NextApiResponse } from 'next';
+import nc from 'next-connect';
+import { Bounty, Prisma, Space } from '@prisma/client';
+import { prisma } from 'db';
+import { onError, onNoMatch, requireUser } from 'lib/middleware';
+import { withSessionRoute } from 'lib/session/withSession';
+import { gettingStartedPageContent } from 'seedData';
```
mattcasey, line 8: please remove unused imports like this one. I believe VS Code highlights them for you?

pages/api/bounties/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).get(getBounties).post(createBounty);
+
+async function getBounties (req: NextApiRequest, res: NextApiResponse<Bounty[]>) {
+
+  const { workspaceId } = req.query;
```
mattcasey, line 16: please call it 'spaceId' to match the rest of the code

pages/api/bounties/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  const bountyListQuery: Prisma.BountyFindManyArgs = workspaceId ? {
```
mattcasey, line 18: this route should honestly just throw a 400 if spaceId is not defined.

pages/api/bounties/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+  } : {} as any;
+
+  const bounties = await prisma.bounty.findMany(bountyListQuery);
+  return res.status(200).json(bounties);
+}
+
+async function createBounty (req: NextApiRequest, res: NextApiResponse<Bounty>) {
+  const data = req.body as Prisma.BountyCreateInput;
+  // add a first page to the space
```
mattcasey, line 30: please remove comment

charmClient.ts
motechFR: Thanks for the tip @mattcasey I will remember this one for future API integrations

components/bounties_v2/BountyEditor.tsx
motechFR: These were default values initially for testing the form.

workspaceId is now spaceId (set by data from useCurrentSpace hook)
createdBy unchanged (set by data from setUser hook)

removed assignee  (we'll populate this later on in the bounty applications feature)

reviewer is removed (we now select one on front-end).
As per your feedback, we collect userID.

See this [change](https://github.com/charmverse/app.charmverse.io/pull/102/commits/b0ae439bf27fe39b8541d93a86d92e9658be0588) for details

components/bounties_v2/BountyList.tsx
motechFR: Makes sense to have spaceId mandatory as everything happens at workspace level
Changed [here](https://github.com/charmverse/app.charmverse.io/pull/102/commits/1a0095e8502ddaa09c8b939161844877a0d6ec3c) to reflect this

pages/api/bounties/index.ts
motechFR: Added this [here](https://github.com/charmverse/app.charmverse.io/pull/102/commits/58c49e105658ba65a0b46a2ada26185c408523c1)

components/bounties_v2/Bounty.tsx
```diff
@@ -42,7 +42,7 @@ export function Bounty ({ bounty }: IBountyInput) {
           <Grid item xs={12} sx={{ minHeight: '90px' }}>
 
             <Typography paragraph={true}>
-              {bounty.description.trim().substring(0, 120)}
+              {bounty.description?.trim().substring(0, 120)}
```
mattcasey, line 45: you can use fancyTrim() or fancyTrimWords() from lib/strings here, it will add the 'ellipsis' only if needed

prisma/migrations/20220221123317_/migration.sql
```diff
@@ -0,0 +1,22 @@
+-- CreateEnum
+CREATE TYPE "BountyStatus" AS ENUM ('open', 'assigned', 'review', 'complete', 'paid');
+
+-- CreateTable
+CREATE TABLE "Bounty" (
+    "id" UUID NOT NULL,
+    "author" UUID NOT NULL,
+    "workspaceId" UUID NOT NULL,
```
mattcasey, line 8: spaceId, not workspaceId please

prisma/schema.prisma
```diff
@@ -136,20 +137,20 @@ enum BountyStatus {
...
-  assignee          String      @db.Uuid
-  rewardAmount      Float
-  rewardToken       String
-  status            BountyStatus
-  title             String
-  linkedTaskId      String      @db.Uuid
+  id               String       @id @default(uuid()) @db.Uuid
+  createdBy        String       @db.Uuid
+  author           User         @relation(fields: [createdBy], references: [id], onDelete: Cascade, onUpdate: Cascade)
+  spaceId          String       @db.Uuid
```
mattcasey, line 144: this should also have the relationship to space defined

prisma/schema.prisma
```diff
@@ -136,20 +137,20 @@ enum BountyStatus {
...
+  createdAt        DateTime     @default(now())
+  updatedAt        DateTime     @default(now())
+  description      String?
+  descriptionNodes Json?
+  reviewer         String?       @db.Uuid
+  assignee         String?       @db.Uuid
+  rewardAmount     Float
```
mattcasey, line 151: i just want to clarify.. is it safe to use Float for these numbers? if not, i think i've also seen a tactic in web3 where it is common to store an integer and the decimal place

prisma/schema.prisma
```diff
@@ -136,20 +137,20 @@ enum BountyStatus {
...
+  rewardToken      String
+  status           BountyStatus @default(open)
+  title            String
+  linkedTaskId     String?      @db.Uuid
```
mattcasey, line 155: this could also have its relation defined

components/bounties_v2/Bounty.tsx
motechFR: Thanks, done in this [change](https://github.com/charmverse/app.charmverse.io/pull/102/commits/12a3a96d88b728df717a8c3cc414ab230c48a637)

mattcasey (approved):

PR #104, opened 2022-02-21
Invite links

mattcasey:


theme/colors.ts
```diff
@@ -23,7 +23,7 @@ export const scrollBarTrackBackgroundColor = darken(backgroundLightColor, 0.1);
 export const scrollBarThumbBackgroundColor = lighten(scrollBarTrackBackgroundColor, 0.5);
 
 // dark mode
-export const primaryTextColorDarkMode = 'rgba(255, 255, 255, 0.9)';
+export const primaryTextColorDarkMode = '#ededed';
```
mattcasey, line 26: cleanup... having 90% opacity has some unexpected effects particularly with icons/emojicons

prisma/schema.prisma
```diff
@@ -128,28 +130,40 @@ model SpaceRole {
...
-  descriptionNodes  Json
-  reviewer          String      @db.Uuid
-  assignee          String      @db.Uuid
-  rewardAmount      Float
-  rewardToken       String
-  status            BountyStatus
-  title             String
-  linkedTaskId      String      @db.Uuid
-}
+// enum BountyStatus {
```
mattcasey, line 146: commented out so as to not conflict with the bounty PR

components/invites/InviteForm.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <Button variant='outlined' color='secondary' onClick={onClose}>
+            Cancel
+          </Button>
+          <PrimaryButton type='submit'>
+            Generate a New Link
+          </PrimaryButton>
+        </Grid>
+        {errors.apiError && (
+          <Grid item>
+            <Typography color='danger'>{errors.apiError.message}</Typography>
```
motechFR, line 128: What do you think of using Material UI alert component, instead of typography?
See [here](https://mui.com/components/alert/) for styling

lib/invites/index.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  });
+  if (!invite) {
+    return {};
+  }
+  if (invite.maxUses > 0 && invite.useCount >= invite.maxUses) {
+    return { invite, expired: true };
+  }
+  else if (invite.maxAgeMinutes > 0) {
+    const timePassed = Date.now() - invite.createdAt.getTime();
```
motechFR, line 24: Could be worth extracting to a utility function for getting time difference.

models/Space.ts
```diff
@@ -2,4 +2,4 @@ import type { Space } from '@prisma/client';
 
 export { Space };
 
-export const DOMAIN_BLACKLIST = ['api', 'login', 'signup', 'createWorkspace'];
+export const DOMAIN_BLACKLIST = ['api', 'invite', 'login', 'signup', 'createWorkspace'];
```
motechFR, line 5: Rather than "Domain_Blacklist" maybe rename to FORBIDDEN_SPACE_NAMES ?

Also, we are referencing top level app-paths in various files (such as Route Guard).

What do you think of having a single source of truth as a const and a type, so we can refer to these easily in future?

PR #107, opened 2022-02-22
Fixed image resize issue on toggle dark/light modes

Devorein:
1. Dropped column drop shadows
2. Images can be resize beyond the width of a column
3. Toggling dark/light theme doesn't reset image size


mattcasey (approved):
noting that the change to Resizer is to control the state ourselves, since toggling light and dark mode seemed to be resetting it

PR #108, opened 2022-02-22
Added placeholder for empty editor

Devorein:
image
image

### DIFF
--- a/components/editor/CharmEditor.tsx
+++ b/components/editor/CharmEditor.tsx
@@ -21,12 +21,12 @@ import { BangleEditor as ReactBangleEditor, useEditorState, useEditorViewContext
 import { table, tableCell, tableHeader, tablePlugins, tableRow } from '@bangle.dev/table';
 import '@bangle.dev/tooltip/style.css';
 import styled from '@emotion/styled';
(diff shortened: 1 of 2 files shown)


components/editor/CharmEditor.tsx
```diff
@@ -94,12 +110,30 @@ const defaultContent: PageContent = {
...
+  const view = useEditorViewContext();
+  const docTextContent = view.state.doc.textContent as string;
+
+  // Only show placeholder if the editor content is empty
+  return docTextContent.length === 0 ? (
+    <Box sx={{
+      // This weird calculation is required to place the placeholder on the same position as the editor
+      top: top ? top - 31 : 30,
+      position: 'relative',
+      color: 'rgba(255, 255, 255, 0.4)',
```
mattcasey, line 123: you need to use the color from our theme, the text is invisible in white mode.

mattcasey (changes requested):
text is invisible in white mode

components/editor/CharmEditor.tsx
mattcasey: i suggest `const theme = useTheme(); .... color: theme.palette.text.secondary`

PR #113, opened 2022-02-22
fix up and down arrow key direction on command palette

mattcasey:
Two problems I fixed, tested in Chrome and Firefox:
- up/down arrows go in reverse direction if the command pallete opened above instead of below the cursor.
- even if the element you are keying down to is still in view, we were scrolling the parent each time.
- same issue was happening on the emoji container, and pressing up/down arrow would go diagonally up/down rather than straight up and  down

The bug stemmed from the fact our elements are embedded inside of a couple parent items (List, ListItem) so `parent` was actually a button, and not the scrolling container as it is on bangle.io.

### DIFF
--- a/components/editor/@bangle.dev/react-emoji-suggest/EmojiSuggest.tsx
+++ b/components/editor/@bangle.dev/react-emoji-suggest/EmojiSuggest.tsx
@@ -48,10 +48,11 @@ export function EmojiSuggest({
     triggerText,
     show: isVisible,
   } = usePluginState(suggestTooltipKey);
-  const width = rowWidth + (parseInt(theme.spacing(1).replace("px", "")) * 2);
+
+  const width = rowWidth + (parseInt(theme.spacing(2).replace("px", "")) * 2);
 
   return reactDOM.createPortal(
-    <StyledEmojiSuggest className="bangle-emoji-suggest">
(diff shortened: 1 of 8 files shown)


components/editor/@bangle.io/js-lib/inline-palette/inline-palette.ts
```diff
@@ -65,7 +65,7 @@ function pluginsFactory({
         safeRequestAnimationFrame(() => {
           view?.focus();
         });
-        if (key === 'UP' ? !getIsTop() : getIsTop()) {
```
mattcasey, line 68: I had trouble connecting where 'counter' is ultimately used, but it seems like whether or not the popup is on top or bottom, the increment/decrement action should be the same

components/editor/@bangle.dev/react-emoji-suggest/EmojiSuggest.tsx
```diff
@@ -48,10 +48,11 @@ export function EmojiSuggest({
     triggerText,
     show: isVisible,
   } = usePluginState(suggestTooltipKey);
-  const width = rowWidth + (parseInt(theme.spacing(1).replace("px", "")) * 2);
+
+  const width = rowWidth + (parseInt(theme.spacing(2).replace("px", "")) * 2);
```
mattcasey, line 52: this fixes the diagonal movement for me... it also allows space for one more emoji

PR #116, opened 2022-02-23
Focalboard integration with Bounty

Devorein:


components/bounties_v2/BountyBadge.tsx
```diff
@@ -1,57 +1,102 @@
...
-          height='30px'
-          src={CryptoLogoPaths[bounty.rewardToken as CryptoCurrency]}
-        />
-        {' '}
-        {' '}
-        {bounty.rewardAmount}
-        {' '}
-        {bounty.rewardToken}
+    <Box sx={{ maxWidth: '400px', background: 'background', borderRadius: theme.spacing(1) }}>
+      <Container sx={{ display: 'flex', alignItems: 'center', padding: theme.spacing(1) }}>
```
mattcasey, line 33: when using the `sx` prop, you dont have to import the theme to define padding... just do `p: 1`. not sure but it may work for borderRadius as well

Devorein: Yeah, Box is a lot better and more generic than Container. I can't exactly pinpoint why I used Container here 🤔

PR #120, opened 2022-02-23
Bounty page

motechFR:
This PR makes the bounty details page interactive.

It brings in an end to end workflow for enabling people to apply to a bounty.

image


components/bounties/BountyApplicantList.tsx
```diff
@@ -0,0 +1,146 @@
...
+    return (
+      <>
+        <Typography>Loading proposals</Typography>
+        <CircularProgress></CircularProgress>
+      </>
+    );
+  }
+
+  async function assignBounty (assignee: string) {
+    const updatedBounty = charmClient.assignBounty(bounty.id, assignee);
```
mattcasey, line 78: did you mean to await on the api call?

components/bounties/BountyApplicantList.tsx
```diff
@@ -0,0 +1,146 @@
...
+export interface IBountyApplicantListProps {
+  bounty: Bounty,
+  bountyReassigned?: () => any
+}
+
+function createData (id: string, message: string, date: string) {
+  return { id, message, date };
+}
+
+const rows = [
```
mattcasey, line 29: this const can be removed

components/bounties/BountyApplicantList.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const minHeight = applicationsMade === 0 ? undefined : Math.min(300, (100 * applicationsMade));
+
+  return (
+    <Box component='div' sx={{ minHeight: `${minHeight}px`, marginBottom: '15px', maxHeight: '70vh', overflowY: 'auto' }}>
+
+      <Table stickyHeader={true} sx={{ minWidth: 650 }} aria-label='bounty applicant table'>
+        <TableHead>
+          <TableRow>
+            <TableCell>
+              <AutorenewIcon onClick={refreshApplications} />
```
mattcasey, line 102: you're going to remove this now or in a separate PR?

components/bounties/BountyApplicantList.tsx
```diff
@@ -0,0 +1,146 @@
...
+              Applicant ID
```
mattcasey, line 103: i think just "Applicant" is enough

components/bounties/BountyApplicantList.tsx
motechFR: Good spot, cleaned up [here](https://github.com/charmverse/app.charmverse.io/pull/120/commits/226b44c3d62e673ff4639b3fddcb81af249f1d81)

components/bounties/BountyApplicantList.tsx
motechFR: Just keeping this for now until parent <> child communication works well

components/bounties/BountyApplicantList.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [applications, setApplications] = useState([] as Application []);
+  const loading = useRef(true);
+
+  useEffect(() => {
+    refreshApplications();
+  }, []);
+
+  useEffect(() => {
+    if (user && space) {
+      const adminRoleFound = user.spaceRoles.findIndex(spaceRole => {
```
mattcasey, line 45: I think Array.some() would read a little better here? it just returns true/false if the iterator ever returns true.

prisma/schema.prisma
```diff
@@ -89,18 +89,19 @@ model Block {
...
+  addresses          String[]
+  spaceRoles         SpaceRole[]
+  favorites          FavoritePage[]
+  pagePermissions    PagePermission[]
+  pages              Page[]
+  blocks             Block[]
+  spaces             Space[]
+  inviteLinks        InviteLink[]
+  bounties           Bounty []
+  Applications          Application []
```
mattcasey, line 104: rename this to "applications" (i believe it was added automatically by prisma)

pages/api/applications/index.ts
```diff
@@ -0,0 +1,54 @@
...
+async function createApplication (req: NextApiRequest, res: NextApiResponse<Application>) {
+  const data = req.body as Application;
+  const ApplicationToCreate = { ...data } as any;
+
+  if (data.createdBy) {
+    (ApplicationToCreate as Prisma.ApplicationCreateInput).applicant = { connect: { id: data.createdBy } };
+    delete ApplicationToCreate.createdBy;
+  }
+  else {
+    return res.status(406).json({ error: 'Please provide an applicant' } as any);
```
mattcasey, line 39: i never used 406 before ... it looks like it's based on 'unsupported protocol' or bad 'user agents'? https://www.exai.com/blog/406-not-acceptable. I think a 400 would be fine here, since we're just checking that the request has a field

pages/api/applications/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    return res.status(406).send({ error: 'Please provide a valid bounty ID' } as any);
+  }
+
+  const ApplicationListQuery: Prisma.ApplicationFindManyArgs = {
+    where: {
+      bountyId: bountyId as string
+    }
+  };
+
+  const bounties = await prisma.application.findMany(ApplicationListQuery);
```
mattcasey, line 26: Should be 'applications' not bounties.
Also for Later: some sort of helper function to validate a user has access to the data they request

lib/dates.ts
```diff
@@ -0,0 +1,84 @@
...
+
+export function convertToLuxonDate (date: DateInput): DateTime {
+  return typeof date === 'number'
+    ? DateTime.fromMillis(date)
+    : typeof date === 'string'
+      ? DateTime.fromISO(date)
+      : DateTime.fromJSDate(date);
+}
+
+function getDaySuffix (dateTime: DateInput): string {
```
mattcasey, line 23: ah i miss this one formatting feature from Moment: `Do` but pretty sure it doesnt exist in luxon

lib/dates.ts
```diff
@@ -0,0 +1,84 @@
+
+import { DateTime, DateTimeUnit as LuxonTimeUnit } from 'luxon';
+
+/**
+ * Datestamps are created as YYYY-MM-DD
+ */
+export type DateStamp = `${number}-${number}-${number}`;
+
+export type DateInput = Date | DateStamp | number;
```
mattcasey, line 9: nice i like that we can take multiple types of input, so you don't have to remember if you have an ISOdate from the server or a date object already

mattcasey (approved):

components/bounties/BountyApplicantList.tsx
motechFR: Replace method by Array.some() [here](https://github.com/charmverse/app.charmverse.io/pull/120/commits/f124ae493832a62de87c0373b34b707db9ae1a1e)

pages/api/applications/index.ts
motechFR: I agree with you, will go with 400 going forward.

I was using 406 "Not Acceptable" in the past as a proxy for "Invalid input." However, re-reading the semantics of the HTTP spec, 400 is indeed the best code for invalid requests

PR #121, opened 2022-02-23
fix state messing with page properties

mattcasey:
This was nuts to fix. Tells me we are doing something wrong, maybe relying on both 'currentPage' and 'pages' in context is wrong.. I just tried to reduce the amount of places we were using the 'currentPage' value. `setState()` for example can either take the new value as input, or a function which then gives you access to the current value:
```
const { currentPage, setCurrentPage } = usePages();

function setPage (updates) {
  // doesnt work
  setCurrentPage({ ...currentPage, ...updates });
  // works
  setCurrentPage(_currentPage => { ..._currentPage, ...updates });
}
```

### DIFF
--- a/components/editor/CharmEditor.tsx
+++ b/components/editor/CharmEditor.tsx
@@ -142,13 +142,11 @@ export default function CharmEditor (
       new Plugin({
         view: () => ({
           update: (view, prevState) => {
-            if (!view.state.doc.eq(prevState.doc)) {
-              if (onPageContentChange) {
-                onPageContentChange({
-                  doc: view.state.doc.toJSON() as PageContent,
-                  rawText: view.state.doc.textContent as string
-                });
-              }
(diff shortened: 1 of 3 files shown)


components/editor/CharmEditor.tsx
```diff
@@ -142,13 +142,11 @@ export default function CharmEditor (
...
-                onPageContentChange({
-                  doc: view.state.doc.toJSON() as PageContent,
-                  rawText: view.state.doc.textContent as string
-                });
-              }
+            if (onPageContentChange && !view.state.doc.eq(prevState.doc)) {
+              onPageContentChange({
+                doc: view.state.doc.toJSON() as PageContent,
+                rawText: view.state.doc.textContent as string
+              });
```
motechFR, line 149: Much cleaner indeed 👍

motechFR (approved):
Read through the fresh version of [pageId]. Much clearer logic and variable naming now

PR #122, opened 2022-02-23
Improved invite link actions

mattcasey:
Made it so there's always a 'share' button on each row, and delete icon appears on hover. Tried to find a good balance of enough but not too much info :)
image


motechFR (approved):
Just tested, looks good!

PR #123, opened 2022-02-24
small tweaks for bangle dev editor

mattcasey:
Some cleanup items/bug fixes:
- Simplified CSS for page title and controls, reducing the need to set multiple 'top' offsets - now we just have an offset whether there's an icon set or not.
- made placeholder  its own component, and now it appears in the correct spot wherever it is used
- the text style cursor was disabled when looking at editor inside focalboard

### DIFF
--- a/components/common/Emoji.tsx
+++ b/components/common/Emoji.tsx
@@ -11,10 +11,9 @@ const Emoji = styled(Box)`
   white-space: nowrap;
   user-select: none;
   cursor: pointer;
-  padding: ${({ theme }) => theme.spacing(1)};
   border-radius: ${({ theme }) => theme.spacing(0.5)};
-  width: 100px;
-  height: 100px;
+  width: 80px;
+  height: 80px;
 
   display: flex;
   align-items: center;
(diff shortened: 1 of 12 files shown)


components/common/Emoji.tsx
```diff
@@ -11,10 +11,9 @@ const Emoji = styled(Box)`
   white-space: nowrap;
   user-select: none;
   cursor: pointer;
-  padding: ${({ theme }) => theme.spacing(1)};
   border-radius: ${({ theme }) => theme.spacing(0.5)};
-  width: 100px;
-  height: 100px;
+  width: 80px;
```
mattcasey, line 15: less padding around icons

components/editor/Placeholder.tsx
```diff
@@ -0,0 +1,27 @@
+import { useMemo } from 'react';
```
mattcasey, line 1: moved placeholder to its own file

theme/@bangle.dev/styles.scss
```diff
@@ -23,10 +24,14 @@
 }
 
 :root {
-  --window-tooltip-zIndex: 5;
+  --window-tooltip-zIndex: 4;
   --inlineUniversalPalette-height: 400px;
 }
 
+.bangle-tooltip {
```
mattcasey, line 31: fixes a bug where the tooltip apears under the top header

Devorein:
One small thing I noticed is that the button for adding Page Cover is overlapping with the Page Icon itself
image

Devorein:
I've also noticed that the placeholder issue still exist for the Bounty form unfortunately 😦 
image

PR #130, opened 2022-02-25
Debounce edits on the page editor

mattcasey:
I picked 500ms as an arbitrary number, open to suggestions

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -106,6 +106,7 @@
       "always"
     ],
     "no-bitwise": "warn",
+    "no-undef": "off",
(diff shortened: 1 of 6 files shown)


.eslintrc.json
```diff
@@ -106,6 +106,7 @@
       "always"
     ],
     "no-bitwise": "warn",
+    "no-undef": "off",
```
mattcasey, line 109: removed this because Typescript namespaces don't work, and TS already takes care of undefined for us.

Devorein (approved):

Devorein (approved):

PR #131, opened 2022-02-25
Multiline page title support

Devorein:
We ended up having to swizzle `@bangle.dev/react/ReactEditor.tsx` component. Previously it used to render the children below the actual editor. In our case since the children were supposed to be placed above the editor, we had to resort to hacky fixes to place them in the right order including using position relative and top all over the place. I managed to fix it by sizzling the component so that the children are placed on top of the editor. This resulted in much cleaner code and less use of dynamic styles for positioning elements.


components/editor/Page/PageTitle.tsx
```diff
@@ -20,12 +41,16 @@ interface PageTitleProps {
 
 export default function PageTitle ({ value, onChange }: PageTitleProps) {
   const view = useContext(EditorViewContext);
+  const theme = useTheme();
```
mattcasey, line 44: this is never used

components/editor/Editor.tsx
```diff
@@ -63,16 +65,16 @@ const EditorHeader = styled.div`
...
-  if (page.icon) {
-    pageTop = 280;
-  }
+  let pageTop = 100;
   if (page.headerImage) {
-    pageTop = 100;
+    pageTop = 50;
     if (page.icon) {
-      pageTop = 120;
+      pageTop = 75;
```
mattcasey, line 72: should be 80px so that the icon is centered on the bottom edge of the banner

hooks/useUser.tsx
```diff
@@ -1,8 +1,8 @@
 import { useWeb3React } from '@web3-react/core';
-import { ReactNode, createContext, useContext, useEffect, useState, useMemo } from 'react';
-import { LoggedInUser } from 'models';
 import charmClient from 'charmClient';
 import useENSName from 'hooks/useENSName';
+import { LoggedInUser } from 'models';
+import { createContext, ReactNode, useContext, useEffect, useMemo, useState } from 'react';
```
mattcasey, line 5: I notice sometimes the imports get switched around in your PRs. Is that a plugin doing it? I usually try to put 3rd party lib imports like react would be at the top of the page.

components/editor/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -0,0 +1,128 @@
+import {
```
mattcasey, line 1: Can you explain why this was copied over and what was changed?

mattcasey (changes requested):
I'm concerned that this solution required pulling in more of bangle.dev code. Please make sure any PRs have adequate explanation when we do this.

mattcasey (approved):

PR #138, opened 2022-02-25
Allow making pages public

motechFR:
Provides a toggle which admin users can enable.

If so, a publicly shareable link can be copied and sent to others.

image

### DIFF
--- a/models/Space.ts
+++ b/models/Space.ts
@@ -2,4 +2,4 @@ import type { Space } from '@prisma/client';
 
 export { Space };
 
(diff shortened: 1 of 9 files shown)


models/Space.ts
```diff
@@ -2,4 +2,4 @@ import type { Space } from '@prisma/client';
 
 export { Space };
 
-export const DOMAIN_BLACKLIST = ['api', 'invite', 'login', 'signup', 'createWorkspace'];
+export const DOMAIN_BLACKLIST = ['api', 'invite', 'login', 'signup', 'createWorkspace', 'share'];
```
mattcasey, line 5: thanks for remembering to update

PR #144, opened 2022-02-27
Create nested page from palette

Devorein:


motechFR (approved):

hooks/useNestedPage.ts
```diff
@@ -0,0 +1,30 @@
+import { Fragment } from '@bangle.dev/pm';
```
mattcasey, line 1: this file does not look like a react hook to me - it doesn't import or export any state. i think it's another good reason to create a new folder in lib/pages where you could just export `addNestedPage`

hooks/useMenu.ts
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+export function useMenu () {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
mattcasey, line 4: This hook is already provided by material-ui-popup-state. There's a `bindMenu` prop you should check out: https://www.npmjs.com/package/material-ui-popup-state#usepopupstate. I prefer we use a single way to do things, and I really have been liking this package which you introduced. I don't think we need a second wrapper

hooks/usePages.tsx
```diff
@@ -1,16 +1,22 @@
...
 import charmClient from 'charmClient';
+import { addBoardClicked } from 'components/databases/focalboard/src/components/sidebar/sidebarAddBoardMenu';
+import { useRouter } from 'next/router';
 import * as React from 'react';
 import { createContext, Dispatch, ReactNode, SetStateAction, useContext, useEffect, useMemo, useState } from 'react';
+import { useIntl } from 'react-intl';
 import { useCurrentSpace } from './useCurrentSpace';
+import { useUser } from './useUser';
 
+type AddPageFn = (space?: Space, page?: Partial<Page>, shouldRoute?: boolean) => Promise<Page>;
```
mattcasey, line 11: Couple comments about this function interface:
1) we should prefer named arguments, a la: `({ space, page, shouldRoute })`. My most basic reason is that from the callsite its impossible to know what this means: `addPage(undefined, undefined, true);` without inspecting the types with an IDE. But it's also more work to add and remove properties without doing large refactors, especially for a utility method
2) 'spaceId' is already an attribute allowed on Page, so it seems unecessary to even offer passing in a 'space' except that this method also lets you redirect to the page. I'd be fine if that was handled elsewhere, and then this method can literally just accept one required input: `page`. How about adding a 2nd method to wrap this one: `addPageAndRedirect(page);` That method could import both `useRouter` and `useSpace` so i dont even have to pass in the space.

components/editor/Editor.tsx
```diff
@@ -118,7 +118,7 @@ export function Editor ({ page, setPage, readOnly = false }: IEditorProps) {
           <EditorHeader>
             {page?.icon && (
               <EmojiContainer updatePageIcon={updatePageIcon}>
-                <Emoji sx={{ fontSize: 78 }}>{page.icon}</Emoji>
+                <Emoji sx={{ fontSize: 60 }}>{page.icon}</Emoji>
```
mattcasey, line 121: 78 is intentional and based on Notion's UI. Please revert it back?

components/editor/@bangle.io/js-lib/inline-palette/hooks.ts
```diff
@@ -65,11 +66,11 @@ export function useInlinePaletteItems<T extends InlinePaletteItem>(
         return () => true;
       }
 
-      return (state: EditorState, dispatch: ((tr: Transaction<any>) => void) | undefined, view: EditorView) => {
-        return item.editorExecuteCommand({
+      return async (state: EditorState, dispatch: ((tr: Transaction<any>) => void) | undefined, view: EditorView) => {
+        return await (item.editorExecuteCommand({
```
mattcasey, line 70: i'm not sure why adding async and await is needed here, when editorExecuteCommand already returns a promise?

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
```diff
@@ -431,17 +433,37 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]>
...
+export function useEditorItems() {
+  const { addNestedPage } = useNestedPage()
+
+  const paletteItems: PaletteItem[] = []
+  Object.entries({...paletteGroupItemsRecord, other: [
+    ...paletteGroupItemsRecord.other,
+    {
+      uid: 'page',
+      title: 'Insert page',
+      icon: <svg viewBox="0 0 32 32" height="1em" width="1em" className="page" style={{ transform: "scale(1.25)", fill: 'currentColor', flexShrink: 0, backfaceVisibility: 'hidden'}}><g> <path d="M16,1H4v28h22V11L16,1z M16,3.828L23.172,11H16V3.828z M24,27H6V3h8v10h10V27z M8,17h14v-2H8V17z M8,21h14v-2H8V21z M8,25h14v-2H8V25z" /> </g></svg>,
```
mattcasey, line 446: 1) SVGs should not be pasted inline, they should be kept with other assets and media in the public folder
2) if you are going to use a new page icon, please update the one we use in the sidebar. right now they look different (i'm also fine if u just use mui's icon for now):
image

components/editor/@bangle.io/extensions/inline-command-palette/InlineCommandPalette.tsx
```diff
@@ -114,7 +114,7 @@ export function InlineCommandPalette() {
   }, [
     view,
     query,
-    editorItems,
+    // editorItems,
```
mattcasey, line 117: why is this commented out? should we just delete it?

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+          <g>
+            <path d='M16,1H4v28h22V11L16,1z M16,3.828L23.172,11H16V3.828z M24,27H6V3h8v10h10V27z M8,17h14v-2H8V17z M8,21h14v-2H8V21z M8,25h14v-2H8V25z' />
+          </g>
+        </svg>
+        )}
+      </Box>
+      <Box
+        fontWeight={600}
+        component='span'
+        onClick={() => {
```
mattcasey, line 93: can't we just use a Nextjs link here and not import router?

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+          transition
+        },
+        '&:hover .actions-menu': {
+          opacity: 1
+        }
+      }}
+    >
+      <Box>
+        {nestedPage?.icon ?? (
+        <svg viewBox='0 0 32 32' height='1em' width='1em' className='page' style={{ transform: 'scale(1.25)', fill: 'currentColor', flexShrink: 0, backfaceVisibility: 'hidden' }}>
```
mattcasey, line 83: please put svg where in the public folder... looks like this may be duplicated from the other line

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        )}
+      </Box>
+      <Box
+        fontWeight={600}
+        component='span'
+        onClick={() => {
+          router.push(`/${(space!).domain}/${node.attrs.path}`);
+        }}
+      >
+        {nestedPage?.title ? nestedPage.title : 'Untitled'}
```
mattcasey, line 97: On Firefox, it looks like title is being squished for some reason: 
image

mattcasey (changes requested):
left comments

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <Box>
+        {nestedPage?.icon ?? (
+        <svg viewBox='0 0 32 32' height='1em' width='1em' className='page' style={{ transform: 'scale(1.25)', fill: 'currentColor', flexShrink: 0, backfaceVisibility: 'hidden' }}>
+          <g>
+            <path d='M16,1H4v28h22V11L16,1z M16,3.828L23.172,11H16V3.828z M24,27H6V3h8v10h10V27z M8,17h14v-2H8V17z M8,21h14v-2H8V21z M8,25h14v-2H8V25z' />
+          </g>
+        </svg>
+        )}
+      </Box>
+      <Box
```
mattcasey, line 90: just another note.. you can just use a div span if you want. The reason to use Box is to have access to all the custom MUI properties like `sx`, but for the most part i don't see you using those. Having less CSS/MUI stuff should help rendering be faster, and make it easier to copy and paste from our editor to someplace else

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+        onClose={hideMenu}
+        onClick={hideMenu}
+      >
+        <MenuItem
+          sx={{ padding: '3px 12px' }}
+          onClick={(e) => {
+            console.log(view, getPos());
+          }}
+        >
+          <ListItemIcon><DeleteIcon fontSize='small' /></ListItemIcon>
```
mattcasey, line 114: for some reason this doesn't actually do anything for me

components/editor/NestedPage.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <Menu
+        anchorEl={anchorEl}
+        open={isOpen}
+        onClose={hideMenu}
+        onClick={hideMenu}
+      >
+        <MenuItem
+          sx={{ padding: '3px 12px' }}
+          onClick={(e) => {
+            console.log(view, getPos());
```
mattcasey, line 111: console.log.. did you forget to add an action?

hooks/usePages.tsx
```diff
@@ -37,14 +48,61 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     }
   }, [space]);
 
-  const value = useMemo(() => ({
+  const addPage: AddPageFn = async (_space, page, shouldRoute) => {
+    shouldRoute = shouldRoute ?? true;
+    const spaceId = (_space?.id ?? space?.id)!;
+    const id = Math.random().toString().replace('0.', '');
+    const pageProperties: Prisma.PageCreateInput = {
```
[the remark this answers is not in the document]
Devorein: Yeah, sure that works. But let's do that in a separate PR if possible.

hooks/useNestedPage.ts
Devorein: Agreed. I will create a new PR with all the refactorings

components/editor/NestedPage.tsx
Devorein: My bad I was about to complete the delete action but forgot to remove the console logs. I will remove it once I'm done adding the delete functionality

hooks/usePages.tsx
Devorein: 1. Agreed. I was honestly confused when I used `addPage(undefined` in one instance. I will switch to named arguments asap.
2. That's a great idea. My only concern is that what if we want to create a page in a separate space. It might not have that good of a use case, but could be important under certain circumstances. Regarding the `addpageAndRedirect` function I can do that. I will place it under `usePages` hook as all the other hooks (`useCurrentSpace`, `useRouter`) are already used there.

components/editor/Editor.tsx
Devorein: But it actually overflows the icon. Let me show you a sample.
This is with 78 size
image
This is with 60 size
image

components/editor/@bangle.io/extensions/inline-command-palette/InlineCommandPalette.tsx
Devorein: Hmm, this caused an infinite loop as the editor items rather than being static object, is being dynamically calculated inside a hook. So the `useEffect` was being called recursively. I actually tried to fix it but couldn't go far. But I think it's best to fix it to avoid rerenders, I will need to check it out again. Adding a TODO comment on top to fix it after completing all the required tasks.

components/editor/NestedPage.tsx
Devorein: Good point. It will help out on SEO. I will see where else I've referenced useRouter and replace those with `next/link`

components/editor/NestedPage.tsx
Devorein: Yep you're right I will replace it will Mui/Icon's appropriate icon component

components/editor/NestedPage.tsx
Devorein: Good point. I will see where I can replace Box with regular span or divs

hooks/useMenu.ts
Devorein: I totally forgot about that package. I guess its becoming a bit hard to keep track of our dependencies as the codebase grows larger

components/editor/Editor.tsx
mattcasey: remove overflow: hidden and add a z-index to page-controls

PR #145, opened 2022-02-28
Notify important events on Discord

motechFR:
Anonymised events that help us serve the Charmverse community


lib/logs/notifyDiscord.ts
```diff
@@ -0,0 +1,47 @@
+import * as http from 'adapters/http';
+
+const webhookBaseUrl = 'https://discord.com/api/webhooks';
```
mattcasey, line 3: just need to update this to an env var: process.env.DISCORD_EVENTS_WEBHOOK. Does that name work?

mattcasey (approved):

lib/logs/notifyDiscord.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  console.log('New event logged', message);
+
+  //  if (isProdEnvironment === true) {
+
+  const webhookIdentifier = process.env.DISCORD_EVENTS_WEBHOOK;
+
+  const requestUrl = `${webhookBaseUrl}${webhookIdentifier}`;
+
+  return http.POST<IDiscordMessage>(requestUrl, { content: message });
```
mattcasey, line 44: one more thing.. i think it's worth catching the promise, though i doubt it will ever fail, just in case to avoid it bubbling up and breaking a request

lib/logs/notifyDiscord.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  let message = `Event: ${eventLog.funnelStage.toUpperCase()} / ${eventLog.eventType}\r\n`;
+
+  const isProdEnvironment = process.env.ENV === 'prod';
+
+  message += eventLog.message;
+
+  console.log('New event logged', message);
+
+  //  if (isProdEnvironment === true) {
```
mattcasey, line 38: don't forget to include this

lib/logs/notifyDiscord.ts
motechFR: As per your suggestion, removed base url and plugged it all into the Environment Variable.
See this [change](https://github.com/charmverse/app.charmverse.io/pull/145/commits/ef8bf88620fbce1d809ca33d12c77559f29b26ea)

lib/logs/notifyDiscord.ts
motechFR: Done, we now catch errors to prevent them bubbling up

PR #151, opened 2022-03-01
Crypto price fix

motechFR:
Adds a secondary source for getting prices
Caches pricing so as to avoid overload of third party APIs
xDAI quotes now supported

image


hooks/usePricing.ts
```diff
@@ -1,21 +1,7 @@
-import { CryptoCurrency, FiatCurrency, IPairQuote } from '../models/Currency';
-import fetch from '../adapters/http/fetch';
+import fetch from 'adapters/http/fetch';
+import { CryptoCurrency, FiatCurrency, IPairQuote } from 'models/Currency';
 
 export function getPricing (base: CryptoCurrency, quote: FiatCurrency): Promise<IPairQuote> {
-  return new Promise((resolve, reject) => {
-    fetch(`https://min-api.cryptocompare.com/data/price?fsym=${base}&tsyms=${quote}`)
```
mattcasey, line 6: this doesn't look like a hook to me, can you put this in lib or something?

lib/utilities/dates.ts
```diff
@@ -12,6 +12,29 @@ export type DateTimeFormat = 'relative' | 'absolute'
 
 export type TimeUnit = 'millisecond' | 'second' | 'minute' | 'hour' | 'day' | 'week' | 'month' | 'year'
 
+export const millisecondsInSecond = 1000;
+
+export const secondsInMinute = 60;
+
+export const minutesInHour = 60;
+
+export const hoursInDay = 24;
```
mattcasey, line 21: these vars dont seem to need to be exported. I try to avoid exporting anything unless it's actually being used, otherwise it makes it harder to refactor later because you assume some other file is using it

pages/api/crypto-price/getters.ts
```diff
@@ -0,0 +1,141 @@
...
+          resolve(pairQuote);
+        }).catch(error => {
+          reject(error);
+        });
+    });
+
+  }
+
+  private getPriceFromCoinMarketCap (base: CryptoCurrency, quote: FiatCurrency): Promise<IPairQuote> {
+    return new Promise((resolve, reject) => {
```
mattcasey, line 115: why do you need to use the Promise constructor, fetch is already returning a promise isn't it?

pages/api/crypto-price/getters.ts
```diff
@@ -0,0 +1,141 @@
+import { CryptoCurrency, FiatCurrency, IPairQuote } from 'models/Currency';
```
mattcasey, line 1: this doesn't look like an endpoint, so it doesn't belong in the /pages/api folder. put in lib?

pages/api/crypto-price/index.ts
```diff
@@ -0,0 +1,24 @@
...
+import { prisma } from 'db';
+import { onError, onNoMatch, requireUser } from 'lib/middleware';
+import { withSessionRoute } from 'lib/session/withSession';
+import { gettingStartedPageContent } from 'seedData';
+import { CryptoCurrency, FiatCurrency } from 'models/Currency';
+import { pricingGetter } from './getters';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).get(getPrice);
```
mattcasey, line 14: nice thanks for thinking to include requireUser.

hooks/usePricing.ts
motechFR: Moved to charmClient in this [change](https://github.com/charmverse/app.charmverse.io/pull/151/commits/db5add29e60dd306bc5cd9ce5ebc3278dad2776a)

lib/utilities/dates.ts
motechFR: Understand your point about not exporting till needed, so I've [removed](https://github.com/charmverse/app.charmverse.io/pull/151/commits/fd051b5c063a6f7c24b5dfab091a1d24686013da) the export statements for now.

We can individually readd them as and when a specific one will be needed.

pages/api/crypto-price/getters.ts
motechFR: Moved out to the lib folder in this [commit](https://github.com/charmverse/app.charmverse.io/pull/151/commits/9f4fb62e93b305e2460110bbf6fec187d73c2c45)

pages/api/crypto-price/getters.ts
motechFR: Thanks for the feedback, updated as per your recommendation in this [change](https://github.com/charmverse/app.charmverse.io/pull/151/commits/59bd2dce5d6c966f49baba332a9573a424b0558f)

mattcasey (approved):

PR #152, opened 2022-03-01
Document environment variables

motechFR:
Lays out which environment variables need to be provided to set up the app

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,23 @@
+# All values below should be replaced by your own
+
+# Link to your postgres instance
+DATABASE_URL=postgres://user:password@postgreshost.com:5432/databasename
+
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -0,0 +1,23 @@
...
+S3_UPLOAD_BUCKET=demo.bucket
+
+# Upload key for AWS S3
+S3_UPLOAD_KEY=AEHHEHHWPPORE
+
+# Region of your bucket
+S3_UPLOAD_REGION=us-west-1
+
+# Secret for AWS upload
+S3_UPLOAD_SECRET=D6gtf5i411kAGG8qinsq
```
mattcasey, line 23: can we just put ??? for values. that way it's obvious this wasnt a real secret?

PR #156, opened 2022-03-02
Update styles based on Notion

mattcasey:


components/databases/focalboard/src/components/dialog.scss
```diff
@@ -14,6 +14,11 @@
         height: 100%;
         background-color: rgba(var(--center-channel-color-rgb), 0.5);
     }
+    @media (prefers-color-scheme: dark) {
+        .wrapper {
+            background-color: rgba(15, 15, 15, 0.8);
```
mattcasey, line 19: this color's taken from Notion.. --center-channel-color-rgb is actually the foreground color, so in dark mode we were adding a lighter overlay

Devorein (approved):

PR #157, opened 2022-03-02
Add more styling to crypto component

mattcasey:
image


motechFR (approved):
Looks much better than previous version!

PR #160, opened 2022-03-02
Add cover image for focalboard card

Devorein:
We might need to add S3 upload support for page cover as well.
image
image


motechFR (approved):
Hey @Devorein I tested this and looks good.

The random cover gradient is a great idea. Looking forward to this being S3-compatible.

mattcasey (approved):
nice work!

Devorein:
Yes, I will make the cover full-width. Also, add cover for the whole focal board rather than an individual card.

PR #161, opened 2022-03-02
Allow crypto api to be accessed from share page

motechFR:


mattcasey (approved):
works for me... the biggest thing i'd wory about is someone using us as an api

PR #163, opened 2022-03-02
Handle empty workspace

motechFR:
Client-side and server-side protection for an empty workspace
Ensures there is always at least 1 empty page


pages/[domain]/index.tsx
```diff
@@ -17,20 +20,22 @@ export const getServerSideProps: GetServerSideProps = async (context) => {
...
   }
-  const firstPage = await prisma.page.findFirst({ where: { spaceId: space.id } });
+  let firstPage = await prisma.page.findFirst({ where: { spaceId: space.id } });
   if (!firstPage) {
-    console.error(`Space has no pages: ${space.id}`);
-    return {
-      notFound: true
-    };
+
+    const userId = context.req.session.user.id;
```
Comment (mattcasey) [inline pages/[domain]/index.tsx:26]: I'm not sure session user is guaranteed here. If I visit app.charmverse.io/charmverse without being logged in, it redirects me to the first page which then redirects me to login (it would be better if we didn't expose the first page at all though..). I know it's super rare to get to this point, but I think a simple fix for now would be to redirect to login if there is no user. We should also check that the user has access to this space before creating a page on their behalf.
Interestingly, if I try to access another workspace while logged in, currently I just get a blank page, which is acceptable for now. If you don't have access to create a page, I'd even be fine if we just threw an error for now

seedData.ts
```diff
@@ -127,6 +127,25 @@ export const pages: Page[] = [
...
+  return {
+    author: { connect: { id: userId } },
+    autoGenerated: true,
+    content: {
+      type: 'doc',
+      content: []
+    },
+    contentText: '',
+    isPublic: false,
+    path: 'getting-started',
```
mattcasey, line 140: since you're putting this in a place where someone might pick it up, can we randomize the path? can just be a uuid

components/common/page-layout/Sidebar.tsx
```diff
@@ -141,6 +142,13 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
...
       await charmClient.deletePage(page.id);
+      if (pages.length === 1) {
+        const newPage = await charmClient.createPage(untitledPage({
+          userId: user!.id,
+          spaceId: space!.id
+        }));
+        setPages([newPage]);
+      }
     }
     setPages(newPages);
```
mattcasey, line 153: I think this would override what you just did on line 150.

seedData.ts
motechFR: Handled in this [change](https://github.com/charmverse/app.charmverse.io/pull/163/commits/353c7d656666a28337734b4e9866278d6033e785)

components/common/page-layout/Sidebar.tsx
```diff
@@ -141,6 +142,13 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
     const newPages = pages.filter(p => p.id !== pageId);
     if (page) {
       await charmClient.deletePage(page.id);
+      if (pages.length === 1) {
+        const newPage = await charmClient.createPage(untitledPage({
+          userId: user!.id,
+          spaceId: space!.id
+        }));
+        setPages([newPage]);
```
mattcasey, line 150: suggestion to add:
```
setPages([newPage]);
router.push(newPage.path);
return;
```

pages/[domain]/index.tsx
Comment (motechFR) [reply pages/[domain]/index.tsx]: I have moved this check into a middleware we can call inside getServerSideProps functions. See this [change](https://github.com/charmverse/app.charmverse.io/pull/163/commits/2087ef8305d461c6993f7dec51f90817f3c1196c)

motechFR:
@mattcasey 
Here is the full list of cases we now support
Logged in user
- If a user deletes the last page in the workspace, a new untitled page is created
- If a user navigates to a workspace without pages, the server will create a new untitled page
- If a user navigates to the root domain of a workspace they do not have access to, they are redirected to their workspace
- If a user navigates to a page in a workspace they have access to that does not exist anymore, they are redirected to the root of the workspace ie. charmverse.io/workspaceDomain

Non logged in user
- If a non logged in user navigates to the root domain of a workspace, they are redirected to the login page

Unhandled edge cases in this PR
- A logged in user tries to access a page in a workspace they do not have access to. A blank screen shows
- A person user tries to access a publicly shared page. A blank screen shows

components/common/page-layout/Sidebar.tsx
motechFR: See l. 169 : 173 for redirect I added. This occurs after the block related logic.

We redirect to the workspace root, and let default behaviour take over. Currently, this consists in navigating you to first page

mattcasey (approved):
nice

PR #165, opened 2022-03-03
Block quote

motechFR:
New block quote element

image
image

### DIFF
--- a/components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
+++ b/components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
@@ -5,6 +5,7 @@ import {
 import { EditorState, Fragment, Node, setBlockType, Transaction } from '@bangle.dev/pm';
 import { rafCommandExec, safeInsert } from '@bangle.dev/utils';
 import ChatBubbleIcon from '@mui/icons-material/ChatBubble';
+import ChatOutlinedIcon from '@mui/icons-material/ChatOutlined';
 import CodeIcon from '@mui/icons-material/Code';
 import DescriptionOutlinedIcon from '@mui/icons-material/DescriptionOutlined';
 import FormatListBulletedIcon from '@mui/icons-material/FormatListBulleted';
@@ -367,7 +368,7 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]> 
     {
       uid: 'quote',
       title: 'Quote',
-      icon: <ChatBubbleIcon sx={{
+      icon: <ChatOutlinedIcon sx={{
         fontSize: 16
       }}/>,
       description: 'Insert a quote in the line below',
--- a/components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
+++ b/components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
@@ -364,6 +364,33 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]> 
         };
       },
     },
+    {
+      uid: 'quote',
+      title: 'Quote',
+      icon: <ChatBubbleIcon sx={{
+        fontSize: 16
+      }}/>,
+      description: 'Insert a quote in the line below',
+      editorExecuteCommand: () => {
+        return (state, dispatch, view) => {
+          rafCommandExec(view!, (state, dispatch) => {
+            return insertNode(state, dispatch, state.schema.nodes.quote.create(
+              undefined,
+              Fragment.fromArray([
+                state.schema.nodes.paragraph.create(undefined, Fragment.fromArray([
+                  state.schema.text("Add a quote")
+                ]))
+              ])
+            ))
+          })
+          return replaceSuggestionMarkWith(palettePluginKey, '')(
+            state,
+            dispatch,
+            view,
+          );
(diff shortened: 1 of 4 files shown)


components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
```diff
@@ -364,6 +364,33 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]>
         };
       },
     },
+    {
+      uid: 'quote',
+      title: 'Quote',
+      icon: <ChatBubbleIcon sx={{
```
mattcasey, line 370: can we use a different icon from callout? How about "Chat" for blockquote: 
image


Or "CalltoAction" for callout:
image

(I think the Chat icon would be my preference)

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
```diff
@@ -364,6 +364,33 @@ const paletteGroupItemsRecord: Record<string, Omit<PaletteItemType, "group">[]>
...
+      }}/>,
+      description: 'Insert a quote in the line below',
+      editorExecuteCommand: () => {
+        return (state, dispatch, view) => {
+          rafCommandExec(view!, (state, dispatch) => {
+            return insertNode(state, dispatch, state.schema.nodes.quote.create(
+              undefined,
+              Fragment.fromArray([
+                state.schema.nodes.paragraph.create(undefined, Fragment.fromArray([
+                  state.schema.text("Add a quote")
```
mattcasey, line 381: I think it's better to leave it empty if we can't have a placeholder, it felt weird having to delete this text

components/editor/Quote.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ReactNode } from 'react';
+import { getSuggestTooltipKey } from './@bangle.dev/react-emoji-suggest/emoji-suggest';
+import { emojiSuggestKey } from './EmojiSuggest';
+
+const StyledBlockQuote = styled.div`
+  border-left: 8px solid ${({ theme }) => theme.palette.quoteMarker.main};
+  font-size: 20px;
+  padding: ${({ theme }) => theme.spacing(1)};
+  margin-top: ${({ theme }) => theme.spacing(1)};
+  border-radius: ${({ theme }) => theme.spacing(0.5)};
```
mattcasey, line 13: Not sure if this was intentional, but I think the left border look better as a long rectangle. Right now, only the left corners are rounded:
image

theme/index.ts
```diff
@@ -196,6 +198,9 @@ export const createThemeLightSensitive = (mode: PaletteMode) => {
       emoji: {
         hoverBackground: mode === 'dark' ? 'rgba(255, 255, 255, 0.08)' : 'rgba(0, 0, 0, 0.04);'
       },
+      quoteMarker: {
+        main: mode === 'dark' ? '#F3EFF5' : '#111111'
```
mattcasey, line 202: The dark mode bar is very hard to see. I don't think we need new colors here.. we could use 'text.secondary' or 'text.primary'? Looking at Notion, they just use the text color but also add a bit more margin which I think helps

mattcasey (changes requested):
style changes

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
Devorein: I agree those looks much more suitable

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
Devorein: One thing I would like to point out is that we also show a default text for callout if we remove it for quote we need to remove it for callout as well

components/editor/Quote.tsx
Devorein: I feel like the width could be decreased a bit, right now takes too much of the attention. Notion uses a lot less wide border.
image

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
motechFR: Handled as of this [change](https://github.com/charmverse/app.charmverse.io/pull/165/commits/34a85ca08c4c18da821548b1a89e6f8d8dbddc6f)

components/editor/@bangle.io/extensions/inline-command-palette/use-editor-items.tsx
motechFR: I've deleted the default text for quote and callout in this [change.](https://github.com/charmverse/app.charmverse.io/pull/165/commits/61b4da157de69b83f102502151950513f7be0c14)

Good to know, we need to provide at least an empty paragraph node, otherwise the quote cannot be modified. See the picture below.

image

components/editor/Quote.tsx
motechFR: I've halved the width of blockquote left border (8 => 4px) and removed radius

image

theme/index.ts
motechFR: Was looking for best contrast, text is indeed the best one! I've updating that and expanded inner padding 

image

PR #166, opened 2022-03-03
Implement sortable navigation

mattcasey:
This works by adding an "index" attribute to every page and sorting on it in the sidebar. Details:
- the default index of all pages is "-1", so by default this means it will always go to the top of a folder.
- We group pages by parentId and then sort by index. (So for example it's possible many pages will have the same index).
- When setting the order of a page, we reset the index value of all its siblings

Other data models I considered:
- a linked list, where pages link to the next page. I'm not a comp sci major, though, and this seems to have some tougher problems:
  - how do you prevent or what happens if two pages link to the same page?
  - have to ensure that pages don't link to a non-existing page (what happens if you delete a page that's linked?)
  - when moving a page from point A to point B, we need to update siblings in A as well as B
- store a sorted array on the parent
  - has similar problems as linked list related to stale references, etc.
- store a 'map document' that just keeps track of all the page ids.
  - while this will also need cleanup when we delete a page, it seems much easier to handle and think about
  - this wouldnt let us have atomic updates, so if two users make updates they could overwrite what the other one is doing


.eslintrc.json
```diff
@@ -81,6 +81,7 @@
     ],
     "object-curly-newline": "off",
     "padded-blocks": "off",
+    "prefer-destructuring": "off",
```
mattcasey, line 84: this is intentional... i don't think it really matters, both are equally readable, but sometimes destructuring still look weird to me

PR #169, opened 2022-03-04
make it so a user can re-login again by going thru signup

mattcasey:
It's possible for the user's server session to conflict with the connected wallet right now, but I'm trying to address one problem which is allowing a user to re-login either from a 2nd browser or if they delete their session cookie.

Current bugs that need to be addressed later:
- there's no way to fully log out of your server session. switching your wallet address doesn't log you out or change the address we're showing in most places
- when you do connect the right address, we dont have a way to update your server session and lookup the user profile

I'm making it so that if you log in on another browser, or delete your session cookie, we'll send you to that same "Create workspace" button once you connect your wallet, but clicking it will log you in and take you to the existing workspace.. thats what i had recommended in the chat

### DIFF
--- a/components/signup/SignupForm.tsx
+++ b/components/signup/SignupForm.tsx
@@ -49,11 +49,12 @@ export default function SignupPageContent () {
   const [user, setUser] = useUser();
   const router = useRouter();
 
+  // redirect user to workspace if they are already logged in
   useEffect(() => {
     if (user) {
       router.push('/');
     }
-  }, [user]);
+  }, []);
 
   async function createAccount () {
     const newUser = await charmClient.createUser({ address: account! });
--- a/hooks/useBounties.tsx
+++ b/hooks/useBounties.tsx
@@ -24,7 +24,7 @@ export function BountiesProvider ({ children }: { children: ReactNode }) {
           setBounties(_bounties);
         });
     }
-  }, [space]);
+  }, [space?.id]);
 
   const value = useMemo(() => ({
     bounties,
--- a/hooks/useSpaces.tsx
+++ b/hooks/useSpaces.tsx
@@ -1,6 +1,7 @@
 import { ReactNode, createContext, useContext, useState, useEffect, useMemo } from 'react';
 import { Space } from 'models';
 import charmClient from 'charmClient';
+import { useUser } from './useUser';
 
 type IContext = [spaces: Space[], setSpaces: (user: Space[]) => void];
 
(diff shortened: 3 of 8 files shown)


components/signup/SignupForm.tsx
```diff
@@ -49,11 +49,12 @@ export default function SignupPageContent () {
   const [user, setUser] = useUser();
   const router = useRouter();
 
+  // redirect user to workspace if they are already logged in
   useEffect(() => {
     if (user) {
       router.push('/');
     }
-  }, [user]);
+  }, []);
```
mattcasey, line 57: This rule was meant to just redirect on page load, but is conflicting with createAccount below when it also sends the user to /createWorkspace.. it just happens to be working because the index page also redirects you to "/createWorkspace" if you don't have one yet

hooks/useBounties.tsx
```diff
@@ -24,7 +24,7 @@ export function BountiesProvider ({ children }: { children: ReactNode }) {
           setBounties(_bounties);
         });
     }
-  }, [space]);
+  }, [space?.id]);
```
mattcasey, line 27: no need to update the state if you just changed the workspace name, etc.

hooks/useSpaces.tsx
```diff
@@ -1,22 +1,26 @@
...
-      .catch(err => {});
-  }, []);
+    if (user) {
+      charmClient.getSpaces()
+        .then(_spaces => {
+          setSpaces(_spaces);
+        })
+        .catch(err => {});
+    }
+  }, [user?.id]);
```
mattcasey, line 23: once you log in, we need to reload the spaces

pages/_app.tsx
```diff
@@ -286,8 +286,8 @@ function FocalBoardProviders ({ children }: { children: ReactNode }) {
 
 function DataProviders ({ children }: { children: ReactNode }) {
   return (
-    <SpacesProvider>
-      <UserProvider>
+    <UserProvider>
+      <SpacesProvider>
```
mattcasey, line 290: spaces needs to depend on users.. this order makes sense to me, i dont think we thought too much about the previous order

pages/createWorkspace.tsx
```diff
@@ -30,6 +31,12 @@ export default function CreateSpace () {
     domain: getDomainFromName(defaultName)
   };
 
+  useEffect(() => {
+    if (spaces.length > 0) {
```
mattcasey, line 35: hack: redirect from this page to your existing space if you have one

components/signup/SignupForm.tsx
motechFR: One suggestion I would have, is having all of our automated redirects at the page level (rather than nested components).

This would simplify our analysis of redirect flows in future.

PR #171, opened 2022-03-04
Implement token gates for workspace

mattcasey:


charmClient.ts
```diff
@@ -349,8 +353,32 @@ class CharmClient {
...
-    return http.POST('/api/transactions', details);
+  // Token Gates
+  getTokenGates (query: { spaceId: string }) {
+    return http.GET<TokenGate[]>('/api/token-gates', query);
+  }
+
+  getTokenGateForSpace (query: { spaceDomain: string }) {
+    return http.GET<TokenGate[]>('/api/token-gates', query).then(result => result[0]);
+  }
+
```
motechFR, line 364: These 2 methods seem redundant, why not have only 1?

PR #174, opened 2022-03-06
Improve join workspace form UI

mattcasey:


theme/styles.scss
```diff
@@ -1,9 +1,23 @@
 /* Make the Next.j wrappers 100% height */
-html,
-body,
-body > div:first-child,
-div#__next {
+
+
+html {
   height: 100%;
+width: 100%;
```
mattcasey, line 6: display:table here and the CSS below was added to fix an issue where the HTML element wouldn't actually go full viewport height.

package.json
```diff
@@ -146,6 +146,7 @@
     "prettier": "^2.5.1",
     "prisma": "^3.9.2",
     "sass": "^1.49.0",
+    "sharp": "^0.30.2",
```
mattcasey, line 149: added by recommendation of the next cli! sharp is an amazing package though

PR #176, opened 2022-03-06
Fix conflict between different @emotion instances

mattcasey:
This was hard to track down.. In only our production build, the lit modal component was destroying our CSS styles. I realized the class definitions from @emotion/react were disappearing. I have a hunch thi sis because react-windows-select also uses @emotion and it was overriding our own. I added a cache for our Emotion styles with its own name (so the classes are all prefixed with "app" now). This seems to work on my laptop and hoping for the best in production!

Related issues:
https://github.com/JedWatson/react-select/issues/3309
https://github.com/emotion-js/emotion/issues/2210

### DIFF
--- a/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
+++ b/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
@@ -73,7 +73,7 @@ function WalletSelectorModal ({
       <DialogTitle onClose={closeModal}>Connect to a wallet</DialogTitle>
       <Error error={error} processError={processConnectionError} />
       <Grid container spacing={2}>
-        <Grid item xs>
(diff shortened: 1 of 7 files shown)


components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
```diff
@@ -73,7 +73,7 @@ function WalletSelectorModal ({
       <DialogTitle onClose={closeModal}>Connect to a wallet</DialogTitle>
       <Error error={error} processError={processConnectionError} />
       <Grid container spacing={2}>
-        <Grid item xs>
+        <Grid item xs={12}>
```
mattcasey, line 76: for some reason these boxes started to collapse with the lit's modal css included but i can't figure out which rule it is.

PR #181, opened 2022-03-07
Enforce board view

motechFR:
Share specific focalboard views

image


pages/api/public/view/[id]/index.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const view = await prisma.block.findFirst({
+    where: {
+      id: id as string,
+      type: 'view'
+    }
+  });
+
+  if (!view) {
+    return res.status(400).json({ error: 'No such view exists' } as any);
```
Comment (mattcasey) [inline pages/api/public/view/[id]/index.ts:25]: i think this should be a 404?

mattcasey (approved):

PR #184, opened 2022-03-07
Replaced Bangle Emoji picker with emoji mart for page icon

Devorein:


components/common/page-layout/PageNavigation.tsx
```diff
@@ -197,16 +198,36 @@ export function PageLink ({ children, href, label, labelIcon, pageId }: PageLink
...
                 return {
                   ...page,
                   icon: emoji
                 };
               }
               return page;
             }));
+
+            if (currentPage && isCurrentPageEdited) {
+              setCurrentPage({
```
mattcasey, line 222: I'm nervous about having this setter here... currentPage should depend on pages, why should you have to know to update 'currentPage' as well? What do you think about adding a useEffect either inside of usePages or inside of `[pageId].tsx` instead, which is currently the only place we set currentPage. it could watch pages and check if the icon changes. just that way we dont have multiple places updating the page, i'm worried the sidebar might end up with a stale state of the pgae or something?
Also, `isCurrentPageEdited` is always going to be true if currentPage exists

mattcasey: So I tried this out and it seems to work in `pageId].tsx`:
```

  useEffect(() => {
    const page = pages.find(p => p.id === currentPage?.id);
    if (page) {
      setCurrentPage(page);
    }
  }, [pages]);
  ```
  It will be faster when we make pages an object map instead of an array

File: components/editor/Editor.tsx [hunk 2]
```diff
@@ -114,9 +123,37 @@ export function Editor ({ page, setPage, readOnly = false }: IEditorProps) {
...
+                    icon={<EmojiIcon />}
+                    name={intl.formatMessage({ id: 'ViewTitle.pick-icon', defaultMessage: 'Pick icon' })}
+                  >
+                    <EmojiPicker onSelect={(emoji) => {
+                      updatePageIcon(emoji);
+                    }}
+                    />
+                  </Menu.SubMenu>
+                  <Menu.Text
+                    id='remove'
```
mattcasey, line 148: Why do we need to define an id on these? I never use id since it's hard t oguarantee it is unique across the page

PR #185, opened 2022-03-07
Get viewId from the query, let DatabaseEditor access it from the query

motechFR:
Revert to previous behaviour of letting focalboard load its view from the url query


pages/share/view/[sharedViewId].tsx
```diff
@@ -1,10 +1,22 @@
...
   const router = useRouter();
 
   const { sharedViewId } = router.query;
 
-  return <BlocksEditorPage publicShare={true} viewId={sharedViewId as string} />;
+  const { viewId } = router.query;
+
+  if (!viewId) {
+    const newPath = generatePath(router.pathname, { ...router.query, sharedViewId });
+    router.replace({
```
Comment (mattcasey) [inline pages/share/view/[sharedViewId].tsx:14]: oh this is brilliant, we dont have to actually trigger a redirect

PR #186, opened 2022-03-07
allow removing properties to group focalboards

mattcasey:


motechFR (approved):
Tested this and works

- Opened card
- Deleted the status
- All cards are now in the same column
- The "Group by" section asks me to add a select type property

PR #191, opened 2022-03-08
Use ENS to resolve wallet addresses in focalboard

mattcasey:
image


motechFR (approved):
Tested, and works great!

Devorein:
That looks great. I just changed one line of CSS to fix the position of the user property value, since it's shifted down a bit.

Devorein (approved):

PR #193, opened 2022-03-08
Fix scrollable viewport for kanban boards

mattcasey:


Devorein:
The first thing I notice is that the banner isn't hidden when we don't have a page icon. But if we do have one it does get hidden.
🤔 Also shouldn't we be scrolling the whole page? It seems we are only doing so for the content rather than the header (page title, icon, description)

image
image
image

its not possible to view the content anymore

pages/_app.tsx
```diff
@@ -157,7 +157,10 @@ import { Provider as ReduxProvider } from 'react-redux';
 import 'react-resizable/css/styles.css';
 import { createThemeLightSensitive } from 'theme';
 import 'theme/@bangle.dev/styles.scss';
-import 'theme/focalboard/styles.scss';
+import 'theme/focalboard/focalboard.typography.scss';
```
mattcasey, line 160: incidentally, Next.js will not autoreload the page if you import an .scss file outside of _app.tsx! @Devorein

motechFR:
Works like a charm! (pun intended)

Tested in board view
- Header row stays fixed even when scrolling vertically
- User does not need to go to bottom of page in order to scroll horizontally

PR #200, opened 2022-03-09
redirect user to join space  if they dont have access

mattcasey:
If I go to any workspace page, for example https://app.charmverse.io/customdomain/foo, i should be redirected to /joinWorkspace?domain=customdomain&returnUrl=https://app.charmverse.io/customdomain/foo

[ ] - tested logged in user
[ ] - tested logged out user


motechFR:
@mattcasey 
Just noticed that a user with an old session but no account may still have a valid cookie. Pushed a fix for this so it doesn't fail

motechFR:
[X] logged in user => Redirects to join workspace
[] non logged in user => Redirects to create workspace / join workspace

motechFR:
[X] logged in user => Redirects to join workspace w/ the token gates
[] non logged in user => Redirects to create workspace / join workspace

PR #202, opened 2022-03-10
Full integration with emoji-mart and phasing out `@bangle.dev/emoji`

Devorein:
Integrated `emoji-mart` package in the following places

1. Page icon
2. Sidebar page navigation icon
3. Callout icon
4. `:` to trigger `emoji-mart`

Please check out these four areas while reviewing


motechFR:
Looks good!

motechFR (approved):

components/common/NewPageMenu.tsx
```diff
@@ -69,11 +74,11 @@ export default function NewPageMenu ({ addPage, tooltip, ...props }: Props) {
         onClose={handleClose}
       >
         <MenuItem onClick={() => createPage({ type: 'page' })}>
-          <ListItemIcon><StyledArticleIcon fontSize='small' /></ListItemIcon>
+          <ListItemIcon><StyledArticleIcon fontSize='small' isLightMode={theme.palette.mode === 'light'} /></ListItemIcon>
           <Typography sx={{ fontSize: 15, fontWeight: 600 }}>Add Page</Typography>
         </MenuItem>
         <MenuItem onClick={() => createPage({ type: 'board' })}>
-          <ListItemIcon><StyledDatabaseIcon fontSize='small' /></ListItemIcon>
+          <ListItemIcon><StyledDatabaseIcon fontSize='small' isLightMode={theme.palette.mode === 'light'} /></ListItemIcon>
```
mattcasey, line 81: we should be able to do this with simple CSS and 'color' property, and 'fill' set to 'currentColor'

mattcasey (changes requested):
Did you try applying colors to icons using CSS first?

components/common/NewPageMenu.tsx
mattcasey: Actually i'm not sure what this is fixing. Icons look fine to me in production, tested in Chrome and Firefox:
image
image

Devorein:
fill="currentColor" would not turn down the opacity in dark mode, it seemed like in light mode the icons were a bit transparent mainly due to the background being white. But I think it's fine to leave it as it is. I've reverted the style changes back to their original state

PR #205, opened 2022-03-10
Remember the last view for a focalboard

Devorein:
Steps to Test
1. Update the view of a focalboard
2. Switch to a different page
3. Navigate back to the focalboard page and see if the view is persisted

1. Move to a regular page (Needed as reloading on the same page would keep the same URL with the `viewId` query parameter)
2. Reload the page
3. Navigate to the focalboard page whose view was changed, and see if it persisted

Areas of Concern
1. Currently we are updating the local storage every time the `pages` state changes. So even if we update the icon or change the content of a page, it gets persisted to local storage which is redundant


hooks/usePages.tsx
```diff
@@ -96,6 +98,14 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     router.push(`/${(space!).domain}/${newPage.path}`);
   };
 
+  const pagesRecord = useMemo(() => {
```
mattcasey, line 101: this really belongs in its own PR where we can update everything at once. I dislike having two ways of doing anything at the same time, unelss its really really important - otherwise we may forget to actually update it everywhere

hooks/useFocalboardViews.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  useEffect(() => {
+    pages.forEach(page => {
+      // Get the value from localstorage first
+      if (page.type === 'board' && page.boardId) {
+        focalboardViewsRecord[page.boardId] = localStorageValue?.[page.boardId] ?? null;
+      }
+    });
+
+    blocks.forEach(block => {
```
mattcasey, line 32: this is an expensive loop just to set defaults, please remove

Devorein: Yes absolutely. We decided on populating the views for each page one by one (when the user visits them) rather than doing it all at the same time.

PR #211, opened 2022-03-11
simplify authentication

mattcasey:
Test Cases to try:

Existing user:
- Log out of Metamask and back in. Should take you to the wallet connect screen and then automatically to your workspace

User with multiple accounts:
- Try switching accounts. Right now I think we'll take you to the /join page... i think we need some UI way of getting to your own workspaces (maybe a link at top left?), or maybe we should be smarter and check if the new address has its own spaces to go to?

New user
- Try using an address that's never had a profile, we should send you to the /signup page to either create or join a new workspace

### DIFF
--- a/components/common/RouteGuard.tsx
+++ b/components/common/RouteGuard.tsx
@@ -77,6 +77,7 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
   async function authCheck (url: string): Promise<{ authorized: boolean, redirect?: UrlObject, user?: User }> {
     const path = url.split('?')[0];
     const spaceDomain = path.split('/')[1];
+    console.log(spaceDomain, path);
 
     // condition: public page
     if (publicPages.some(basePath => path.startsWith(basePath))) {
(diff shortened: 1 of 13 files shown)


Devorein:
For multiple accounts sometimes I'm redirected to my workspace and other time I get redirected to the `/join` page. But I do think it's best if we first check if the current switched address has any spaces to go to and redirect the user there, as they might not always want to join a workspace.

For a completely new user, I get redirected to either the `/join` page or the `/createWorkspace` page, but not to the `/signup` page

motechFR:
All below tests performed starting from an active workspace

**Existing user:**

Log out of Metamask and back in. Should take you to the wallet connect screen and then automatically to your workspace

[X] Works perfect

**User with multiple accounts:**

Try switching accounts. Right now I think we'll take you to the /join page... i think we need some UI way of getting to your own workspaces (maybe a link at top left?), or maybe we should be smarter and check if the new address has its own spaces to go to?

[X] Switch to account that is a contributor to current space. I can see my spaces. I am on current space

[X] Switch to account that is NOT a contributor to current space. I am now in the join workspaces screen with the current space prefilled
[X] On the join workspaces page, I switch to the previous account with access to this space. I am redirected to that space


[X] Create a new, non connected Metamask account. Nothing happens, and no account switch fires off. Once I click connect, I am rerouted to the "join workspace" for that page


New user

[]Try using an address that's never had a profile, we should send you to the /signup page to either create or join a new workspace  ==> I am sent directly to the create workspace page

Devorein (approved):

PR #212, opened 2022-03-11
Link to page using `/` shortcut

Devorein:
1. Type `/` to view the `Link to Page` option
2. Click on the option to view a list of all pages in the workspace
3. Click on any of the pages to see it get embedded in the editor

I tried to implement @motechFR suggestion to allow navigation via arrow key and select page on pressing enter and was almost close to completing it, but just couldn't seem to get it working. I've added that to the backlog as that's quite important for UX

### DIFF
--- a/components/common/page-layout/PageNavigation.tsx
+++ b/components/common/page-layout/PageNavigation.tsx
@@ -135,14 +135,14 @@ const PageAnchor = styled.a`
   }
 `;
 
-const PageIcon = styled(EmojiCon)`
+const StyledPageIcon = styled(EmojiCon)`
   height: 24px;
   width: 24px;
   margin-right: 4px;
   color: ${({ theme }) => theme.palette.secondary.light};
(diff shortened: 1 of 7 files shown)


components/common/page-layout/PageNavigation.tsx
```diff
@@ -183,9 +183,9 @@ export function PageLink ({ children, href, label, labelIcon, pageId }: PageLink
     <Link href={href} passHref>
       <PageAnchor onClick={stopPropagation}>
         {labelIcon && (
-          <PageIcon {...bindTrigger(popupState)}>
+          <div {...bindTrigger(popupState)}>
```
mattcasey, line 186: Not sure if this was by accident, but PageIcon has some CSS we need, such as right margin and the hover effect to let you know you can change the icon.

mattcasey:
Approved with comment!

PR #224, opened 2022-03-13
Import notion page to charmverse

Devorein:
1. Create an integration for your notion workspace, give it read permission to shared pages
2. Go to `.env.local` and add two env variables
	1. `NOTION_OAUTH_CLIENT_ID`
	2. `NOTION_OAUTH_SECRET`
3. Click on settings, and under Workspace tab you'll find the import section where the Import Workspace button is present
4. For now I've managed to create a mapping for the following notion blocks to that of CharmEditor
`heading 1, heading 2, heading 3, paragraph, bullet list, todo list, numbered list, callout, quote, code, divider, video, embed, bookmark, image, table`
5. Note that blocks like `bullet list, callout, quote` can contain any level of nested content, I've limited it to 10 levels of nested content for now. 
6. Along with the following blocks I've also managed to carry on all sorts of inline formatting for text as well (code, underline, bold, strike, and so on)
	
image


pages/[domain]/settings/workspace.tsx
```diff
@@ -94,6 +99,36 @@ export default function WorkspaceSettings () {
...
+              </g>
+            </svg>
+          )}
+        >
+          <Typography
```
Comment (mattcasey) [inline pages/[domain]/settings/workspace.tsx:122]: The text is too large. Maybe a card format with a button inside would be better, but for now can you just use the 'large' size button with normal text, so it will look like:
image

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
+/* eslint-disable no-await-in-loop */
```
mattcasey, line 1: this file is enormous, please remove all the logic about formatting, etc. and put it in a new lib/notion folder. this file can mostly just be the wiring up of that logic to the request

motechFR (approved):

pages/api/notion/callback.ts
mattcasey, line 1: also just add "no-await-in-loop": "off" to our eslint. i dont see why that's a big deal

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+  plain_text: string;
+  href: string | null;
+};
+
+type GetDatabaseResponse = {
+  title: Array<RichTextItemResponse>;
+  icon: {
+    type: 'emoji';
+    emoji: string;
+  } | null | {
```
mattcasey, line 49: i think this would be more readable if each type in the union was a separate type. i'm not sure why we have so many nulls in these unions either..

```
type FileBlock = { ... }
type CoverBlock = { ... }
type GetDatabaseResponse = FileResponse | CoverBlock | null;
```

mattcasey: Or for now, i'd even say just put these in their own file "lib/notion/types.ts" and we can clean up later

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+    redirect: string
+    spaceId: string
+    userId: string
+  } = {} as any;
+  try {
+    state = JSON.parse(decodeURIComponent(req.query.state as string));
+  }
+  catch (e) {
+    console.error('Error parsing state notion callback', e);
+    res.status(400).send('Invalid state');
```
mattcasey, line 1839: let's always send JSON back for errors... send({ error: "Invalid state" }); it makes it easier to handle errors on the frontend if they always have the same shape

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+
+  await importFromWorkspace({
+    accessToken: token.access_token,
+    spaceId: state.spaceId,
+    userId: state.userId,
+    workspaceName: token.workspace_name,
+    workspaceIcon: token.workspace_icon
+  });
+  const cookies = new Cookies(req, res);
+  cookies.set('notion-user', userId, {
```
mattcasey, line 1864: do we actually need to set this cookie? It doesn't seem like it

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+      content: [{
+        type: 'paragraph',
+        content: []
+      }]
+    },
+    headerImage: block.cover?.type === 'external' ? block.cover.external.url : null,
+    icon: block.icon?.type === 'emoji' ? block.icon.emoji : null,
+    title,
+    type: 'board',
+    pageId: v4(),
```
mattcasey, line 1442: please put any default values inside of createPrismaPage, including content -- i actually believe we do not need to provide any content to begin with.

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+  });
+
+  let searchResult = await notion.search({
+    page_size: 100
+  });
+
+  const searchResults = searchResult.results;
+
+  // While there are more pages the integration has access to
+  while (searchResult.has_more && searchResult.next_cursor) {
```
mattcasey, line 1468: can you add a limit here? so if searchResults.length > 1000 then we exit the loop.

pages/api/notion/callback.ts
```diff
@@ -0,0 +1,1871 @@
...
+          ...commonBlockData
+        }]
+      });
+    }
+
+    return createdPages[blockId];
+  }
+
+  const workspacePageId = v4();
+  await createPrismaPage({
```
mattcasey, line 1774: theres no reason to pre-generate the pageid. just do `page = await createPrismaPage(...)` and you'll have access to the new page id. just trying to help keep this code simple.. shouldn't need to pass 'null' values in either

charmClient.ts
```diff
@@ -186,7 +195,7 @@ class CharmClient {
   private blockToFBBlock (block: Block): FBBlock {
     return {
       ...block,
-      deleteAt: block.deletedAt ? new Date(block.deletedAt).getTime() : 0,
+      deletedAt: block.deletedAt ? new Date(block.deletedAt).getTime() : 0,
```
mattcasey, line 198: I didn't realize you wanted to change the field on focalboard... please confirm if we have to migrate existing data in production

mattcasey:
@Devorein one more thing, can you add the new env vars to .env.example?

pages/[domain]/settings/workspace.tsx
Comment (Devorein) [reply pages/[domain]/settings/workspace.tsx:122]: Yes absolutely. Doing it right away. But shouldn't it say Import workspace from Notion rather than pages?

pages/api/notion/callback.ts
Devorein: I agree. The types themselves take up quite a lot of lines. I wasn't sure where to transfer them to. I will move the business logic to `/lib/notion` that way we can also write unit tests for it quite easily

Devorein: Agreed but I guess sometimes it's better to aggregate all those promises inside an array and use `Promise.all` to resolve all of them concurrently, but its definitely something the programmer has better knowledge of when and when not to do than eslint. Disabling seems to be the correct option.

pages/api/notion/callback.ts
Devorein: Notion's official api has a huge type. They tried to play it too clever (by automatically generating types from their code) and ended up writing way too much type-related code. I just imported those from their library as they don't explicitly export those for some unknown reason. The repo already has a pretty big issue thread regarding this
https://github.com/makenotion/notion-sdk-js/issues/219

pages/api/notion/callback.ts
Devorein: My bad. That makes sense.

pages/api/notion/callback.ts
Devorein: Right now no we don't actually use it anywhere. I will remove it asap

pages/api/notion/callback.ts
Devorein: Makes sense. Just fixed it

pages/api/notion/callback.ts
Devorein: Yeah, I've added a few global variables to manipulate the whole import. One of them is `IMPORTED_PAGES_LIMIT` which does exactly what you just said.

pages/api/notion/callback.ts
Devorein: Updated

charmClient.ts
```diff
@@ -131,6 +131,15 @@ class CharmClient {
     return http.POST<InviteLinkPopulated[]>(`/api/invites/${id}`);
   }
 
+  // Notion Import
+  importFromNotion ({ spaceId }: {spaceId: string}) {
```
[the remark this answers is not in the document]
Devorein: Good catch. I used it during the initial stages but forgot to remove it.

charmClient.ts
Devorein: Yeah, that's why I went with the change. Our Prisma model uses `deletedAt` so not sure why we used `deleteAt` all this time

Devorein:
Yes absolutely. I should've added that eariler

mattcasey (approved):

PR #225, opened 2022-03-14
fix flash of login page on first load

mattcasey:
There's a flash of the login page when you go to https://app.charmverse directly, that's because the redirect inside `useEffect()` is happening after the component mounts. The fix is to rely on a different value to trigger the login view which is set in the useEffect method

### DIFF
--- a/components/common/base-layout/getLayout.tsx
+++ b/components/common/base-layout/getLayout.tsx
@@ -10,3 +10,15 @@ export default function getLayout (page: ReactElement) {
     </PageWrapper>
   );
 }
+
+/**
+ * Use as an alternative to getLayout when you need to control if the whole page should appear
+ */
+export function BaseLayout ({ children }: {children: ReactElement []}) {
+  return (
+    <PageWrapper>
+      <Header />
+      {children}
+    </PageWrapper>
+  );
+}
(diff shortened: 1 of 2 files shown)


motechFR:
Hey @mattcasey I was still getting a flash on my end.

After testing, it turns out this was the base layout displaying. In this [commit](https://github.com/charmverse/app.charmverse.io/pull/225/commits/cfac097936270b567e309f754f397e9eb332ce4c), I provided a way for the root page to control showing the layout itself.

Let me know if this approach works!

motechFR (approved):

components/common/base-layout/getLayout.tsx
```diff
@@ -10,3 +10,15 @@ export default function getLayout (page: ReactElement) {
     </PageWrapper>
   );
 }
+
+/**
+ * Use as an alternative to getLayout when you need to control if the whole page should appear
+ */
+export function BaseLayout ({ children }: {children: ReactElement []}) {
```
mattcasey, line 17: this is functionally the same thing as getLayout. you can just put this in index.tsx:
```

  return (
    getLayout(
      <>
        <LoginPageContent account={account} />
        <Footer />
      </>
    )
  );
  ```

PR #233, opened 2022-03-15
Erc20 payments

motechFR:
This pull request enables DAOs to pay their community with ERC20 tokens

- Add custom tokens across all supported blockchains
- Retrieve data automatically where available
- Select custom tokens to pay a bounty

image


charmClient.ts
```diff
@@ -389,6 +390,22 @@ class CharmClient {
...
+
+  getTokenMetaData ({ chainId, contractAddress }: ITokenMetadataRequest): Promise<ITokenMetadata> {
+    return http.GET('/api/tokens/metadata', { chainId, contractAddress });
+  }
+
+  createPaymentMethod (paymentMethod: Partial<PaymentMethod>): Promise<PaymentMethod> {
+    return http.POST('/api/payment-methods', paymentMethod);
+  }
+
+  listPaymentMethods (spaceId: string): Promise<PaymentMethod []> {
```
mattcasey, line 402: there shouldn't be a space between the element name and the `[]`

components/bounties/BountyBadge.tsx
```diff
@@ -30,9 +32,11 @@ export function BountyBadge ({ truncate = false, bounty, direction = 'row', hide
   const [space] = useCurrentSpace();
   const theme = useTheme();
 
-  const bountyLink = `/${space!.domain}/bounty/${bounty.id}`;
+  const [paymentMethods] = usePaymentMethods();
+
+  const tokenInfo = getTokenInfo(paymentMethods, bounty.rewardToken);
 
-  const imageLogo = CryptoLogoPaths[bounty.rewardToken as CryptoCurrency];
+  const bountyLink = `/${space!.domain}/bounty/${bounty.id}`;
```
mattcasey, line 39: can you change this to /bounties/<bounty_id>? I didn't notice we make it singular when there's an ID but i think we should just stick with the REST design, where the resource part doesn't change

components/common/form/CustomErcTokenForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    try {
+      const createdPaymentMethod = await charmClient.createPaymentMethod(paymentMethod);
+      refreshPaymentMethods();
+      onSubmit(createdPaymentMethod);
+    }
+    catch (error: any) {
+      setFormError({
+        message: error.message,
```
mattcasey, line 129: in our middleware, we actually return a field `error` instead of `message`

lib/middleware/requireKeys.ts
```diff
@@ -0,0 +1,27 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import { IApiError } from 'lib/utilities/errors';
+
+/**
+ * Generates a request handler that checks for target keys
+ */
+export function requireKeys<T> (keys: Array<keyof T>, location: 'body' | 'query') {
```
mattcasey, line 7: Seems useful, also I believe it's possible to re-use our yup schemas inside of Next.js api calls. I forget the library right now that i used

lib/middleware/requireSpaceMembership.ts
```diff
@@ -0,0 +1,36 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import { prisma } from 'db';
+
+/**
+ * Allow an endpoint to be consumed if it originates from a share page
+ */
+export async function requireSpaceMembership (req: NextApiRequest, res: NextApiResponse, next: Function) {
```
mattcasey, line 7: i wrote something very similar in lib/users called hasAccessToSpace():
```
min
  const { error } = await hasAccessToSpace({
    userId: req.session.user.id,
    spaceId: gate.spaceId,
    role: 'admin'
  });
  if (error) {
    return res.status(401).json({ error });
  }
  ```

**Review by mattcasey — APPROVED** [review]
approved! with some style comments

File: components/settings/payment-methods/CompositePaymentMethodList.tsx [hunk 6]
```diff
@@ -0,0 +1,125 @@
+import DeleteIcon from '@mui/icons-material/Delete';
```
mattcasey, line 1: I would like to be consistent with our delete UI, for example whether we use an "X" or Trash can. Please use "@mui/icons-material/Close" to match the other pages, with a tooltip "Delete" around it. The delete/edit icons should also be hidden until you hover over the row.

components/common/form/CustomErcTokenForm.tsx
```diff
@@ -0,0 +1,278 @@
...
+              Contract address
+            </InputLabel>
+            <Input
+              {...register('contractAddress')}
+              type='text'
+              fullWidth
+            />
+            {
+              errors?.contractAddress && (
+              <Alert severity='error'>
```
mattcasey, line 165: This is a good case for Textfield, it has an 'error' attribute which is not as jarring to see:
image
```
<TextField
              {...register('contractAddress')}
              type='text'
              error={!!errors.contractAddress?.message}
              helperText={errors.contractAddress?.message}
              fullWidth
            />
            ```

File: components/settings/payment-methods/CompositePaymentMethodList.tsx [hunk 8]
```diff
@@ -0,0 +1,125 @@
...
+
+                </Box>
+
+              </TableCell>
+              <TableCell width={200} sx={{ px: 0 }}>
+                <Typography>
+                  <a href={getChainExplorerLink(row.chainId, row.contractAddress, 'token')} target='_blank' rel='noreferrer'>
+                    <LaunchIcon sx={{ mr: 1 }} />
+                  </a>
+                  <strong>{getChainById(row.chainId)?.chainName}</strong>
```
mattcasey, line 103: i think the launch icon is unecessary.. no one's really going to need to go visit the network. could you just wrap the label itself? and you should use common/components/Link.tsx. it has some basic styling for links. if you add `external` property it will use MUI's regular link component instead of the Next link component.

components/common/form/CustomErcTokenForm.tsx
```diff
@@ -0,0 +1,278 @@
+import { yupResolver } from '@hookform/resolvers/yup';
```
mattcasey, line 1: this form seems like it belongs with the other components in payment-methods, not sure i see a case where we'd need it more than once and I want to include Gnosis wallet here

components/settings/payment-methods/CompositePaymentMethodList.tsx
```diff
@@ -0,0 +1,125 @@
...
+                </Typography>
+              </TableCell>
+              <TableCell width={150} sx={{ px: 0 }} align='center'>
+
+                {
+                isAdmin && (
+                  <DeleteIcon
+                    onClick={() => {
+                      setPaymentMethodIdToDelete(row.contractAddress);
```
mattcasey, line 112: we should use row.id and not contract address

pages/api/payment-methods/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const existingPaymentMethod = await prisma.paymentMethod.findFirst({
+    where: {
+      chainId,
+      contractAddress,
+      spaceId
+    }
+  });
+
+  if (existingPaymentMethod) {
```
mattcasey, line 58: Is there an issue if a user creates a 2nd payment with the same contract address? They will see it in the UI. Logically, it doesn't make sense today, but that could change.. just want to be cautious about adding too much checks, it increases the amount of code we have to maintain. we should validate to prevent errors in the system, i care less about the user doing silly things

charmClient.ts
motechFR: Fixed in this [commit](https://github.com/charmverse/app.charmverse.io/pull/233/commits/655bd8ecd09af857bd001d1b5c22e56a14ee680a)

It would be great to enforce this across our codebase. I searched for a native ESLint rule, but didn't come across one that fits.

See this link from [Prettier playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLWcBOAzAhmOAAgFkBPASQx3yOAB0pDDsIIAKASiUIGcZN0AcwA0DJgCNcmbgEFMmXKQA8UAK4BbcVgB8oqAF8GDGKQAORACZx1EOQtKEAvCQpU8BQgG0AukahwAD1MITBhCXHE+BTAwsAAbXB4eQgAxVkJUdVM46wQYZLJKeGoPekZmVk5CMqZautqAegbCABUAeQARNu5MOBhVTEZcIZjVXDjCADdx1SIACyw4MXqmXv7BwgByBbi4iE3lusNywT7CSUwqmtq1gcZPJgBGPUPagCZhQgBmPTrfcuOx2MZiIACU4ABHWZ8FogpxbADiAFEWptCAAfLYACSRMg6aMxmwACm0AMqojFbIkAVQphLaRJa5DaADlSQStgBhVkspGculbDpIgAyKKRHM2LVBMk54pAwhAEFMMFQ0B4yFAUkwEAA7kSpAh1ShxjrFOqFeJogBrPqk3DqODC9BwZB4OI8OAW6220z4ITIfizBXWLQWKwWYXDQRjU5pTDqXAwFVQQTIRXYbDykBzGDqOIAdTmqHgPF9BFJhuLqEmxdIabASSz6A9oSJCkECdd4w9CoAVjwAqShDkAIqqCDwLvuz0gX2YFtpkzmHhgATKrOmASwfOoCwwObIAAcAAYFZuIB78wpTGnN3AW5MXQqoRO4G2lUaQIkALT+ODhrNeihVBejbXAO1wKcexAD11FQANMCDGDhzgMdXygmcYAiHc9wPJAPhAfhcFQOIhE5CB1E7FB7wAVizVQPRaCIjTdaDJlmSgrFgUlV1QZUZCgCxSRMHIMP0fQgA), which does seem to allow us to enforce this rule, albeit still need to find which rule does that

components/bounties/BountyBadge.tsx
motechFR: Fixed in this [commit](https://github.com/charmverse/app.charmverse.io/pull/233/commits/7657d594701ca44e6d116b8cc226ff67b0c78542)

components/common/form/CustomErcTokenForm.tsx
motechFR: Handled in this [commit](https://github.com/charmverse/app.charmverse.io/pull/233/commits/5698cc470a0f18f0e2b3a07c0cbe6c20fb6e8462) as per your feedback.

We could definitely gain from standardising errors.

See lib/utilities/errors.ts which I added as part of this PR for an initial go at this. It would be worth a chat to decide how we model errors and then build this into the system.
I have some previous research I've done on the topic which would be relevant

lib/middleware/requireSpaceMembership.ts
motechFR: I noticed that hasAccessToSpace existed. 2 challenges were the fact that it also includes some roles logic, and it must also be manually called (although I get your point about not having spaceId in original request). 

The route handler approach helps us focus the API controller only on what's needed.

Agree with you, we should come up with a DRY approach that has the best of both worlds.

One other approach, could be expanding the sessions to also include current space of the session, and compare that with spaceId references in the request.

components/settings/payment-methods/CompositePaymentMethodList.tsx
motechFR: Changed in this [commit](https://github.com/charmverse/app.charmverse.io/pull/233/commits/a4f2dcc59ff1cf09bf97ddf52a8fbef472181ee3)

I copied an existing implementation you had and wrapped that into a reusable ElementDeleteIcon, so we can guarantee same behaviour going forward.

components/common/form/CustomErcTokenForm.tsx
motechFR: Updated the fields in this [change](https://github.com/charmverse/app.charmverse.io/pull/233/commits/3bf719eca116cfc8aa4f6e0cc4e10f2628db86e9)

components/settings/payment-methods/CompositePaymentMethodList.tsx
motechFR: Thanks for the tip, I've made the change [here](https://github.com/charmverse/app.charmverse.io/pull/233/commits/97a0b278bb19fa9f2abf316b49757282ff96ab2c)

components/common/form/CustomErcTokenForm.tsx
motechFR: Moved to folder in this [commit](https://github.com/charmverse/app.charmverse.io/pull/233/commits/fb0e74cdb4c88fa968c3d387aa95f28753673dfe)

pages/api/payment-methods/index.ts
motechFR: I think a good philosophy, is to prevent users from accidentally doing stuff they'd most likely cancel after.

Also, the payment methods do contain the token decimals property, which we wouldn't want 2 versions of.

Generally, it's good to have only 1 Symbol + token name for the payment method

components/settings/payment-methods/CompositePaymentMethodList.tsx
motechFR: Changed [here](https://github.com/charmverse/app.charmverse.io/pull/233/commits/a9ae44308aff58feaafd6d55fc83c6b634b740ed)

PR #236, opened 2022-03-17
Custom crypto price

motechFR:
Display the price of custom cryptos added to payment methods

image


components/editor/CryptoPrice.tsx
```diff
@@ -65,6 +69,34 @@ export function CryptoPrice ({ preset, onQuoteCurrencyChange, onBaseCurrencyChan
   const [selectionList, setSelectionList] = useState<null | OptionListName>(null);
   const [error, setError] = useState<null | string>(null);
 
+  const [paymentMethods] = usePaymentMethods();
+
+  const [cryptoList, setCryptoList] = useState<string []>([]);
```
mattcasey, line 74: please no space between an interface and `[]`

components/editor/CryptoPrice.tsx
```diff
@@ -65,6 +69,34 @@ export function CryptoPrice ({ preset, onQuoteCurrencyChange, onBaseCurrencyChan
...
+
+  useEffect(() => {
+    const baseCurrencies: string [] = CryptoCurrencies.slice();
+
+    if (paymentMethods) {
+      Object.entries(paymentMethods)
```
mattcasey, line 80: paymentMethods is now a list, not a map.. i suppose this would not compile once merged to main?

components/editor/CryptoPrice.tsx
```diff
@@ -74,10 +106,14 @@ export function CryptoPrice ({ preset, onQuoteCurrencyChange, onBaseCurrencyChan
 
   function refreshPrice () {
     setLoadingState(true);
-    charmClient.getPricing(baseCurrency, quoteCurrency)
+
+    const symbol = getTokenInfo(paymentMethods, baseCurrency).tokenSymbol as CryptoCurrency;
```
mattcasey, line 110: please use caution when asserting types. Really I think it should only be used if it is 100% true, otherwise we're just covering over what TS is trying to protect us from. In this case,  baseCurrency is passed into the component so there's no guarantee it is valid. It's better to just throw an error in this case, since it's probably not likely but the sooner we can alert to incorrect input the easier it will be to fix

components/editor/CryptoPrice.tsx
mattcasey: I think this whole block of code can just be written as:
```
import { isTruthy } from 'lib/utilities/types';
...
const [paymentMethods] = usePaymentMethods();
const cryptoList = CryptoCurrencies.concat(
  paymentMethods.map(method => method.contractAddress).filter(isTruthy)
);
```
I also think it's fine to include testnets.. Are they not supported?

lib/crypto-price/dataSources/coinMarketCap.ts
```diff
@@ -0,0 +1,49 @@
...
+const apiServiceToken = process.env.CMC_API_TOKEN!;
+
+export function getPriceFromCoinMarketCap (
+  base: CryptoCurrency,
+  quote: FiatCurrency,
+  apiToken = apiServiceToken
+): Promise<IPairQuote> {
+  return new Promise((resolve, reject) => {
+    const cmcCryptoCurrencyId = CoinMarketCapCryptoMapping[base];
+    fetch(`https://pro-api.coinmarketcap.com/v2/cryptocurrency/quotes/latest?id=${cmcCryptoCurrencyId}&convert=${quote}`, {
```
mattcasey, line 24: Sent you that article this week about organization, and I think this file could go into 'adapters' folder, which captures the code to talk to external layers. Thoughts? I'm not sure i am in love with the idea, but i guess it's nice to know what the external layers are. I'm fine if u wanna leave this here until we have time to discuss general principles

mattcasey (changes requested):

components/editor/CryptoPrice.tsx
motechFR: Handled in this [change](https://github.com/charmverse/app.charmverse.io/pull/236/commits/09b6af07707e659c3b590b53546b0c82c2b5a360)

components/editor/CryptoPrice.tsx
motechFR: I have simplified as per your suggestion in this [commit](https://github.com/charmverse/app.charmverse.io/pull/236/commits/c73da265b5c3d7c249a8c17ffc03e395ff5ebe3f)

Testnets ignored as these tokens don't have an inherent market value

components/editor/CryptoPrice.tsx
motechFR: I've changed the signature of getPricing to a string in this [commit](https://github.com/charmverse/app.charmverse.io/pull/236/commits/5aba60edad6e50131e1cc32c1fadd8dfe0d2d260)

As we now are passing an undetermined range of crypto symbols to the pricing API, it makes more sense to use this Crypto | string type. See here for changes across the [stack](https://github.com/charmverse/app.charmverse.io/pull/236/commits/5c9c5449d7b7a1f1cea8f64b4f29005f7efdae96)

PR #239, opened 2022-03-18
add a tool to run locally with production Env vars

mattcasey:
Now, you can just run `npm run start:remote` this will run a dev env that's talking to our production DB. (you still need your ip address whitelisted though). This was helpful to me today because i could see the original source code where an exception occurred. I am using `react-env` instead of `dotenv` because it provides more flexibility, allowing us to change the env vars at runtime instead of build time.


package.json
```diff
@@ -29,6 +30,7 @@
     "@bangle.dev/react-menu": "^0.28.4-alpha.2",
     "@bangle.dev/table": "^0.28.4-alpha.2",
     "@bangle.dev/tooltip": "^0.28.4-alpha.2",
+    "@beam-australia/react-env": "^3.1.1",
```
mattcasey, line 33: Note that there's also a `react-env` node module, but it's literally just a package.json and hasn't been updated in 7 years. I reviewed this guy's source code as well since it accesses sensitive information!

PR #256, opened 2022-03-22
Markdown export

motechFR:
Add the export to markdown capability to pages in Charmverse

image

See below for the source Charmverse page, and the outputted markdown as well as how it renders 
image

image


components/common/page-layout/Header.tsx
```diff
@@ -113,6 +154,41 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
               </Tooltip>
             </>
           )}
+
+          {isPage && (
+            <Box sx={{ ml: 1 }} component='div' ref={pageMenuAnchor}>
```
mattcasey, line 159: image
This icon doesn't line up vertically with the ones next to it. i think it may just need to be warpped with "IconButton"? and fontSize='medium' is redundant i think

motechFR: Thanks for the feedback! Fixed in this [commit](https://github.com/charmverse/app.charmverse.io/pull/256/commits/fa4adb2bd2f3917571193a70d450403084251feb)

PR #268, opened 2022-03-23
Notion import v2

Devorein:
This PR introduces the following changes

1. Create the representation of the user's workspace in memory rather than creating pages as soon as we find them. There are certain advantages to this approach
	1. Its easier to handle the errors
	2. Its easier to report the errors
	3. Its easier to test
2. Previously we used to fail the whole import process even if one single block of a page failed to import. Now we handle the error more gracefully by
	1. Skipping over the block that failed to import
	2. Skipping over the pages that failed to import
3. Better error reporting when a page/database fails to import
	1. Show its id and title
	2. Show all the blocks that failed, this will be an array of array, for example, blocks like `callout` can contain nested blocks so the inner array contains the trails of blocks which led to the block that failed and the outer array contains all of the block trails that failed for the given database/page
4. For pages whose parent page failed we make the workspace page its parent
5. For database pages where the database wasn't created we don't add the pages (doesn't make any sense to add it to the workspace page as it's a regular page, not a database one), if we create those cards then they would only consume space and not be visible to the user as they've not been connected with a focalboard board (database page)
6. If the user's code is invalid or it's ill-formatted we show the correct error
7. We return the created pages and the errors that occurred during import from the API request. That way we don't need to reload the page. The issue with reloading the page is that the user might not be able to view all the errors that happened and we do create some pages so there is no concept for complete success or failure its most likely going to be partial

Sample error message
image


components/editor/NestedPage.tsx
```diff
@@ -44,6 +44,7 @@ const NestedPageContainer = styled((props: any) => <div {...props} />)`
   &:hover .actions-menu {
     opacity: 1;
   }
+  margin: ${({ theme }) => theme.spacing(1, 0)};
```
mattcasey, line 47: we shouldn't have any margin on top/bottom of components (see Notion as example.. all the lines are always the same height)

pages/[domain]/settings/workspace.tsx
```diff
@@ -154,11 +159,43 @@ export default function WorkspaceSettings () {
         >
           {isImportingFromNotion ? 'Importing pages from Notion' : 'Import pages from Notion'}
         </Button>
-        {notionError && (
+        {notionFailedImports.length !== 0 && (
           <Alert severity='error' sx={{ mt: 2 }}>
```
Comment (mattcasey) [inline pages/[domain]/settings/workspace.tsx:163]: make the severity 'warning' instead of error

pages/api/notion/import.ts
```diff
@@ -4,6 +4,8 @@ import * as http from 'adapters/http';
 import { NextApiRequest, NextApiResponse } from 'next';
 import { importFromWorkspace } from 'lib/notion/importFromWorkspace';
 import { withSessionRoute } from 'lib/session/withSession';
+import { Page } from '@prisma/client';
+import { FailedImportsError } from 'pages/[domain]/settings/workspace';
```
mattcasey, line 8: Should add `import type { FailedImportsError } ...` here just to be safe. I am not sure if webpack would be able to tree-shake from the frontend code or it would need to pull in all the other dependencies

pages/[domain]/settings/workspace.tsx
```diff
@@ -24,48 +24,53 @@ import CircularProgress from '@mui/material/CircularProgress';
...
-        .then(() => {
+        .then(({ failedImports, importedPages }) => {
           setIsImportingFromNotion(false);
           showMessage('Successfully imported');
           mutate(`pages/${space.id}`);
-          const baseUrl = `${router.asPath.split('?')[0]}?success=true`;
-          router.replace(baseUrl, undefined, { shallow: true });
+          showMessage('Notion workspace successfully imported');
+          setNotionFailedImports(failedImports);
+          setPages((pages) => ({
```
Comment (mattcasey) [inline pages/[domain]/settings/workspace.tsx:62]: you don't need to update the context, we are already calling mutate() above. I would also avoid sending the pages back in response, that could be a large payload

components/editor/NestedPage.tsx
Devorein: Alright removing it right away

pages/api/notion/import.ts
Devorein: Better be safe than sending typedefs on clients PC.

pages/[domain]/settings/workspace.tsx
Comment (Devorein) [reply pages/[domain]/settings/workspace.tsx]: Thank you. I was a bit confused with `mutate` but it makes total sense now. Yep, it could definitely be quite large. So better to not send all of them after import.

mattcasey (approved):

pages/[domain]/settings/workspace.tsx
Comment (Devorein) [reply pages/[domain]/settings/workspace.tsx]: Done

PR #272, opened 2022-03-24
Redirect user to previous url for failed discord connection

Devorein:
1. Redirect the user where they came from after connecting to discord, rather than leaving the user on our `discord/callback` API endpoint showing a json payload
2. We append `discord=fail` or `discord=success` to the query params of the redirected URL for the client to know which ux to show
3. Remove `discord` query parameter from the redirected URL as it could be different
4.  We show a snackbar based on the value of the `discord` query parameter
image


pages/_app.tsx
```diff
@@ -169,6 +169,8 @@ import {
 import 'theme/lit-modal/styles.scss';
 import { setTheme as setLitProtocolTheme } from 'theme/lit-modal/theme';
 import 'theme/styles.scss';
+import Snackbar from 'components/common/Snackbar';
```
mattcasey, line 172: I think this snackbar "could" be added with a global context instead, that way you only need to import useSnackbar. Right now, useSnackbar is just reducing a small amount of logic, but you're still doing most of the work

pages/api/discord/callback.ts
```diff
@@ -67,29 +67,43 @@ handler.get(async (req, res) => {
...
   const { id, ...rest } = discordAccount;
 
-  await prisma.discordUser.create({
-    data: {
-      account: rest as any,
-      discordId: id,
-      user: {
-        connect: {
-          id: state.userId
+  const url = new URL(state.href);
```
mattcasey, line 70: what is "state.href"? Is that the redirect? "href" is an anchor attribute and does'nt provide any context here, a better name could be "redirect"

pages/index.tsx
```diff
@@ -8,6 +8,8 @@ import Footer from 'components/login/Footer';
 import { useSpaces } from 'hooks/useSpaces';
 import { Web3Connection } from 'components/_app/Web3ConnectionManager';
 import { useUser } from 'hooks/useUser';
+import useSnackbar from 'hooks/useSnackbar';
+import Snackbar from 'components/common/Snackbar';
```
mattcasey, line 12: these are unused imports

mattcasey (changes requested):
one request before approval: don't put the discord code inside of _app.tsx

pages/_app.tsx
Devorein: Thats a good point. We are only ever going to use a single Snackbar at most so no point in using that component in multiple places and not using a hook/context to call it

pages/_app.tsx
```diff
@@ -259,6 +275,7 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
                       <RouteGuard>
                         <ErrorBoundary>
                           {getLayout(<Component {...pageProps} />)}
+                          <Snackbar severity={severity} handleClose={handleClose} isOpen={isOpen} message={message ?? ''} />
```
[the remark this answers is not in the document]
Devorein: Yep did that on [this commit](https://github.com/charmverse/app.charmverse.io/pull/272/commits/64d265ab319c16ac0776002d868041217138cb9d)

pages/index.tsx
Devorein: Good catch. Thank you. Fixed

pages/api/discord/callback.ts
Devorein: Good point. I thought we could use `redirect` but it might be confused with the redirect URL the oauth service would send after completing the flow but I guess that's called `callback` urkl. Fixed it

pages/_app.tsx
```diff
@@ -231,9 +234,22 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
 
   // wait for router to be ready, as we rely on the URL to know what space to load
   const router = useRouter();
+
+  // We might get redirected after connection with discord, so check the query param if it has a discord field
+  // It can either be fail or success
+  useEffect(() => {
+    if (router.query.discord === 'fail') {
```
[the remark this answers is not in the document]
Devorein: Yeah, I was thinking what could be a good place to put that logic. Thank you. Moved to the correct place.

Devorein:
Made all the changes you requested. Will merge as soon as the checks pass.

PR #280, opened 2022-03-25
use twemoji for icons and fix alignment

mattcasey:
According to Safwan, on Windows OS, Notion replaces the system emojis with Twitter emojis. On Mac, we see the Apple emojis, which I think is what we also see in the emoji picker. So, this is just making life better for non-Mac users, and more consistent w/Notion.

I also made the icons align when selecting or viewing 'nested pages' in Charmeditor


Devorein (approved):

motechFR (approved):
Using Ubuntu + Chrome, I see the twitter emoji in icon section

PR #281, opened 2022-03-25
Removed unnecessary snackbars

Devorein:
1. Removed local `Snackbar` to opt for a single global `Snackbar`
2. Close the snackbar using `setIsOpen` as soon as we change route
3. Test the following places where we used local snackbars
	1. Select a text and see if the floating menu appears or not
	2. Select a text and click on link format, see if a new floating menu appears or not
	3. Iframe and image blocks
	4. Create a child page `Insert Page` and click the right-most icon (to delete, duplicate and copy link)
	5. Click on `Link to Page` and see if the dropdown appears or not

### DIFF
--- a/components/common/Snackbar.tsx
+++ b/components/common/Snackbar.tsx
@@ -2,25 +2,36 @@ import MuiAlert, { AlertColor, AlertProps } from '@mui/material/Alert';
 import Snackbar, { SnackbarProps } from '@mui/material/Snackbar';
 import Stack from '@mui/material/Stack';
 import { useSnackbar } from 'hooks/useSnackbar';
-import * as React from 'react';
+import { useRouter } from 'next/router';
+import { forwardRef, useEffect } from 'react';
 
-const Alert = React.forwardRef<HTMLDivElement, AlertProps>((props, ref) => {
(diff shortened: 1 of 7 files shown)


components/common/Snackbar.tsx
```diff
@@ -2,25 +2,36 @@ import MuiAlert, { AlertColor, AlertProps } from '@mui/material/Alert';
...
 
 export default function CustomizedSnackbar (props: CustomizedSnackbarProps) {
-  const { severity, message, handleClose, isOpen, showMessage } = useSnackbar();
+  const { setIsOpen, severity, message, handleClose, isOpen } = useSnackbar();
 
-  const { autoHideDuration = 5000 } = props;
+  // Close the snackbar if we change url
+  useEffect(() => {
+    setIsOpen(false);
+  }, [window.location.href]);
```
mattcasey, line 26: this seems like an optimization we may not need, and expensive since it gets triggered all the time, even if the snackbar isn't being shown 99% of the time. Is it bad to let it hang around as you navigate?

mattcasey (approved):
approved but i dont think it should watch the window URL

PR #282, opened 2022-03-25
Show ens name in Bounty Reviewer

Devorein:
1. If the contributor has an ens name use that inside the Contributor text field
2. If the contributor has a username use that otherwise use their wallet address
When the contributor has discord connected (username) and ens name
image

When the contributor has only discord connected (username
image

### DIFF
--- a/hooks/useENSName.tsx
+++ b/hooks/useENSName.tsx
@@ -2,7 +2,7 @@ import type { Web3Provider } from '@ethersproject/providers';
 import { useWeb3React } from '@web3-react/core';
 import useSWRImmutable from 'swr/immutable';
 
-export const fetchENSName = async (_: any, library: Web3Provider, address: string) => library.lookupAddress(address);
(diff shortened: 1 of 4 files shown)


hooks/useENSName.tsx
```diff
@@ -2,11 +2,10 @@ import type { Web3Provider } from '@ethersproject/providers';
 import { useWeb3React } from '@web3-react/core';
 import useSWRImmutable from 'swr/immutable';
 
-const fetchENSName = (_: any, library: Web3Provider, address: string) => library.lookupAddress(address);
+export const fetchENSName = async (_: any, library: Web3Provider, address: string) => library.lookupAddress(address);
```
mattcasey, line 5: remove async and no need to export

Devorein: Thank you for catching that. I forgot to remove those after testing.

PR #286, opened 2022-03-28
Discord roles import

Devorein:


components/common/DiscordServersModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+  {
+    isImportRolesFromServerLoading, isListDiscordServersLoading,
+    discordServers, onImportingDiscordRoles, onClose, isOpen
+  }:
+  {
+    isImportRolesFromServerLoading: boolean, discordServers: DiscordUserServer[], isListDiscordServersLoading: boolean,
+    isOpen: boolean, onImportingDiscordRoles: (guildId: string) => Promise<void>, onClose: () => void
+  }
+) {
+  const sortedServers = useMemo(() => discordServers
```
mattcasey, line 17: Avoid over-optimizing with useMemo unless you're certain it helps. An array sort is extremely fast since it's a native JS operation, and my guess is this component only draws once.

mattcasey (approved):

components/common/DiscordServersModal.tsx
Devorein: Makes sense

PR #287, opened 2022-03-28
Discord roles import matt

mattcasey:
- removed special error handling for db ops. These types of errors should not happen unless we lost a db connection
- tried to keep code together, and moved the 'import roles' button to the Roles settings page, i think it feels more natural to be able to see the roles you just imported
- imported updates from main branch so that the 'id' on a spaceRole is now NOT NULL (fixing some errors i got while trying to import roles to existing accounts)
- Re-organized components for settings pages so that we have them grouped by page


.vscode/settings.json
```diff
@@ -1,12 +1,12 @@
 {
   "[typescriptreact]": {
-    "editor.defaultFormatter": "vscode.typescript-language-features",
+    "editor.defaultFormatter": "dbaeumer.vscode-eslint",
```
mattcasey, line 3: this typescript formatter was trying to remove spaces between function name and parens, so i reverted it to vscode-eslint

charmClient.ts
```diff
@@ -162,10 +162,10 @@ class CharmClient {
   }
 
   listDiscordServers (payload: {code: string, spaceId: string}) {
-    return http.POST<{servers: DiscordUserServer[]}>('/api/discord/listServers', payload);
+    return http.GET<{servers: DiscordUserServer[]}>('/api/discord/listServers', payload);
```
mattcasey, line 165: no need for POST here

.vscode/settings.json
Devorein: My bad I changed that without realizing that it would affect our current `settings.json`

charmClient.ts
Devorein: This endpoint actually had some db manipulation code in it before changing it so used POST, but forgot to revert it back to GET after I removed them

PR #291, opened 2022-03-29
Roles UI

mattcasey:
With this commit, users will be able to add/remove roles, and assign users to them (I think we're missing a "rename" action?). 

I left the page hidden until I can fix our Discord permissions, currently it fails when you try to import roles


components/settings/roles/components/ImportDiscord/DiscordServersModal.tsx
```diff
@@ -0,0 +1,56 @@
+
```
mattcasey, line 1: this file and the others inside ImportDiscord were just moved to 'components' subfolder

Devorein:
This bug is not related to commits made in this PR but it's located in one of the files from this PR. We are not resetting the `name`  state after creating a role. As such the previously created role was being shown. Its located inside of `RoleSettings.tsx` on line `64` right after closing the popupState.

We can add a very simple `resetField('name')` to reset the state to its empty initial value

PR #298, opened 2022-03-31
Page permissions ui cleanup

mattcasey:
UI cleanup for the page permissions form fields.

Left some feedback in my comments below, but please consider:
- try not to abstract too early - generalizing logic and components takes time to both write and read. it's not always a net gain
- careful when copying code, remove stuff that we don't need
- write clean code and follow the existing code style and patterns we have. For example, camel case filenames, pay attention to whitespace when indenting code


.eslintrc.json
```diff
@@ -8,7 +8,8 @@
   "extends": [
     "plugin:react/recommended",
     "airbnb",
-    "plugin:@next/next/recommended"
+    "plugin:@next/next/recommended",
+    "plugin:@typescript-eslint/recommended"
```
mattcasey, line 12: i tried adding this but it still doesn't remove the whitespace in array types. Please just remember to remove it manually for now (there's also a way to write custom rules if you want to experiment: https://typescript-eslint.io/docs/development/custom-rules/

mattcasey: leaving it in because it adds some extra things we can use

charmClient.ts
```diff
@@ -475,7 +475,7 @@ class CharmClient {
    * Get full set of permissions for a specific user on a certain page
    */
   computeUserPagePermissions (request: IPagePermissionUserRequest): Promise<IPagePermissionFlags> {
-    return http.POST('/api/permissions/query', request);
+    return http.GET('/api/permissions/query', request);
```
mattcasey, line 478: use GET if you're just retrieving data

components/common/PageLayout/components/ShareButton/components/PagePermissions/AddPagePermissionsForm.tsx
```diff
@@ -86,8 +78,8 @@ export function AddPagePermissionsForm ({ pageId, existingPermissions = [], perm
     ]).then(() => permissionsAdded());
   }
 
-  console.log('To hide', userIdsToHide);
-
+  const { custom, ...permissionsWithoutCustom } = permissionLevels as Record<string, string>;
```
mattcasey, line 81: a little trick with spread operator, but was able to replace `filterObjectKeys`

components/common/PageLayout/components/ShareButton/components/PagePermissions/PagePermissions.tsx
```diff
@@ -1,31 +1,54 @@
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import Grid from '@mui/material/Grid';
-import Input from '@mui/material/Input';
+import Input from '@mui/material/OutlinedInput';
```
mattcasey, line 4: this is the 'outlined' variant of Input. we should always use outlined inputs

components/common/PageLayout/components/ShareButton/components/PagePermissions/PagePermissions.tsx
```diff
@@ -1,31 +1,54 @@
...
-import { InputEnumToOptions } from 'components/common/form/InputEnumToOptions';
-import { filterObjectKeys } from 'lib/utilities/objects';
+import InputEnumToOptions from 'components/common/form/InputEnumToOptions';
 import { PagePermission } from '@prisma/client';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
 import { ElementDeleteIcon } from 'components/common/form/ElementDeleteIcon';
 import { bindPopover, bindTrigger, usePopupState } from 'material-ui-popup-state/hooks';
 import { usePages } from 'hooks/usePages';
-import { AddPagePermissionsForm } from './AddPagePermissionsForm';
+import AddPagePermissionsForm from './AddPagePermissionsForm';
```
mattcasey, line 20: React component files should export the component as default

components/common/PageLayout/components/ShareButton/components/PagePermissions/PagePermissions.tsx
```diff
@@ -54,21 +77,18 @@ function sortPagePermissions (pagePermissions: IPagePermissionWithAssignee[]):
   });
 }
 
-interface IProps {
+interface Props {
```
mattcasey, line 80: you don't need to add "I" to an interface name. This was something I saw when TS first came out but everyone has moved away from it

components/common/PageLayout/components/ShareButton/components/PagePermissions/PagePermissions.tsx
```diff
@@ -101,18 +121,81 @@ export function PagePermissions ({ pageId }: IProps) {
...
-  const popupState = usePopupState({ variant: 'popover', popupId: 'add-a-permission' });
+    */
+
+  const { custom, ...permissionsWithoutCustom } = permissionLevels as Record<string, string>;
 
   return (
-    <Box>
+    <Box padding={1}>
+
+      <Typography gutterBottom>Page Permissions</Typography>
```
mattcasey, line 138: text should always be wrapped in 'Typography' component, or some other MUI component

components/common/form/InputEnumToOptions.tsx
```diff
@@ -1,41 +1,39 @@
...
+  useEffect(() => {
+    if (defaultValue && !value) {
+      setValue(defaultValue);
+    }
+  }, [defaultValue]);
+
   return (
     <FormControl fullWidth>
       {
-        title && (<InputLabel id='demo-simple-select-label'>{title}</InputLabel>)
```
mattcasey, line 30: careful when copying examples, it's not valid HTML to have "id" multiple times on a page

components/common/form/InputSearchRole.tsx
```diff
@@ -1,12 +1,6 @@
-import { Autocomplete, Box, TextField, Typography } from '@mui/material';
```
mattcasey, line 1: i feel like you could have just used Autocomplete directly. This is copying a really awkward abstraction I made for the Contributors input just to get the types to work, but we're only using the multi-select version of this input, and it's querying the roles a second time.

lib/permissions/pages/page-permission-mapping.ts
```diff
@@ -1,7 +1,7 @@
 import { PageOperations, PagePermissionLevel } from '@prisma/client';
 import { PageOperationType } from './page-permission-interfaces';
 
-export const PagePermissionLevelTitle: Record<keyof typeof PagePermissionLevel, string> = {
+export const permissionLevels: Record<keyof typeof PagePermissionLevel, string> = {
```
mattcasey, line 4: updated to match the format of other constants in this file

seedData.ts
```diff
@@ -1,10 +1,6 @@
-import { Page, Prisma, Space, SpaceRole, User } from '@prisma/client';
```
mattcasey, line 1: just getting rid of some of this code we don't use to avoid Typescript conflicts

components/common/form/InputSearchContributor.tsx
```diff
@@ -29,27 +29,29 @@ function filterContributors (contributors: Contributor [], filter: IContributors
...
-    <Autocomplete<Contributor>
-      defaultValue={defaultContributor}
-      loading={contributors.length === 0}
+  return (
+    <Autocomplete
+      {...props}
+      disabled={filteredOptions.length === 0}
+      loading={options.length === 0}
       sx={{ minWidth: 150 }}
-      disableCloseOnSelect={disableCloseOnSelect}
```
mattcasey, line 50: this felt kinda awkard that i always have to close the popup, so removed it.

components/common/form/InputEnumToOptions.tsx
```diff
@@ -1,41 +1,41 @@
...
 
-export function InputEnumToOptions ({ onChange = () => {}, defaultValue, title, keyAndLabel }: Props) {
+export default function InputEnumToOptions ({ onChange = () => {}, defaultValue, title, keyAndLabel }: Props) {
 
   const options = Object.entries(keyAndLabel);
 
+  const [value, setValue] = useState<string | null>('');
+
+  useEffect(() => {
+    if (defaultValue && !value) {
```
mattcasey, line 21: This avoids passing a defaultValue directly to the component, since 'defaultValue' is supposed to be set when the component mounts and never changes.

.eslintrc.json
motechFR: Will do, added a note to myself to explore the writing of custom ESLint rules

charmClient.ts
motechFR: Noted

components/common/PageLayout/components/ShareButton/components/PagePermissions/AddPagePermissionsForm.tsx
motechFR: Reading this, filterObjectKeys still seems an objectively better solution for following reasons
- In-built type safety, whereas here we revert to Record<string, string>
- Repeatable behaviour, we can be sure of what include / exclude will do
- Better readability. I had to catch what was going here with the destructuring

In terms of the const exports, it would be better to only ever have a single source of truth for all the available permission levels.

That said, I can see a case where "custom" is in fact a special case to exclude, and we have
- One const with custom
- One const without custom

components/common/PageLayout/components/ShareButton/components/PagePermissions/PagePermissions.tsx
motechFR: I've been wondering about this, will apply this pattern going forward

components/common/form/InputEnumToOptions.tsx
motechFR: Noted

components/common/form/InputSearchRole.tsx
motechFR: I've made a note to request further clarification about this on our next chat

components/common/form/InputEnumToOptions.tsx
motechFR: Will add this to our developer patterns

PR #299, opened 2022-03-31
Discord login matt

mattcasey:


pages/api/discord/connect.ts
```diff
@@ -39,10 +39,10 @@ export interface ConnectDiscordResponse {
 
 // TODO: Add nonce for oauth state
 async function connectDiscord (req: NextApiRequest, res: NextApiResponse<ConnectDiscordResponse | {error: string}>) {
-  const { code, spaceId } = req.body as ConnectDiscordPayload;
```
mattcasey, line 42: i realized we should connect your discord roles for all the spaces you have access to, not just the first one you log in to. So spaceId is no longer an input, and we loop through them below.

components/common/PageLayout/components/Account/Account.tsx
```diff
@@ -10,16 +10,12 @@ import Avatar from 'components/common/Avatar';
 import { useUser } from 'hooks/useUser';
 import { Web3Connection } from 'components/_app/Web3ConnectionManager';
 import { Chains, RPC } from 'connectors';
-import { useContext, useEffect, useState } from 'react';
```
mattcasey, line 13: moved all the discord logic out of this file and into the AccountModal. the modal uses query params to know if it should be open, and when you close it we remove the query params

components/common/PageLayout/components/Account/components/AccountModal/AccountModal.tsx
```diff
@@ -132,24 +169,34 @@ function AccountModal ({ isConnectDiscordLoading, isOpen, onClose }:
           </div>
         </Tooltip>
       </Stack>
-      <StyledButton
```
mattcasey, line 135: i hid "Logout" if your wallet is connected, because we just log you back in automatically

hooks/useDiscordLogin.ts
```diff
@@ -1,30 +1,29 @@
...
 
-export function useDiscordLogin (cleanRoutes?: boolean) {
+export function useDiscordLogin () {
   const router = useRouter();
-  const isLogInWithDiscord = typeof router.query.code === 'string' && router.query.discord === '1' && router.query.type === 'login';
   const { showMessage } = useSnackbar();
-  const [user, setUser] = useUser();
+  const [_, setUser] = useUser();
 
-  useSWRImmutable(isLogInWithDiscord ? [router.query.code, router.query.discord, router.query.type] : null, async () => {
```
mattcasey, line 13: I think SWR should only be used if we do something with the result. otherwise, useEffect() serves the same purpose of running code once. And the router.query params always exist on first render, based on my experience

components/common/PageLayout/components/Account/components/AccountModal/AccountModal.tsx
```diff
@@ -132,24 +169,31 @@ function AccountModal ({ isConnectDiscordLoading, isOpen, onClose }:
...
+        </Alert>
+      )}
+      {/* user cant be logged out so long as tehir wallet is connected (TODO: fix!) */}
+      {!account && (
+        <Box display='flex' justifyContent='flex-end' mt={2}>
+          <StyledButton
+            size='small'
+            variant='outlined'
+            color='secondary'
+            loading={isLoginOut}
```
mattcasey, line 184: note that our Button component has a 'loading' prop which disables and shows a loading icon

components/login/LoginPageContent.tsx
```diff
@@ -60,13 +61,14 @@ export default function LoginPageContent ({ account }: { account: string | null
...
               Tasks, docs, bounties, and more
             </Typography>
-            <Box display='flex' gap={1}>
+            <Box display='flex' gap={2} alignItems='center'>
               <PrimaryButton size='large' loading={!triedEager} onClick={openWalletSelectorModal}>
                 Connect Wallet
               </PrimaryButton>
-              <PrimaryButton size='large' loading={!triedEager} href={`/api/discord/oauth?redirect=${encodeURIComponent(window.location.href.split('?')[0])}&type=login`}>
+              <Typography color='secondary' variant='body2'>or</Typography>
+              <Button variant='outlined' size='large' loading={!triedEager} href={`/api/discord/oauth?redirect=${encodeURIComponent(window.location.href.split('?')[0])}&type=login`}>
```
mattcasey, line 69: i made 'connect discord' slightly less emphasized, so the two buttons are not visually competing

components/settings/payment-methods/components/PaymentMethodForm.tsx
```diff
@@ -343,7 +342,7 @@ export default function PaymentForm ({ onSubmit, defaultChainId = 1 }: Props) {
             )
           }
           <Grid item>
-            <Button type='submit' disabled={!isValid || getValues('contractAddress') === ''}>Create payment method</Button>
+            <Button type='submit' disabled={!isValid || (values.contractAddress === '' && values.gnosisSafeAddress === '')}>Create payment method</Button>
```
mattcasey, line 345: careful adding validation rules. if you look at how the form components are used, contractAddress only is used for metamask wallet types

lib/discord/handleDiscordResponse.ts
```diff
@@ -34,7 +34,7 @@ export async function handleDiscordResponse<T> (endpoint: string): Promise<{stat
     else if (err.code === 50001) {
       return {
         status: 400,
-        error: 'Charmverse does not have access to the server'
+        error: 'CharmVerse does not have access to the server'
```
mattcasey, line 37: the "V" in CharmVerse should be capitalized

pages/api/profile/index.ts
```diff
@@ -55,7 +55,7 @@ async function createProfile (req: NextApiRequest, res: NextApiResponse<LoggedIn
   }
 }
 
-async function getProfile (req: NextApiRequest, res: NextApiResponse<LoggedInUser | { error: any }>) {
+async function getUser (req: NextApiRequest, res: NextApiResponse<LoggedInUser | { error: any }>) {
```
mattcasey, line 58: i made this more of an update route

components/login/LoginPageContent.tsx
```diff
@@ -60,13 +61,14 @@ export default function LoginPageContent ({ account }: { account: string | null
...
               Tasks, docs, bounties, and more
             </Typography>
-            <Box display='flex' gap={1}>
+            <Box display='flex' gap={2} alignItems='center'>
               <PrimaryButton size='large' loading={!triedEager} onClick={openWalletSelectorModal}>
                 Connect Wallet
               </PrimaryButton>
-              <PrimaryButton size='large' loading={!triedEager} href={`/api/discord/oauth?redirect=${encodeURIComponent(window.location.href.split('?')[0])}&type=login`}>
+              <Typography color='secondary' variant='body2'>or</Typography>
+              <Button variant='outlined' size='large' href='/api/discord/oauth?type=login'>
```
mattcasey, line 69: i removed the redirect. the backend will just send them to the root page "/" and we'll re-use the logic in RouteGuard to send them to the right place

hooks/useUser.tsx
```diff
@@ -15,50 +15,17 @@ export function UserProvider ({ children }: { children: ReactNode }) {
...
           setUser(_user);
-          setIsLoaded(true);
         })
-        .catch(err => {
-          // probably needs to log in
+        .finally(() => {
           setIsLoaded(true);
         });
     }
-    else if (account && user) {
```
mattcasey, line 30: To keep the app simple to understand, logging users in should be handled in a single place, RouteGuard

hooks/useUser.tsx
```diff
@@ -15,50 +15,17 @@ export function UserProvider ({ children }: { children: ReactNode }) {
...
-            setIsLoaded(true);
-          });
-        });
-      }
-    }
-    // user disconnects their wallet
-    else if (!account && !user) {
-      setIsLoaded(false);
-      // Log the user from session
-      charmClient.getUser().then(_user => {
```
mattcasey, line 56: i wasn't sure what this is for.. but now we always check for the user on page load, whether or not a wallet address exists

pages/api/discord/createAccount.ts
```diff
@@ -16,7 +16,7 @@ const handler = nc({
 
 export async function createAccountWithDiscord (req: NextApiRequest, res: NextApiResponse<LoggedInUser | { error: any }>) {
   const { code } = req.body as {code: string};
-  const discordAccount = await getDiscordAccount(code as string, req.headers.host!.startsWith('localhost') ? `http://${req.headers.host}/api/discord/callback` : 'https://app.charmverse.io/api/discord/callback');
+  const discordAccount = await getDiscordAccount(code as string, req.headers.host?.startsWith('localhost') ? `http://${req.headers.host}/api/discord/callback` : 'https://app.charmverse.io/api/discord/callback');
```
mattcasey, line 19: don't use "!" assertions unless absolutely sure

components/common/RouteGuard.tsx
```diff
@@ -122,7 +121,7 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
       return { authorized: false, user: _user };
     }
     // condition: trying to access a space without access
-    else if (!isViewingBountyDetails && !isImportingRolesFromDiscord && isSpaceDomain(spaceDomain) && !spaces.some(s => s.domain === spaceDomain)) {
+    else if (isSpaceDomain(spaceDomain) && !spaces.some(s => s.domain === spaceDomain)) {
```
mattcasey, line 124: i don't see why these exceptions were added. if you want to access a space, but don't have access, you go to the join page, no matter what.

components/common/RouteGuard.tsx
```diff
@@ -95,11 +94,6 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
     const spaceDomain = path.split('/')[1];
     // condition: public page
     if (publicPages.some(basePath => firstPathSegment === basePath)) {
-      // Connected wallet in / route
-      if (account && !user) {
```
mattcasey, line 99: if it's a public page, we don't need to log the user in or create an account

pages/api/discord/connect.ts
Devorein: Ok I assume you talked with Alex about this as it might not be intuitive to most users

components/common/PageLayout/components/Account/Account.tsx
Devorein: Makes sense

hooks/useDiscordLogin.ts
Devorein: Yeah I initially used it since we were calling the `/login` multiple times using `useEffect`, but it doesn't matter now anyway since everything has been moved to the backend

components/common/PageLayout/components/Account/components/AccountModal/AccountModal.tsx
Devorein: Thats wonderful. I never knew that. That will shave off a few extra lines of code at least

components/common/PageLayout/components/Account/components/AccountModal/AccountModal.tsx
Devorein: Yeah, that makes sense.

components/login/LoginPageContent.tsx
Devorein: I like this a lot lot better. Along with the small or. Much better UI

components/settings/payment-methods/components/PaymentMethodForm.tsx
Devorein: Oh sorry, I didn't know that. I was trying to create a payment method without being connected to my wallet and faced a parameter missing error from backend so thought I might as well add the validation logic on the client side as well

pages/api/discord/createAccount.ts
Devorein: That header should always be present, but its better to be safe than blowing up production

components/common/RouteGuard.tsx
Devorein: I wasn't being logged in to my account even after connecting my wallet without it so I thought to add the logic underneath inside this place, but its handled properly now

components/common/RouteGuard.tsx
Devorein: This became redundant after implementing getting the current user from the session on initial load. Previously it was throwing blowing up the bounties and importing from discord page

Devorein (approved):

PR #303, opened 2022-03-31
Connect discord for invite page

Devorein:
**Session Exist**
1. We get the user from the session. So no matter which wallet address you switch to it won't have any effect on the invite flow. We use the user from the session
2. After a successful discord connect, we log you as that user so clicking on accept Invite would add that user to the workspace

**Session Doesn't Exist**
1. Changing the wallet address will have an impact as before accepting the invite we create the user with that account (it won't create one if a user with that wallet address already exists and log you as that user)
2. If you connect to discord we fallback to **Session Exist**

### DIFF
--- a/components/invite/InviteLinkPage.tsx
+++ b/components/invite/InviteLinkPage.tsx
@@ -1,5 +1,4 @@
 import { useContext } from 'react';
-import styled from '@emotion/styled';
 import Box from '@mui/material/Box';
 import Card from '@mui/material/Card';
 import Typography from '@mui/material/Typography';
@@ -10,17 +9,9 @@ import { InviteLinkPopulated } from 'lib/invites';
 import PrimaryButton from 'components/common/PrimaryButton';
 import { useUser } from 'hooks/useUser';
 import charmClient from 'charmClient';
+import { CenteredBox } from './components/CenteredBox';
 
-const CenteredBox = styled.div`
-  position: absolute;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
-  width: 400px;
-  max-width: 100%;
-`;
-
-export default function InvitationPage ({ error, invite }: { error?: string, invite?: InviteLinkPopulated }) {
+export default function InvitationPage ({ invite }: { invite: InviteLinkPopulated }) {
 
   const [user] = useUser();
   const { openWalletSelectorModal, triedEager } = useContext(Web3Connection);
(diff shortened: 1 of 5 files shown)


components/invite/InviteLinkPage.tsx
```diff
@@ -27,13 +28,21 @@ export default function InvitationPage ({ error, invite }: { error?: string, inv
...
+    }
+    if (invite) {
+      await charmClient.acceptInvite({ id: invite.id });
+      window.location.href = `/${invite.space.domain}`;
     }
-    await charmClient.acceptInvite({ id: invite!.id });
-    window.location.href = `/${invite!.space.domain}`;
   }
 
+  const spaceRole = user?.spaceRoles.find(_spaceRole => _spaceRole.spaceId === invite?.spaceId);
```
mattcasey, line 40: why is this needed?

components/invite/InviteLinkPage.tsx
```diff
@@ -68,12 +77,12 @@ export default function InvitationPage ({ error, invite }: { error?: string, inv
             <PrimaryButton size='large' loading={!triedEager} onClick={openWalletSelectorModal}>
               Connect Wallet
             </PrimaryButton>
-            {/* <PrimaryButton size='large' loading={!triedEager} href={`/api/discord/oauth?redirect=${encodeURIComponent(window.location.href.split('?')[0])}&type=login`}>
+            <PrimaryButton size='large' loading={!triedEager} href={`/api/discord/oauth?redirect=${encodeURIComponent(window.location.href.split('?')[0])}&type=login`}>
```
mattcasey, line 80: remove the loading prop here

components/invite/InviteLinkPage.tsx
```diff
@@ -49,17 +58,17 @@ export default function InvitationPage ({ error, invite }: { error?: string, inv
       </CenteredBox>
     );
   }
-  return (
+  return invite ? (
```
mattcasey, line 61: it's either error or invite that will exist. i just couldn't express it in the types, but i prefer not to add this logic check

components/invite/InviteLinkPage.tsx
```diff
@@ -10,6 +10,7 @@ import { InviteLinkPopulated } from 'lib/invites';
 import PrimaryButton from 'components/common/PrimaryButton';
 import { useUser } from 'hooks/useUser';
 import charmClient from 'charmClient';
+import { useSpaces } from 'hooks/useSpaces';
```
mattcasey, line 13: this is unused

components/invite/InviteLinkPage.tsx
Devorein: If they have quite a lot of spaces I assume, but it's definitely a very rare case. Would cut off one extra API request to our backend and arguably a better UX IMO

components/invite/InviteLinkPage.tsx
Devorein: Yep moved it with error handling to make it more readable

mattcasey (approved):

PR #305, opened 2022-03-31
Return only pages a user has access to

motechFR:
Only show the pages a user has access to

### DIFF
--- a/pages/api/spaces/[id]/pages.ts
+++ b/pages/api/spaces/[id]/pages.ts
@@ -69,10 +69,6 @@ async function getPages (req: NextApiRequest, res: NextApiResponse<Page[]>) {
     }
   });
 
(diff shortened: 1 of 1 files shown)


motechFR:
@mattcasey this automagically works!

- Pages filtered out at the Prisma query level, so no additional load on server
- The existing client side code works perfect, and repositions a nested page at the root of the directory if no parent is available

pages/api/spaces/[id]/pages.ts
```diff
@@ -11,14 +11,68 @@ const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
...
+        }
+      ]
+
     },
     include: {
       permissions: true
     }
   });
+
+  if (spaceId === 'eed82490-f0bd-4cf8-8651-775b666947e5') {
```
Comment (mattcasey) [inline pages/api/spaces/[id]/pages.ts:72]: dont forget to remove this

mattcasey (approved):

PR #306, opened 2022-03-31
Fix permissions in notion import

mattcasey:


lib/notion/importFromWorkspace.ts
```diff
@@ -365,37 +377,45 @@ async function createPrismaPage ({
...
     },
     headerImage,
     icon,
-    title,
+    title: title || '',
     type,
     boardId,
-    parentId
+    parentId,
+    permissions: {
```
mattcasey, line 410: @motechFR here's the fix, adding space-level permissions when we create a page

motechFR: Looks good to me @mattcasey . Glad to see our new permission syntax in use by dev team

lib/notion/importFromWorkspace.ts
```diff
@@ -5,17 +5,18 @@ import { Prisma } from '@prisma/client';
...
+import log from 'lib/log';
 import { BlockObjectResponse, GetDatabaseResponse, GetPageResponse, RichTextItemResponse } from './types';
 
 // Limit the highest number of pages that can be imported
-const IMPORTED_PAGES_LIMIT = 1000; const
-  BLOCKS_FETCHED_PER_REQUEST = 100; const
-  MAX_CHILD_BLOCK_DEPTH = 5;
+const IMPORTED_PAGES_LIMIT = 1000;
+const BLOCKS_FETCHED_PER_REQUEST = 100;
+const MAX_CHILD_BLOCK_DEPTH = 10;
```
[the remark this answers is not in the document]
Devorein: No problem. I don't think it will be an issue. But it's for nested content. For example, if we have 6 Callout blocks nested inside of one another, only 5 of them would be created, this has nothing to do with child pages as most of them are located at the top level or at most 1-2 level in depth

PR #314, opened 2022-04-03
Show views of database inside PageNavigation

Devorein:


components/common/PageLayout/components/PageNavigation.tsx
```diff
@@ -501,7 +507,7 @@ function RenderDraggableNode ({ item, onDropAdjacent, onDropChild, pathPrefix, a
           <Typography variant='caption' className='MuiTreeItem-content' sx={{ display: 'flex', alignItems: 'center', color: `${greyColor2} !important`, ml: 3 }}>
             No pages inside
           </Typography>
-        )}
+        ) : views.map(view => <PageTreeItem labelIcon={iconForViewType(view.fields.viewType)} label={view.title} href={`${pathPrefix}/${item.path}${item.type === 'board' ? `?viewId=${view.id}` : ''}`} id={view.id} />)}
```
mattcasey, line 510: needs a 'key' prpoerty on each, and can u break this up into shorter lines?

Devorein: Oh my I completely forgot about that 👀

Devorein: Done

pages/_app.tsx
```diff
@@ -305,6 +306,13 @@ function FocalBoardProviders ({ children }: { children: ReactNode }) {
 }
 
 function DataProviders ({ children }: { children: ReactNode }) {
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    const loadAction: any = initialLoad; /* eslint-disable-line @typescript-eslint/no-explicit-any */
```
mattcasey, line 312: I think this method would need to be called every time the user changes workspaces

Devorein: Thanx for the changes in your PR. This would dramatically reduce the number of times we need to fetch the blocks. There are probably on avg 5-10x more blocks than pages, so even one less API request counts.

PR #319, opened 2022-04-04
Show breadcrumb of pages on header

Devorein:
1. Each link is separated by a / (backslash)
2. If there are more than 3 links, just show the first, the one before the last and the last one (current page), the rest are shown via an ... (ellipsis)
3. Otherwise show all of them (for 1, 2, and 3 levels of nesting)
4. Each of the previous links can be navigated by clicking on them
5. We show the icon and title of each link and it kept up to date with the state

### DIFF
--- a/components/common/PageLayout/components/Header.tsx
+++ b/components/common/PageLayout/components/Header.tsx
@@ -158,7 +158,7 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
       }
     }
     return currentPageParentChain.reverse();
-  }, [pages]);
+  }, [pages, currentPage]);
 
   const isFavorite = currentPage && user?.favorites.some(({ pageId }) => pageId === currentPage.id);
 
--- a/components/common/PageLayout/components/Header.tsx
+++ b/components/common/PageLayout/components/Header.tsx
@@ -119,7 +119,7 @@ function PageBreadcrumbs ({ pages }: {pages: Page[]}) {
     id: '123'
   }, pages[pages.length - 2], pages[pages.length - 1]] : pages).map(page => ({
     title: page.title,
-    link: `/${router.query.domain}/${page.path}`,
+    link: page.path ? `/${router.query.domain}/${page.path}` : null,
     icon: page.icon,
     id: page.id
   }));
@@ -130,7 +130,7 @@ function PageBreadcrumbs ({ pages }: {pages: Page[]}) {
 export default function Header ({ open, openSidebar }: { open: boolean, openSidebar: () => void }) {
   const router = useRouter();
   const colorMode = useColorMode();
-  const { pages, linkedPages, currentPageId, isEditing } = usePages();
+  const { pages, currentPageId, isEditing } = usePages();
   const [user, setUser] = useUser();
   const theme = useTheme();
   const [pageMenuOpen, setPageMenuOpen] = useState(false);
@@ -141,13 +141,18 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
 
   const currentPageChain = useMemo(() => {
     const currentPageParentChain: Page[] = [];
-    let currentPageInChain: LinkedPage | null = linkedPages[currentPageId];
+    let currentPageInChain: Page | null | undefined = pages[currentPageId];
     while (currentPageInChain) {
       currentPageParentChain.push(currentPageInChain);
-      currentPageInChain = currentPageInChain.parent;
+      if (currentPageInChain.parentId) {
+        currentPageInChain = pages[currentPageInChain.parentId];
+      }
+      else {
+        currentPageInChain = null;
+      }
     }
     return currentPageParentChain.reverse();
-  }, [currentPage]);
+  }, [pages]);
 
   const isFavorite = currentPage && user?.favorites.some(({ pageId }) => pageId === currentPage.id);
 
--- a/components/common/PageLayout/components/Header.tsx
+++ b/components/common/PageLayout/components/Header.tsx
@@ -26,11 +26,10 @@ import charmClient from 'charmClient';
 import { charmEditorPlugins, specRegistry } from 'components/common/CharmEditor/CharmEditor';
 import { useColorMode } from 'context/color-mode';
 import { LinkedPage, usePages } from 'hooks/usePages';
-import { usePageTitle } from 'hooks/usePageTitle';
 import { useUser } from 'hooks/useUser';
 import { PageContent } from 'models';
 import { useRouter } from 'next/router';
-import { useMemo, useRef, useState } from 'react';
+import { Fragment, useMemo, useRef, useState } from 'react';
 import Account from './Account';
 import ShareButton from './ShareButton';
 import { StyledPageIcon } from './PageNavigation';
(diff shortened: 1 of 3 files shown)


components/common/PageLayout/components/Header.tsx
```diff
@@ -69,18 +69,68 @@ function PageBreadcrumbs () {
   return null;
 }
 
+function PageBreadcrumbs ({ pages }: {pages: Page[]}) {
+  const router = useRouter();
+  const trimTrails = pages.length > 3;
+  const trailsInfo = (trimTrails ? [pages[0], {
```
mattcasey, line 75: cleaner way to write this:
```
const splitCrumb = {
  title: '...',
  path: null,
  icon: null,
  id: '123'
}
const displayedPages = trimTrails ? [pages[0], splitCrumb, pages[pages.length - 2], pages[pages.length - 1]] : pages;
return ( ... );
```

components/common/PageLayout/components/Header.tsx
```diff
@@ -69,18 +69,68 @@ function PageBreadcrumbs () {
...
+    title: '...',
+    path: null,
+    icon: null,
+    id: '123'
+  }, pages[pages.length - 2], pages[pages.length - 1]] : pages).map(page => ({
+    title: page.title,
```
mattcasey, line 81: this map looks unecessary ( all the properties are the same?)

components/common/PageLayout/components/Header.tsx
```diff
@@ -69,18 +69,68 @@ function PageBreadcrumbs () {
...
+        const trailTitle = trailInfo.title ?? 'Untitled';
+        const pageDeco = (
+          <Box display='flex'>
+            {trailInfo.icon && <StyledPageIcon icon={trailInfo.icon} />}
+            {trailTitle}
+          </Box>
+        );
+
+        return trailInfoIndex !== trailsInfo.length - 1 ? (
+          <Fragment key={trailInfo.id}>
```
mattcasey, line 99: this should be the Breadcrumb component

components/common/PageLayout/components/Header.tsx
```diff
@@ -69,18 +69,68 @@ function PageBreadcrumbs () {
...
+    path: page.path,
+    icon: page.icon,
+    id: page.id
+  }));
+
+  return (
+    <Box gap={1} display='flex'>
+      {trailsInfo.map((trailInfo, trailInfoIndex) => {
+        const trailTitle = trailInfo.title ?? 'Untitled';
+        const pageDeco = (
```
mattcasey, line 91: i wasn't sure what pageDeco meant. A better name for this could be "currentPageCrumb" or something like that?

mattcasey (changes requested):
- use the existing breadcrumb components to maintain style consistency
- simplify linked pages logic and put into Header file
- bug: when I create a new page, the new page doesn't appear in the list (i think this will be solved if you replace the linked pages logic)

components/common/PageLayout/components/Header.tsx
```diff
@@ -69,18 +69,68 @@ function PageBreadcrumbs () {
...
+          <Box display='flex'>
+            {trailInfo.icon && <StyledPageIcon icon={trailInfo.icon} />}
+            {trailTitle}
+          </Box>
+        );
+
+        return trailInfoIndex !== trailsInfo.length - 1 ? (
```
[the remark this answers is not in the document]
Devorein: You are right. I've fixed it in the latest commits

components/common/PageLayout/components/Header.tsx
Devorein: Done

components/common/PageLayout/components/Header.tsx
Devorein: Yes I replaced it in the latest commit

components/common/PageLayout/components/Header.tsx
Devorein: Makes sense. I've changed it in the latest commit

Devorein:
You are right about the algorithm to find the trail of parents of a page, I went overboard with that logic. Fixed it to be much much simpler in the latest commit.

hooks/usePages.tsx
```diff
@@ -53,6 +56,39 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     setPages(data?.reduce((acc, page) => ({ ...acc, [page.id]: page }), {}) || {});
   }, [data]);
 
+  const linkedPages = useMemo(() => {
```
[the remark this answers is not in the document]
Devorein: You are right about the algorithm to find the trail of parents of a page, I went overboard with that logic. Fixed it to be much much simpler in the latest commit.

components/common/PageLayout/components/Header.tsx
```diff
@@ -43,44 +44,121 @@ const StyledToolbar = styled(Toolbar)`
...
+        // Last item doesn't have any /
+        if (itemIndex !== items.length - 1) {
+          return (
+            <Fragment key={item.id}>
+              {item.link ? (
+                <Link href={item.link}>
+                  {currentPageCrumb}
+                </Link>
+              ) : <div>{currentPageCrumb}</div>}
+              <span className='breadcrumb-slash'>/</span>
```
mattcasey, line 84: why did you need to add this? the "/" was being added by CSS already

components/common/PageLayout/components/Header.tsx
```diff
@@ -43,44 +44,121 @@ const StyledToolbar = styled(Toolbar)`
...
+      {items.map((item, itemIndex) => {
+        const itemTitle = item.title || 'Untitled';
+        const currentPageCrumb = (
+          <Box display='flex'>
+            {item.icon && <StyledPageIcon icon={item.icon} />}
+            {itemTitle}
+          </Box>
+        );
+
+        // Last item doesn't have any /
```
mattcasey, line 75: i originally just printed out the last item (the current page). i think that's cleaner than adding to the list here,there's no reason it should link to itself, have a breadcrumb slash, etc. can we keep it how it was?

components/common/PageLayout/components/Header.tsx
```diff
@@ -43,44 +44,121 @@ const StyledToolbar = styled(Toolbar)`
...
-      <BreadCrumb>
-        <Link href={`/${router.query.domain}/bounties`}>
-          Bounties
-        </Link>
-      </BreadCrumb>
-    );
-  }
-  return null;
+  const bountyId = router.query.bountyId as string;
+  const { bounties } = useBounties();
```
mattcasey, line 98: there is'nt a need to get the bounties. we have the current page title already

components/common/PageLayout/components/Header.tsx
```diff
@@ -163,8 +241,7 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
         }}
         >
           <Typography noWrap component='div' sx={{ fontWeight: 500, maxWidth: 500, textOverflow: 'ellipsis' }}>
-            <PageBreadcrumbs />
-            {pageTitle}
+            {router.route === '/[domain]/bounties/[bountyId]' ? <BountyBreadcrumbs /> : router.route === '/[domain]/[pageId]' ? <PageBreadcrumbs pages={currentPageChain} /> : null}
```
mattcasey, line 244: i prefered this inside a component like i originally had

components/common/PageLayout/components/Header.tsx
```diff
@@ -25,14 +25,15 @@ import { Page } from '@prisma/client';
 import charmClient from 'charmClient';
 import { charmEditorPlugins, specRegistry } from 'components/common/CharmEditor/CharmEditor';
 import { useColorMode } from 'context/color-mode';
-import { usePages } from 'hooks/usePages';
-import { usePageTitle } from 'hooks/usePageTitle';
```
mattcasey, line 29: we need to keep using this, it's used by settings and possible other pages

PR #321, opened 2022-04-04
fix React error about passing a ref down to our custom buttons

mattcasey:
This removes an error that appears on every page load.


motechFR:
Looks good!

Adding [relevant documentation](https://blog.logrocket.com/cleaning-up-the-dom-with-forwardref-in-react/#:~:text=React%20forwardRef%20is%20a%20method,anywhere%20it%20is%20being%20used.) about forwardRef for anyone reading this in future

PR #322, opened 2022-04-04
fix error related to board page items in sidebar menu

mattcasey:
- There was an error from MUI because we were not passing a 'nodeId' property to the TreeItem used for board views. I also split it into its own component from PageTreeItem because they don't really need to share the same code.
```
Warning: Failed prop type: The prop `nodeId` is marked as required in `ForwardRef(TreeItem)`, but its value is `undefined`
```
- Also: highlight the view row when you're on the page (see screenshot)
image


motechFR (approved):
Confirmed, this works for me

PR #327, opened 2022-04-06
telegram integration

Devorein:
Telegram integration PR like discord. Note that our prisma schema has been updated so its paramount to deploy the latest schema to prod. Similar to discord a `telegramUser` field is added to `User` and `TelegramUser` table is also created. The UI for connecting/disconnecting with telegram is the same as discord.

### DIFF
--- a/charmClient.ts
+++ b/charmClient.ts
@@ -60,9 +60,7 @@ class CharmClient {
   }
 
   updateUser (payload: { username?: string, avatar?: string, addresses?: string[] }) {
-    return http.PUT<LoggedInUser>('/api/profile', {
-      payload
-    });
+    return http.PUT<LoggedInUser>('/api/profile', payload);
   }
 
   async createSpace (spaceOpts: Prisma.SpaceCreateInput) {
--- a/charmClient.ts
+++ b/charmClient.ts
@@ -161,6 +161,10 @@ class CharmClient {
     return http.POST<TelegramUser>('/api/telegram/connect', telegramAccount);
   }
 
+  disconnectTelegram () {
+    return http.POST('/api/telegram/disconnect');
+  }
+
   disconnectDiscord () {
     return http.POST('/api/discord/disconnect');
   }
--- a/charmClient.ts
+++ b/charmClient.ts
@@ -59,9 +59,9 @@ class CharmClient {
     });
   }
 
-  updateUser ({ addresses }: { addresses?: string[] }) {
+  updateUser (payload: { username?: string, avatar?: string, addresses?: string[] }) {
     return http.PUT<LoggedInUser>('/api/profile', {
-      addresses
+      payload
     });
   }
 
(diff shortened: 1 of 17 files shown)


pages/api/profile/index.ts
```diff
@@ -35,7 +35,8 @@ async function createUser (req: NextApiRequest, res: NextApiResponse<LoggedInUse
           }
         }
       },
-      discordUser: true
+      discordUser: true,
+      telegramUser: true
```
mattcasey, line 39: make sure to remove this field when we write the user object to the session, like we do with discord

pages/api/telegram/connect.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  const { id, ...rest } = telegramAccount;
+  const userId = req.session.user.id;
+  let telegramUser: TelegramUser;
+
+  try {
+    telegramUser = await prisma.telegramUser.create({
+      data: {
+        account: rest as any,
+        telegramId: String(id),
```
mattcasey, line 35: curious why do you have to cast this as a string? i think it's ok to use a number if that's what telegram uses

charmClient.ts
```diff
@@ -59,10 +60,8 @@ class CharmClient {
     });
   }
 
-  updateUser ({ addresses }: { addresses?: string[] }) {
-    return http.PUT<LoggedInUser>('/api/profile', {
-      addresses
-    });
+  updateUser (payload: { username?: string, avatar?: string, addresses?: string[] }) {
```
mattcasey, line 63: just want to confirm that we're not actually updating username/avatar from what i can tell, altho this change doesn't hurt anything :)

mattcasey (changes requested):
one change request: don't write telegramuser data to session
one data model design suggestion: store telegram user id in its native type (number). this could get confusing if we are doing any other api calls, i think

components/common/PageLayout/components/Account/components/AccountModal/AccountModal.tsx
```diff
@@ -19,10 +19,14 @@ import { useUser } from 'hooks/useUser';
 import styled from '@emotion/styled';
 import { CircularProgress, Tooltip } from '@mui/material';
 import charmClient from 'charmClient';
+// @ts-ignore
+import TelegramLoginButton from 'react-telegram-login';
```
mattcasey, line 23: is it missing types? we should add a .d.ts file, i can help if you want

Devorein: Wow I can do that. Never done that before so should be a learning experience. Good thing is that they have everything in proptypes so shouldn't take too long

pages/api/profile/index.ts
Devorein: Oh nice call. Thank you.

pages/api/telegram/connect.ts
Devorein: I was trying to play it safe as we don't have any idea about the range of this field. But I guess they are using a numeric id as well so its safe

Devorein:
Both the changes have been implemented in the latest commits. I also made sure we don't allow the user to update their username and avatar using the API. I will be done via a first successful connection to discord/telegram only

mattcasey (approved):

PR #330, opened 2022-04-06
fix: focalboard loses scroll position when opening a card

mattcasey:
It turns out, when we change the URL parameter in Next.js, it re-mounts all the components which loses the scroll position. I finally found a mechanism to update the URL without Next.js knowing it, and instead of relying on router, we just update the visible card using useState.

### DIFF
--- a/components/[pageId]/BoardPage/BoardPage.tsx
+++ b/components/[pageId]/BoardPage/BoardPage.tsx
@@ -1,18 +1,21 @@
 import { generatePath } from 'lib/utilities/strings';
 import { Page } from 'models';
 import { useRouter } from 'next/router';
-import { useCallback, useEffect } from 'react';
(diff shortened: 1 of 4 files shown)


components/[pageId]/BoardPage/BoardPage.tsx
```diff
@@ -135,14 +140,28 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
...
           cards={cards}
-          shownCardId={router.query.cardId as string}
           showCard={showCard}
           activeView={activeView}
           groupByProperty={property}
           dateDisplayProperty={displayProperty}
           views={viewsToProvide}
           showShared={clientConfig?.enablePublicSharedBoards || false}
         />
+        {typeof shownCardId === 'string' && (
```
Comment (mattcasey) [inline components/[pageId]/BoardPage/BoardPage.tsx:150]: moved this outside so CenterPanel actually doesn't redraw if you open a card anymore

PR #333, opened 2022-04-06
Telegram login: fix for Firefox

mattcasey:
I was not able to connect to Telegram in Firefox. It works if I remove the "opacity: 0" setting, but then the ugly button appears. I found an SO post where someone figured out how to call the login method without using the iframe. Only thing is we need the bot id, not the bot name. it's the first part of the token that the BotFather gives


components/common/PageLayout/components/Account/components/AccountModal/components/TelegramLoginIframe.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export function loginWithTelegram (callback: (user: TelegramAccount) => void) {
+  // @ts-ignore - defined by the script: https://telegram.org/js/telegram-widget.js
+  window.Telegram.Login.auth(
+    { bot_id: TELEGRAM_BOT_ID, request_access: true },
+    callback
+  );
+}
+
+export default class TelegramLoginButton extends React.Component<{ widgetVersion?: string }> {
```
mattcasey, line 15: I Moved this to a sub-component of Account... it doesn't belong in the general folder since it's a one-time use.

Devorein (approved):

components/common/PageLayout/components/Account/components/AccountModal/components/TelegramLoginIframe.tsx
Devorein: Makes sense

PR #343, opened 2022-04-10
Fixed not being able to detect content emptyness

Devorein:
We couldn't detect page content emptiness if there was a non-content node. See below (a page node). I encountered this while testing out the notion import
image
1. Correctly detect content emptiness for non-content nodes (mention, pages, cryptoPrice and so on)
2. Show the correct icon in the sidebar if the page is empty (couldn't detect correctly earlier)

### DIFF
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -258,7 +258,7 @@ export function convertPageContentToMarkdown (content: PageContent, title?: stri
   return markdown;
 }
 
-export function checkForEmpty (content: PageContent) {
+export function checkForEmpty (content: PageContent | null) {
   return !content?.content
   || content.content.length === 0
   || (content.content.length === 1
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -260,12 +260,12 @@ export function convertPageContentToMarkdown (content: PageContent, title?: stri
 }
 
 function checkForEmpty (content: PageContent) {
-  return content === null || !content?.content
+  return !content?.content
   || content.content.length === 0
   || (content.content.length === 1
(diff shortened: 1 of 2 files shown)


components/common/PageLayout/components/PageNavigation.tsx
```diff
@@ -490,9 +491,7 @@ function RenderDraggableNode ({ item, onDropAdjacent, onDropChild, pathPrefix, a
 
   const { focalboardViewsRecord } = useFocalboardViews();
 
-  const docContent = (item.content as PageContent)?.content;
-  const isEmptyContent = docContent && (docContent.length <= 1
-    && (!docContent[0] || (docContent[0] as PageContent)?.content?.length === 0));
+  const isEmptyContent = checkForEmpty(item.content as PageContent);
```
mattcasey, line 494: remember item.content can be null too, not sure you need to cast it here

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -278,10 +284,7 @@ function CharmEditor (
 
   function _onContentChange (view: EditorView) {
     // @ts-ignore missing types from the @bangle.dev/react package
-    const docContent: { content: { size: number } }[] = view.state.doc.content.content;
```
mattcasey, line 281: i am pretty sure thtis is is a different structure than what we write to our page

mattcasey (approved):

components/common/PageLayout/components/PageNavigation.tsx
Devorein: Yeah you are right. Thanks for the edit

components/common/CharmEditor/CharmEditor.tsx
Devorein: This is the structure before converting to json. We are converting it to json before passing to the emptiness detection function

PR #344, opened 2022-04-10
Fixed discord import roles of same name

Devorein:
Not the best solution as I couldn't figure out how to reproduce it more than one time but it will stop breaking the whole import if one role can't be imported. I've also added log to let us know which role we failed to import and all the roles the user was trying to import.


mattcasey:
I think, since it doesn't matter actually if it's there or not, we should just replace the create with an upsert

mattcasey (approved):
approved but i think it should be logged as a warning unless we are going to respond to this still

PR #345, opened 2022-04-11
Automate api testing

motechFR:
Provide automated health checks for our API


README.md
```diff
@@ -93,6 +93,67 @@ npx prisma migrate dev --name blocks_title_required
...
+- Integration tests: Covering the expected responses from API endpoints
+
+We use Jest to run these tests.
+
+You can run individual tests, or test the whole system.
+
+Before running tests, configure a .env.test.local file.
+
+Set the following value inside the .env.test.local file
+DATABASE_URL=postgres://postgres:postgres@localhost:5432/charmversetest
```
mattcasey, line 114: I think we should just commit this file. The db create script has "charmversetest" hard-coded as well, and I don't think anyone will need to override this for the time being

lib/utilities/errors.ts
```diff
@@ -1,13 +1,59 @@
-import { AlertColor } from '@mui/material/Alert';
 
-export interface IError {
+// Subset of error types from @mui/material/Alert
+export type ErrorSeverity = 'warning' | 'error';
+
+const ErrorCodes = {
+  Unknown: 0,
+  'Invalid input': 1,
```
mattcasey, line 7: Should these just be http status codes? We have 'errorType' on the error class which is already unique. This way the general error handler can just use the code it gets passed.

testing/setupDatabase.ts
```diff
@@ -0,0 +1,68 @@
+import { createUserFromWallet } from 'lib/users/createUser';
```
mattcasey, line 1: kind of feels like these utilities should go into /lib/testing , what do you think?

jest.config-server.ts
```diff
@@ -0,0 +1,225 @@
...
+  // testEnvironmentOptions: {},
+
+  // Adds a location field to test results
+  // testLocationInResults: false,
+
+  // The glob patterns Jest uses to detect test files
+  testMatch: [
+    '**/pages/api/**/?(*.)+(spec|test).[tj]s?(x)',
+    '**/public-api/**/?(*.)+(spec|test).[tj]s?(x)',
+    '**/testing/**/setupDatabase.spec.ts'
```
mattcasey, line 172: FYI this is missing regex for the lib folder (array.spec.ts for example).
Also, I would like to enforce consistency - all test files should end in spec.ts OR test.ts. I think I also prefer .spec, and am fine to just add special rules for the files inside focalboard for now or just rename those files

README.md
```diff
@@ -93,6 +93,67 @@ npx prisma migrate dev --name blocks_title_required
...
+$ dotenv -e .env.test.local -- npx prisma studio
+
+# Execute tests
+$ npm run test-server
+
+```
+
+
+__Info for running individual tests__
+If you are using VSCode, install the (Jest Runner plugin)[https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner]
```
mattcasey, line 146: i tried this out, pretty neat!

README.md
motechFR: Makes sense, added that [here](https://github.com/charmverse/app.charmverse.io/pull/345/commits/c646e945d726039a3f3792d436333849b97e4689)

package.json
```diff
@@ -10,7 +10,10 @@
     "start:remote": "react-env --path .env.production -- next dev",
     "prepare": "husky install",
     "db-tool": "dotenv -e .env.production -- npx prisma studio",
-    "db-tool:dev": "dotenv -e .env.local -- npx prisma studio"
+    "db-tool:dev": "dotenv -e .env.local -- npx prisma studio",
+    "test-setup-db": "./testing/configure-db.sh",
+    "test-client": "npm run test-setup && jest --config='./jest.config-client.ts'",
```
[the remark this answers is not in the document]
motechFR: Updated docs and package.json as per your feedback in this [commit](https://github.com/charmverse/app.charmverse.io/pull/345/commits/12424ce4e0056cc26ce89da04faa7d1d154b3262)

lib/users/getUser.ts
```diff
@@ -0,0 +1,53 @@
+import { LoggedInUser } from 'models';
+import { User, Prisma } from '@prisma/client';
+import { prisma } from 'db';
+
+type UserIdentifiers = Extract<keyof User, 'id' | 'addresses'>
+
+export type UserProfileQuery = Partial<Record<UserIdentifiers, string>>
```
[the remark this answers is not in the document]
motechFR: Good spot, cleaned up [here](https://github.com/charmverse/app.charmverse.io/pull/345/commits/bfafe05c30f97b4e6a2f6f6ca9e209f0f09290a3)

The type is still used so external consumers can specify if they're using an ID or an address

We could also simply make the method more intelligent, so it searches by ID or address internally

lib/users/getUser.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export async function getUserProfile (key: UserIdentifiers, value: string): Promise<LoggedInUser> {
```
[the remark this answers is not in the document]
motechFR: Going off our previous chat, we want to extract business logic from API controllers.

This piece of functionality can now be independantly tested.

lib/utilities/array.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { uniqueValues, sortArrayByObjectProperty } from './array';
```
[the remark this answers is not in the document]
motechFR: Cool, sounds like we've got our standard for inline tests. Updated [here](https://github.com/charmverse/app.charmverse.io/pull/345/commits/de1e1d34e90d715062ce701d5fdb059ea5ad2cb5)

lib/utilities/errors.ts
motechFR: Good idea! Made changes in this [commit](https://github.com/charmverse/app.charmverse.io/pull/345/commits/ca466cd88908f6e58171d647972f671590c5c178)

testing/setupDatabase.ts
motechFR: I think it's preferable to keep lib focused solely on application code, hence why I went for this approach.

Let me know if that makes sense

jest.config-server.ts
motechFR: Updated the [regex file](https://github.com/charmverse/app.charmverse.io/pull/345/commits/9e875bedf39c0f5e112f82937c72013e3da820a8) in alignment with this

For focalboard, that'll be handled separately in the jest.config-client.ts when we get to front end testing

README.md
motechFR: Thanks! Very handy indeed

testing/setupDatabase.ts
mattcasey: i'm fine with that

PR #357, opened 2022-04-13
Fixes around notion-import

Devorein:
1. Retry on all requests with `err.status > 499` error
2. No ungrouped page if there is none
3. Create default permission for pages created from cards (Was already added)
4. Hide the View as a Page button


lib/notion/importFromWorkspace.ts
```diff
@@ -1230,6 +1282,13 @@ export async function importFromWorkspace ({ workspaceName, workspaceIcon, acces
       await createCharmversePageFromNotionPage(block);
     }
   }
+
+  if (totalUngroupedPages > 0) {
+    await createPrismaPage(ungroupedPageInput);
+  }
+
+  log.info('[notion] Failed imports', { failedImportsRecord });
```
mattcasey, line 1290: you should just add failedImportsRecord to the existing log, that way there's only one log event with related ametadata

PR #360, opened 2022-04-13
Add page for each card

Devorein:
1. Create pages for the 3 cards (when creating a board)
2. Create a page for card when clicking on + icon in page navigation board pages
3. Create a page when clicking on + New underneath each kanban column

Make relation between Page and Block (Card) so that deleting a card deletes the page


package.json
```diff
@@ -156,6 +156,7 @@
     "@types/swagger-ui-react": "^4.1.1",
     "@typescript-eslint/eslint-plugin": "^5.17.0",
     "@typescript-eslint/parser": "^5.17.0",
+    "dotenv-cli": "^4.1.1",
```
mattcasey, line 159: please remove this, its already listed in prod dependencies and it needs to not be a dev dependency

mattcasey: you need to uninstall it completely and reinstall since its marked as a dev depdnency now in the package-lock.json

pages/api/blocks/index.ts
```diff
@@ -82,9 +82,31 @@ async function createBlocks (req: NextApiRequest, res: NextApiResponse<Block[]>)
         createdBy: req.session.user.id,
         updatedBy: req.session.user.id
       }));
+      const cardBlocks = newBlocks.filter(newBlock => newBlock.type === 'card');
+
+      const cardPages: Prisma.PageCreateManyInput[] = cardBlocks.map(cardBlock => ({
+        createdBy: cardBlock.createdBy,
+        updatedBy: cardBlock.updatedBy,
```
mattcasey, line 89: dont forget updatedAt

mattcasey (approved):

package.json
Devorein: Alright done

pages/api/blocks/index.ts
Devorein: Thanks for catching that. But prisma uses a default value of `now()` for `createdAt` and `updatedAt` fields

PR #362, opened 2022-04-14
Delete card pages and card blocks when deleting board

Devorein:
Right now if we delete the focalboard page, all the card pages get moved to the root level of the navigation
image


components/common/PageLayout/components/Sidebar.tsx
```diff
@@ -226,7 +230,7 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
         );
       }
     }
-
+    setPages(newPages);
```
mattcasey, line 233: You could just refresh the backend next time with the mutate() method since we use swr

mattcasey (approved):

PR #378, opened 2022-04-18
Sync card block and card page

Devorein:
1. Each card block can be viewed as a standalone page
2. Updating the icon, header image, and title of the card block updates that of the connected page for all views (board, table, calendar, and gallery, they had separate logic, unfortunately)
3. No charm_text block is created when adding a new card block
4. Fixed duplicating card block (to include creating a card page and not creating the `charm_text` block)
5. Fixed public API (to get content from page rather than charm_text) and relevant tests
6. Make sure notion import doesn't create charm_text


components/[pageId]/BoardPage/BoardPage.tsx
```diff
@@ -105,6 +105,7 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   });
 
   const showCard = useCallback((cardId?: string) => {
+    console.log(cardId);
```
Comment (mattcasey) [inline components/[pageId]/BoardPage/BoardPage.tsx:108]: console.log :)

pages/api/pages/[id]/index.ts
```diff
@@ -41,16 +41,71 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
+      if (!blockUpdateInput.fields) {
+        blockUpdateInput.fields = {};
+      }
+      (blockUpdateInput.fields as any).headerImage = req.body.headerImage;
+      shouldUpdate = true;
+      updatingFields = true;
+    }
+    if (shouldUpdate) {
+      if (updatingFields) {
+        const cardBlockFields = await prisma.block.findUnique({
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:84]: remember to remove this completely, shouldn't need to update cards anymore

theme/styles.scss
```diff
@@ -40,4 +40,15 @@ blockquote .bangle-nv-child-container {
 
 .telegram-login-CharmverseBot {
   width: 100%;
+}
+
+.CardDetail {
```
mattcasey, line 45: please move this to theme/focalboard

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts]: Yes I've removed that in the latest commits. We don't need to keep the page metadata in two places (page and block), just page is enough

.env.test.local
```diff
@@ -1,3 +1,3 @@
 DATABASE_URL=postgres://postgres:postgres@localhost:5432/charmversetest
-PORT=3335
+PORT=3000
```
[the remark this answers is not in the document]
Devorein: Make sense. I thought we are supposed to close our dev server and run the test server.

components/common/BoardEditor/focalboard/src/components/kanban/kanban.tsx
```diff
@@ -80,8 +74,9 @@ type Props = {
 
 const Kanban = (props: Props) => {
     const {board, activeView, cards, groupByProperty, visibleGroups, hiddenGroups} = props
+    
     const popupState = usePopupState({ variant: 'popper', popupId: 'new-group' });
-
+    const {pages} = usePages();
```
mattcasey, line 79: doesnt seem like this is used

components/common/BoardEditor/focalboard/src/mutator.ts
```diff
@@ -738,6 +749,9 @@ class Mutator {
         newCard.fields.isTemplate = asTemplate
         newCard.rootId = board.id
         newCard.parentId = board.id
+        if (cardPage) {
+          newCard.fields.content = cardPage.content
```
mattcasey, line 753: we aren't relying on this field anymore right?

pages/api/blocks/index.ts
```diff
@@ -84,29 +84,58 @@ async function createBlocks (req: NextApiRequest, res: NextApiResponse<Block[]>)
...
+              id: cardBlock.id
+            }
+          },
+          createdAt: cardBlock.createdAt,
+          path: `page-${Math.random().toString().replace('0.', '')}`,
+          title: cardBlock.title,
+          icon: cardBlock.fields.icon,
+          type: 'card',
+          headerImage: cardBlock.fields.headerImage,
+          contentText: '',
```
mattcasey, line 112: we should pull the text in as well from the existing page

pages/api/pages/[id]/index.ts
```diff
@@ -41,16 +40,30 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
+      id: pageId
     },
     data: req.body,
     include: {
       permissions: true
     }
   });
-  return res.status(200).json(space);
+
+  // Making sure the card page and card block title stays in sync
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:53]: is this still needed?

pages/api/public/pages/[pageId]/index.ts
```diff
@@ -28,7 +29,49 @@ async function getPublicPage (req: NextApiRequest, res: NextApiResponse<Page>) {
     return res.status(404).json({ error: 'Page not found' } as any);
   }
 
-  return res.status(200).json(page);
+  pages.push(page);
+
+  if (page.type === 'board') {
+    // For publicly shared board page fetch all of its card pages
```
Comment (mattcasey) [inline pages/api/public/pages/[pageId]/index.ts:35]: i would think it'd be cleaner to make the context in 'usePages' grab the correct pages instead?

pages/api/v1/databases/[id]/pages.ts
```diff
@@ -48,7 +47,13 @@ export async function createPage (req: NextApiRequest, res: NextApiResponse) {
     throw new DatabasePageNotFoundError(id as string);
   }
 
-  validateCreationData(req.body);
+  try {
```
Comment (motechFR) [inline pages/api/v1/databases/[id]/pages.ts:50]: We don't need to catch and send the error since this is now handled by onError middleware.

theme/styles.scss
Devorein: Yes thank you. I've relocated the styles and its working correctly

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -56,7 +56,7 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
 
   const isPage = router.route.includes('pageId');
 
-  const isExportablePage = (currentPage as Page)?.type === 'page';
+  const isExportablePage = (currentPage as Page)?.type.match(/(page|card)/);
```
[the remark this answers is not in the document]
Devorein: Sure reverting back to be more readable

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -80,6 +78,7 @@ export default function Header ({ open, openSidebar }: { open: boolean, openSide
       });
 
       let markdown = serializer.serialize(state.pmState.doc);
+      console.log({ markdown });
```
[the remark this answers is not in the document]
Devorein: Resolved

components/common/BoardEditor/focalboard/src/components/kanban/kanban.tsx
Devorein: Great catch. Thank you. Removed it.

components/common/BoardEditor/focalboard/src/mutator.ts
Devorein: No, we remove the content on the backend. This is a little hack I had to do in order to make it work with our existing API. I know it's not the most intuitive but in the backend, we use it to extract the content from the card block, place it on the card page that is being duplicated and remove the content from the card block.

pages/api/blocks/index.ts
Devorein: Any example of how we do that. We need to replace that one in a few other places as well.

pages/api/blocks/index.ts
```diff
@@ -84,29 +84,58 @@ async function createBlocks (req: NextApiRequest, res: NextApiResponse<Block[]>)
...
+          permissions: {
+            create: [
+              {
+                permissionLevel: 'full_access',
+                spaceId: cardBlock.spaceId
+              }
+            ]
+          }
+        };
+        if (cardBlock.fields.content) {
```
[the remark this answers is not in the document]
Devorein: Good point. Added it in the latest commits

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:53]: Good catch. I forgot to remove it.

pages/api/v1/databases/[id]/pages.ts
Comment (Devorein) [reply pages/api/v1/databases/[id]/pages.ts:50]: Alright that makes sense

components/common/BoardEditor/focalboard/src/mutator.ts
Devorein: After discussing we decided that its best to keep it the way it is.

pages/api/blocks/index.ts
Devorein: After discussing we decided that since the `contentText` is already present in the card block just pass it in the payload and delete it from the backend

PR #382, opened 2022-04-20
fix error logging server-side

mattcasey:
i noticed the error log is not showing the proper error anymore. i didn't test this, it also looks like we are now sending "message" instead of "error" property to the front-end. but i dont think we have updated all our front-end to look for that?


motechFR:
Hey @mattcasey just tweaked slightly how we build these errors.

Below are screenshots of a native JS error with Stacktrace, and one of our internal System Errors

System Error
image

Native error
image

mattcasey:
LGTM tahnks @motechFR

PR #385, opened 2022-04-20
Remove some circular dependencies

mattcasey:
This fixes 5 dependencies, mostly in our CharmEditor, from our code. there's 25 remaining but they are all inside focalboard


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -16,7 +16,7 @@ import {
 } from '@bangle.dev/base-components';
 import debounce from 'lodash/debounce';
 import { NodeView, Plugin, SpecRegistry, BangleEditorState } from '@bangle.dev/core';
-import { columnResizing, EditorView, Node } from '@bangle.dev/pm';
```
mattcasey, line 19: 'columnResizing' is part of skiff-org/tables... i think my VS Code added this import automatically, and i only saw a warning during the next build:
```
./components/common/CharmEditor/CharmEditor.tsx
Attempted import error: 'columnResizing' is not exported from '@bangle.dev/pm' (imported as 'columnResizing')
```

PR #393, opened 2022-04-22
Fix sort by checkbox in focalboard

mattcasey:


components/common/BoardEditor/focalboard/src/store/cards.ts
```diff
@@ -188,10 +187,7 @@ function sortCards(cards: Card[], board: Board, activeView: BoardView, usersById
             }
             Utils.log(`Sort by property: ${template?.name}`)
             sortedCards = sortedCards.sort((a, b) => {
-                // Always put cards with no titles at the bottom, regardless of sort
```
mattcasey, line 191: this seems like a dumb rule, why should cards with no title be handled differently?

components/common/BoardEditor/focalboard/src/store/cards.ts
```diff
@@ -211,39 +207,52 @@ function sortCards(cards: Card[], board: Board, activeView: BoardView, usersById
...
+                        result = -1;
+                    }
+                    else {
+                        result = titleOrCreatedOrder(a, b)
+                    }
                 } else {
                     // Text-based sort
 
                     if (aValue.length > 0 && bValue.length <= 0) {
-                        return -1
```
mattcasey, line 232: there were a couple places focalboard was just returning a result, but this means we'd ignore the 'reverse' option below. so i think it was an oversight

PR #394, opened 2022-04-23
Upload notion's images to our s3

mattcasey:
Notion's images expire so we need to re-host them when we import. The s3 path uses the spaceId to group images. if a space is ever deleted, we can delete the whole folder to save space.


lib/notion/importFromWorkspace.ts
```diff
@@ -1259,3 +1275,19 @@ export async function importFromWorkspace ({ workspaceName, workspaceIcon, acces
...
   return Object.values(failedImportsRecord).slice(0, 25);
 }
+
+// if image is stored in notion s3, it will expire so we need to reupload it to our s3
+function getPersistentImageUrl ({ image, spaceId }: { image: NotionImage, spaceId: string }): Promise<string | null> {
+  const url = image.type === 'external' ? image.external.url : image.type === 'file' ? image.file.url : null;
+  const isNotionS3 = url?.includes('amazonaws.com/secure.notion-static.com');
+  if (url && isNotionS3) {
+    const fileName = getFilePath({ url, spaceId });
+    return uploadToS3({ fileName, url }).then(r => r.url).catch(error => {
```
mattcasey, line 1285: added a catch here ... if you are running this locally without AWS cred set up, it will fail. otherwise i don't expect the error to happen

lib/aws/uploadToS3Server.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  await s3Upload.done();
+  const location = `https://s3.amazonaws.com/${bucket}/${fileName}`;
+  return {
+    url: location
+  };
+}
+
+export function getFilePath ({ spaceId, url }: { spaceId: string, url: string }) {
```
mattcasey, line 43: we're uploading images to /spaces/<spaceId>/<random>/<filename>

lib/notion/importFromWorkspace.ts
Devorein: Makes sense.

Devorein (approved):

PR #397, opened 2022-04-24
Recursively delete all child pages/blocks when deleting parent

Devorein:
1. Find all the child ids of the page/block recursively
2. Check if the user has delete permission for the page
3. If not remove it from the final set of ids to be "deleted"
4. Deleted doesn't mean removing from the database, we have the `deletedAt` field to indicate when it's been deleted.

There are two places where I had to implement this logic.


charmClient.ts
```diff
@@ -277,10 +277,7 @@ class CharmClient {
   }
 
   async deleteBlock (blockId: string, updater: BlockUpdater): Promise<void> {
-    const deletedBlock = await http.DELETE<Block>(`/api/blocks/${blockId}`);
-    const fbBlock = this.blockToFBBlock(deletedBlock);
-    fbBlock.deletedAt = new Date().getTime();
-    updater([fbBlock]);
+    await http.DELETE<{deletedCount: number}>(`/api/blocks/${blockId}`);
```
mattcasey, line 280: are you sure we don't need to return the blocks still? I copied this code from original focalboard.

components/common/PageLayout/components/Sidebar.tsx
```diff
@@ -200,34 +201,38 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
...
+            async () => {
+              // success
+            },
+            async () => {
+              // error
+            }
+          );
+        }
+      }
+      if (deletedCount !== 0) {
```
mattcasey, line 234: it feels unecessary to check if deletedCount is !== 0, since it should always be 1 unless someone else is trying to delete the same page at the same time, right?

components/common/PageLayout/components/Sidebar.tsx
```diff
@@ -200,34 +201,38 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
...
+        await mutate(`pages/${space?.id}`);
       }
     }
     setPages(newPages);
```
mattcasey, line 238: we are calling setPages, but also mutate on line 235.. we should only do one or the other

lib/api/deleteNestedChild.ts
```diff
@@ -0,0 +1,68 @@
...
+      },
+      select: {
+        id: true,
+        type: true
+      }
+    }));
+
+    childPageIds = [];
+    for (const _childPage of _childPages) {
+      const pagePermission = await computeUserPagePermissions({
```
mattcasey, line 25: these async functions can be done in parallel. we should update computePermissions to take a list of page ids. i think this is an expensive call

scripts/assignNull.ts
```diff
@@ -0,0 +1,27 @@
+import { prisma } from 'db';
+
+async function main () {
+  await prisma.page.updateMany({
+    where: {
+      deletedAt: {
+        not: null
```
mattcasey, line 7: what is the current value of these?

pages/api/pages/[id]/index.ts
```diff
@@ -54,13 +54,19 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
+    userId
   });
-  return res.status(200).json({ ok: true });
+
+  if (!permissionsSet.delete) {
+    return res.status(401).json({
+      error: 'You are not allowed to perform this action'
+    });
+  }
+  return res.status(200).json({ deletedCount: (await deleteNestedChild(pageId, userId)).length });
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:69]: put thins in its own line

components/common/PageLayout/components/Sidebar.tsx
Devorein: Thank you I feel that's an edge case that the user won't be able to get to. I will remove that asap.

components/common/PageLayout/components/Sidebar.tsx
Devorein: Yeah good point, no need to mutate the state twice

lib/api/deleteNestedChild.ts
Devorein: We decided to completely scratch the permission checking feature for nested child pages, so this has been removed in the latest commits.

charmClient.ts
Devorein: After discussion, we decided that this is a worthy feature to keep. So reverting back the changes

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:69]: Alright done

mattcasey (approved):

PR #399, opened 2022-04-26
clean up layout of view titles and enable toggling for public boards

mattcasey:
The original story was to enable toggling views in public. I went a little overboard and updated our UI to match Notion's :) i like the tabs a lot more

image


pages/api/blocks/[id]/index.ts
```diff
@@ -1,32 +1,61 @@
...
+  if (!rootBlock) {
+    throw new ApiError({
+      message: 'Block not found',
+      errorType: 'Data not found'
     });
   }
-  const { deletedChildPageIds, rootBlock } = await deleteNestedChild(pageId, userId);
-  return res.status(200).json({ deletedCount: deletedChildPageIds.length, rootBlock });
+
+  if (rootBlock.type === 'card' || rootBlock.type === 'board') {
```
Comment (mattcasey) [inline pages/api/blocks/[id]/index.ts:34]: Checking the block type is a bug fix - we could no longer delete view blocks, since there's no page permissions for them!

mattcasey:
@Devorein there's an issue with this, i'm trying to get double-click support working for the links but the links completely refresh the page

Devorein (approved):

PR #401, opened 2022-04-26
fix scrolling in focalboard cards

mattcasey:
This fixes a bug: https://app.charmverse.io/charmverse/page-49366552253645923?cardId=370458ce-83d1-4f52-aeb5-2fb06d2ba257&viewId=d15c9b90-8918-44da-bbd8-db8712faa723.

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.scss
+++ b/components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.scss
@@ -1,5 +1,4 @@
 .CardDetail {
-    overflow-x: hidden;
     
     .title {
         width: 100%;


components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.scss
```diff
@@ -1,5 +1,4 @@
 .CardDetail {
-    overflow-x: hidden;
```
mattcasey, line 2: This was added by you 5 days ago, but I'm not sure why. Doesn't seem like we should be hiding content, and it is triggering scrollbars on the y axis.

Devorein (approved):

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.scss
Devorein: Yeah, I'm not sure why I added that in the first place but there must've been a reason. I don't think it will impact anything significant

PR #408, opened 2022-04-26
Inheritance migrate script

motechFR:


scripts/inheritPagePermissions.ts
```diff
@@ -2,17 +2,19 @@ import { prisma } from 'db';
 import { setupPermissionsAfterPageRepositioned } from 'lib/permissions/pages/triggers';
 import { cond } from 'lodash';
 import { resolve } from 'node:path/win32';
-import { resolveChildPages } from '../lib/pages/server';
+import { IPageWithPermissions, resolveChildPages } from '../lib/pages/server';
 
-async function recursiveRebuild (pageId: string): Promise<true> {
+async function recursiveRebuild (pageId: string | IPageWithPermissions, level = 0, sourcePageNumber = 0): Promise<true> {
   await setupPermissionsAfterPageRepositioned(pageId);
```
mattcasey, line 6: why do we call this method? It seems to mostly focus on updates based on a parent, but all of these pages are roots (based on the query parentId: null).

mattcasey: i think u can skip if level === 0, might save some time

PR #409, opened 2022-04-27
more optimizations

mattcasey:
My approach was again to profile the React components while typing in the focalboard title input. most of it now looks like it's inside the MUI TreeView component. the only way we can get away w/o rendering all of that would be to separate the page content state from the page tree info, both of which we store in the PageProvider currently..

### DIFF
--- a/components/[pageId]/BoardPage/BoardPage.tsx
+++ b/components/[pageId]/BoardPage/BoardPage.tsx
@@ -41,7 +41,7 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   const dispatch = useAppDispatch();
   const [shownCardId, setShownCardId] = useState(router.query.cardId);
   const { pages } = usePages();
-  const accessibleCards = useMemo(() => cards.filter(card => pages[card.id]), [Object.keys(pages).toString()]);
+  const accessibleCards = useMemo(() => cards.filter(card => pages[card.id]), [cards, Object.keys(pages).toString()]);
 
   useEffect(() => {
     const boardId = page.boardId!;
--- a/components/[pageId]/BoardPage/BoardPage.tsx
+++ b/components/[pageId]/BoardPage/BoardPage.tsx
@@ -1,7 +1,7 @@
 import { generatePath } from 'lib/utilities/strings';
 import { Page } from 'models';
 import { useRouter } from 'next/router';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useState, useMemo } from 'react';
 import { useHotkeys } from 'react-hotkeys-hook';
 import CenterPanel from 'components/common/BoardEditor/focalboard/src/components/centerPanel';
 import { sendFlashMessage } from 'components/common/BoardEditor/focalboard/src/components/flashMessages';
@@ -41,7 +41,7 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   const dispatch = useAppDispatch();
   const [shownCardId, setShownCardId] = useState(router.query.cardId);
   const { pages } = usePages();
-  const accessibleCards = cards.filter(card => pages[card.id]);
+  const accessibleCards = useMemo(() => cards.filter(card => pages[card.id]), [Object.keys(pages).toString()]);
 
   useEffect(() => {
     const boardId = page.boardId!;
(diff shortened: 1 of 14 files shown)


components/[pageId]/BoardPage/BoardPage.tsx
```diff
@@ -41,7 +41,7 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   const dispatch = useAppDispatch();
   const [shownCardId, setShownCardId] = useState(router.query.cardId);
   const { pages } = usePages();
-  const accessibleCards = cards.filter(card => pages[card.id]);
+  const accessibleCards = useMemo(() => cards.filter(card => pages[card.id]), [cards, Object.keys(pages).toString()]);
```
Comment (mattcasey) [inline components/[pageId]/BoardPage/BoardPage.tsx:44]: accessibleCards depends on pages, but should only depend on the page ids.

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -106,31 +96,31 @@ function CenterPanel(props: Props) {
     }
   }
 
-  const addCardFromTemplate = async (cardTemplateId: string) => {
```
mattcasey, line 109: this and duplicateCardFromTemplate rely on pages, but i doubt anyone knows about the shortcut and the templates are currently hidden in the UI

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
```diff
@@ -173,8 +173,6 @@ interface PageLinkProps {
 }
 
 export function PageLink ({ showPicker = true, children, href, label, labelIcon, pageType, pageId }: PageLinkProps) {
-  console.log('redner page link');
-  const { setPages } = usePages();
```
mattcasey, line 177: moved this into EmojiMenu when it is visible

components/common/PageLayout/PageLayout.tsx
```diff
@@ -77,26 +77,29 @@ const HeaderSpacer = styled.div`
...
+const LayoutContainer = styled.div`
+  display: flex;
+  height: 100%;
+`;
+
+function PageLayout ({ children }: { children: React.ReactNode }) {
   const [open, setOpen] = React.useState(true);
   const [user] = useUser();
-  const { currentPageId, pages } = usePages();
-  const currentPage = pages[currentPageId];
```
mattcasey, line 84: i moved this call to usePages into CurrentPageFavicon component - we now re-render only that component instead of the whole page!

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
Devorein: Yeah that makes sense. I was cautious about removing it

components/common/PageLayout/PageLayout.tsx
Devorein: Thats great!!!

components/[pageId]/BoardPage/BoardPage.tsx
Comment (Devorein) [reply components/[pageId]/BoardPage/BoardPage.tsx:44]: is `Object.keys(pages).toString()` guaranteed to be the same on every render? As far as I remember the order is random and that would be a very large string of ids to check for equality....

components/common/PageLayout/PageLayout.tsx
Devorein: Yeah localized state ftw

components/[pageId]/BoardPage/BoardPage.tsx
Comment (mattcasey) [reply components/[pageId]/BoardPage/BoardPage.tsx:44]: I could add a sort.. but i think it's actually pretty predictable today.

I googled this interesting post https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/:
> What is interesting is that there was a spec change of e.g. Object.keys from [ES5](https://www.ecma-international.org/ecma-262/5.1/#sec-15.2.3.14) to [ES6](http://www.ecma-international.org/ecma-262/6.0/#sec-object.keys). The ES6 spec defines that Object.keys also relies on ownPropertyKeys which makes it predictable in today's browsers, too.

Devorein (approved):

components/[pageId]/BoardPage/BoardPage.tsx
Comment (Devorein) [reply components/[pageId]/BoardPage/BoardPage.tsx:44]: Thats nice. I guess it will work fine then

PR #411, opened 2022-04-27
Fix dependency of useContributors hook value so it updates consumers

motechFR:
Consumers of the useContributors hook were receiving an empty list of contributors

I have updated the useMemo dependency to ensure updates are properly cascaded.


mattcasey:
thanks!

PR #413, opened 2022-04-27
Make entering a title in focalboard cards faster + make FB table view faster

mattcasey:


components/bounties/BountyModal.tsx
```diff
@@ -20,7 +20,7 @@ export default function BountyModal (props: Props) {
   const { open, onClose, onSubmit, mode = 'create', bounty } = props;
   return (
     <Modal size='large' title={modalTitles[mode]} open={open} onClose={onClose} sx={{ margin: 'auto', maxHeight: '80vh' }}>
-      <BountyEditorForm onSubmit={onSubmit} bounty={bounty} mode={mode} />
+      {open && <BountyEditorForm onSubmit={onSubmit} bounty={bounty} mode={mode} />}
```
mattcasey, line 23: don't render charmeditor if it's not visible

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.tsx
```diff
@@ -59,12 +64,34 @@ const CardDetail = (props: Props): JSX.Element|null => {
...
+    }, [card]);
+
+    const page = pages[card?.id];
+    if (!card || !page) {
       return null
     }
-
     return (
-      <EditorPage pageId={card.id} publicShare={isSharedPage} shouldLoadPublicPage={false}/>
+      <DocumentPage page={page} setPage={setPage} readOnly={readonly} />
```
mattcasey, line 94: Didn't seem like we needed any logic in EditorPage except for setPage, which i just copied here for now because it's pretty straight-forward

components/common/BoardEditor/focalboard/src/components/table/tableRow.tsx
```diff
@@ -45,14 +41,10 @@ export const columnWidth = (resizingColumn: string, columnWidths: Record<string,
     return Math.max(Constants.minColumnWidth, columnWidths[templateId] || 0)
 }
 
-const TableRow = React.memo((props: Props) => {
-    const {board, activeView, onSaveWithEnter, columnRefs, card} = props
-    const contents = useAppSelector(getCardContents(card.id || ''))
-    const comments = useAppSelector(getCardComments(card.id))
```
mattcasey, line 51: i removed context for contents and comments. it was only being used to show lastUpdatedBy and lastUpdatedAt! the way these were written, it was always causing an update any time card were updated. i had a fix but then realized we could just get rid of this

components/common/BoardEditor/focalboard/src/store/comments.ts
```diff
@@ -53,3 +55,16 @@ export function getCardComments(cardId: string): (state: RootState) => CommentBl
             sort((a, b) => a.createdAt - b.createdAt)
     }
 }
+
+// optimized version. see: https://react-redux.js.org/api/hooks
+// export function getCardCommentsMemoFriendly (): (state: RootState, cardId: string) => CommentBlock[] {
```
mattcasey, line 60: left this here in case we need it to optimize someplace else

package.json
```diff
@@ -194,6 +195,7 @@
     "prettier": "^2.5.1",
     "prisma": "^3.9.2",
     "sass": "^1.49.0",
-    "supertest": "^6.2.2"
+    "supertest": "^6.2.2",
+    "ts-jest": "^27.1.4"
```
mattcasey, line 199: this is required for compiling ts

pages/_app.tsx
```diff
@@ -248,47 +247,45 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
   }
 
   return (
-    <CacheProvider value={createCache({ key: 'app' })}>
```
mattcasey, line 251: i removed the 'emotion cache provider'.. it seemed to have no effect. if we notice styles flickering on production, i'll bring this back.

components/[pageId]/DocumentPage/components/PageTitleInput.tsx
```diff
@@ -46,14 +46,24 @@ interface PageTitleProps {
 
 export default function PageTitle ({ value, onChange, readOnly }: PageTitleProps) {
   const view = useContext(EditorViewContext);
+  const [title, setTitle] = useState(value);
```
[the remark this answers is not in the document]
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/PageTitleInput.tsx:49]: Nice one! Small suggestion, how about decoupling the parent onChange call from local _onChange ?

We could link the parent onChange call to an onBlur event instead

PR #414, opened 2022-04-28
Bounty filters

motechFR:
When opening bounty list, show only active bounties.

Allow users to filter bounties.

image


components/common/form/InputBountyStatus.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <Grid item xs={12}>
+        <Select
+          multiple
+          value={selectedChoices as any}
+          onChange={selectOption}
+          displayEmpty={true}
+          renderValue={(selectedValues: any) => (
+
+            (renderSelected === false || !selectedValues || selectedValues.length === 0) ? (
+              'Select status'
```
mattcasey, line 62: Maybe "Include status" would be more helpful?

components/bounties/BountyStatusBadge.tsx
```diff
@@ -65,6 +67,55 @@ export interface IBountyBadgeProps {
...
+  return (
+    <Chip
+      label={BOUNTY_LABELS[status]}
+      variant='filled'
+      onDelete={onDelete}
+      sx={{
+        fontWeight: 'bold',
+        backgroundColor: () => {
+          // @ts-ignore
+          return theme.palette[BountyStatusColours[status]].main;
```
mattcasey, line 83: should make sure this color is set, or it could blow up the UI

components/bounties/BountyStatusBadge.tsx
```diff
@@ -65,6 +67,55 @@ export interface IBountyBadgeProps {
...
 
+export function BountyStatusChip ({ status, onDelete }: {status: BountyStatus, onDelete?: (status: BountyStatus) => void}) {
+
+  const theme = useTheme();
+
+  return (
+    <Chip
+      label={BOUNTY_LABELS[status]}
+      variant='filled'
+      onDelete={onDelete}
```
mattcasey, line 78: I suggest using 'size="small" here... the chips looked pretty big at first.
image

components/common/form/InputBountyStatus.tsx
```diff
@@ -0,0 +1,98 @@
+import Box from '@mui/material/Box';
```
mattcasey, line 1: This feels more like a "molecule" than an "atom". i dont think its necessary to put all our form molecules in the common folder, this seems like a one-off for the bounty page.

mattcasey (approved):
Works as expected! Design-wise, I think the filters are a little too prominent. I don't think people will need to change them often, and I wasn't sure at first what I am looking at. Some ideas - make them smaller, align on the right side, also include a small grey label: "Bounty Status", "Filter by Status"... not sure. I think this is fine as-is for v1 though

components/common/form/InputBountyStatus.tsx
motechFR: Good feedback, I've split up the options for rendering selected options in available options, and rendering selected options locally in the component in this [commit](https://github.com/charmverse/app.charmverse.io/pull/414/commits/134ffb0a3b269f37f000a06b2fa1f4210e3cb4e8)

components/common/form/InputBountyStatus.tsx
motechFR: Yeah makes sense, it's highly focused on bounties after all. Moved it over to bounties folder.

motechFR:
I've made the changes you suggested (reposition chips to right + reduce size), apart from adding the label, to not clutter up UI

I think definitely once we have multiple filters might make sense to add label in

PR #416, opened 2022-04-29
Inline comments matt

mattcasey:
- mostly CSS changes
- put the comment actions under a popup menu

### DIFF
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -273,7 +273,7 @@ const StyledReactBangleEditor = styled(ReactBangleEditor)`
   }
 `;
 
-const PageThreadListBox = styled(Box)`
+const PageThreadListBox = styled.div`
   position: fixed;
   right: 0px;
   width: 400px;
--- a/components/common/CharmEditor/InlineCharmEditor.tsx
+++ b/components/common/CharmEditor/InlineCharmEditor.tsx
@@ -109,19 +109,10 @@ const StyledReactBangleEditor = styled(ReactBangleEditor)`
   }
 `;
 
-const defaultContent: PageContent = {
-  type: 'doc',
-  content: [
-    {
-      type: 'paragraph'
-    }
-  ]
-};
-
 export type UpdatePageContent = (content: ICharmEditorOutput) => any;
 
 interface CharmEditorProps {
-  content?: PageContent;
+  content?: PageContent | null;
   children?: ReactNode;
   onContentChange?: UpdatePageContent;
   readOnly?: boolean;
@@ -129,7 +120,7 @@ interface CharmEditorProps {
 }
 
 export default function CharmEditor (
-  { content = defaultContent, children, onContentChange, style, readOnly = false }:
+  { content, children, onContentChange, style, readOnly = false }:
   CharmEditorProps
 ) {
   const onContentChangeDebounced = onContentChange ? debounce((view: EditorView) => {
--- a/components/common/CharmEditor/components/@bangle.dev/react-menu/Menu.tsx
+++ b/components/common/CharmEditor/components/@bangle.dev/react-menu/Menu.tsx
@@ -1,10 +1,8 @@
 import styled from "@emotion/styled";
 import React from 'react';
+import Paper from '@mui/material/Paper';
 
-const StyledMenu = styled.div`
-  background-color: ${({ theme }) => theme.palette.background.light};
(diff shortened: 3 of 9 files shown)


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -273,7 +273,7 @@ const StyledReactBangleEditor = styled(ReactBangleEditor)`
   }
 `;
 
-const PageThreadListBox = styled(Box)`
+const PageThreadListBox = styled.div`
```
mattcasey, line 276: Please don't use "Box" if you are not using any Box properties - it's just extra overhead, classes, etc. we don't need

components/common/CharmEditor/InlineCharmEditor.tsx
```diff
@@ -109,27 +109,18 @@ const StyledReactBangleEditor = styled(ReactBangleEditor)`
   }
 `;
 
-const defaultContent: PageContent = {
```
mattcasey, line 112: there's no need for default Content for the charm editor

components/common/CharmEditor/components/@bangle.dev/react-menu/Menu.tsx
```diff
@@ -1,10 +1,8 @@
 import styled from "@emotion/styled";
 import React from 'react';
+import Paper from '@mui/material/Paper';
 
-const StyledMenu = styled.div`
-  background-color: ${({ theme }) => theme.palette.background.light};
-  box-shadow: 0 10px 15px -3px rgb(0 0 0 / 0.1), 0 4px 6px -4px rgb(0 0 0 / 0.1);
-  border-radius: ${({ theme }) => theme.spacing(0.5)};
+const StyledMenu = styled(Paper)`
```
mattcasey, line 5: Paper gives us standard box-shadow, border-radius, and background color. ALso I've been trying to use Paper because it look better - white background for popups instead of the dark grey

components/common/CharmEditor/components/inlineComment/InlineComment.components.tsx
```diff
@@ -18,9 +18,8 @@ import PageThread from '../PageThread';
 import { SuggestTooltipPluginKey, SuggestTooltipPluginState, hideSuggestionsTooltip } from '../@bangle.dev/tooltip/suggest-tooltip';
 import { updateInlineComment } from './inlineComment.utils';
 
-const ThreadContainerBox = styled(Box)`
+const ThreadContainer = styled.div`
   max-height: 400px;
-  overflow: auto;
```
mattcasey, line 23: i had to remove this rule so that the shadow would appear, not sure if it was used for something?

theme/@bangle.dev/styles.scss
```diff
@@ -353,7 +353,7 @@ ol li {
 }
 
 .bangle-editor p {
-  font-size: 1rem;
+  //font-size: 1rem;
```
mattcasey, line 356: not 100% sure what effect this will have on our regular documents, but i think it looks a little better and it allowed me to make the font-size smaller for comments

components/common/CharmEditor/components/@bangle.dev/react-menu/Menu.tsx
Devorein: Yeah we should refrain from our custom style and use a MUI when ever possible

components/common/CharmEditor/InlineCharmEditor.tsx
Devorein: Yeah checked out it. Its not necessary

components/common/CharmEditor/CharmEditor.tsx
Devorein: Great catch. I will try to replace it in all places I used it unintentionally.

theme/@bangle.dev/styles.scss
Devorein: It looks fine to me :D

components/common/CharmEditor/components/inlineComment/InlineComment.components.tsx
Devorein: This is actually quite important as otherwise when showing the inline thread box it would not overflow correctly

Devorein (approved):

components/common/CharmEditor/components/inlineComment/InlineComment.components.tsx
Devorein: This is actually quite important. It allows the inline threads to not overflow. 
image
image

PR #418, opened 2022-04-29
edit comments inline

mattcasey:


components/common/CharmEditor/InlineCharmEditor.tsx
```diff
@@ -119,10 +122,11 @@ interface CharmEditorProps {
   onContentChange?: UpdatePageContent;
   readOnly?: boolean;
   style?: CSSProperties;
+  noPadding?: boolean;
```
mattcasey, line 125: the placeholder looked off.. i think the padding we were passing in via style and padding set on the placeholder need to be in sync, so this was my solution

Devorein (approved):

Devorein:
I merged it using git rather than in github. Closing the PR.

PR #421, opened 2022-04-29
Separate component with localized state to not rerender parent component

Devorein:
1. Move the state from parent `PageThread` component to more localized child components
2. Previously since the state was in the parent, anytime I wrote something in the Add comment charmeditor the time data of the comment changed, which lead to a very bad UX.
3. It's still there in some places (Clicking on edit comment from the ellipsis or even clicking on it), but mostly hidden


components/common/CharmEditor/components/PageThread.tsx
```diff
@@ -402,36 +425,28 @@ const PageThread = forwardRef<HTMLDivElement, PageThreadProps>(({ showFindButton
...
-                resetState();
-              }}
-            >
-              Add
-            </Button>
-          </Box>
-        </Box>
+        <AddCommentCharmEditor
+          key={`${Boolean(editedCommentId)}.${thread.comments[thread.comments.length - 1]?.id}`}
+          readOnly={Boolean(editedCommentId)}
```
mattcasey, line 430: we dont use this input to edit comments so i think you can remove the references to editedCommentId?

lib/inline-comments/client/scrollToThread.ts
```diff
@@ -0,0 +1,44 @@
+import { highlightDomElement } from 'lib/dom/client/highlight';
```
mattcasey, line 1: i dont think we should put everything inclient/server folders.. is there a reason this has to be in a client folder? Keep in mind since we do SSR, this is getting imported on the server as well anyway

mattcasey: i also think this belongs with the inline comment feature, not in lib

components/common/CharmEditor/components/PageThread.tsx
Devorein: This doesn't make the CharmEditor disabled (which it should since we are passing `readOnly` prop) when we are editing a comment. I've tried it out without the editedCommentId key and it doesn't work as it uses the same component as the key didn't change.

lib/inline-comments/client/scrollToThread.ts
Devorein: Since it was using a specific dom api I thought we shouldn't keep it inside the `lib/[folder]/client` folder. But you are right, I will move it to `inlineCOmment.utils.ts` since its more suitable there

Devorein: See there are a few more files, three of which are client/server agnostic, but `scrollToThread` seemed to be a client specific thing.

components/common/CharmEditor/components/PageThread.tsx
```diff
@@ -82,61 +83,122 @@ function ThreadHeaderButton ({ disabled = false, onClick, text, startIcon, ...pr
...
+            style={{
+              fontSize: 14,
+              width: '100%'
+            }}
+          />
+        </Box>
+      </Box>
+      <Collapse
+        sx={{
+          pl: isEditable ? 4 : 1
```
mattcasey, line 160: i think this should always be pl: 4. i dont know why i had it one and then the other

Devorein: No I actually edited that, in order to align the edit buttons alongside the charmeditor

mattcasey (approved):

PR #422, opened 2022-04-29
Add delete support to all rows in CharmEditor

mattcasey:
I added "Duplicate" too, since I wanted to have a context menu. 

image


components/common/CharmEditor/components/rowActions/rowActions.ts
```diff
@@ -0,0 +1,106 @@
...
+                rowNodeOffset: dom.offset && dom.node.childNodes[dom.offset] ? dom.offset : 0
+              };
+              view.dispatch(view.state.tr.setMeta(key, newState));
+            }
+          }
+        }
+
+        const throttled = throttle(onMouseOver, 100);
+
+        view.dom.addEventListener('mousemove', throttled);
```
mattcasey, line 96: I could have used 'mouseover' which happens much less often here, but I want to pick up your cursor moving in the left gutter of the page, where it's just padding from the container. At first, I tried shifting over all elements with margin-left: -50, padding-left: 50, but noticed immediately that it might break styling of some elements, for example list items have a padding already set. But if the elements are shifted over, we would be able to just listen for mouseover.

components/common/CharmEditor/components/CryptoPrice.tsx
```diff
@@ -42,7 +42,7 @@ export function cryptoPriceSpec () {
       group: 'block',
       parseDOM: [{ tag: 'div.charm-crypto-price' }],
       toDOM: (): DOMOutputSpec => {
-        return ['div.charm-crypto-price', 0];
+        return ['div.charm-crypto-price'];
```
mattcasey, line 45: this '0' was actually throwing an error, so i removed it. you're not supposed to have a "0" if there is no content in a component.

Devorein: Nice catch

components/common/CharmEditor/components/rowActions/rowActions.ts
Devorein: I can take a look into it. I just used `margin-left: -50px, padding-left: 100px` and it doesn't seem like its breaking the UL tag around list items 🤔

Devorein (approved):

PR #423, opened 2022-04-29
Sidebar improvements

motechFR:
Hide access to settings section if user is not a space admin

Add in a link to the Charmverse discord for support

image

### DIFF
--- a/components/common/CreateSpaceForm.tsx
+++ b/components/common/CreateSpaceForm.tsx
@@ -75,6 +75,7 @@ export default function WorkspaceSettings ({ defaultValues, onSubmit: _onSubmit,
         spaceRoles: {
           create: [{
             role: 'admin',
+            isAdmin: true,
             user: {
               connect: {
                 id: user!.id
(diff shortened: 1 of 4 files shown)


components/common/CreateSpaceForm.tsx
```diff
@@ -75,6 +75,7 @@ export default function WorkspaceSettings ({ defaultValues, onSubmit: _onSubmit,
         spaceRoles: {
           create: [{
             role: 'admin',
+            isAdmin: true,
```
mattcasey, line 78: nice catch

components/common/PageLayout/components/Sidebar.tsx
```diff
@@ -167,6 +168,8 @@ export default function Sidebar ({ closeSidebar, favorites }: SidebarProps) {
   const [isScrolled, setIsScrolled] = useState(false);
   const favoritePageIds = favorites.map(f => f.pageId);
 
+  const isAdmin = user?.spaceRoles.some(role => role.spaceId === space?.id && (role.isAdmin === true || role.role === 'admin'));
```
mattcasey, line 171: can we deprecate/delete the use of "role" now? why is it still a valid way to be admin?

Devorein: I have the same question. Why do we have two ways to check if someone is an admin 😕

motechFR:
Closed as #424 handled this in further detail

PR #426, opened 2022-05-01
Archive and permanently delete pages

Devorein:
1. A simple modal view to show deleted pages sorted by their deleted date in descending order
2. Two buttons at the rightmost to allow restoring and permanently deleting a page
image
3. When visiting a deleted page in the trash a banner is shown at the top to restore or permanently delete the page
image
4. Permanently deleting a page would permanently delete all of its child pages, and the same goes for restoring a page
5. Deleted pages can't be publicly shared nor can they be publicly viewed (this is what Notion does)

**NOTE** Separately deleted card will be shown inside a deleted board, this is because we can't figure out if the card was deleted manually or was it deleted by the board


charmClient.ts
```diff
@@ -117,8 +117,16 @@ class CharmClient {
...
+  archivePage (pageId: string) {
+    return http.DELETE<{deletedCount: number}>(`/api/pages/${pageId}?action=archive`);
+  }
+
+  async deletePage (pageId: string) {
+    return http.DELETE<{deletedCount: number}>(`/api/pages/${pageId}?action=delete`);
+  }
+
+  async restorePage (pageId: string) {
+    return http.DELETE<{deletedCount: number}>(`/api/pages/${pageId}?action=restore`);
```
mattcasey, line 129: this should not be a DELETE method. We should only use DELETE for actual deletPage

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -70,7 +71,7 @@ function Editor ({ page, setPage, readOnly = false }: IEditorProps) {
 
   const card = cards.find(_card => _card.id === page.id);
 
-  const comments = card ? useAppSelector(getCardComments(card.id)) : [];
+  const comments = useAppSelector(getCardComments(card?.id));
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:74]: i actually dont think this is very clean. you're putting the onus on getCardComments to handle someone calling it with undefined.

mattcasey (changes requested):
- when deleting a page, we should no longer redirect you anywhere - the Deleted banner should just appear
- need to filter out deleted pages from /pages/[domain]/index.ts. right now it sends me to a deleted page every time i open my workspace, because it happens to be the first one

components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+    // Route to the first alive page
+    router.push(`/${router.query.domain}/${Object.values(pages).find(page => page?.deletedAt === null)?.path}`);
+    setIsMutating(false);
+  }
+
+  async function restorePage () {
+    const page = pages[pageId];
+    setIsMutating(true);
+    await charmClient.restorePage(pageId);
+    await mutate(`pages/${space?.id}`);
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx:45]: why do we need to re-query all pages? remember this is pretty slow

components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const dispatch = useAppDispatch();
+  const { pages } = usePages();
+  const router = useRouter();
+
+  const isShowingCard = new URLSearchParams(window.location.search).get('cardId');
+
+  async function deletePage () {
+    setIsMutating(true);
+    await charmClient.deletePage(pageId);
+    await mutate(`pages/${space?.id}`);
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx:35]: just use setPages instead

components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx:45]: originally i thought we didnt need to call mutate, now i remember there's children that might be affected. wo[uld it be hard to have restorePage return a list of pages to "restore"? that would be much faster, u could just loop over them and fix it with setPages..

lib/pages/modifyChildPages.ts
```diff
@@ -0,0 +1,89 @@
+import { Prisma } from '@prisma/client';
+import { prisma } from 'db';
+
+export type ChildModificationAction = 'delete' | 'restore' | 'archive'
+
+export async function modifyChildPages (parentId: string, userId: string, action: ChildModificationAction) {
```
mattcasey, line 6: great candidate for tests

Devorein: Absolutely agree with that. I will add some.

components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx:45]: Its been implemented in the latest commit.

charmClient.ts
Devorein: Good point. Went with `/api/pages/[id]/archive` endpoint to support archiving and restoring pages.

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:74]: I reverted it back since it caused a nasty and really hard-to-find issue when I tried to restore an archived page.
`useAppSelector` is a hook, and we are conditionally calling the hook. This violates the rule of using hooks and leads to this error
`Uncaught Invariant Violation: Rendered more hooks than during the previous render`

mattcasey:
Looks good! I have one final suggestion:
image


Replace "Page 1" placeholder with "Filter by page title..." and we don't need a description below. (Inspired by Notion)

mattcasey (approved):

PR #429, opened 2022-05-02
Admin status fix

motechFR:
1. Create/Update/Delete roles
- Only admin can delete, create and update roles
- Only admin can assign or unassign roles
- If discord roles are imported, and I connect my discord account as a non admin user, I will still auto-receive my Discord roles

2. Create/Update/Delete invites
- Any user (including non space members) can accept invites
- Any space member can see invites
- Only admin can create or delete invites

3. Update/Delete space
- Only admin can update or delete space

4. Update contributors (admin or contributor role)
- admin can delete any space membership
- user can only delete their own space membership
- only admin can change contributor / admin status
- if the last admin will be made a contributor, reject this operation so the space has at least 1 admin
- if the last admin will be deleted, reject this operation so the space has at least 1 admin

5. Add/Remove token gates
- anyone can get token gates (useful for connecting to a space)
- only admin can add or remove token gates

6. Add/Remove/Update payment methods
- Any space member can list or add payment methods
- Only admin can delete payment methods

### DIFF
--- a/components/settings/payment-methods/PaymentMethods.tsx
+++ b/components/settings/payment-methods/PaymentMethods.tsx
@@ -4,17 +4,20 @@ import { usePaymentMethods } from 'hooks/usePaymentMethods';
 import Button from 'components/common/Button';
 import Typography from '@mui/material/Typography';
 import { usePopupState } from 'material-ui-popup-state/hooks';
+import useIsAdmin from 'hooks/useIsAdmin';
 import CustomERCTokenForm from './components/CustomERCTokenForm';
 import GnosisSafeForm from './components/GnosisSafeForm';
 import Legend from '../Legend';
 import PaymentMethodList from './components/PaymentMethodList';
 
-export default function PaymentMethods ({ isAdmin = true }) {
+export default function PaymentMethods () {
 
   const gnosisPopupState = usePopupState({ variant: 'popover', popupId: 'gnosis-popup' });
   const ERC20PopupState = usePopupState({ variant: 'popover', popupId: 'ERC20-popup' });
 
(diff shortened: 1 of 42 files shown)


components/settings/payment-methods/PaymentMethods.tsx
```diff
@@ -4,18 +4,21 @@ import { usePaymentMethods } from 'hooks/usePaymentMethods';
...
 
-export default function PaymentMethods ({ isAdmin = true }) {
+export default function PaymentMethods () {
 
   const gnosisPopupState = usePopupState({ variant: 'popover', popupId: 'gnosis-popup' });
   const ERC20PopupState = usePopupState({ variant: 'popover', popupId: 'ERC20-popup' });
 
   const [paymentMethods] = usePaymentMethods();
 
+  const isAdmin = useIsAdmin();
```
mattcasey, line 20: we actually allow non-admins to create payment methods. the default was true before.. i dont see a reason to remove access (i mentioned this in the bugs channel too).

lib/users/errors.ts
```diff
@@ -0,0 +1,21 @@
+import { SystemError } from 'lib/utilities/errors';
+
+export class AdministratorOnlyError extends SystemError {
```
mattcasey, line 3: i think all of these should be in lib/middlewre. its too hard to go find what errors are available all the time

motechFR: I was going with the approach of library-specific error classes.

However, you raise a good point. Maybe we could promote lib/utilities/error to its own folder, and have different error families there in files
ie
errors/user
errors/space
errors/api

PR #430, opened 2022-05-02
Cron job script to delete archived pages

Devorein:
1. A simple script (using `node-cron`) to schedule a cron job every hour to check for archived pages and delete them
2. The archived pages must be at least 30 days old to be deleted
3. Updated `start:prod` npm script to use `concurrently` to run both cron script and next.js prod server simultaneously


package.json
```diff
@@ -10,9 +10,10 @@
     "lint": "eslint -c .eslintrc.json . --ext .ts --ext .tsx",
     "lint:prod": "npm run lint -- --quiet",
     "start": "next dev",
+    "start:cron": "dotenv -e .env.local -- ts-node scripts/deleteArchivedPages.ts",
```
mattcasey, line 13: maybe add 'start:cron:prod'

scripts/deleteArchivedPages.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export async function main () {
+  log.debug('[cron]: Starting delete-archived-pages cron job');
+  // Cron job that runs every hour
+  cron.schedule('0 * * * *', async () => {
+    const pages = await prisma.page.findMany({
+      where: {
+        deletedAt: {
+          not: null
```
mattcasey, line 20: lte 30daysago

package.json
Devorein: Yep I've added the necessary steps for both prod, dev and staging environment

scripts/deleteArchivedPages.ts
Devorein: Great idea. Added that logic in the latest commit

lib/pages/__tests__/deleteArchivedPages.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+import { v4 } from 'uuid';
+import { deleteArchivedPages } from '../server/deleteArchivedPages';
+
+describe('deleteArchivedPages', () => {
+  it('Should delete the archived pages older than 30 days', async () => {
+    const space = await prisma.space.findFirst();
+    const user = await prisma.user.findFirst();
+    if (space && user) {
+      // A page archived just now
+      const firstPage = await createPage({
```
mattcasey, line 13: please put this into a 'before' method, since it's test setup. and we should delete any pages not used in an "after" method

motechFR: Ah yes, indeed, since whole test suite runs, you will inevitably be affected by global state.

The "generateUserAndSpaceWithApiToken" provides you with a guaranteed new space and user pair, which will be scoped to that test suite or event individual test, depending on where to use it.

This way you can deterministically provide a spaceId in all tests instead of returning a random spaceId as you are currently doing.

```typescript
// Add this to the top of your test
let user: User;
let space: Space;
let apiToken: SpaceApiToken;

beforeAll(async () => {
  const generated = await generateUserAndSpaceWithApiToken(v4());

  user = generated.user;
  space = generated.space;
  apiToken = generated.apiToken;
});


```

lib/pages/__tests__/deleteArchivedPages.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      const pages = await prisma.page.findMany({
+        select: {
+          id: true
+        }
+      });
+      const blocks = await prisma.block.findMany({
+        select: {
+          id: true
+        }
+      });
```
motechFR, line 72: Update this query to use the spaceId from the space you created at the top of test suite.

BeforeAll will set up this space for you and make it accessible to all tests in that file. No further action needed

mattcasey: that would work except we are also evaluating the deletedBlocksCount and deletedPagesCount. IMO, while it wouldn't hurt to do here, it is on other tests to be cleaning up their state, since not all actions we have can be guaranteed to be space-specific. and if we don't start cleaning up state now, it will be much harder in the future.

PR #437, opened 2022-05-03
Fix z-index of wallet connect for Lit modal

mattcasey:
I started replacing all our custom z-indexes with standard ones from MUI. Also, using css variables because these are much easier to use than importing the theme object...!

image


components/bounties/[bountyId]/components/BountyApplicantList.tsx
```diff
@@ -116,8 +116,7 @@ export function BountyApplicantList ({
           background: theme.palette.background.dark,
           '.MuiTableCell-root': {
             background: theme.palette.settingsHeader.background
-          },
-          zIndex: 9000
```
Comment (mattcasey) [inline components/bounties/[bountyId]/components/BountyApplicantList.tsx:120]: this one is suspicious but i don't know why we would need to set a zIndex here. maybe if the header was sticky? I tested and it seems to be fine

Devorein (approved):

PR #438, opened 2022-05-03
UI updates to trash modal

mattcasey:
- more room for scrollable area
- used MenuItem + next/link components so you can click the whole row to navigate, and it highlights on hover
image


Devorein (approved):
Great everything looks fine. I just added a new commit to make sure we add left padding when there are no archived pages.
image

PR #441, opened 2022-05-04
Update README.md

AndreiMitrea:


mattcasey (approved):
FYI we should not be having to run npx patch-package, its in the postinstall. haven't figured out what's going on

mattcasey:
I think this can be closed after https://github.com/charmverse/app.charmverse.io/pull/444 is merged, user shouldn't have to patch packages manually

PR #442, opened 2022-05-04
Publish proposals

motechFR:
This PR allows a user to export proposals to Snapshot

1. Connect your snapshot domain
- From the workspace settings
- Or inline, from the "publish to Snapshot" section

2. Validate you can publish
- (auto) Checks Charmverse space is connected to Snapshot space
- (auto) Checks the user is allowed to post
- (auto) Checks to ensure the user doesn't exceed the snapshot max length for proposal

Good to know: UI says max 14 400 characters, but here [says](https://docs.snapshot.org/proposals/create#proposals-limitations) 6 400 characters

3. Publish to snapshot
- Sign request from your Metamask wallet
- Link to snapshot proposal shows in context menu


components/snapshot/PublishToSnapshot.tsx
```diff
@@ -0,0 +1,119 @@
+import ExitToAppIcon from '@mui/icons-material/ExitToApp';
```
mattcasey, line 1: there is no "snapshot" page, so these components belong someplace else. We organize components by heierarchy, so this seems to fit inside of PageLayout, where we have the dropdown to publish to Snapshot

components/snapshot/PublishToSnapshot.tsx
```diff
@@ -0,0 +1,119 @@
...
+      {
+
+      checkingProposal && (
+        <>
+          <LoadingIcon size={18} sx={{ mr: 1 }} />
+          <ListItemText primary='Checking proposal' />
+          {
+
+          /**
+           *   <LoadingComponent>
```
mattcasey, line 67: looks like this can be removed?

mattcasey (changes requested):
looks good! only thing is i think the snapshot React components belong in the header component according to their heierachy

components/snapshot/PublishToSnapshot.tsx
motechFR: Sounds good, relocated it there

pages/_app.tsx
```diff
@@ -183,6 +183,8 @@ import 'theme/styles.scss';
 import Snackbar from 'components/common/Snackbar';
 import { SnackbarProvider } from 'hooks/useSnackbar';
 import { initialLoad } from 'components/common/BoardEditor/focalboard/src/store/initialLoad';
+import { LocalizationProvider } from '@mui/x-date-pickers/LocalizationProvider';
+import { AdapterLuxon } from '@mui/x-date-pickers/AdapterLuxon';
```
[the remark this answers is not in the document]
motechFR: Gotcha about the providers.

In this case, since this enables us to use MUI DatePickers anywhere in the app, I think we can keep it here.

PR #447, opened 2022-05-05
Updated crypto price format

AndreiMitrea:
Updated the format in which the crypto price is displayed in the crypto module. The two scenarios are:

- if price > 0 : display two digits after the decimal point
- if price < 0: display four digits after the decimal point

image

image

### DIFF
--- a/lib/utilities/__tests__/formatting.spec.ts
+++ b/lib/utilities/__tests__/formatting.spec.ts
@@ -1,0 +1,16 @@
+import { formatMoney } from '../formatting';
+
+describe('amount formating', () => {
+
+  it('should be correct for values larger than 1', () => {
+    const amount = formatMoney(2932.12345, 'USD');
+
+    expect(amount).toBe('$2,932.12');
+  });
+
+  it('should be correct for values less than 1', () => {
+    const amount = formatMoney(0.12345, 'USD');
+
+    expect(amount).toBe('$0.1234');
+  });
+});
--- a/lib/utilities/formatting.ts
+++ b/lib/utilities/formatting.ts
@@ -6,9 +6,11 @@ export function formatMoney (amount: number, currency: FiatCurrency): string {
 
   const formatter = new Intl.NumberFormat(userLocale, {
     style: 'currency',
-    currency
+    currency,
+    ...(amount < 1 && {
+      minimumFractionDigits: 4
+    })
   });
 
   return formatter.format(amount);
-
 }


lib/utilities/__tests__/formatting.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { formatMoney } from '../formatting';
+
+describe('amount formatting', () => {
```
motechFR, line 3: Good reflex adding tests!

components/common/CharmEditor/components/CryptoPrice.tsx
```diff
@@ -189,7 +189,7 @@ export function CryptoPrice ({ preset, onQuoteCurrencyChange, onBaseCurrencyChan
           )}
 
           <Typography component='div' align='center' sx={{ fontSize: 36, lineHeight: 1, mt: 2 }}>
-            {loading === false && !error && formatMoney(lastQuote.amount, quoteCurrency)}
+            {loading === false && !error && formatMoney(lastQuote.amount, quoteCurrency, window.navigator.language)}
```
motechFR, line 192: Good one moving this here, that way the lib method is reusable between server and client

lib/utilities/formatting.ts
```diff
@@ -1,14 +1,14 @@
...
-  const userLocale = window.navigator.language;
+export function formatMoney (amount: number, currency: FiatCurrency, userLocale: string): string {
 
   const formatter = new Intl.NumberFormat(userLocale, {
     style: 'currency',
-    currency
+    currency,
+    ...(amount < 1 && {
+      minimumFractionDigits: 4
+    })
```
motechFR, line 10: **Should**

Inline destructuring assertion + conditional adds extra effort to see what's going on.

We do use inline destructuring, but often only for copying an object and updating some keys.

Also just realised this will break on negatives.

How about below? (Would also ensure the 2 digits are always present for .1 => .10 or int => int.00 situations


```typescript
const minimumFractionDigits = (amount < 1 && amount > -1 ) ? 4 : 2;

Intl.NumberFormat(userLocale, {
    style: 'currency',
    currency,
    minimumFractionDigits
})
```

motechFR (changes requested):
Hey @AndreiMitrea looks good. Just added some details of a change I think is needed on the lib method

components/common/CharmEditor/components/CryptoPrice.tsx
mattcasey: remember the client-side is SSR so no matter where you put it, you may need to check window is defined first. it probably just doesn't hit this code  path because of the condition on loading

lib/utilities/formatting.ts
AndreiMitrea: @motechFR Thank you. Is it possible for amount to have negative values?

motechFR: For the use case of crypto price, not really.

But if we want to display negative fiat amounts in future, this function will be ready to go!

motechFR:
Looks good!

PR #453, opened 2022-05-07
Get pending transactions from Gnosis Safe + auto-import Gnosis safes

mattcasey:
These are pulled in live:
image

There's still a lot of work to do for UX, but for separate PRs:
- we should group ones with the same nonce, since you can only choose one to execute
- some rows have more than one action. should expand to show the actions
- need to query transactions to find the relevant bounty

And we can get your safes automatically for an address. So there's just a 'sync' button now, and you can add a custom name:
image


components/bounties/[bountyId]/components/BountyPaymentButton.tsx
```diff
@@ -128,7 +128,7 @@ export default function BountyPaymentButton ({
         await switchActiveNetwork(chainToUse.chainId);
       }
 
-      const signer = await library.getSigner(account);
+      const signer = library.getSigner(account);
```
Comment (motechFR) [inline components/bounties/[bountyId]/components/BountyPaymentButton.tsx:131]: Good spot!

components/profile/integrations/MultiSigWallets.tsx
```diff
@@ -38,63 +42,81 @@ const gnosisUrl = (address: string) => `https://gnosis-safe.io/app/rin:${address
...
+      setIsLoadingSafes(true);
+      try {
+        const safes = await getSafesForAddresses(gnosisSigner, user.addresses);
+        const safesData = safes.map(safe => ({
+          address: safe.address,
+          chainId: safe.chainId,
+          name: getWalletName(safe.address) // get existing name if user gave us one
+        }));
+        await charmClient.setUserMultiSigs(safesData);
+        await mutate('/profile/multi-sigs');
```
motechFR, line 62: It seems SWR provides us with a mutate function from the hook itself bound to the same key as per this doc on [bound mutate](https://swr.vercel.app/docs/mutation#bound-mutate)

So we could call mutate as below without the key, which removes risk of typos in passing the path

```typescript
const {data, mutate} = useSWR()
// Some code
mutate()
```

components/profile/integrations/MultiSigWallets.tsx
```diff
@@ -38,63 +42,81 @@ const gnosisUrl = (address: string) => `https://gnosis-safe.io/app/rin:${address
...
+  function getWalletName (address: string) {
+    return data?.find(wallet => wallet.address === address)?.name;
   }
 
   if (!data) {
     return null;
   }
 
+  // sort the rows to prevent random order
+  const sortedRows = data.sort((a, b) => a.address < b.address ? -1 : 1);
```
motechFR, line 79: Stylistic suggestion, I think it's a good team convention to always have clearer notations for ordering and filtering functions, so this could be

```typescript
const sortedRows = data.sort((walletA, walletB) => walletA.address < walletB.address ? -1 : 1);
```

components/profile/tasks/TasksList.tsx
```diff
@@ -63,20 +37,16 @@ export default function TasksList () {
...
-                ))}
+                {task.nonce} - {task.description} <br />
+                Gnosis <Link target='_blank' href={task.gnosisUrl} external><OpenInNewIcon /></Link>
               </TableCell>
-              <TableCell><Chip label={task.type} variant='outlined' /></TableCell>
-              <TableCell>{task.workspace}</TableCell>
-              <TableCell><Chip label='Sign' variant='outlined' onClick={handleSign} /></TableCell>
+              <TableCell><Chip label='Multisig' variant='outlined' /></TableCell>
+              <TableCell>CharmVerse</TableCell>
+              <TableCell><Chip component={Link} label='Sign' variant='outlined' href={task.gnosisUrl} external /></TableCell>
```
motechFR, line 49: I think a target=_blank would be handy here, as otherwise I get kicked out of Charmverse and into the Gnosis app

motechFR:
Documenting the flow here for myself and others:

- useGnosisTasks
_Gets a list of transactions to sign from Gnosis based on the active account and available gnosis wallets_ 

-- getPendingTasks
_Generates getter request based on the account and list of wallets_

--- getTransactionsforSafe
_Instantiantes Gnosis service instance based on account and chain ID of gnosis wallet (optional)_

---- getGnosis service
_Wrapper around Gnosis service instantiation to configure gnosis RPC Url and ethers adapter_

lib/gnosis/gnosis.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+async function getTransactionsforSafe (signer: Signer, wallet: UserMultiSigWallet): Promise<GnosisTransaction[]> {
+  const service = getGnosisService({ signer, chainId: wallet.chainId });
+  if (service) {
+    // const transactions = await service.getMultisigTransactions(wallet.address);
+    // return transactions.results;
+    const transactions = await service.getPendingTransactions(wallet.address);
+    console.log(transactions);
+    // TODO: not sure if we actually need to get transaction info again
```
motechFR, line 78: Just ran a test, I think we don't need to get the transactions again. The shape of objects returned in both cases seems to be the same


image

image

pages/api/profile/multi-sigs/index.ts
```diff
@@ -24,22 +23,27 @@ async function list (req: NextApiRequest, res: NextApiResponse<UserMultiSigWalle
...
+async function setWallets (req: NextApiRequest, res: NextApiResponse<UserMultiSigWallet>) {
+
+  const walletsInput = req.body as (Pick<UserMultiSigWallet, 'address' | 'chainId' | 'name'>)[];
+  const walletsData = walletsInput.map(wallet => ({
+    ...wallet,
+    userId: req.session.user.id,
+    walletType: 'gnosis' as const
+  }));
+
+  await prisma.$transaction([
```
motechFR, line 35: I wasn't aware of this pattern, but that's really good to know!

Definitely worth sharing in our next team standup. It replaces Promise.all and also avoids creating an inconsistent database if the set of operations only succeeds half-way.

motechFR (approved):

components/bounties/[bountyId]/components/BountyPaymentButton.tsx
Comment (mattcasey) [reply components/bounties/[bountyId]/components/BountyPaymentButton.tsx:131]: i spent a bunch of time wrapping it in an async hook too lol

components/common/LoadingComponent.tsx
```diff
@@ -1,5 +1,5 @@
 
-import React, { ReactNode } from 'react';
+import React from 'react';
```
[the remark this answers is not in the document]
mattcasey: Default to importing what you need. There was other code that is using React which I didn't change, I was replacing ReactNode here with JSX.Element

components/profile/integrations/MultiSigWallets.tsx
mattcasey: that is better, thanks

components/profile/integrations/MultiSigWallets.tsx
mattcasey: i'm not sure what is easier to read in your example

components/common/LoadingComponent.tsx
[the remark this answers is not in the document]
motechFR: Understood, makes sense as a principle

components/profile/integrations/MultiSigWallets.tsx
motechFR: Just the part where we say walletA , walletB instead of a, b within the sorting function

components/profile/tasks/TasksList.tsx
mattcasey: agreed

lib/gnosis/gnosis.ts
mattcasey: thanks.. i figure i can leave 'em out too. removed

components/profile/integrations/MultiSigWallets.tsx
mattcasey: I personally find this enough:
```
const walletsSorted = wallets.sort((a, b) => a.address > b.address : -1 : 1);
```

You already read 'wallets' once. (in my example it was more ambiguous, so i gave the array a name). When I'm inside the sort logic, i actually don't care what it is, just two objects with 'address' attached. It should be obvious to the reader they are both wallets, and not one a wallet, the other.. a user or something.  It also helps keep the line length shorter too. But I'm fine with both really and don't think we need a hard rule.
Similar examples:
```
const truthyWallets = wallets.filter(a => Boolean(a.address));
const mappedWallets = wallets.map(a => a.address);
```
I think I picked this up from Scala, it's normal to see a _ completely replace the variable name when its not necessary:
```
const mappedWallets = wallets.map(_ => _.address)
```
It's all on one line, if the method was multiple lines i would probably expect better names 
so i dont feel strongly either way, but open to seeing if others want to make a rule about it

PR #454, opened 2022-05-08
Extended markdown export

motechFR:
Added support for exporting
- Tables
- Nested pages

Added initial code for exporting columns, but further research needed


components/common/CharmEditor/components/nestedPage/nestedPage.ts
```diff
@@ -13,6 +14,67 @@ export interface NestedPagePluginState {
...
+ * Returns markdown content with page data interpolated
+ * @param convertedToMarkdown
+ */
+export async function replaceNestedPages (convertedToMarkdown: string): Promise<string> {
+
+  const nestedPageMarkers = parseNestedPagesToReplace(convertedToMarkdown);
+
+  const isServer = typeof window === 'undefined';
+  // Dynamic import allows this function to be loaded in the client-side without triggering a server-side import
+  const linkGetter = isServer ? ((await import('lib/pages/server/generatePageLink')).generatePageLink) : charmClient.getPageLink;
```
Devorein, line 53: I don't think we need to create a separate endpoint for getting the page link. We can embed that information inside the dom as attributes and extract those attributes while exporting. For example inside `NestedPage.tsx` we can change this line

```jsx
<NestedPageContainer>
```

to this

```jsx
<NestedPageContainer id={`page-${nestedPage?.id}`} data-title={nestedPage?.title} data-path={nestedPage?.path}>
```

and use something like this when generating the link of the page

```ts
const linkGetter = isServer ? ((await import('lib/pages/server/generatePageLink')).generatePageLink) : (pageId: string) => {
  const documentNode = document.querySelector(`#page-${pageId}`) as HTMLDivElement;
  const info: PageLink = {
    title: '',
    url: ''
  };
  if (documentNode) {
    info.title = documentNode.getAttribute('data-title') as string;
    // It should be part of the same space domain, so no need to remove that, only remove the page path
    const urlWithSpaceDomain = window.location.href.split('/').slice(0, 4).join('/');
    info.url = `${urlWithSpaceDomain}/${documentNode.getAttribute('data-path')}`;
  }
  return info;
};
```

pages/api/pages/[id]/link.ts
```diff
@@ -0,0 +1,48 @@
+
+import { prisma } from 'db';
```
Comment (Devorein) [inline pages/api/pages/[id]/link.ts:2]: I think we can keep it, but we have removed its usage as per the above review. It's best if we reduce the number of API calls if there is an alternate

pages/api/pages/[id]/link.ts
```diff
@@ -0,0 +1,48 @@
...
+    userId: req.session.user.id,
+    spaceId: page.spaceId as string,
+    adminOnly: false
+  });
+
+  if (error) {
+    throw error;
+  }
+
+  const pageLink = await generatePageLink(pageId);
```
Comment (Devorein) [inline pages/api/pages/[id]/link.ts:42]: Note that we are already querying the page inside this function, the above query would've already fetched all the information required to generate the link.

pages/api/v1/pages/[pageId].ts
```diff
@@ -155,13 +156,13 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
     where: {
       id: card.id
-    },
-    select: {
-      content: true
     }
   });
 
   const cardToReturn = new PageFromBlock(updatedPage, boardSchema, cardPage?.content as PageContent);
+  if (cardPage) {
```
Comment (Devorein) [inline pages/api/v1/pages/[pageId].ts:163]: We are exporting to markdown in a lot of places, if you search by `PageFromBlock` in vsc we should be updating the markdown in all of those places.

Devorein (changes requested):

PR #462, opened 2022-05-10
Updated UI for Gnosis tasks

mattcasey:
- added 'owners' and 'threshold' to the gnosis safes collection
- renamed "userMultiSigWallet" table to "userGnosisSafe". This felt like an early abstraction, I don't even know of any other multi-sig wallets. Also, semantically your "safe" can be accessed through many "wallets" such as metamask, according to Gnosis Safe's home page. So a safe != wallet.
- group tasks by safe, then by nonce
- moved API to get tasks to the back-end because we have a bunch of data to join on in addition to what we get from Gnosis's API (users, bounties, safe address metadata..) and I can see us wanting to re-use a lot of it for the email notifications
- allow user to connect safes from the Tasks page

Still to do in another PR:
- join data with bounties and addresses with users in our system
- tests on backend logic

image


components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -0,0 +1,164 @@
...
+        <GridColumn sx={{ justifyContent: 'flex-end' }}>
+          <Chip
+            clickable
+            component='a'
+            label={transaction.myAction || 'Waiting for others'}
+            href={transaction.myActionUrl}
+            target='_blank'
+            color={transaction.myAction ? 'primary' : undefined}
+            variant={transaction.myAction ? 'filled' : 'outlined'}
+          />
```
motechFR, line 56: Since we need to execute the lowest nonce transaction at all times, could we disable the actions for greater nonces?

pages/api/profile/gnosis-safes/index.ts
```diff
@@ -13,32 +13,32 @@ handler
...
     where: {
       userId: req.session.user.id
     }
   });
   return res.status(200).json(wallets);
 }
 
-async function setWallets (req: NextApiRequest, res: NextApiResponse<UserMultiSigWallet>) {
+async function setWallets (req: NextApiRequest, res: NextApiResponse<UserGnosisSafe>) {
+
```
motechFR, line 27: I think it would be good to normalise the GET PUT, DELETE and POST methods to
listSafes

updateSafe

deleteSafe

setSafes

motechFR (approved):

components/bounties/components/MultiPaymentButton.tsx
```diff
@@ -24,7 +24,7 @@ export default function MultiPaymentButton ({ chainId, safeAddress, transactions
   const [safe] = useGnosisSafes([safeAddress]);
   const network = getChainById(chainId);
   if (!network?.gnosisUrl) {
-    throw new Error(`Invalid network: ${chainId}`);
+    throw new Error(`Unsupported Gnosis network: ${chainId}`);
```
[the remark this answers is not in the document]
mattcasey: this is not in the back-end, so i dont thin ksystem errors are as helpful

components/profile/integrations/GnosisSafes.tsx
```diff
@@ -148,30 +145,30 @@ function WalletRow ({ wallet, updateWallets }: { wallet: Wallet, updateWallets:
...
-  async function deleteWallet (_wallet: Wallet) {
-    await charmClient.deleteUserMultiSig(_wallet.id);
+  async function deleteSafe (_safe: Safe) {
+    await charmClient.deleteMyGnosisSafe(_safe.id);
     updateWallets();
     deleteConfirmation.close();
   }
 
-  async function saveWalletName ({ name }: { name: string }) {
+  async function safeSafeName ({ name }: { name: string }) {
```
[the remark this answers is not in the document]
mattcasey: yes thanks!

components/profile/tasks/GnosisTasksList.tsx
mattcasey: we could and should.. i'll save it for when we are actually executing transactions. i was thinking we also might add a line similar to Gnosis to say "these tasks cant be executed now..."

pages/api/profile/gnosis-safes/index.ts
mattcasey: agreed and fixed

PR #463, opened 2022-05-11
Add workspace custom image

AndreiMitrea:
Added functionality to add workspace avatar image.


components/bounties/[bountyId]/BountyDetails.tsx
```diff
@@ -14,7 +14,7 @@ import { Application, Bounty } from '@prisma/client';
 import { usePopupState } from 'material-ui-popup-state/hooks';
 import charmClient from 'charmClient';
 import ScrollableWindow from 'components/common/PageLayout/components/ScrollableWindow';
-import Avatar from 'components/common/Avatar';
+import { Avatar } from 'components/common/Avatar';
```
Comment (mattcasey) [inline components/bounties/[bountyId]/BountyDetails.tsx:17]: we don't have a standard pattern really to group similar components (I just ran into this creating the common/UserDisplay component) but i think we should keep the normal case as a default export still. Thoughts?

components/common/Avatar/AvatarWithIcons.tsx
```diff
@@ -0,0 +1,41 @@
...
+    display: inline-block;
+    position: relative;
+`;
+
+export type AvatarWithIconsProps = InitialAvatarProps & {
+    icons?: ReactNode;
+    isHovered: boolean;
+};
+
+export function AvatarWithIcons (props: AvatarWithIconsProps) {
```
mattcasey, line 32: component files should always export the component as the default export

prisma/migrations/20220511204834_added_space_image/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
```
mattcasey, line 1: i re-created this migration since I deployed a migration to main earlier today and it was out of order

components/common/CreateSpaceForm.tsx
```diff
@@ -106,7 +108,16 @@ export default function WorkspaceSettings ({ defaultValues, onSubmit: _onSubmit,
...
       <Grid container direction='column' spacing={2}>
         <Grid item display='flex' justifyContent='center'>
-          <Avatar name={watchName} variant='rounded' />
+          <Avatar
+            name={watchName}
+            variant='rounded'
+            spaceImage={watchSpaceImage}
+            updateImage={(url) => setValue('spaceImage', url, { shouldDirty: true })}
+          />
+          <TextField
```
mattcasey, line 117: I tested it without the TextField and it seems to work fine

components/common/Avatar/AvatarWithIcons.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const { icons, isHovered } = props;
+
+  return (
+    <StyledBox>
+      { isHovered && <StyledStack direction='row' spacing={2} p={0.5}>{ icons }</StyledStack> }
```
mattcasey, line 37: can you just use CSS for this?

components/settings/workspace/LargeAvatar.tsx
```diff
@@ -1,18 +1,91 @@
...
+      onMouseEnter={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
+    >
+      <input
+        type='file'
+        hidden
+        accept='image/*'
+        ref={inputFile}
+        onChange={async (e) => {
+          const firstFile = e.target.files?.[0];
```
mattcasey, line 55: could we include some sane limit on image size? maybe 1MB limit. i think safwan implemented something like this already

mattcasey: hm you can ignore this comment for now. we could use the 'sharp' module to resize images on the backend, might be a better fix anyway

components/bounties/[bountyId]/BountyDetails.tsx
Comment (AndreiMitrea) [reply components/bounties/[bountyId]/BountyDetails.tsx:19]: Moving both AvatarWithIcons  and Avatar directly in the common folder sounds good. Will do that. Yes, I agree that exporting all components from a barrel file would be good.

components/common/Avatar/AvatarWithIcons.tsx
AndreiMitrea: You mean for spacing, padding, potentialy the direction? I personally prefer to use the component props as much as possible, I find it easier to maintain consistency across the app.

components/common/CreateSpaceForm.tsx
AndreiMitrea: Removed, thanks!

components/common/Avatar/AvatarWithIcons.tsx
AndreiMitrea: I misunderstood. Replaced he js hover logic with CSS. Thanks!

mattcasey (approved):

mattcasey:
Approved.. and I had to update the migration one more time :) so your local db needs to be cleared out again

PR #464, opened 2022-05-11
Guild.xyz integration

Devorein:
1. Button for importing roles from guild.xyz, its mixed with the button for importing roles from discord to save space. The current import role source is saved in localstorage
image
2. If the current user doesn't have any wallet address associated with their account (doesn't matter if the wallet is currently connected or not), we show all the guilds in the import modal, otherwise we show only the guilds that the current user is part of
image
	2.1 Users can select the guilds that they want to import roles from
	2.2 We show the number of roles present in that guild, thought about showing the member counts, but its not returned when fetching a guild by its id.
	2.3 We're using react-virtualized to make the UX as smooth and performant as possible
	2.4 Clicking on the guild name would navigate the user to the guild page
3. When importing roles from guild, we keep track of the corresponding guild.xyz role
4. Nobody can rename, assign/unassign members to a role imported from guild.xyz only the admin can delete it. This validation is done both on client-side and server-side. A small tooltip is shown to make the user aware of the fact
image
5. Automatically create and assign roles to all members in the workspace (those who have the same guild.xyz role, we do this checking via the `sourceId` column on `role` table, rather than naive name equality) when importing roles
6. Any user who login will have their roles updated based on the imported guild.xyz roles, if they have access to a new role they will be assigned that in charmverse, if they lost a role due to requirement issue they will be removed off it in charmverse


lib/guild-xyz/assignGuildRolesForUser.ts
```diff
@@ -0,0 +1,118 @@
+import { prisma } from 'db';
+import { user } from '@guildxyz/sdk';
+import { SpaceRole } from '@prisma/client';
+
+export async function assignGuildRolesForUser (userId: string, firstAddress?: string, _spaceRoles?: SpaceRole[]) {
```
mattcasey, line 5: use object notation for arguments: function ({ userId, address, spaceRoles }). I think 1-2 args is fine for a list but anything else and now you have to start passing in undefiends, etc. to pass in just spaceRoles

mattcasey: we're only using this method in one place. There shouldn't be an optional way to call it. Just take a userId, or take the addresses.

prisma/schema.prisma
```diff
@@ -297,6 +297,8 @@ model Role {
   space            Space             @relation(fields: [spaceId], references: [id], onDelete: Cascade)
   spaceRolesToRole SpaceRoleToRole[]
   permissions      PagePermission[]
+  source           String?
+  sourceRoleId     String?
```
mattcasey, line 301: 'role' is redundant in this name. 'sourceId' woulda been fine too

lib/guild-xyz/assignGuildRolesForSpace.ts
```diff
@@ -0,0 +1,82 @@
...
+  rolesImportedWithGuild.forEach(roleImportedWithGuild => {
+    if (roleImportedWithGuild.sourceRoleId) {
+      guildRoleCharmverseRoleRecord[roleImportedWithGuild.sourceRoleId] = roleImportedWithGuild.id;
+    }
+  });
+
+  const workspaceRoleIdsSet = new Set(rolesImportedWithGuild.map(roleImportedWithGuild => roleImportedWithGuild.sourceRoleId as string));
+
+  for (const spaceRole of spaceRoles) {
+    const charmverseUser = await prisma.user.findUnique({
```
mattcasey, line 39: you should select the 'user' when querying spaceRoles so you dont have to query it here.

lib/guild-xyz/assignGuildRolesForSpace.ts
```diff
@@ -0,0 +1,82 @@
...
+      const memberships = await user.getMemberships(charmverseUser.addresses[0]);
+      const roleIdsSet: Set<string> = new Set();
+      memberships?.forEach(membership => {
+        membership.roleids.forEach(roleid => roleIdsSet.add(String(roleid)));
+      });
+
+      // Filter the roles that are part of the workspace only
+      const roleIdsInWorkspace = Array.from(roleIdsSet).filter(roleId => workspaceRoleIdsSet.has(roleId));
+      for (const roleIdInWorkspace of roleIdsInWorkspace) {
+        const roleId = guildRoleCharmverseRoleRecord[roleIdInWorkspace];
```
mattcasey, line 55: There's a lot of variables for me while trying to follow this code. I think less is more... you could get rid of lines 29-34 `guildRoleCharmverseRoleRecord`  and just replace this line with:
```
const roleId = rolesImportedWithGuild.find(role => role.sourceId === roleIdInWorkspace);
```

lib/guild-xyz/assignGuildRolesForSpace.ts
```diff
@@ -0,0 +1,82 @@
...
+  const workspaceRoleIdsSet = new Set(rolesImportedWithGuild.map(roleImportedWithGuild => roleImportedWithGuild.sourceRoleId as string));
+
+  for (const spaceRole of spaceRoles) {
+    const charmverseUser = await prisma.user.findUnique({
+      where: {
+        id: spaceRole.userId
+      }
+    });
+    // If the user has a wallet address
+    if (charmverseUser?.addresses?.[0]) {
```
mattcasey, line 45: we should be mapping over all the addresses here... maybe it could be something like:
```
const memberships = await Promise.all(charmVerseUser.addresses.map(address => user.getMemberships(address)))
const rolesToAssign = memberships.map(guildRole => rolesInWorksapce.find(role => role.sourceRoleId === guildRole.id).filter(isTruthy)
await prisma.$transaction([
   prisma.delete({ where: { userId, source: 'guild.xyz', spaceId }}) // delete exsiting guild roles
   ...rolesToAssign.map(...)
]);
```

lib/guild-xyz/assignGuildRolesForUser.ts
```diff
@@ -0,0 +1,118 @@
...
+      const guildRolesUserLostAccess = Array.from(
+        workspaceGuildRoleIdsSet
+      ).filter(
+        workspaceGuildRoleId => !guildUserRoleIds.includes(workspaceGuildRoleId)
+      );
+
+      for (const guildRoleId of guildUserRoleIds) {
+        const roleId = guildRoleCharmverseRoleRecord[guildRoleId];
+        if (roleId) {
+          await prisma.spaceRoleToRole.upsert({
```
mattcasey, line 67: it's only copy-pasted twice but seems like some of this code could be shared betweenthe two functions

lib/guild-xyz/assignGuildRolesForUser.ts
```diff
@@ -0,0 +1,118 @@
+import { prisma } from 'db';
+import { user } from '@guildxyz/sdk';
+import { SpaceRole } from '@prisma/client';
+
+export async function assignGuildRolesForUser (userId: string, firstAddress?: string, _spaceRoles?: SpaceRole[]) {
+  firstAddress = firstAddress ?? (await prisma.user.findUnique({
```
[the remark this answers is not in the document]
Devorein: Ow, it's just because we don't have multi-address for a user at the moment, but since it's a potential feature I think you're right, I will make the update

lib/guild-xyz/assignGuildRolesForUser.ts
Devorein: Yeah I agree, I thought I might need it in some places where I only pass the `userId` but to save two queries one for getting the address of the user and the other for the getting spaceRoles, I decided to pass them as optional arguments, I will update it to take both of those as required arguments

prisma/schema.prisma
Devorein: Yeah I thought of changing it to `sourceId` but wanted to keep `role` as what if we need to store the guild id as well. It would be great to have two clear columns `sourceRoleId` and `sourceGuildId` then. And making the `source` an enum is a great idea. I will do that

lib/guild-xyz/assignGuildRolesForSpace.ts
```diff
@@ -0,0 +1,82 @@
...
+      const memberships = await user.getMemberships(charmverseUser.addresses[0]);
+      const roleIdsSet: Set<string> = new Set();
+      memberships?.forEach(membership => {
+        membership.roleids.forEach(roleid => roleIdsSet.add(String(roleid)));
+      });
+
+      // Filter the roles that are part of the workspace only
+      const roleIdsInWorkspace = Array.from(roleIdsSet).filter(roleId => workspaceRoleIdsSet.has(roleId));
+      for (const roleIdInWorkspace of roleIdsInWorkspace) {
```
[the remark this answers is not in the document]
Devorein: Good catch. I've also added a try/catch block with appropriate log to capture failed role assignment for a user

lib/guild-xyz/assignGuildRolesForSpace.ts
Devorein: Ah. 🤦🏼 Thank you. I completely forgot about that. Updated in the latest commit

lib/guild-xyz/assignGuildRolesForSpace.ts
Devorein: Hmm, I'm not sure that as for a workspace with a lot of users, we could potentially make the process a lot faster using a simple record. I have however removed a lot of "unnecessary" code by removing a few of the variables.

lib/guild-xyz/assignGuildRolesForSpace.ts
Devorein: Yes that does make sense. I've updated it to make use of all the addresses of a user and used the transaction API to remove roles for users who lost access.

lib/guild-xyz/assignGuildRolesForUser.ts
Devorein: You are right. I've updated the code to keep it DRY.

mattcasey (approved):

PR #467, opened 2022-05-11
Better @ mention UX

Devorein:
1. Filter out the archived pages for `/link to page` and `@mention`
2. Replace inconsistent `@mention` UI with that of `/link to page`
3. Up/Down arrow to navigate between items
4. Press enter to select item


components/common/CharmEditor/components/nestedPage/components/NestedPagesList.tsx
```diff
@@ -37,43 +33,11 @@ function NestedPagesList () {
   return (
     <PopoverMenu container={tooltipContentDOM} isOpen={isVisible} onClose={onClose} width={460}>
       <GroupLabel>Select a page</GroupLabel>
-      {isVisible && <PagesList onSelectPage={onSelectPage} />}
+      {isVisible && (
+      <PagesList onSelectPage={onSelectPage} />
```
motechFR, line 37: Nice one Safwan, this is a good example of cleanup

motechFR (approved):
Hey @Devorein this PR works well. 

One issue I have, is the lack of filtering as I type out.

Could you double check with team to see if we can get this into the PR?

Otherwise, for UX-only aspect looks good

components/common/CharmEditor/components/nestedPage/components/NestedPagesList.tsx
Devorein: Thank you. We were using pretty much the same component in two places so made the decision to refactor

Devorein:
I think you are right. I will try to get the text search feature on this pr. They are both related to good UX

components/common/CharmEditor/components/inlinePalette/useEditorItems.tsx
```diff
@@ -632,12 +632,18 @@ export function useEditorItems () {
           description: 'Link to a new page',
           editorExecuteCommand: (() => {
             return (async (state, dispatch, view) => {
-              renderSuggestionsTooltip(NestedPagePluginKey, { component: 'nestedPage', threadIds: [] })(state, dispatch, view);
-              return replaceSuggestionMarkWith(palettePluginKey, '')(
-                state,
-                dispatch,
-                view
-              );
+              if (nestedPagePluginKey) {
```
mattcasey, line 635: This should be required, it doesn't look like we ever call useEditorItems withought it. I would prefer we actually just import it like we had before, since it's a static variable and not dynamic

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -59,6 +58,8 @@ export interface ICharmEditorOutput {
 const emojiSuggestPluginKey = new PluginKey('emojiSuggest');
 const mentionSuggestPluginKey = new PluginKey('mentionSuggest');
 const floatingMenuPluginKey = new PluginKey('floatingMenu');
+const nestedPagePluginKey = new PluginKey('nestedPage');
```
mattcasey, line 61: Since keys are being generated statically, I think they should just live next to the plugin/component. It doesn't make sense to pass them around as props unless we were builidng a 3rd party lib like bangle.dev

mattcasey (changes requested):
the page search should appear below my text, not above it
image

components/common/BoardEditor/focalboard/src/components/table/tableRow.tsx
```diff
@@ -48,7 +51,7 @@ function TableRow (props: Props) {
     const isManualSort = activeView.fields.sortOptions.length === 0
     const isGrouped = Boolean(activeView.fields.groupById)
     const [isDragging, isOver, cardRef] = useSortable('card', card, !props.readonly && (isManualSort || isGrouped), props.onDrop)
-
+    const {pages} = usePages()
```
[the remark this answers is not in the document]
Devorein: Yes, thank you. I've removed it, now we are passing a prop for the content

components/common/CharmEditor/components/inlinePalette/useEditorItems.tsx
Devorein: The plugin keys had to be created for each charmeditor and they have to be unique, I had to go this route when writing the inline charm editor as each editor instance had to have their own plugins. I just updated the `nestedPage` and `inlineComment` plugins to be separate for each charmeditor as there could be cases where we might have more than one charmeditor rendered on the screen.

components/common/CharmEditor/CharmEditor.tsx
Devorein: The same response as above

mattcasey (approved):

PR #468, opened 2022-05-12
Default stablecoin

motechFR:
All spaces are created with default USDC for ETH mainnet, Rinkeby and Polygon mainnet
image


motechFR:
@mattcasey I'd like your input

I added a unique constraint for the [spaceId, contractAddress, chainId] combo to avoid duplicate payment methods.

This then allows upsert behaviour, prevents us accidentally adding duplicates if we rerun the script, and also adds UI protections.

I've added some better error handling in API and form to deal with this
image


I'm not sure how this might affect Gnosis Safe, so wanted to highlight this to you.

mattcasey:
looks good, one comment is to remove '.server' from the file name

motechFR:
Thanks @mattcasey , done the renaming.

PR #471, opened 2022-05-12
Bounty UI fixes

motechFR:
- Hide applications + reviewer + assignee when bounty is in 'suggestion' status

- If amount is 0, open the edit box and highlight amount

- Change label to "save"

- Move "approve suggestion" to after the description

- Vertical align "reward not set" as well as whole row

- Add "suggested by" field

image

### DIFF
--- a/components/bounties/[bountyId]/BountyDetails.tsx
+++ b/components/bounties/[bountyId]/BountyDetails.tsx
@@ -33,6 +33,7 @@ import { useRouter } from 'next/router';
 import { useEffect, useMemo, useState } from 'react';
 import useIsAdmin from 'hooks/useIsAdmin';
 import BountyModal from '../components/BountyModal';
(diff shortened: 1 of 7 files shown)


Devorein:
Everything looks great. But not sure what you meant by your second point. I still see the label being "Edit Bounty" 🤔 
image

Devorein:
Also for suggest_by field, we don't seem to be showing that at any place? Shouldn't the admin be able to see who suggested it?

Devorein (approved):

motechFR:
> Also for suggest_by field, we don't seem to be showing that at any place? Shouldn't the admin be able to see who suggested it?

Thanks for feedback Safwan, I'll integrate this into the updated bounties v2

motechFR:
> Everything looks great. But not sure what you meant by your second point. I still see the label being "Edit Bounty" thinking image

For now, I left it as is, we can come back to this later

PR #472, opened 2022-05-12
Show a visible box when the permission is hovered

motechFR:
This PR is in response to an issue reported by a user who thought they couldn't change their permissions.

It displays a box when a permission level is hovered.

highlight-permission


AndreiMitrea (approved):
Code review only.

PR #476, opened 2022-05-14
add joined date to contributors

mattcasey:
Small PR to show the date someone joined. I was having trouble telling who was who based on wallet addreess, and I think this context is helpful. 

image


components/settings/contributors/ContributorList.tsx
```diff
@@ -42,18 +43,28 @@ export default function ContributorList ({ isAdmin, spaceId, spaceOwner }: Props
...
-          />
-        )
-
-      ))}
+      <Table>
+        <TableHead>
+          <TableRow>
+            <TableCell></TableCell>
+            <TableCell>Join date</TableCell>
+            <TableCell></TableCell>
```
motechFR, line 51: Quick feedback, the column headers look a bit weird with just "Join date" in the middle

What about adding a "Username" on the left? (and leave admin / contributor empty)

lib/utilities/dates.ts
```diff
@@ -58,15 +58,22 @@ export function getTimeDifference (
   return timeDifference[`${timeUnit}s`];
 }
 
-function getDaySuffix (dateTime: DateInput): string {
+export function getDaySuffix (dateTime: DateInput): string {
```
motechFR, line 61: Stroke of luck, when testing I had Friday 13th May as join date on local DB.

It came out as 13rd. I've fixed this and added tests

motechFR (approved):

components/settings/contributors/ContributorList.tsx
mattcasey: ya i agree

lib/utilities/dates.ts
mattcasey: thanks

PR #477, opened 2022-05-15
Lit modal fix (temporary)

mattcasey:
The ERC20 token dropdown select is broken according to Lit. They are going to fix it n the next 1-2 weeks by rewriting it. In the meantime, we just want to hide it.

I tried using CSS to hide it, but  there's no way to only target the correct elements.


motechFR (approved):
Approved! Note to team if testing npm package patches,

Make sure to delete .next build folder, run npx patch-package, after which the effects of patch apply

PR #478, opened 2022-05-16
remove etags to force page reloading

mattcasey:
Alex has said a few times that he's had to hard-refresh the page to get updates, using Chrome. This is hard to reproduce but it looks like Next.js uses ETags to tell browsers when to refresh the page. Perhaps the Etags are not always updated properly - by removing them, we force the browser to always load the first page. With Etags, I get a 304 Status code after the first load, without them, it's a 200 each time (re-download the page). I believe this only affects the first request and not our other static assets.

https://stackoverflow.com/questions/69838808/next-js-is-caching-my-server-side-pages-how-to-disable-that

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,6 +1,7 @@
 const path = require('path');
 
 const config = {
+  generateEtags: false,
   poweredByHeader: false,
   webpack5: true,
   experimental: {


motechFR:
hey @mattcasey just adding deploy tag so we can see how this works in staging

Apparently from the stackoverflow link you shared, there tends to be differences with localhost

motechFR:
Hey @mattcasey just checked this out by doing a hard reload on prod and staging.

As expected, we get a 200 on staging (full load), and a 304 (page cached) on prod on the HTML corresponding to target page. Hopefully this should fix the issue!

etags

motechFR (approved):

PR #481, opened 2022-05-17
Snoozing gnosis safe transactions

Devorein:


charmClient.ts
```diff
@@ -135,6 +136,10 @@ class CharmClient {
     return http.PUT<Page>(`/api/pages/${pageOpts.id}`, pageOpts);
   }
 
+  snoozeTransactions (payload: UpdateGnosisSafeState) {
+    return http.PUT('/api/profile/snoozeTransactions', payload);
```
mattcasey, line 140: can you put this under api/profile/gnosis-safes/snoozeTransactions?

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -26,10 +34,25 @@ const GridColumn = styled((props: any) => <Grid item xs {...props} />)`
...
-function TransactionRow ({ transaction }: { transaction: GnosisTransactionPopulated }) {
+type UserWithGnosisSafeState = User & {userGnosisSafeState: UserGnosisSafeState | null}
 
+function TransactionRow (
+  { firstNonce, isSnoozed, snoozedUsers, transaction }:
+  { firstNonce: number, isSnoozed: boolean, snoozedUsers: UserWithGnosisSafeState[], transaction: GnosisTransactionPopulated }
+) {
   const [expanded, setExpanded] = useState(false);
   const isReadyToExecute = transaction.confirmations?.length === transaction.threshold;
+  const filteredSnoozedUsers: UserWithGnosisSafeState[] = [];
```
mattcasey, line 45: Why not put this on `GnosisTransactionPopulated`? I designed it so that we do all the data joining in one place on the backend. I'll need this info for email notifications as well, which won't be importing this React component

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -95,19 +143,44 @@ function TransactionRow ({ transaction }: { transaction: GnosisTransactionPopula
...
+  { isSnoozed, snoozedUsers, address, safeName, safeUrl, tasks }:
+  { isSnoozed: boolean, snoozedUsers: UserWithGnosisSafeState[], address: string, safeName: string | null, safeUrl: string, tasks: GnosisTask[] }
+) {
 
   return (
     <>
-      <Typography color='inherit'>
-        Tasks from safe: <Link href={safeUrl} external target='_blank'>{safeName || shortenHex(address)} <OpenInNewIcon fontSize='small' /></Link>
-      </Typography>
+      <Box display='flex' justifyContent='space-between'>
```
mattcasey, line 153: i think this Box element is unecessary since it has one child only

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -95,19 +143,44 @@ function TransactionRow ({ transaction }: { transaction: GnosisTransactionPopula
...
+          </Link>
+        </Typography>
+      </Box>
       {
         tasks.map((task: GnosisTask) => (
           <Card key={task.nonce} sx={{ my: 2, borderLeft: 0, borderRight: 0 }} variant='outlined'>
             <Box display='flex'>
-              <Box px={3} height={rowHeight} display='flex' alignItems='center'>
-                {task.transactions[0].nonce}
+              <Box px={3} height={rowHeight} display='flex' alignItems='center' gap={1}>
```
mattcasey, line 180: this 'gap' rule doesn't seem to affect the layout?

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -129,13 +208,205 @@ function SafeTasks ({ address, safeName, safeUrl, tasks }: { address: string, sa
...
+    await charmClient.snoozeTransactions({
+      snoozeFor: newSnoozedForDate.toJSDate(),
+      snoozeMessage
+    });
+    mutateTasks();
+    resetState();
+  }
+
+  return (
+    <Box sx={{ float: 'right' }}>
```
mattcasey, line 301: <Box float='right'>. sx properties are much slower than normal props. not a big deal, but if you're gonna use Box you might as well use the built-in props

lib/users/getUser.ts
```diff
@@ -37,7 +37,8 @@ export async function getUserProfile (key: UserIdentifiers, value: string): Prom
         }
       },
       discordUser: true,
-      telegramUser: true
+      telegramUser: true,
+      userGnosisSafeState: true
```
mattcasey, line 41: i dont think this belongs on the user profile. it's hard to maintain that we update the profile everywhere. it could jsut be part of the gnosis tasks query

prisma/schema.prisma
```diff
@@ -115,27 +115,35 @@ model Block {
...
+  applications              Application[]
+  username                  String?
+  avatar                    String?
+  isBot                     Boolean?              @default(false)
+  discordUser               DiscordUser?
+  telegramUser              TelegramUser?
+  comments                  Comment[]
+  threads                   Thread[]
+  gnosisSafes               UserGnosisSafe[]
+  userGnosisSafeState       UserGnosisSafeState?
```
mattcasey, line 139: this could be called "gnosisSafeState"

mattcasey (changes requested):
please put the logic to know who snoozed into gnosis.tasks, or i'll have to rewrite it later. other than that, some style comments that aren't blocking

charmClient.ts
Devorein: I agree I thought of doing that. Let me do that asap.

components/profile/tasks/GnosisTasksList.tsx
Devorein: Thank you for catching that. I initially placed the snooze button there but after removing it I forgot to delete the `Box`

components/profile/tasks/GnosisTasksList.tsx
Devorein: Yep, same mistake as above. I placed the snooze icon beside the nonce as we thought of showing it for each task, but forgot to remove it. Also since we aren't using any special `Typography` specific stuff over where I removed that as well (`Typography` for showing nonce value)

prisma/schema.prisma
Devorein: oops, you're right. I updated the casing but totally forgot to remove the redundant user prefix

components/profile/tasks/GnosisTasksList.tsx
Devorein: I thought we would be duplicating information then since it's not a per-transaction level snoozing. But considering your use case I've moved it inside `GnosisTransactionPopulated`

lib/users/getUser.ts
Devorein: Yeah, we talked about that during the meeting but I wasn't sure how to do so. Maybe a separate endpoint as Mo mentioned that returns the user's gnosis-safe state is different from the one we use for `/profile`, but for now, I've just added `user` field when retrieving the tasks just as you mentioned above.

PR #485, opened 2022-05-18
Dev tools for prosemiror

mattcasey:
icon appears at bottom right of page


motechFR (approved):
Hey @mattcasey looks great!

Just tested, and this tool will be very useful

Works on normal pages
Works on card pages when viewed as a page

mattcasey:
I fixed the dependency to be a production one, merging as part of a separate PR

PR #487, opened 2022-05-19
Persisting last viewed view of a board

Devorein:
1. Storing the currently visible focalboard view in local storage so that the next time we visit the same board it navigates to the last view
2. Only show 3 views on the view header, following how notion does it the rest are shown in a separate button. Clicking on it will show the rest of the views and clicking on that view will navigate to that particular view
3. only a single click is required to view options


components/common/BoardEditor/focalboard/src/components/addViewMenu.tsx
```diff
@@ -186,7 +188,7 @@ function AddViewMenu (props: AddViewProps) {
         startIcon={<AddIcon />}
         color='secondary'
         size='small'
-        sx={{ px: 1.5 }}
+        sx={{ px: 1.5, ...(props.sx ?? {}) }}
```
AndreiMitrea, line 191: I don't think `props` is defined here. I think something like this would work
`
const AddButton = styled(Button)<{ sx?: SxProps }>`
....
`
and you pass `sx` on the component.

components/common/BoardEditor/focalboard/src/components/viewHeader/viewTabs.tsx
```diff
@@ -21,46 +21,79 @@ import ListItemIcon from '@mui/material/ListItemIcon';
...
-function ViewTabs ({ intl, readonly, showView, views }: ViewTabsProps) {
+const SHOWN_VIEWS = 3;
+
+function ViewTabs({ board, activeView, intl, readonly, showView, views }: ViewTabsProps) {
   const router = useRouter()
   const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
   const [currentView, setCurrentView] = useState<BoardView | null>(null);
-  const popupState = usePopupState({ variant: 'popover', popupId: 'rename-view-popup' });
+  const renameViewPopupState = usePopupState({ variant: 'popover', popupId: 'rename-view-popup' });
+  const showViewsPopupState = usePopupState({variant: "popover", popupId: "show-views-popup"});
```
AndreiMitrea, line 48: Weird that the `"` were not automatically replaced with apostrophes. Just a curiosity.

components/common/BoardEditor/focalboard/src/components/addViewMenu.tsx
Devorein: You don't need to explicitly set `{sx?: SxProps}` for that component as the types are automatically inferred by `styled`. But I'm not sure what you mean by **I don't think props is defined here** 😕

AndreiMitrea: Nevermind, I misread the code context.

AndreiMitrea (approved):
Code review only.

PR #489, opened 2022-05-20
small ui fixes

mattcasey:
- revert the width of pages

### DIFF
--- a/theme/@bangle.dev/styles.scss
+++ b/theme/@bangle.dev/styles.scss
@@ -467,7 +467,7 @@ details {
   position: absolute;
   width: 30px;
   height: 24px;
-  margin-top: 4px;
+  margin-top: 5px;
   opacity: 0;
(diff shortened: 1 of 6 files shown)


Devorein (approved):

theme/@bangle.dev/styles.scss
```diff
@@ -467,7 +467,7 @@ details {
   position: absolute;
   width: 30px;
   height: 24px;
-  margin-top: 4px;
+  margin-top: 5px;
```
Devorein, line 470: Our material UI theme has a spacing of `8px`, I think we decided on 4px since it's a 1/2 of that.

mattcasey: this is not for spacing, it's meant to align the handlebar that appears next to the line. i didn't do the math, but visually this makes the dots centered

PR #491, opened 2022-05-20
speed up time to create a card

mattcasey:
I Also sped up creating a page. Deployed to staging for your convenience! https://pr-491-faster-cards.charmverse.co/login

### DIFF
--- a/components/[pageId]/BoardPage/BoardPage.tsx
+++ b/components/[pageId]/BoardPage/BoardPage.tsx
@@ -110,7 +110,10 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   const showCard = useCallback((cardId?: string) => {
     const newPath = generatePath(router.pathname, router.query);
 
-    const query: any = { cardId, viewId: router.query.viewId };
+    const query: any = { viewId: router.query.viewId };
+    if (cardId) {
+      query.cardId = cardId;
+    }
 
     if (readonly) {
       query.r = Utils.getReadToken();
--- a/components/[pageId]/BoardPage/BoardPage.tsx
+++ b/components/[pageId]/BoardPage/BoardPage.tsx
@@ -149,7 +149,7 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
           views={boardViews}
           showShared={clientConfig?.enablePublicSharedBoards || false}
         />
-        {typeof shownCardId === 'string' && shownCardId.length !== 0 && shownCardId !== 'undefined' && (
+        {typeof shownCardId === 'string' && shownCardId.length !== 0 && (
           <RootPortal>
             <CardDialog
               board={board}
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
@@ -5,12 +5,12 @@ import { Box } from '@mui/system'
 import PageBanner, { randomBannerImage } from 'components/[pageId]/DocumentPage/components/PageBanner'
 import PageDeleteBanner from 'components/[pageId]/DocumentPage/components/PageDeleteBanner'
 import { useCurrentSpace } from 'hooks/useCurrentSpace'
-import { Page } from 'models'
+import { Page } from '@prisma/client'
 import React, { useState } from 'react'
 import Hotkeys from 'react-hot-keys'
+import { mutate } from 'swr'
 import { injectIntl, IntlShape } from 'react-intl'
 import { connect } from 'react-redux'
-import { mutate } from 'swr'
 import { BlockIcons } from '../blockIcons'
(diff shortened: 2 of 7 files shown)


components/[pageId]/BoardPage/BoardPage.tsx
```diff
@@ -110,7 +110,10 @@ export default function BoardPage ({ page, setPage, readonly }: Props) {
   const showCard = useCallback((cardId?: string) => {
     const newPath = generatePath(router.pathname, router.query);
 
-    const query: any = { cardId, viewId: router.query.viewId };
+    const query: any = { viewId: router.query.viewId };
```
Comment (mattcasey) [inline components/[pageId]/BoardPage/BoardPage.tsx:113]: this doesn't affect anything, but if cardId is undefined, it actually sets cardId=undefined in the URL.

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -157,7 +158,12 @@ function CenterPanel(props: Props) {
         'add card',
         async (block: Block) => {
           if (space) {
-            await mutate(`pages/${space.id}`)
+            await mutate(`pages/${space.id}`, async (pages: Page[]): Promise<Page[]> => {
+              const newPage = await charmClient.getPage(block.id);
+              return [...pages, newPage];
+            }, {
+              revalidate: false
```
mattcasey, line 165: setting 'revalidate = false' allows us to update the cache data w/o calling the api again

hooks/usePages.tsx
```diff
@@ -15,10 +15,13 @@ import { useUser } from './useUser';
 import useIsAdmin from './useIsAdmin';
 
 export type LinkedPage = (Page & {children: LinkedPage[], parent: null | LinkedPage});
+
+type PagesMap = Record<string, Page | undefined>;
```
mattcasey, line 19: some cleanup on types that i ended up not using

pages/api/pages/[id]/index.ts
```diff
@@ -29,27 +63,26 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
+    throw new ActionNotPermittedError('You do not have permission to reposition this page');
   }
 
-  // eslint-disable-next-line
-  if (updateContent.isPublic != undefined && permissions.edit_isPublic !== true) {
-    return res.status(401).json({
-      error: 'You cannot update the public status of this page'
-    });
+  if (updateContent.hasOwnProperty('isPublic') && permissions.edit_isPublic !== true) {
+    throw new ActionNotPermittedError('You do not have permission to update the public status of this page');
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:70]: @motechFR please check i'm diong this right, not sure if we just hadn't gotten around to updating these responses or there's a reason we were still returning them manually.

pages/api/pages/[id]/index.ts
```diff
@@ -29,27 +63,26 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse) {
...
+    throw new ActionNotPermittedError('You do not have permission to update this page');
   }
   const pageWithPermission = await prisma.page.update({
     where: {
       id: pageId
     },
     data: req.body,
     include: {
-      permissions: true
+      permissions: {
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:81]: not sure if this was causing a bug, but i noticed that we are including 'sourcePermissions' when we get all pages, so i added it here too in case we use the response to update the pages cache

components/[pageId]/BoardPage/BoardPage.tsx
Comment (Devorein) [reply components/[pageId]/BoardPage/BoardPage.tsx:113]: Great catch. We can remove the cardId === 'undefined' condition as well

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
Devorein: Nice. That will definitely speed everything up. Previously we were updating the 'cache' using the `setPages` but this is much cleaner and avoids using the `usePages` hook

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:71]: We are using the response from updating the page to update the cache/state so that is the correct approach imo

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:60]: Yes I believe thats the correct error we should be throwing for permission related issues

Devorein (approved):

pages/api/pages/[id]/index.ts
Comment (motechFR) [reply pages/api/pages/[id]/index.ts:60]: Hey @mattcasey looks good, this is indeed the correct way to assert if a permission is available.

That's also the pattern we've been using for throwing API Errors.

pages/api/pages/[id]/index.ts
Comment (motechFR) [reply pages/api/pages/[id]/index.ts:71]: This code path has been updated in PR #500

PR #492, opened 2022-05-20
Profile details section

AndreiMitrea:
Added user details section. The following functionalities have been implemented as part of this PR:
- upload user avatar image
- add user description
- add user social media information: Twitter, LinkedIn, GitHub URLs, Discord user

image

Unit tests will be added as part of: https://app.charmverse.io/charmverse/page-3071344310753725


components/profile/public/components/IdentityModal.tsx
```diff
@@ -0,0 +1,115 @@
...
+    formState: { errors }
+  } = useForm<FormValues>({
+    defaultValues,
+    resolver: yupResolver(schema)
+  });
+
+  function onSubmit (values: FormValues) {
+  }
+
+  const useIntegration = (type: IntegrationType) => {
```
Devorein, line 53: I assume this is WIP

components/profile/public/components/SocialModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+const StyledButton = styled(Button)`
+    border-radius: 7px;
+`;
+
+export const schema = yup.object({
+  twitterURL: yup.string().notRequired().ensure().trim()
+    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?(?:mobile\.)?twitter\.com\/([a-zA-Z0-9_]+)/i, 'Invalid Twitter link'),
+  githubURL: yup.string().notRequired().ensure().trim()
+    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?github\.([a-z])+\/([A-Za-z0-9]{1,})+\/?$/i, 'Invalid GitHub link'),
+  discordUsername: yup.string().notRequired().ensure(),
```
Devorein, line 24: I like that we are validating these URLs but should we add one for discord as well? Every valid discord id must have the # (hash) and 4 digits with it

components/profile/public/components/SocialModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+    save(values as Social);
+  }
+
+  useEffect(() => {
+    reset(social);
+  }, [social]);
+
+  return (
+
+    <Modal open={isOpen} onClose={() => {}} size='large'>
```
Devorein, line 60: Maybe call `close` over there to close the modal when clicked away?

components/profile/public/components/SocialModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+                fullWidth
+                error={!!errors.githubURL}
+                helperText={errors.githubURL?.message}
+                placeholder='https://github.com/charmverse'
+              />
+            </Grid>
+          </Grid>
+          <Grid item container direction='row' alignItems='center'>
+            <Grid item container direction='row' xs={12} md={3}>
+              <SvgIcon viewBox='0 -10 70 70' sx={{ color: '#000000', height: '22px' }}>
```
Devorein, line 96: `fill`: `inherit` might be better as the icons are quite difficult to see in dark mode. But if that doesn't work we might have to use `theme.palette.mode === "dark" ? '#fff' : '#000'`

components/profile/public/components/SocialModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+                fullWidth
+                error={!!errors.twitterURL}
+                helperText={errors.twitterURL?.message}
+                placeholder='https://mobile.twitter.com/charmverse'
+              />
+            </Grid>
+          </Grid>
+          <Grid item container direction='row' alignItems='center'>
+            <Grid item container direction='row' xs={12} md={3}>
+              <GitHubIcon style={{ color: '#000000', height: '22px' }} />
```
Devorein, line 81: Same issue as below

Devorein (changes requested):

components/profile/public/components/SocialModal.tsx
AndreiMitrea: This is no longer needed, made other updates.

Devorein (approved):

prisma/migrations/20220520130853_added_user_details/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- CreateTable
```
mattcasey, line 1: You'll need to re-create this migration file since new migrations have been added to main branch

pages/profile/public.tsx
```diff
@@ -5,7 +5,7 @@ import PublicProfile from 'components/profile/public';
 
 export default function PublicProfilePage () {
 
-  setTitle('Public Profile');
+  setTitle('');
```
mattcasey, line 8: please add this back. we should always have a page title

components/profile/public/components/DescriptionModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+            {...register('description')}
+            fullWidth
+            error={!!errors.description}
+            helperText={errors.description?.message}
+            placeholder='Tell the world a bit more about yourself ...'
+            multiline
+            minRows={10}
+          />
+          <Box justifyContent='end' mt={3} sx={{ display: 'flex' }}>
+            <StyledButton type='submit'>
```
mattcasey, line 67: please use the standard "components/common/Button". if we change border-radius of buttons, it should be across the app

components/profile/public/components/DescriptionModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <TextField
+            {...register('description')}
+            fullWidth
+            error={!!errors.description}
+            helperText={errors.description?.message}
+            placeholder='Tell the world a bit more about yourself ...'
+            multiline
+            minRows={10}
+          />
+          <Box justifyContent='end' mt={3} sx={{ display: 'flex' }}>
```
mattcasey, line 66: with Box you can just write: "display='flex'"

components/profile/public/enums/index.ts
```diff
@@ -0,0 +1 @@
+export * from './IntegrationType';
```
mattcasey, line 1: Please note: we don't really use enums. It's usually better to have an actual array and grab types from that instead:
```
  const INTEGRATION_TYPES = ['Metamask', 'Discord', 'Telegram'] as const;
  type IntegrationType = typeof INTEGRATION_TYPES[number[;
```

components/profile/public/types/Social.ts
```diff
@@ -0,0 +1,7 @@
+export type Social = {
```
mattcasey, line 1: you should just put types and enums in a standanrd 'interfaces.ts' file. i think splitting up types so much would be a hard pattern to follow everywhere

components/profile/public/enums/IntegrationType.ts
```diff
@@ -0,0 +1,5 @@
+export enum IntegrationType {
```
mattcasey, line 1: Please note: we should default to use type unions vs enums. It's usually better to have an actual array and grab types from that instead, so consumers can use it as they want (you might want to loop through the options for example):
```
  const INTEGRATION_TYPES = ['Metamask', 'Discord', 'Telegram'] as const;
  type IntegrationType = typeof INTEGRATION_TYPES[number[;
```

components/profile/public/components/UserDetails.tsx
```diff
@@ -0,0 +1,197 @@
...
+  height: 36px;
+`;
+
+export default function UserDetails () {
+  const { account } = useWeb3React();
+  const [user, setUser] = useUser();
+  const { userDetails, setUserDetails } = useUserDetails();
+  const ENSName = useENSName(account);
+  const [isDiscordUsernameCopied, setIsDiscordUsernameCopied] = useState(false);
+  const [isDescriptionModalOpen, setIsDescriptionModalOpen] = useState(false);
```
mattcasey, line 40: check out some examples in our code using `material-ui-popup-state`. This version is fine, but I've found the library saves me a couple lines and is easier to read.

components/profile/public/components/UserDetails.tsx
```diff
@@ -0,0 +1,197 @@
...
+`;
+
+const StyledDivider = styled(Divider)`
+  height: 36px;
+`;
+
+export default function UserDetails () {
+  const { account } = useWeb3React();
+  const [user, setUser] = useUser();
+  const { userDetails, setUserDetails } = useUserDetails();
```
mattcasey, line 37: i would have just called charmClient here with useSWR:
```
const { data: userDetails, mutate } = useSWR('userDetails', () => charmClient.getUserDetails());
```
You can call mutate() to refresh the data. I usually wouldn't write a hook unless there was more logic around the satte or I needed it elsewhere

components/profile/public/components/UserDetails.tsx
```diff
@@ -0,0 +1,197 @@
...
+import { getDisplayName } from 'lib/users';
+import useENSName from 'hooks/useENSName';
+import charmClient from 'charmClient';
+import { DescriptionModal, IdentityModal, SocialModal } from '.';
+import { Social } from '../types';
+
+const StyledBox = styled(Box)`
+
+  svg {
+    cursor: pointer;
```
mattcasey, line 26: is this style necessary? the edit svg's are wrapped with IconButton which makes the cursor a pointer already

components/profile/public/components/SocialModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+const StyledButton = styled(Button)`
+    border-radius: 7px;
+`;
+
+export const schema = yup.object({
+  twitterURL: yup.string().notRequired().ensure().trim()
+    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?(?:mobile\.)?twitter\.com\/([a-zA-Z0-9_]+)/i, 'Invalid Twitter link'),
+  githubURL: yup.string().notRequired().ensure().trim()
+    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?github\.([a-z])+\/([A-Za-z0-9]{1,})+\/?$/i, 'Invalid GitHub link'),
+  discordUsername: yup.string().notRequired().ensure()
```
mattcasey, line 24: i'm not sure why, but if i just open the modal and click "Save", discord username is the only one that throws an error when the input is empty.
image

mattcasey (changes requested):
needs a new migration file, and some style requests otherwise looks good!

AndreiMitrea:
> needs a new migration file, and some style requests otherwise looks good!

Made the changes. In the new migration file this was also generated:
https://github.com/charmverse/app.charmverse.io/pull/492/files#diff-da24d60257ebf85b963fceedc503d6a6340de7fd9a3024ad69e1ad3946eb5ba8R2

PR #494, opened 2022-05-21
Fallback to raw emoji if twoemoji doesn't exist

Devorein:


mattcasey:
The emojis work now but they are too small. Why did we need tweomji?

image

mattcasey (approved):

PR #495, opened 2022-05-21
Upload discord & telegram avatar to s3

Devorein:


pages/api/discord/connect.ts
```diff
@@ -76,7 +77,12 @@ async function connectDiscord (req: NextApiRequest, res: NextApiResponse<Connect
     return;
   }
 
-  const avatar = discordAccount.avatar ? `https://cdn.discordapp.com/avatars/${discordAccount.id}/${discordAccount.avatar}.png` : undefined;
+  const avatarUrl = discordAccount.avatar ? `https://cdn.discordapp.com/avatars/${discordAccount.id}/${discordAccount.avatar}.png` : undefined;
+  let avatar: string | null = null;
+  if (avatarUrl) {
+    const { url } = await uploadToS3({ fileName: `user-content/${userId}/${uuid()}/${decodeURIComponent(new URL(avatarUrl).pathname.split('/').pop() || '')?.replace(/\s/g, '-') || uuid()}`, url: avatarUrl });
+    avatar = url;
```
mattcasey, line 84: you could be fancy here: `({ url: avatar } = await ...);`

pages/api/telegram/connect.ts
```diff
@@ -41,13 +43,14 @@ async function connectTelegram (req: NextApiRequest, res: NextApiResponse<Telegr
       }
     });
 
+    const avatar = (await uploadToS3({ fileName: `user-content/${userId}/${uuid()}/${decodeURIComponent(new URL(telegramAccount.photo_url).pathname.split('/').pop() || '')?.replace(/\s/g, '-') || uuid()}`, url: telegramAccount.photo_url })).url;
```
mattcasey, line 46: this line and the other one should be broken up. We're at a good point to standardize the user folder name too, with a new method for example `getUserS3Folder(userId)`

mattcasey (changes requested):
please fix the long lines, otherwise I tested and it looks good!

pages/api/discord/connect.ts
Devorein: I didn't know you could do that 👀 Updated.

pages/api/telegram/connect.ts
Devorein: I've added `getUserS3Folder({userId, url})` and used that function to generate the S3 filename for the users rather copying it over in all places.

Devorein:
Yep all of them have been refactored to use a function so the length has condensed quite a bit

PR #496, opened 2022-05-21
Scroll to the correct item in page navigation

Devorein:
Scroll to the current page in the page navigation for a long list. Currently, we keep the scroll bar at the top which leads to a suboptimal UX. The user has to manually scroll to get to the page.


mattcasey (changes requested):
+1 I think we should make sure the current page is visible, nice catch! I tested and it kind of works but it also is distracting if the item is already visible. See my loom: https://www.loom.com/share/fd413fa1a6ed41518a93f2bc22eab4f5.

Some changes I think we should make:
- on page load, the animation takes focus away from the page content. Can you try setting scrollTop directly instead, so that when the view loads, navigation is already in the correct position? (check out 'useLayoutEffect'?)
- the menu should not move if the item is already visible. maybe this would address the bug I shared in Loom?

mattcasey (approved):

PR #498, opened 2022-05-22
Update updatedAt field

Devorein:
Please try out the following scenarios and check if the `updated at` property is being updated correctly or not
1. Updating the content of the page
2. Add/Remove/Update the icon
3. Add/Remove/Update the headerimage
4. Update property value


pages/api/blocks/index.ts
```diff
@@ -162,7 +164,17 @@ async function updateBlocks (req: NextApiRequest, res: NextApiResponse<Block[]>)
...
+  const pageBlocks = cardBlocks.map(cardBlock => prisma.page.update({
+    where: { id: cardBlock.id },
+    data: {
+      updatedAt: new Date(),
+      updatedBy: req.session.user.id
+    }
+  }));
+
+  const updated = await prisma.$transaction(blockOps);
+  await prisma.$transaction(pageBlocks);
```
mattcasey, line 177: these transactions should be combined

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -41,6 +44,8 @@ type Props = {
 const PropertyValueElement = (props:Props): JSX.Element => {
     const [value, setValue] = useState(props.card.fields.properties[props.propertyTemplate.id] || '')
     const [serverValue, setServerValue] = useState(props.card.fields.properties[props.propertyTemplate.id] || '')
+    const {pages, setPages} = usePages();
```
mattcasey, line 47: propertyValueElement is used for all the table rows in table view, and I don't think we can afford to pull in the page context here. you sould be passing in updatedAt like we do for pageTitle, etc.

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -51,6 +56,17 @@ const PropertyValueElement = (props:Props): JSX.Element => {
 
     const editableFields: Array<PropertyType> = ['text', 'number', 'email', 'url', 'phone']
 
+    function updatePage() {
```
mattcasey, line 59: a better name for this function would be 'updatePageTimestamp'. Do we really need this, though? Can't we look at the updatedAt from the card instead when using focalboard?

mattcasey (commented):
Thanks for jumping on this... but where in the UI was the user seeing a date?  This update tries to make sure we update both all the time, but this is adding extra coupling to our data models that I want to be sure we need. There's no guarantee we always remember to update them both, and it feels "wrong" to be updating pages in the blocks endpoint, and block in the pages endpoint. 

We could also just say that the 'updatedAt' on page is related to fields on the page table, and 'updatedAt' on the card is for properties of the card table.  The reader/UI can decide which date to show. Did you consider that as an option?

pages/api/blocks/index.ts
Devorein: It would return us both the pages and the blocks and there is no definitive way to differentiate them. I guess I can use the `schema` or `fields` property which is only available in block. Will make the changes

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: There are properties like icons, header images, and content that are page-specific so we update the page, but for changes to a property value, it's card-specific so we update the card. More often than not it's the content that changes so we need the updated fields from the page, but that might not be accurate all the time. Plus now we need to update the card's state any time the user enters some text. That's why I believe we went with page's updated fields rather than card's

pages/api/blocks/index.ts
mattcasey: i thought you would filter the results:
```
const results = await prisma.$transaction(pageBlocks);
return results.filter(a => Boolean(a.cardId));
```

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
mattcasey: This makes total sense, thanks. I was lacking context about what date in the UI was wrong

components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -94,13 +93,20 @@ export default function EditorPage (
       .finally(() => {
         setIsEditing(false);
       });
-  }, [currentPageId, publicShare]);
+  }, [currentPageId, publicShare, currentPage]);
```
Comment (mattcasey) [inline components/[pageId]/EditorPage/EditorPage.tsx:96]: i'm curious was there a case where we were not updating setPage?

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -50,6 +48,7 @@ const PropertyValueElement = (props:Props): JSX.Element => {
     const finalDisplayValue = displayValue || emptyDisplayValue
 
     const editableFields: Array<PropertyType> = ['text', 'number', 'email', 'url', 'phone']
+    const latestUpdated = (new Date(updatedAt)).getTime() > (new Date(card.updatedAt)).getTime() ? "page" : "card";
```
mattcasey, line 51: ISODates are naturally sorted so you should be able to just compare the string values
```
     const latestUpdated = updatedAt > card.updatedAt ? "page" : "card";
     ```

mattcasey (approved):

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Nice one. I will change that asap.

components/[pageId]/EditorPage/EditorPage.tsx
Comment (Devorein) [reply components/[pageId]/EditorPage/EditorPage.tsx:96]: Good catch, I just checked if everything works without adding the `currentPage` inside the dependency array and it works, but it wasn't working without adding it at one point.

PR #505, opened 2022-05-24
delete toggle wrapper with backspace key

mattcasey:


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -255,12 +255,6 @@ const StyledReactBangleEditor = styled(ReactBangleEditor)`
     }
     cursor: pointer;
   }
-
-  .charm-paragraph {
-    .bangle-nv-child-container {
-      width: 100%;
```
mattcasey, line 261: this is also doing nothing because container is an inline element

PR #510, opened 2022-05-25
fix bug deleting a node

mattcasey:
Not 100% sure why, but when deleting a paragraph content inside a list item, it crashes the editor. This is serious because the user can't undo. Another fix would be to not use "span" as the contentDOM for Paragraph tags, but this seems less intrusive.


motechFR (approved):
Works on staging, thanks

PR #512, opened 2022-05-26
Public identity

AndreiMitrea:


prisma/migrations/20220526203229_added_user_identity_type/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "User" ADD COLUMN     "identityType" TEXT;
```
mattcasey, line 2: can we change this to an enum? there's a few examples in the .prisma file

AndreiMitrea: Will do, thank you.

mattcasey (approved):
approved, just one request: make 'identity type' an enum in prisma

prisma/schema.prisma
```diff
@@ -113,6 +113,13 @@ model Block {
   Page      Page[]
 }
 
+enum IdentityType {
+  Wallet
+  Discord
+  Telegram
+  Name
```
mattcasey, line 120: "Name" is a little vague, RandomName maybe? or Anonymous?

mattcasey (approved):

PR #513, opened 2022-05-26
Bounty payments

Devorein:


lib/applications/actions/reviewSubmission.ts
```diff
@@ -24,6 +25,10 @@ export async function reviewSubmission ({ submissionId, decision }: SubmissionRe
     throw new WrongStateError('Submissions must be in review for you to approve them');
   }
 
+  if (submission.status !== 'complete' && decision === 'pay') {
+    throw new WrongStateError('Submissions must be completed for you to make payment');
+  }
+
```
motechFR, line 31: Please extract this to its own method

reviewSubmission is completely independent from logic of paying

lib/applications/actions/listSubmissions.ts
```diff
@@ -1,14 +1,16 @@
-import { Application, Bounty } from '@prisma/client';
+import { Application, Transaction } from '@prisma/client';
 import { prisma } from 'db';
-import { DataNotFoundError } from '../../utilities/errors';
+
+export interface ApplicationWithTransactions extends Application {
+  transactions: Transaction[]
+}
```
motechFR, line 6: Please move to interfaces for this lib.

Implementation should not have its interfaces.

This also prevents us accidentally importing server code to the client

lib/applications/actions/reviewSubmission.ts
```diff
@@ -6,7 +6,8 @@ import { ReviewDecision, SubmissionReview } from '../interfaces';
 
 const submissionStatusAfterDecision: Record<ReviewDecision, ApplicationStatus> = {
   approve: 'complete',
-  reject: 'rejected'
+  reject: 'rejected',
+  pay: 'paid'
 };
```
motechFR, line 11: 'pay' should not be a review decision

lib/applications/shared.ts
```diff
@@ -24,7 +24,7 @@ export function submissionIsEditable ({ submission, bounty }: {submission: Appli
   return true;
 }
 
-export function moveUserApplicationToFirstRow (submissions: Application[], userId: string): Application[] {
+export function moveUserApplicationToFirstRow<Data extends ApplicationWithTransactions | Application> (submissions: Data[], userId: string): Data[] {
```
motechFR, line 27: Make this type ApplicationWithTransactions only, to avoid future data undefined issues

lib/transactions/createTransaction.ts
```diff
@@ -0,0 +1,29 @@
+import { Transaction } from '@prisma/client';
+import { prisma } from 'db';
+import { DataNotFoundError } from 'lib/utilities/errors';
+
+export type TransactionCreationData = Pick<Transaction, 'transactionId' | 'chainId'> & {applicationId: string}
+
```
motechFR, line 6: Please move this to separate interfaces file in lib/transactions/interfaces

lib/applications/actions/reviewSubmission.ts
Devorein: yep done in the latest commits

lib/transactions/createTransaction.ts
Devorein: yep done in the latest commits

lib/applications/actions/reviewSubmission.ts
Devorein: yep done in the latest commits

lib/applications/shared.ts
Devorein: I made it Application to make it compatible with every calls

PR #518, opened 2022-05-27
Small tweaks to Personal Nexus and integrations

mattcasey:
- de-dupe gnosis safes for users that have more than one wallet in control of a safe
- use purple color for discord icon on Integrations page
- better layout for Integrations
- most of this PR turned into a refactor of Discord connect. I split it up from the main page and we're now using cookies instead of the URL - I left the URL alone because some components are still using it. Please check it works on your end!

New layout:
image

### DIFF
--- a/components/profile/integrations/components/IdentityProviders.tsx
+++ b/components/profile/integrations/components/IdentityProviders.tsx
@@ -1,26 +1,22 @@
 
 import styled from '@emotion/styled';
-import { Alert, CircularProgress, Divider, Stack, SvgIcon, Tooltip, Typography } from '@mui/material';
+import { Alert, Box, Card, Divider, Grid, Stack, SvgIcon, Tooltip, Typography } from '@mui/material';
 import { useWeb3React } from '@web3-react/core';
 import charmClient from 'charmClient';
 import Button from 'components/common/Button';
 import { injected, walletConnect, walletLink } from 'connectors';
-import { ReactNode, useContext, useEffect, useState } from 'react';
+import { ReactNode, useContext, useState } from 'react';
 import { Web3Connection } from 'components/_app/Web3ConnectionManager';
-import useENSName from 'hooks/useENSName';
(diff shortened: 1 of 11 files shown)


components/profile/integrations/components/IdentityProviders.tsx
```diff
@@ -163,67 +107,59 @@ export default function IdentityProviders () {
...
-
-      {discordError && (
-        <Alert severity='error'>
-          {discordError}
-        </Alert>
-      )}
-
-      <Divider />
+      <DiscordProvider>
+        {({ isConnected, isLoading, connect, error }) => (
```
mattcasey, line 123: I got this idea from react-virtualized: DiscordProvider acts like a hook, instead of a react component as its children, it takes a function which it can pass state to. It's very similar to React context but without using React context

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -571,24 +571,33 @@ export default function GnosisTasksSection () {
 
   const safesWithTasks = tasks?.gnosis;
 
+  if (!safesWithTasks) {
```
mattcasey, line 574: making sure the loading component doesn't appear at the same time as 'connect your safes'

components/profile/tasks/IntegrationCard.tsx
```diff
@@ -40,7 +40,7 @@ export default function IntegrationCard () {
     <IntegrationCardContainer>
       <Grid container spacing={3}>
         <Grid item xs>
-          <Link href='profile/integrations'>
+          <Link href='/profile/integrations'>
```
mattcasey, line 43: this is the most important part of this PR because the link is broken

Devorein (approved):

components/profile/tasks/IntegrationCard.tsx
Devorein: Great catch

PR #523, opened 2022-05-27
fix stale cache of pages

mattcasey:
Still getting the hang of state in React... I assumed the "closures" are being regenerated so we'd be fine, but apparently not. I even removed all the debounce logic, but the only thing that works is to capture a ref to the original pages state.

https://dmitripavlutin.com/react-hooks-stale-closures/

### DIFF
--- a/components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
+++ b/components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
@@ -1,46 +1,39 @@
 import { useEditorViewContext } from '@bangle.dev/react';
 import { rafCommandExec } from '@bangle.dev/utils/pm-helpers';
-import { Page } from '@prisma/client';
 import { useCallback } from 'react';
 import { usePages } from 'hooks/usePages';
 import { addPage } from 'lib/pages';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
 import { useUser } from 'hooks/useUser';
+import { v4 } from 'uuid';
+import { useRouter } from 'next/router';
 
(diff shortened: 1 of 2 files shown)


components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
```diff
@@ -1,46 +1,39 @@
...
-
-    rafCommandExec(view, (state, dispatch) => {
-      if (!page) {
-        return false;
-      }
-      const nestedPageNode = state.schema.nodes.page.create({
-        id: page.id
+        if (dispatch) {
+          dispatch(state.tr.replaceSelectionWith(nestedPageNode));
+          router.push(`/${router.query.domain}/${newPage.path}`);
```
Devorein, line 32: Creating the now before routing shows the `Page Not Found` temporarily leading to a suboptimal UX, and calling it outside causes a editor to crash owing to mismatch transaction

components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
```diff
@@ -1,46 +1,39 @@
...
-      if (user && space) {
-        page = await addPage({
-          createdBy: user.id,
-          parentId: currentPageId,
-          spaceId: space.id
+  const router = useRouter();
+  const addNestedPage = useCallback(async () => {
+    if (user && space) {
+      const pageId = v4();
+      const newPage = await addPage({
```
Devorein, line 20: `addPageAndRedirect` redirects right after creating the page, we need to interject and create the node before we can route, thus the reason behind using `addPage`

PR #525, opened 2022-05-28
fix widths of modals

mattcasey:
This reverts a change to the Modal component from the profile section. It causes the success modal used on token gates to stretch full width:

image

@AndreiMitrea can you let me know which modal relied on this update?

### DIFF
--- a/components/common/Modal/SuccessModal.tsx
+++ b/components/common/Modal/SuccessModal.tsx
@@ -9,7 +9,7 @@ export default function SuccessModal ({ title, ...props }: { title?: string } & 
       <DialogTitle onClose={props.onClose} sx={{ padding: 0 }}>
         <Box display='flex' gap={1} width={300} alignItems='center'>
           <CheckCircleOutlineIcon color='success' fontSize='large' />
-          {typeof title === 'undefined' ? 'Success' : title}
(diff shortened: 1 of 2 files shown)


components/common/Modal/SuccessModal.tsx
```diff
@@ -9,7 +9,7 @@ export default function SuccessModal ({ title, ...props }: { title?: string } &
       <DialogTitle onClose={props.onClose} sx={{ padding: 0 }}>
         <Box display='flex' gap={1} width={300} alignItems='center'>
           <CheckCircleOutlineIcon color='success' fontSize='large' />
-          {typeof title === 'undefined' ? 'Success' : title}
+          {typeof title === 'undefined' ? 'Success' : 'title'}
```
AndreiMitrea, line 12: Title here should be used as variable, not string.

AndreiMitrea (approved):

AndreiMitrea:
@mattcasey The change was indeed for profile section,  I think it was for having custom modal widths (not the standard 670px, for example). Modals look good with this revert. Sorry for introducing the bug.

components/common/Modal/SuccessModal.tsx
mattcasey: yikes, thanks for catching this

PR #528, opened 2022-05-29
Dont fetch child pages for public boards

Devorein:


motechFR (approved):
Looks good @Devorein , approved

PR #536, opened 2022-05-30
Clean up nexus header

mattcasey:
- re-enable the link to nexus page
- add a consistent header/title on each nexus page to facilitate navigation
- Hide more information for the task header on mobile:
image

### DIFF
--- a/components/profile/components/NexusPageTitle.tsx
+++ b/components/profile/components/NexusPageTitle.tsx
@@ -2,17 +2,31 @@ import Link from 'components/common/Link';
 import { Box, Divider, Typography } from '@mui/material';
 import ChevronLeft from '@mui/icons-material/ChevronLeft';
 
-export default function PageTitle ({ title }: { title: string }) {
+export default function PageTitle ({ subPage }: { subPage?: string }) {
 
   return (
-    <Typography variant='h1' sx={{ display: 'flex', alignItems: 'flex-end', gap: 1, mb: 3 }}>
+    <Typography
+      variant='h1'
+      fontWeight='bold'
+      sx={{
+        display: 'flex',
+        alignItems: 'flex-end',
+        gap: 1,
+        mb: 3
+      }}
+    >
       <Link color='inherit' href='/profile'>
         <ChevronLeft />
         My Nexus
-      </Link><Divider orientation='vertical' flexItem color='secondary' />
-      <Box alignItems='center' style={{ position: 'relative', top: '-2px', fontSize: '.7em' }}>
-        {title}
-      </Box>
+      </Link>
+      {subPage && (
+        <>
+          <Divider orientation='vertical' flexItem color='secondary' />
+          <Box alignItems='center' style={{ position: 'relative', top: '-2px', fontSize: '.7em' }}>
+            {subPage}
+          </Box>
+        </>
+      )}
     </Typography>
   );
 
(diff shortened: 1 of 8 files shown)


components/profile/components/NexusPageTitle.tsx
```diff
@@ -0,0 +1,38 @@
...
+        }
+      }}
+    >
+      {subPage ? (
+        <>
+          <Link color='secondary' href='/profile/tasks' sx={{ display: 'flex', alignItems: 'center' }}>
+            <ChevronLeft />
+            {MyNexus}
+          </Link>
+          <Divider orientation='vertical' flexItem color='secondary' />
```
Devorein, line 29: image
I'm not sure why but the divider is showing a very thing blue vertical line

components/profile/tasks/GnosisTasksList.tsx
```diff
@@ -105,7 +105,7 @@ function TransactionRow (
           <Grid container spacing={1}>
             <Grid item xs={5}>
               {transaction.actions.map(action => (
-                <Box py={1}>
+                <Box py={1} key={action.to.address}>
```
Devorein, line 108: Great catch

components/profile/tasks/TasksPageHeader.tsx
```diff
@@ -75,29 +76,24 @@ export default function IntegrationCard () {
...
+                      fontWeight: 500
                     }}
                   >Integrations
                   </Typography>
                 </Box>
-                <Divider sx={{ borderRightWidth: 2 }} orientation='vertical' variant='middle' flexItem />
                 <Box
-                  display='flex'
+                  display={{
+                    xs: 'none',
```
Devorein, line 94: Maybe make it hide on `sm`? As we are hiding the integration icons even when ample space is available
image

Devorein (changes requested):

components/profile/components/NexusPageTitle.tsx
mattcasey: gross, thanks for catching!

components/profile/tasks/TasksPageHeader.tsx
mattcasey: sure

components/profile/tasks/GnosisTasksList.tsx
mattcasey: Thanks, it's still not perfect. i have some gnosis tasks sending +1 payment to the same address

components/profile/components/NexusPageTitle.tsx
mattcasey: it's because i added color='secondary', which is uneeded

components/profile/tasks/TasksPageHeader.tsx
mattcasey: I used 'md' because with 'sm' sometimes the right column falls below the left

components/profile/tasks/GnosisTasksList.tsx
Devorein: Maybe use a combination of `action.friendlyValue`, `transaction.id` & `action.to.address`?

components/profile/components/NexusPageTitle.tsx
Devorein: Awesome

Devorein (approved):

PR #537, opened 2022-05-31
Small changes for snapshot modal ui

Devorein:
Related [card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=c0659288-95de-451c-bc56-95a315e399fc)

### DIFF
--- a/components/common/PageLayout/components/Header/snapshot/PublishToSnapshot.tsx
+++ b/components/common/PageLayout/components/Header/snapshot/PublishToSnapshot.tsx
@@ -67,7 +67,7 @@ export default function PublishToSnapshot ({ page }: {page: Page}) {
       }
 
       {
-        !checkingProposal && currentSpace?.snapshotDomain && !proposal && (
+        !checkingProposal && !proposal && (
           <>
             <IosShareIcon
               fontSize='small'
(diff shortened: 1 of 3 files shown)


components/common/PageLayout/components/Header/snapshot/PublishToSnapshot.tsx
```diff
@@ -65,7 +67,7 @@ export default function PublishToSnapshot ({ page }: {page: Page}) {
       }
 
       {
-        !checkingProposal && !proposal && (
+        !checkingProposal && currentSpace?.snapshotDomain && !proposal && (
```
motechFR, line 70: Please remove current space and this condition.

The reason we don't check for space.snapshotDomain is that the PublishingForm will handle this for us, and give you an inline way to connect to snapshot

motechFR (approved):
Approved, just please change that 1 condition that doesn't need to be there

components/common/PageLayout/components/Header/snapshot/PublishToSnapshot.tsx
Devorein: Great catch. I have removed that in the latest commit.

PR #538, opened 2022-05-31
Updates to migration script

mattcasey:
see diff

### DIFF
--- a/scripts/updateUserIdentity.ts
+++ b/scripts/updateUserIdentity.ts
@@ -22,21 +22,32 @@ import { prisma } from '../db';
   let ensFound = 0;
 
   await Promise.all(users.map(async user => {
-    if (!user.addresses) {
+    if (!user.addresses?.length) {
       return;
     }
     const address = user.addresses[0];
-    const ens: string | null = await getENSName(address);
+    const ens = await getENSName(address);
 
-    user.ensName = ens || undefined;
-    ensFound += 1;
+    if (ens) {
+      user.ensName = ens || undefined;
+      ensFound += 1;
+    }
   }));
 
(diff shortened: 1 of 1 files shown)


scripts/updateUserIdentity.ts
```diff
@@ -8,25 +8,46 @@ import { prisma } from '../db';
 (async () => {
 
   const users: Array<Partial<LoggedInUser>> = await prisma.user.findMany({
+    where: {
+      identityType: null
+    },
```
mattcasey, line 13: decided to just filter these out here

scripts/updateUserIdentity.ts
```diff
@@ -8,25 +8,46 @@ import { prisma } from '../db';
...
   });
 
-  users.forEach(async user => {
-    if (!user.addresses) {
+  console.log('found', users.length, 'users missing identity type');
+
+  let ensFound = 0;
+
+  await Promise.all(users.map(async user => {
+    if (!user.addresses?.length) {
```
mattcasey, line 25: user.addresses ALWAYS exists because it's always an array (even though the TS says different)

scripts/updateUserIdentity.ts
```diff
@@ -8,25 +8,46 @@ import { prisma } from '../db';
...
     }
   });
 
-  users.forEach(async user => {
-    if (!user.addresses) {
+  console.log('found', users.length, 'users missing identity type');
+
+  let ensFound = 0;
+
+  await Promise.all(users.map(async user => {
```
mattcasey, line 24: needed to add Promise.all() here, before it was not waiting for ens to resolve

AndreiMitrea (approved):

PR #541, opened 2022-06-01
Set space default page permission level

Devorein:


seedData.ts
```diff
@@ -21,6 +21,7 @@ function MockSpace (partial: Partial<Space>): Space {
     updatedBy: '0x87ddfh6g435D12CE393aBbA3f81fe6C594543sdw',
     snapshotDomain: null,
     defaultVotingDuration: null,
+    defaultPagePermissionGroup: null,
```
motechFR, line 24: I like your exhaustiveness, being explicit about the default.

FYI, as per Prisma [docs](https://www.prisma.io/docs/concepts/components/prisma-schema/data-model#:~:text=Note%3A%20The%20default%20value%20of%20an%20optional%20field%20is%20null%20.) the default value of an optional field is null.

charmClient.ts
```diff
@@ -617,6 +617,12 @@ class CharmClient {
     return http.PUT(`/api/spaces/${spaceId}/snapshot`, data);
   }
 
+  setDefaultPagePermission (spaceId: string, pagePermissionLevel: PagePermissionLevel | null) {
```
motechFR, line 620: Could also be an object argument with destructured props

seedData.ts
Devorein: Thank you. But it was actually typescript that threw an error first 😛

charmClient.ts
Devorein: yep agreed. From now on if there are more than one parameters I will switch to an object.

lib/permissions/pages/triggers/__tests__/page-created.spec.ts
```diff
@@ -1,46 +1,87 @@
...
+        permissionLevel: spaceWithDefaultPagePermissionGroup.defaultPagePermissionGroup as PagePermissionLevel
+      })
+    );
+  });
 
+  it('should assign full_access to a root page if it doesn\'t exists', async () => {
+
+    const page = await createPage({
+      createdBy: user.id,
+      spaceId: spaceWithoutDefaultPagePermissionGroup.id
```
motechFR, line 65: Nice one Safwan, spaceWithoutDefaultPagePermissionGroup and spaceWithDefaultPagePermissionGroup are very good variables names, instantly understand what they are about

lib/spaces/__tests__/setSpaceDefaultPagePermission.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    try {
+      await setSpaceDefaultPagePermission(v4(), 'editor');
+      throw new ExpectedAnError();
+    }
+    catch (err) {
+      expect(err).toBeInstanceOf(SpaceNotFoundError);
+    }
+  });
+
+  it('Should return space with permissions after the default page permission has been set', async () => {
```
motechFR, line 55: Careful, how we phrase our unit tests is one of the most vitally important pieces of code we write. It's literally the "truth" about what our application does or not.

As much as possible, the "it should" statement needs to stand on its own without reading the code.

As permissions is already a well known concept (and a table in Prisma), a better way to write this would be

"Should set the default page permission level for a space and return the updated space"

lib/spaces/setSpaceDefaultPagePermission.ts
```diff
@@ -0,0 +1,27 @@
+import { PagePermissionLevel } from '@prisma/client';
+import { prisma } from 'db';
+import { SpaceNotFoundError } from 'lib/public-api';
+
+export async function setSpaceDefaultPagePermission (spaceId: string, defaultPagePermissionGroup: PagePermissionLevel) {
```
motechFR, line 5: This function is missing an explicit return type. It should be just a Space.

lib/spaces/setSpaceDefaultPagePermission.ts
```diff
@@ -0,0 +1,27 @@
...
+    where: {
+      id: spaceId
+    },
+    data: {
+      defaultPagePermissionGroup
+    },
+    include: {
+      permissions: true
+    }
+  });
```
motechFR, line 26: Not needed. Unless there's a reason I'm not aware of.

(Page) permissions can be linked to a space, user, role or public. So just returning a list of page permissions linked to a space isn't very useful

pages/[domain]/settings/workspace.tsx
```diff
@@ -1,30 +1,39 @@
...
+import { useRouter } from 'next/router';
+import { ReactElement } from 'react';
+import { useForm } from 'react-hook-form';
+
+const spaceDefaultPagePermissionLabel: Record<Exclude<PagePermissionLevel, 'custom'>, string> = {
+  full_access: 'Full Access',
+  editor: 'Editor',
+  view: 'View',
+  view_comment: 'View & Comment'
+};
```
Comment (motechFR) [inline pages/[domain]/settings/workspace.tsx:36]: This would be much better suited in our lib folder at lib/permissions/pages/page-permission-mapping.ts

motechFR (approved):
Approved! Please handle the changes and we are good to go

testing/setupDatabase.ts
```diff
@@ -91,6 +70,31 @@ export async function generateUserAndSpaceWithApiToken (walletAddress: string =
...
+          isAdmin
+        }
+      }
+    },
+    include: {
+      apiToken: true
+    }
+  });
+}
+
```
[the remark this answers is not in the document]
Devorein: Yes, you are right. I have removed `createSpace` and resorted to `generateUserAndSpaceWithApiToken`

lib/permissions/pages/triggers/__tests__/page-created.spec.ts
Devorein: Thank you 😄

lib/spaces/__tests__/setSpaceDefaultPagePermission.spec.ts
Devorein: Nice that's a lot more readable and closer to what we are actually doing.

lib/spaces/setSpaceDefaultPagePermission.ts
Devorein: I've added a separate type called `SpaceWithPermissions` inside our `lib/spaces/interfaces.ts` file and used that inside `charmClient`, `handler` and the lib function

lib/spaces/setSpaceDefaultPagePermission.ts
Devorein: Yeah I was thinking the same. We did add a point inside the card to return the permission but I might have mistaken that for the actual permissions rather than the default page permission level. Reverted back the changes I said in the previous review.

pages/[domain]/settings/workspace.tsx
Comment (Devorein) [reply pages/[domain]/settings/workspace.tsx:36]: Oh great catch. I was aware of that file before but it didn't cross my mind at all while I was trying to write the UI/ Big thanks 😄

PR #542, opened 2022-06-01
install dev tools from zip file

mattcasey:
Problem:
Currently, our fork of prosemirror-dev-tools includes another fork of a library called "unstated". I'm using the because the npm version has an old version of React, causing errors during install. Because of this, npm needs to run a build command on unstated because the compiled code is not included on github.

Fix:
With this PR, I have packaged our fork of PMT using `npm pack` and linked to it on S3 instead of using a github URL. Our fork comes with its dependencies already compiled, and this seems to work fine without yarn installed on my dev environment.


AndreiMitrea (approved):
WORKS! THank you, Matt!

PR #551, opened 2022-06-03
update typescript to 4.7

mattcasey:
Released May 24: https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -20,5 +20,6 @@
   "editor.codeActionsOnSave": {
     "source.organizeImports": false
   },
-  "jestrunner.configPath": "jest.config-server.ts"
+  "jestrunner.configPath": "jest.config-server.ts",
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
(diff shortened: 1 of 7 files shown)


lib/browser.ts
```diff
@@ -30,8 +30,7 @@ export function isTouchDevice () {
   }
   else {
     const mQ = typeof window !== 'undefined'
-      && window.matchMedia
-      && matchMedia('(pointer:coarse)');
+      && window.matchMedia('(pointer:coarse)');
```
mattcasey, line 33: This is a code change but should be OK. According to https://caniuse.com/matchmedia, every browser supports it

.vscode/settings.json
```diff
@@ -20,5 +20,6 @@
   "editor.codeActionsOnSave": {
     "source.organizeImports": false
   },
-  "jestrunner.configPath": "jest.config-server.ts"
+  "jestrunner.configPath": "jest.config-server.ts",
+  "typescript.tsdk": "node_modules/typescript/lib"
```
mattcasey, line 24: make VS Code use the correct version

Devorein: Nice

Devorein (approved):

lib/browser.ts
Devorein: Nice. good on ts for detecting a condition that will always be truthy

PR #557, opened 2022-06-03
Migration from GnosisSafeState to NotificationState

Devorein:


scripts/gnosisSafeStateToNotification.ts
```diff
@@ -0,0 +1,18 @@
+import { prisma } from 'db';
+
+export async function main () {
```
mattcasey, line 3: lets just skip migration, i'm the only one using notifications in prod right now

PR #562, opened 2022-06-03
Validate your nexus path in realtime

mattcasey:
image

### DIFF
--- a/components/profile/components/PoapSection.tsx
+++ b/components/profile/components/PoapSection.tsx
@@ -1,5 +1,5 @@
 import { useContext } from 'react';
-import useSWR from 'swr';
+import useSWRImmutable from 'swr/immutable';
 import charmClient from 'charmClient';
 import { Box, Grid, Link, Stack, SvgIcon, Typography } from '@mui/material';
 import EditIcon from '@mui/icons-material/Edit';
@@ -34,7 +34,7 @@ function PoapSection (props: PoapSectionProps) {
   const managePoapModalState = usePopupState({ variant: 'popover', popupId: 'poap-modal' });
   const { openWalletSelectorModal } = useContext(Web3Connection);
   const isPublic = isPublicUser(user);
-  const { data: poapData, mutate: mutatePoaps } = useSWR(`/poaps/${user.id}`, () => {
+  const { data: poapData, mutate: mutatePoaps } = useSWRImmutable(`/poaps/${user.id}`, () => {
     return isPublicUser(user) ? Promise.resolve({ visiblePoaps: [], hiddenPoaps: [] }) : charmClient.getUserPoaps();
   });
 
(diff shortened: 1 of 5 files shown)


components/profile/components/PoapSection.tsx
```diff
@@ -34,7 +34,7 @@ function PoapSection (props: PoapSectionProps) {
   const managePoapModalState = usePopupState({ variant: 'popover', popupId: 'poap-modal' });
   const { openWalletSelectorModal } = useContext(Web3Connection);
   const isPublic = isPublicUser(user);
-  const { data: poapData, mutate: mutatePoaps } = useSWR(`/poaps/${user.id}`, () => {
+  const { data: poapData, mutate: mutatePoaps } = useSWRImmutable(`/poaps/${user.id}`, () => {
```
mattcasey, line 37: I made this immutable to reduce the load on people w/bad internet. It just means we won't refresh it every time you come back to the page.

components/profile/components/UserDetails.tsx
```diff
@@ -44,7 +43,7 @@ const isPublicUser = (user: PublicUser | LoggedInUser): user is PublicUser => us
 export default function UserDetails ({ readOnly, user, updateUser }: UserDetailsProps) {
   const { account } = useWeb3React();
 
-  const { data: userDetails, mutate } = useSWR(`/userDetails/${user.id}`, () => {
+  const { data: userDetails, mutate } = useSWRImmutable(`/userDetails/${user.id}`, () => {
```
mattcasey, line 46: made this immutable because it's not likely to change often

pages/api/profile/checkPathAvailability.ts
```diff
@@ -0,0 +1,27 @@
...
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireUser)
+  .get(checkPathExists);
+
+export async function checkPathExists (req: NextApiRequest, res: NextApiResponse<{ available: boolean}>) {
+  const path = req.query.path as string;
+  const existing = await prisma.user.findUnique({
```
Devorein, line 16: Maybe add a `select: { id: true }` as we are calling this endpoint on each path text input change.

components/profile/components/UserPathModal.tsx
```diff
@@ -1,16 +1,36 @@
...
+}
+
+const debouncedValidate = debouncePromise(validatePath, 500);
 
 export const schema = yup.object({
-  path: yup.string().ensure().trim()
+  path: yup.string().ensure().trim().lowercase()
     .min(3)
     .max(50)
+    .matches(/^[a-zA-Z0-9-]+$/g, 'Only alphanumeric characters and hyphens are allowed')
```
Devorein, line 24: We are converting it to lowercase so `A-Z` is not necessary

Devorein (approved):

components/profile/components/UserPathModal.tsx
mattcasey: i didn't see the conversion happen until i saved the form. you might be right but i'm choosing to allow uppercase anyway so we don't accidentally call it an error

PR #566, opened 2022-06-05
Only show show actions if the editor content isn't empty

Devorein:
We are showing the row action component even when the editor is empty. This PR hide the row-action component unless the editor has content
Now
image
Previously
image

### DIFF
--- a/components/common/CharmEditor/components/rowActions/rowActions.ts
+++ b/components/common/CharmEditor/components/rowActions/rowActions.ts
@@ -43,7 +43,16 @@ export function plugins ({ key }: { key: PluginKey }) {
         view.dom.parentNode?.appendChild(tooltipDOM);
 
         function onMouseOver (e: MouseEventInit) {
-
+          const isEditorEmpty = Boolean(document.querySelector('.empty-editor'));
+          const rowHandleElement = document.querySelector('.row-handle') as HTMLDivElement;
+          if (rowHandleElement) {
+            if (!isEditorEmpty) {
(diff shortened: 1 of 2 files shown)


components/common/CharmEditor/components/rowActions/rowActions.ts
```diff
@@ -43,7 +43,16 @@ export function plugins ({ key }: { key: PluginKey }) {
         view.dom.parentNode?.appendChild(tooltipDOM);
 
         function onMouseOver (e: MouseEventInit) {
-
+          const isEditorEmpty = Boolean(document.querySelector('.empty-editor'));
+          const rowHandleElement = document.querySelector('.row-handle') as HTMLDivElement;
+          if (rowHandleElement) {
+            if (!isEditorEmpty) {
+              rowHandleElement.style.display = 'initial';
```
mattcasey, line 50: please change this to CSS instead

mattcasey (changes requested):
Querying the DOM is expensive. Is there a reason we can't use CSS to apply styles to the handle?

components/common/CharmEditor/components/rowActions/rowActions.ts
Devorein: You are right 🤦🏼 Fixed in the latest commit.

Devorein:
Not really. That just hadn't crossed my mind. Fixed in the latest commit.

mattcasey (approved):

PR #568, opened 2022-06-06
Hide snooze info icon on smaller screen

Devorein:


mattcasey (approved):
thanks for using the 'sx' prop!

PR #574, opened 2022-06-06
Fix for incorrect page ordering

AndreiMitrea:
Fixes for incorrect page ordering on drag and drop.

Some scenarios fixed by this PR that can be reproduced in production:

1. Dragging as child page
1.1 Make sure there are at least 2 pages in the workspace.
1.2 Drag one as the child page of the other.
1.3 Notice the console error:  _Uncaught Invariant Violation: Expected to find a valid target. targetId=_

2. Order shuffle for new pages
https://www.loom.com/share/d4f085548c9b42318d8705672ca00f8f
2.1 Create a new workspace.
2.2 Add 5 new pages.
2.3 Click randomly on the screen.
2.4 Notice the order changes.

3. Three is a problem
3.1 Create a new workspace.
3.2 Create two pages and a board.
3.3 Drag the second page above the first page.
3.4 Drag the board in between the two pages.
3.5 Notice the board is put first.

### DIFF
--- a/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
+++ b/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
@@ -16,6 +16,7 @@ import { ComponentProps, Dispatch, ReactNode, SetStateAction, SyntheticEvent, us
 import { useDrop } from 'react-dnd';
 import { checkForEmpty } from 'components/common/CharmEditor/utils';
 import { addPageAndRedirect, NewPageInput } from 'lib/pages';
+import sortBy from 'lodash/sortBy';
 import TreeNode, { MenuNode, ParentMenuNode } from './components/TreeNode';
 
 const StyledTreeRoot = styled(TreeRoot)<{ isFavorites?: boolean }>`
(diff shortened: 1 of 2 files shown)


Devorein:
@AndreiMitrea Do you face this issue when you drag and drop on a page? [Here](https://www.loom.com/share/21db8e5e26694908bde1e3f143d67fdd) is a loom recording
image

Devorein:
Another issue. [Here](https://www.loom.com/share/15bc01b8b32546aa946d73fb72963c3f) is a loom recording. I dragged Page 2.1 and dropped it above Page 3, but it got re-ordered to be underneath it.

AndreiMitrea:
> @AndreiMitrea Do you face this issue when you drag and drop on a page? [Here](https://www.loom.com/share/21db8e5e26694908bde1e3f143d67fdd) is a loom recording image

Yes, ran into that error too. Will look into it. Thank you!

mattcasey (changes requested):
please provide a better explanation of the bug so we can reproduce. I tested this PR but it seems to make drag/drop never work.. the page always jumps to the t

Not sure if the error @Devorein is posting is related, necessarily. I see that error sometimes even though the drag-and-drop worked... but it would be good to have an explanation at least

mattcasey (approved):

PR #575, opened 2022-06-07
Multi bounty payment

Devorein:


mattcasey (approved):
tested! I added a tooltip for the button when it's disabled because you aren't connected

mattcasey (approved):

PR #577, opened 2022-06-07
Embed bounty video when no bounties exist

motechFR:


mattcasey (commented):
full width is enormous on a widescreen, it needs a max-width
i think a max-width of 700 pixels is fine (the same width we use for document view) and height should be proportional to the width of the video, not the page 
our vide components use an aspect ratio of 1.77, you could try that

mattcasey (changes requested):
also, i think we should still show 'no bounties found' if there are hidden bounties due to the filter. this is supposed to be just for onboarding. it's helpful to know that my filter is returning no results

motechFR:
All done @mattcasey 

Thanks for suggestion about nuance of 0 count due to filters versus no bounties at all, great UX. Have handled this

mattcasey (changes requested):
See the loom.. it still is acting wonky: https://www.loom.com/share/f18a23e31ed44e9ab40a6599408e5399

Please center the video and make it 700px wide and 395px tall with a max-width of 100%. There's no need for 'vh' or percents or anything else.

motechFR:
Much simpler indeed!

Full HD screen (1920 * 1080)
image

PC 1366 * 768
image


Samsung Galaxy S8 (emulated 360 * 740)

PR #579, opened 2022-06-08
Fix for uploading cover image

AndreiMitrea:
Fix for uploading cover image.

image

### DIFF
--- a/components/[pageId]/DocumentPage/components/PageBanner.tsx
+++ b/components/[pageId]/DocumentPage/components/PageBanner.tsx
@@ -3,6 +3,7 @@ import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
 import { ListItem, Typography } from '@mui/material';
 import { Box } from '@mui/system';
+import { usePopupState } from 'material-ui-popup-state/hooks';
 import ImageSelector from 'components/common/ImageSelector/ImageSelector';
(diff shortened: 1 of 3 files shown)


components/[pageId]/DocumentPage/components/PageBanner.tsx
```diff
@@ -81,6 +81,7 @@ function PageBanner ({ focalBoard, headerImage, setPage }: PageBannerProps) {
         className='page-cover-controls'
       >
         <ImageSelector
+          autoOpen={true}
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/components/PageBanner.tsx:84]: This will automatically open the cover when we visit the page. We need to open it when the button is clicked

Devorein (approved):

components/[pageId]/DocumentPage/components/PageBanner.tsx
```diff
@@ -81,6 +84,7 @@ function PageBanner ({ focalBoard, headerImage, setPage }: PageBannerProps) {
         className='page-cover-controls'
       >
         <ImageSelector
+          autoOpen={coverImageModalState.isOpen}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageBanner.tsx:87]: autoOpen should be a 'open by default' setting, if we have an 'open' state it should be called 'open'. cc @Devorein

mattcasey (approved):

mattcasey (approved):

PR #583, opened 2022-06-08
Use metamask to determine gnosis safe for bulk payments

Devorein:


components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -37,6 +41,26 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
...
+  const gnosisSigner = useGnosisSigner();
+  const [currentUser] = useUser();
+  const [safeInfos, setSafeInfos] = useState<({
+    chainId: number;
+  } & SafeInfoResponse)[]>([]);
+
+  useEffect(() => {
+    async function main () {
+      if (gnosisSigner && currentUser) {
+        setSafeInfos(await getSafesForAddresses(gnosisSigner, currentUser.addresses));
```
mattcasey, line 53: please use the _connected address_, not the one we have on the current user. you should be using `useWeb3React` to retrieve the address that my metamask.

mattcasey: i think useSWR would be a little cleaner here, too, but not necessary:
```
const { address } = useWeb3React();
const signer = useGnosisSigner();
const safeInfo = useSWR(signer && '/connected-gnosis-safes', () => getSafesForAddress({ signer, chainId, address });
```

mattcasey (changes requested):
use connected wallet instead of user profile

lib/gnosis/gnosis.ts
```diff
@@ -45,7 +45,10 @@ interface GetSafesForAddressProps {
   chainId: number;
 }
 
-async function getSafesForAddress ({ signer, chainId, address }: GetSafesForAddressProps): Promise<({ chainId: number } & SafeInfoResponse)[]> {
+type Result = ({ chainId: number } & SafeInfoResponse)[];
```
mattcasey, line 48: just shortening the line, it was over the max length

mattcasey (approved):

.github/actions/install/action.yml
```diff
@@ -8,7 +8,7 @@ runs:
 
     - uses: actions/setup-node@v3
       with:
-        node-version: "16"
+        node-version: "16.14"
```
mattcasey, line 11: 16.15 was recently released and updated npm. I believe this is why I started getting an error during npm install

PR #584, opened 2022-06-08
Organize images under /images folder and de-dupe crypto config

mattcasey:


charmClient.ts
```diff
@@ -15,7 +15,7 @@ import { IUser, UserWorkspace } from 'components/common/BoardEditor/focalboard/s
 import { IWorkspace } from 'components/common/BoardEditor/focalboard/src/blocks/workspace';
 import { OctoUtils } from 'components/common/BoardEditor/focalboard/src/octoUtils';
 import { InviteLinkPopulated } from 'pages/api/invites/index';
-import { FiatCurrency, IPairQuote } from 'models/Currency';
+import { FiatCurrency, IPairQuote } from 'connectors';
```
mattcasey, line 18: 'Currency' was duplicating a bunch of config from 'connectors'. I could see moving 'connectors' to like a lib/blockchains folder, but for now the simplest thing is to delete the file in models. The stuff in models really belongs in lib anyway because we don't keep most of our models there - i created this folder before we started using prisma!

connectors.ts
```diff
@@ -236,10 +235,44 @@ export type Blockchain = keyof typeof RPC;
 
 export const RPCList = Object.values(RPC);
 
+export type CryptoCurrency = typeof RPC[Blockchain]['nativeCurrency']['symbol'];
```
motechFR, line 238: Really cool deduplication!

lib/utilities/formatting.ts
```diff
@@ -1,4 +1,4 @@
-import { FiatCurrency } from '../../models/Currency';
+import { FiatCurrency } from '../../connectors';
```
motechFR, line 1: Can be cleaned up to a node-like import

```typescript
import { FiatCurrency } from 'connectors';
```

motechFR (approved):
Approved, tested crypto price widget, looks good

PR #588, opened 2022-06-09
require and save wallet address for bounty submissions

mattcasey:
turns out we were not passing back the wallet address from the form

### DIFF
--- a/components/bounties/components/BountySubmissionContent.tsx
+++ b/components/bounties/components/BountySubmissionContent.tsx
@@ -6,11 +6,10 @@ import { useContributors } from 'hooks/useContributors';
 import Typography from '@mui/material/Typography';
 
 interface Props {
-  bounty: Bounty
-  submission: Application
+  submission: Application;
 }
 
-export default function BountySubmissionContent ({ bounty, submission }: Props) {
+export default function BountySubmissionContent ({ submission }: Props) {
 
   const [contributors] = useContributors();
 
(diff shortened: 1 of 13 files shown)


lib/applications/interfaces.ts
```diff
@@ -27,7 +27,7 @@ export interface ApplicationWithBounty extends Application {
   bounty: Bounty
 }
 
-export type SubmissionContent = Partial<Pick<Application, 'submission' | 'submissionNodes' | 'walletAddress'>>
+export type SubmissionContent = Pick<Application, 'submission' | 'submissionNodes'> & { walletAddress: string }
```
mattcasey, line 30: Should we be pulling these values from Application at all, since they are optional/nullable?

components/bounties/components/BountySubmissionContent.tsx
```diff
@@ -6,11 +6,10 @@ import { useContributors } from 'hooks/useContributors';
 import Typography from '@mui/material/Typography';
 
 interface Props {
-  bounty: Bounty
-  submission: Application
+  submission: Application;
 }
```
motechFR, line 10: Good spot, we don't need this for now

motechFR:
Looks good to me @mattcasey 

One small note, since we're fixing bounties, I made a small commit to trigger rollupBountyStatus each time the bounty is updated.

This was to handle the case where for example:
- Bounty has 1/1 submissions in progress
- Cap is raised to 2
- Bounty should return to open status, but wasn't

PR #592, opened 2022-06-09
clean up swr + dont use user session for data

mattcasey:
- cleans up the code around useSWR 
- we should never rely on 'req.session.user' to be up-to-date. We really should just replace this with a userId


AndreiMitrea (approved):
Code review only.

PR #594, opened 2022-06-10
Add breadcrumbs for public page

mattcasey:
This was trickier than it looked because I had to make sure all the context loads properly, including with SWR refreshes. The flow now looks like this:
1) the PublicPage component determines what space and page we're looking at by calling getPublicPage(). This  api call now supports a path as well as a pageId.
2) usePages() is used to grab the rest of the pages. It depends on useSpace(), but that depends on the /domain URL. So I updated usePages() to check for the /share/[pageId] route and in that case, just assume it's the first workspace available (since we only load one for public pages).

image

### DIFF
--- a/components/common/PageLayout/components/Header/Header.tsx
+++ b/components/common/PageLayout/components/Header/Header.tsx
@@ -25,14 +25,14 @@ import { usePages } from 'hooks/usePages';
 import { useUser } from 'hooks/useUser';
 import { generateMarkdown } from 'lib/pages/generateMarkdown';
 import { useRouter } from 'next/router';
-import React, { useRef, useState } from 'react';
+import { useRef, useState } from 'react';
 import Account from '../Account';
 import ShareButton from '../ShareButton';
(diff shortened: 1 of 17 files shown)


components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -115,15 +117,16 @@ export default function Header (
         edge='start'
         sx={{
           display: {
-            xs: hideSidebarOnSmallScreen ? 'none' : 'block',
-            md: 'block'
+            xs: hideSidebarOnSmallScreen ? 'none' : 'inline-flex',
+            md: 'inline-flex'
```
mattcasey, line 121: this was messing up the shape of the hover background (looks like an oval instead of a circle)

Devorein (approved):

Devorein:
Looks great. One small issue I found was viewing a public card page. Initially the card properties and breadcrumbs aren't shown but after refocus they are shown.

PR #596, opened 2022-06-10
Simplify loading of permissions and fix for loading context for public pages

mattcasey:
The fix was to remove teh call to 'refreshPage' inside of usePAges but it required a number of other changes which led to some nice cleanup.. thanks @motechFR

### DIFF
--- a/charmClient.ts
+++ b/charmClient.ts
@@ -153,7 +153,7 @@ class CharmClient {
   }
 
   createPage (pageOpts: Prisma.PageCreateInput) {
-    return http.POST<Page>('/api/pages', pageOpts);
+    return http.POST<IPageWithPermissions>('/api/pages', pageOpts);
   }
 
   getPage (pageId: string, spaceId?:string) {
--- a/charmClient.ts
+++ b/charmClient.ts
@@ -141,7 +141,7 @@ class CharmClient {
   }
 
   getPages (spaceId: string) {
-    return http.GET<Page[]>(`/api/spaces/${spaceId}/pages`);
+    return http.GET<IPageWithPermissions[]>(`/api/spaces/${spaceId}/pages`);
   }
 
   getArchivedPages (spaceId: string) {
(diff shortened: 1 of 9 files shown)


charmClient.ts
```diff
@@ -141,7 +141,7 @@ class CharmClient {
   }
 
   getPages (spaceId: string) {
-    return http.GET<Page[]>(`/api/spaces/${spaceId}/pages`);
+    return http.GET<IPageWithPermissions[]>(`/api/spaces/${spaceId}/pages`);
```
mattcasey, line 144: this turned out to be unecessary but i believe it's more accurate

motechFR: I've also updated the corresponding typing in the api in this [commit](https://github.com/charmverse/app.charmverse.io/pull/596/commits/b75f6b5817d0ed0503ab50bff4eacb1773f0029b)

Always good to have our types be accurate

components/common/PageLayout/components/ShareButton/components/ShareToWeb.tsx
```diff
@@ -41,50 +41,35 @@ const CopyButton = styled((props: any) => <Button color='secondary' variant='out
...
-export default function ShareToWeb ({ pageId }: Props) {
+export default function ShareToWeb ({ pageId, pagePermissions }: Props) {
 
   const router = useRouter();
-  // What user can do
-  const { pages, setPages, getPagePermissions, refreshPage } = usePages();
+  const { pages, getPagePermissions, refreshPage } = usePages();
   const [copied, setCopied] = useState<boolean>(false);
   const [space] = useCurrentSpace();
+  const publicPermission = pagePermissions.find(publicPerm => publicPerm.public === true) ?? null;
```
motechFR, line 54: The amount of deletions going on in this whole file is extremely pleasing

motechFR (approved):

PR #598, opened 2022-06-11
Add workspace  domain to public page path

mattcasey:
I ran into this because I was sharing the "getting started" page, otherwise no one would run into this right now. But it will be important when we let people customize the path anyway.


__integration-tests__/server/pages/api/public/pages/[pageId].spec.ts
```diff
@@ -0,0 +1,79 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/public/pages/[pageId].spec.ts:1]: @mattcasey Finally found a decent simple set of tests for page data API.

This section makes sure we support either pageId or space domain + page path

pages/api/public/pages/[...pageId]/index.ts
```diff
@@ -22,19 +22,34 @@ handler.get(getPublicPage);
 
 async function getPublicPage (req: NextApiRequest, res: NextApiResponse<PublicPageResponse>) {
 
-  const { pageId } = req.query;
+  const { pageId: pageIdOrPath } = req.query as {pageId: string[]};
```
Comment (motechFR) [inline pages/api/public/pages/[...pageId]/index.ts:25]: Didn't know about spread syntax in the API till I worked with this

pages/api/public/pages/[...pageId]/index.ts
```diff
@@ -22,19 +22,34 @@ handler.get(getPublicPage);
...
+          deletedAt: null,
+          spaceId: space.id,
+          space: {
+            domain: spaceDomain
+          },
+          path: pagePath.trim()
+        }
+      });
+    }
+  }
```
Comment (motechFR) [inline pages/api/public/pages/[...pageId]/index.ts:52]: Corrected this lookup so we are comparing spaceDomain and not spaceId

testing/setupDatabase.ts
```diff
@@ -61,7 +61,8 @@ export async function generateUserAndSpaceWithApiToken (walletAddress: string =
     space = await prisma.space.create({
       data: {
         name: 'Example space',
-        domain: v4(),
+        // Adding prefix avoids this being evaluated as uuid
+        domain: `domain-${v4()}`,
```
motechFR, line 65: Allows us to differentiate between UUID and non-UUID  url segment

motechFR (approved):

pages/api/public/pages/[...pageId]/index.ts
Comment (mattcasey) [reply pages/api/public/pages/[...pageId]/index.ts:52]: i'm already querying the space by domain, so i don't think we need that line?

__integration-tests__/server/pages/api/public/pages/[pageId].spec.ts
Comment (mattcasey) [reply __integration-tests__/server/pages/api/public/pages/[pageId].spec.ts:1]: that's great... why is "no-unused-vars" necessary here?

__integration-tests__/server/pages/api/public/pages/[pageId].spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+const exampleText = 'Content to compare';
+
+const pagePath = 'page-path';
+
+// These API calls should succeed without needed a user account
+describe('GET /api/public/pages/[pageId] - Load public page', () => {
+  it('should return the public page if provided the page ID and respond 200', async () => {
+
+    const page = await createPage({
```
Comment (mattcasey) [inline __integration-tests__/server/pages/api/public/pages/[pageId].spec.ts:26]: is this page being cleaned up somewhere?

Comment (motechFR) [reply __integration-tests__/server/pages/api/public/pages/[pageId].spec.ts:26]: I've added an inline cleanup for now in each test

pages/api/public/pages/[...pageId]/index.ts
Comment (motechFR) [reply pages/api/public/pages/[...pageId]/index.ts:52]: Thanks for spotting that, agree it's redundant.

Let's go with spaceDomain + pagePath as it's more meaningfully related to the original API path query

PR #600, opened 2022-06-12
Logout button in nexus page

Devorein:


components/nexus/components/NexusPageTitle.tsx
```diff
@@ -21,18 +28,33 @@ export default function PageTitle ({ subPage }: { subPage?: string }) {
...
-          <Box component='span' alignItems='center' sx={{ fontWeight: 'bold' }}>
-            {subPage}
+        <Box component='span' alignItems='center' sx={{ fontWeight: 'bold' }}>
+          {subPage}
+        </Box>
+      )
+        : (
+          <Box display='flex' justifyContent='space-between' width='100%' alignItems='center'>
+            <Box component='span' sx={{ fontWeight: 'bold' }}>{MyNexus}</Box>
+            {/* user cant be logged out so long as their wallet is connected (TODO: fix!) */}
```
mattcasey, line 38: can you change this to a tooltip and disable the button? it was confusing in the old UI that it would just disapper. and Tooltips will be invisible if the 'label/title' is an empty string

PR #601, opened 2022-06-13
Page full width switch

Devorein:


components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -188,7 +194,29 @@ export default function Header ({ open, openSidebar, hideSidebarOnSmallScreen }:
...
+                        try {
+                          await charmClient.updatePage({
+                            id: currentPage.id,
+                            fullWidth: !isFullWidth
+                          });
+                          setPages((_pages) => ({ ..._pages, [currentPageId]: { ...currentPage, fullWidth: !isFullWidth } }));
+                          setIsFullWidth(!isFullWidth);
+                        }
+                        catch (_) {
+                          //
```
mattcasey, line 215: what's the point in a try/catch if we don't do anything?

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -66,20 +67,26 @@ interface HeaderProps {
...
-  const { pages, currentPageId } = usePages();
+  const { pages, currentPageId, setPages } = usePages();
   const [user, setUser] = useUser();
   const theme = useTheme();
   const [pageMenuOpen, setPageMenuOpen] = useState(false);
   const [pageMenuAnchorElement, setPageMenuAnchorElement] = useState<null | Element>(null);
   const pageMenuAnchor = useRef();
   const currentPage = currentPageId ? pages[currentPageId] : undefined;
-
+  const [isFullWidth, setIsFullWidth] = useState(currentPage?.fullWidth ?? false);
```
mattcasey, line 77: maybe this is not necessary since we are updating the pages context?

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -188,7 +189,23 @@ export default function Header ({ open, openSidebar, hideSidebarOnSmallScreen }:
                   {/* Publishing to snapshot */}
 
                   <PublishToSnapshot page={currentPage as Page} />
-
+                  <Divider />
+                  <ListItemButton>
+                    <ListItemText>
```
mattcasey, line 194: clicking the label doesn't do anything, i think it should or at least the cursor should not make it look clickable:
image

mattcasey (approved):

PR #603, opened 2022-06-13
Trigger lint error for use of console.log

mattcasey:


.eslintrc.json
```diff
@@ -64,10 +67,10 @@
         "ignoreTemplateLiterals": true
       }
     ],
-    "no-alert": "off",
+    "no-alert": "warn",
```
mattcasey, line 70: no-alert should be  error but we have one alert still and i don't want to change it in this PR

PR #605, opened 2022-06-13
fix navigation in bounties

mattcasey:


components/common/PageLayout/getLayout.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactElement } from 'react';
+import PageLayout from './PageLayout';
+
+export default function getLayout (page: ReactElement) {
```
mattcasey, line 4: I noticed when I'm navigating around, the entire page redraws - I even see a white background for an instant. Next.js isn't supposed to rerender the layout, I'm wondering if we need to use the exact same method everywhere: https://nextjs.org/docs/basic-features/layouts

components/bounties/components/BountyCard.tsx
```diff
@@ -23,24 +24,26 @@ export function BountyCard ({ truncate = true, bounty }: IBountyInput) {
...
-      >
-        <CardHeader title={bounty.title} titleTypographyProps={{ sx: { fontSize: '1rem', fontWeight: 'bold' } }} />
-        <Box p={2} width='100%' display='flex' flex={1} flexDirection='column' justifyContent='space-between'>
-          <Typography paragraph={true}>
-            {fancyTrim(bounty.description, 120)}
-          </Typography>
-          <BountyStatusBadge truncate={truncate} bounty={bounty} />
-        </Box>
-      </CardActionArea>
+      <Link href={bountyUrl} passHref>
```
mattcasey, line 27: it's important to always use next/link when linking to pages, or else it will reload the whole app

hooks/useBounties.tsx
```diff
@@ -79,6 +80,11 @@ export function BountiesProvider ({ children }: { children: ReactNode }) {
     setCurrentBountyId(id);
 
     if (id) {
+      const bountyFromCache = bountiesRef.current.find(b => b.id === id);
+      if (bountyFromCache) {
+        setCurrentBounty(bountyFromCache);
```
mattcasey, line 85: use the current global state. I assume this is only invalid if someone else is making edits to the same bounty at the moment.. left the refresh in but it'll run after we show the current state

components/common/PageLayout/getLayout.tsx
motechFR: Nothing to add, but what's the relationship with raw anchor links and getLayout?

components/bounties/[bountyId]/BountyDetails.tsx
```diff
@@ -4,14 +4,13 @@ import { BountyApplicantList } from 'components/bounties/[bountyId]/components/B
...
 import BountySubmissions from 'components/bounties/[bountyId]/components_v3/BountySubmissions';
-import { Container } from 'components/[pageId]/DocumentPage/DocumentPage';
 import { useBounties } from 'hooks/useBounties';
 import { usePageTitle } from 'hooks/usePageTitle';
 import { useEffect } from 'react';
 
 export default function BountyDetails () {
 
-  const [_, setPageTitle] = usePageTitle();
+  const [, setPageTitle] = usePageTitle();
```
Comment (motechFR) [inline components/bounties/[bountyId]/BountyDetails.tsx:13]: Will remember this one for next time

PR #606, opened 2022-06-13
Fix links to board views on public pages

mattcasey:
We were using some interpolation logic from react-dom-router, which doesn't support syntax for `[...pageId]` in the path. Unfortunately Next.js doesn't export any utilities, AFAICT. So I replaced as many as I can with next/link components, and created a new helper `getUriWithParam` for one place where we just want to keep the URL up to date.


mattcasey:
optimistic merging for the bug fix

PR #609, opened 2022-06-14
Bounty application comments

AndreiMitrea:
Added comment section for reviewer and submitter in the bounty application modal.

image


prisma/migrations/20220614035657_added_application_thread/migration.sql
```diff
@@ -0,0 +1,19 @@
+/*
```
mattcasey, line 1: remember to regenerate this please before merging. We've had at least one prisma schema update in the past day

mattcasey (changes requested):
I think we should split up threads for bounty apps and pages. It might be saving a bit of time writing some CRUD ops but otherwise the actual use cases and UI are very different, and we're introducing a number of if/else statements to make it work. What happens when we add comments for a 3rd or 4th type of entity?

PR #611, opened 2022-06-14
Mentioned tasks in personal nexus

Devorein:


components/nexus/TasksPage.tsx
```diff
@@ -1,19 +1,20 @@
...
+import AlternateEmailIcon from '@mui/icons-material/AlternateEmail';
 import GnosisTasksList from './GnosisTasksList';
+import MentionedTasksList from './MentionedTasksList';
 import TasksPageHeader from './TasksPageHeader';
 import NexusPageTitle from './components/NexusPageTitle';
 import NotifyMeButton from './components/NotifyMeButton';
 import SnoozeButton from './components/SnoozeButton';
 
-type TaskType = 'multisig' | 'bounty' | 'proposal' | 'discussion'
+type TaskType = 'multisig' | 'bounty' | 'proposal' | 'discussion' | 'mentioned'
```
mattcasey, line 17: this actually belongs under Discussion. Discussion would also include comments, responses to comments, etc. They are similar but different from bounties, multisig and proposals.

components/nexus/TasksPage.tsx
```diff
@@ -98,16 +102,16 @@ export default function TasksPage () {
                 variant='text'
                 size='small'
                 onClick={() => setCurrentTab(task.type)}
-                color={currentTab === task.type ? 'textPrimary' : 'secondary'}
+                color={currentTab === task.type ? 'primary' : 'secondary'}
```
mattcasey, line 105: please copy the style from workspace setting tabs (text is black, the underline is blue):
image

components/nexus/TasksPage.tsx
```diff
@@ -39,6 +40,7 @@ const tabStyles = {
 
 const TASK_TYPES = [
   { icon: <KeyIcon />, type: 'multisig' },
+  { icon: <AlternateEmailIcon />, type: 'mentioned' },
   { icon: <BountyIcon />, type: 'bounty' },
```
mattcasey, line 44: please remove the tasks we still don't support yet

lib/mentions/extractMentions.ts
```diff
@@ -0,0 +1,39 @@
+import { PageContent } from 'models';
+
+/**
+ * Extract all mention ids from page content (json)
+ * @param content Page content as raw json
+ * @returns An array of mention ids
+ */
+export function extractMentions (content: PageContent) {
```
mattcasey, line 8: can you write this in a non-mutative way? something like:
function extractMentions (node): Mention[] {
  if (node.content) {
     return node.content.map(_node => extractMentions({ ..._node, parent: node })).flat()
  }
  else if (node.parent && node.type === 'mention') {
     const mention = { ... }
     return [mention];
  }
  else {
    return [];
  }
}

lib/mentions/extractMentions.ts
```diff
@@ -0,0 +1,39 @@
+import { PageContent } from 'models';
```
mattcasey, line 1: Since this code is related to prosemirror, it belongs in the CharmEditor folder

lib/mentions/getMentionedTasks.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (markedMentionTaskIds.has(mentionedTask.mentionId)) {
+      mentionedTasks.marked.push(mentionedTask);
+    }
+    else {
+      mentionedTasks.unmarked.push(mentionedTask);
+    }
+  });
+
+  return {
+    marked: mentionedTasks.marked.sort((mentionTaskA, mentionTaskB) => mentionTaskA.createdAt > mentionTaskB.createdAt ? -1 : 1),
```
mattcasey, line 110: if you add an `orderBy: createdAt` to the query for mentions, I don't think you have to sort them manually

lib/mentions/interfaces.ts
```diff
@@ -0,0 +1,16 @@
...
+  spaceId: string
+  spaceDomain: string
+  pagePath: string
+  spaceName: string
+  createdBy: User | null
+  pageTitle: string
+  text: string
+}
+
+export type MarkMentionTask = Pick<MentionedTask, 'mentionId' | 'createdAt'>
```
mattcasey, line 16: i think this belongs in the MarkMentionTask file, next to the method that is using it. generally we keep interfaces close to the code. i think using an interfaces.ts file makes sense when there isn't an obvious owner of an interface..

lib/mentions/markMentionedTasks.ts
```diff
@@ -0,0 +1,24 @@
...
+  const taskIds = new Set(userNotifications.map(userNotification => userNotification.taskId));
+
+  const mentionsNotNotified = mentions.filter(mention => !taskIds.has(mention.mentionId));
+
+  await prisma.userNotification.createMany({
+    data: mentionsNotNotified.map(mention => ({
+      taskId: mention.mentionId,
+      type: 'mention' as NotificationType,
+      userId,
+      createdAt: mention.createdAt
```
mattcasey, line 21: this should just be the current time, no?

components/nexus/MentionedTasksList.tsx
```diff
@@ -0,0 +1,125 @@
...
+  const { tasks, error } = useTasks();
+
+  useEffect(() => {
+    async function main () {
+      if (tasks?.mentioned && tasks.mentioned.unmarked.length !== 0) {
+        await charmClient.markMentions(tasks.mentioned.unmarked);
+      }
+    }
+
+    if (tasks?.mentioned) {
```
mattcasey, line 101: change this to `(tasks?.mentioned && tasks.mentioned.unmarked.length !== 0)` and then you don't need a second conditional inside main()

pages/api/tasks/mark-mentions.ts
```diff
@@ -0,0 +1,18 @@
+import { onError, onNoMatch, requireUser } from 'lib/middleware';
```
mattcasey, line 1: can we make this more generic? just "tasks/mark-seen" maybe. There's nothing specific about it other than the 'type' inherited from UserNotification. we actually should be (or will be) doing the same op for gnosis tasks, bounties, etc.

components/nexus/MentionedTasksList.tsx
```diff
@@ -0,0 +1,125 @@
...
+            </Grid>
+          </Grid>
+        </Card>
+      </Link>
+    </Box>
+  );
+}
+
+export default function MentionedTasksList () {
+  const { tasks, error } = useTasks();
```
mattcasey, line 92: Because tasks includes data for gnosis and mentions already, we should move this call to TasksPage and just pass the tasks down. OR we should break up the queries, so request mentions and tasks separately

components/nexus/TasksPage.tsx
Devorein: Ow good point. I will have to update the prisma schema as well then.

components/nexus/TasksPage.tsx
Devorein: Yup absolutely. On it.

components/nexus/TasksPage.tsx
Devorein: Yes just hid them

lib/mentions/extractMentions.ts
Devorein: Wouldn't it be a bit difficult to do unit tests then?

lib/mentions/getMentionedTasks.ts
Devorein: Unfortunately, it's a bit more complicated than that. Since we have to fetch and join the user that created the mention, the order could potentially get a bit messed up and we can only do database-level sorting for the marked mention tasks

lib/mentions/interfaces.ts
Devorein: There is a pretty significant reason I didn't place it inside `markMentionTasks.ts` file as the type `MarkMentionTask` is used in our charmClient as well. And if we just import that type we might end up importing the server-side only code. But I guess we can do `import type {MarkMentionTask} from ...`. I will give that a shot

lib/mentions/markMentionedTasks.ts
Devorein: Oops you are right.

components/nexus/MentionedTasksList.tsx
Devorein: Done

pages/api/tasks/mark-mentions.ts
Devorein: Great point. I just made it generic to handle any form of tasks. The lib `markTasks` reside inside `lib/tasks` and the endpoint is not `tasks/mark`

components/nexus/MentionedTasksList.tsx
Devorein: Ok having a bit of an issue with this one. Lets talk about it during the meeting

lib/mentions/extractMentions.ts
Devorein: Alright done. I've changed my code to be non-mutative but followed a bit different approach closely resembling the snippet you shared

lib/mentions/extractMentions.ts
Devorein: Moved it inside `lib/prosemirror`

mattcasey (approved):

PR #618, opened 2022-06-16
Added functionality for detecting old build

AndreiMitrea:
Added functionality for detecting when a new version is available.

image


next.config.js
```diff
@@ -63,6 +64,11 @@ const config = {
         });
       };
     }
+    _config.plugins.push(
+      new webpack.DefinePlugin({
+        'process.env.NEXT_PUBLIC_BUILD_ID': JSON.stringify(buildId)
```
mattcasey, line 69: why do we need to use JSON.stringify here? According to Typescript, buildId is a string already

mattcasey, line 69: i think `"${buildId}"` would be clearer than JSON.stringify, at first i didn't know why we needed it but it's to append quotes on the end

pages/_app.tsx
```diff
@@ -259,6 +272,14 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
                         <Global styles={cssVariables} />
                         <RouteGuard>
                           <ErrorBoundary>
+                            {
+                              isOldBuild && (
+                              <RefreshPageAlert
+                                clear={() => setIsOldBuild(false)}
+                                content='You are viewing an out of date version, please refresh the page.'
```
mattcasey, line 279: this phrase is one word too long, and it wraps to a 2nd line. can we change it to something shorter?

image

components/common/RefreshPageAlert.tsx
```diff
@@ -0,0 +1,48 @@
...
+import RefreshIcon from '@mui/icons-material/Refresh';
+
+const StyledAlert = styled(Alert)`
+    position: absolute;
+    margin: auto;
+    left: 0;
+    right: 0;
+    max-width: 590px;
+    width: 100%;
+    z-index: 2000;
```
mattcasey, line 14: always use CSS variables for z-index

components/common/RefreshPageAlert.tsx
```diff
@@ -0,0 +1,48 @@
+import styled from '@emotion/styled';
+import { Alert, Typography } from '@mui/material';
+import IconButton from '@mui/material/IconButton';
+import ClearIcon from '@mui/icons-material/Clear';
+import RefreshIcon from '@mui/icons-material/Refresh';
+
+const StyledAlert = styled(Alert)`
+    position: absolute;
+    margin: auto;
```
mattcasey, line 9: please use MUI snackbar as discussed.

components/common/RefreshPageAlert.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+function RefreshPageAlert (props: RefreshPageAlertProps) {
+  const { clear, content } = props;
+
+  return (
+    <StyledAlert
+      severity='warning'
+      action={(
+        [
+          <IconButton onClick={() => window.location.reload()}>
```
mattcasey, line 34: these buttons are not vertically centered:
image

package.json
```diff
@@ -232,6 +233,8 @@
     "prisma": "^3.9.2",
     "sass": "^1.49.0",
     "supertest": "^6.2.2",
-    "ts-jest": "^28.0.5"
+    "ts-jest": "^28.0.5",
+    "webpack": "^5.73.0",
+    "webpack-dev-server": "^4.9.2"
```
mattcasey, line 238: i don't think this module is necessary?

mattcasey (changes requested):
see comments

mattcasey:
I added the "deploy" label here, please make sure this works with Beanstalk by testing in staging first

PR #626, opened 2022-06-19
Fixed discord import roles with improved UX

Devorein:


motechFR (approved):
@Devorein Approving this PR as it fixes lost functionality.

However, we will need to come back to this component.

You are adding Discord-specific logic to the RolesSettings component, whose main goal is to federate specialist role components

PR #628, opened 2022-06-20
Moved discord roles import logic inside a hook

Devorein:
Discord roles import logic was previously written inside `RoleSettings`, while it doesn't introduce any bug it's not a good practice. Now the whole logic is located inside a local hook, making it much easier to reuse and reason about.

### DIFF
--- a/components/settings/roles/hooks/useImportDiscordRoles.ts
+++ b/components/settings/roles/hooks/useImportDiscordRoles.ts
@@ -9,6 +9,14 @@ import { useEffect } from 'react';
 import useSWRImmutable from 'swr/immutable';
 import useRoles from './useRoles';
 
+function routerQueryIsDiscordCallback () {
+  const urlSearchParams = new URLSearchParams(window.location.href);
+  return {
+    shouldImportDiscordRoles: typeof urlSearchParams.get('guild_id') === 'string' && urlSearchParams.get('discord') === '1' && urlSearchParams.get('type') === 'server',
+    serverConnectFailed: urlSearchParams.get('discord') === '2' && urlSearchParams.get('type') === 'server'
+  };
+}
+
 export function useImportDiscordRoles () {
   const { showMessage } = useSnackbar();
   const isAdmin = useIsAdmin();
@@ -18,11 +26,10 @@ export function useImportDiscordRoles () {
     refreshRoles
   } = useRoles();
 
-  const shouldRequestServers = isAdmin && space && typeof router.query.guild_id === 'string' && router.query.discord === '1' && router.query.type === 'server';
-  const serverConnectFailed = router.query.discord === '2' && router.query.type === 'server';
+  const { serverConnectFailed, shouldImportDiscordRoles } = routerQueryIsDiscordCallback();
   const guildId = router.query.guild_id as string;
   // Using immutable version as otherwise its called twice
-  const { data, isValidating, error } = useSWRImmutable(shouldRequestServers && space ? 'discord-roles-import' : null, async () => {
+  const { data, isValidating, error } = useSWRImmutable(isAdmin && shouldImportDiscordRoles && space ? 'discord-roles-import' : null, async () => {
     return charmClient.importRolesFromDiscordServer({
       guildId,
       spaceId: (space as Space).id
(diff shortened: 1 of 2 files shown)


components/settings/roles/hooks/useImportDiscordRoles.ts
```diff
@@ -0,0 +1,50 @@
...
+export function useImportDiscordRoles () {
+  const { showMessage } = useSnackbar();
+  const isAdmin = useIsAdmin();
+  const router = useRouter();
+  const [space] = useCurrentSpace();
+  const {
+    refreshRoles
+  } = useRoles();
+
+  const shouldRequestServers = isAdmin && space && typeof router.query.guild_id === 'string' && router.query.discord === '1' && router.query.type === 'server';
```
motechFR, line 21: It seems that you are querying params corresponding to a discord callback URL. I suggest making this clearer.

You could define a routerQueryIsDiscordCallback(): boolean method, and then use that as part of shouldRequestServers.

Could you also add a comment explaining what shouldRequestServers means?

components/settings/roles/hooks/useImportDiscordRoles.ts
```diff
@@ -0,0 +1,50 @@
...
+  const serverConnectFailed = router.query.discord === '2' && router.query.type === 'server';
+  const guildId = router.query.guild_id as string;
+  // Using immutable version as otherwise its called twice
```
motechFR, line 24: Great example of a useful comment that explains the why behind a decision.

As much as possible, we want our code to do what it says, and only use comments when going meta, and explaining why we are doing something

components/settings/roles/hooks/useImportDiscordRoles.ts
```diff
@@ -0,0 +1,50 @@
...
+      refreshRoles();
+    }
+    else if (error) {
+      // Major failure while trying to import discord server role
+      showMessage(error.message || error.error || 'Something went wrong. Please try again', 'error');
+    }
+    else if (serverConnectFailed) {
+      showMessage('Failed to connect to Discord', 'warning');
+    }
+    silentlyUpdateURL(window.location.href.split('?')[0]);
```
motechFR, line 44: Here, I would defined what [0] is supposed to be, so the reader understands what we are passing to silentlyUpdateUrl

motechFR:
@Devorein Really great that you refactored the logic so quick after initial hotfix. This looks much better

I also like how this component is now telling you Discord is importing compared to a couple versions ago.

I had issues performing the import, will contact you about this.

components/settings/roles/hooks/useImportDiscordRoles.ts
Devorein: Yes absolutely. Doing it right now.

components/settings/roles/hooks/useImportDiscordRoles.ts
Devorein: Thank you 😄

components/settings/roles/hooks/useImportDiscordRoles.ts
Devorein: Yeah you are right. Its a bit difficult to understand what's going on right now. I will use an appropriate variable to store that.

components/settings/roles/hooks/useImportDiscordRoles.ts
```diff
@@ -0,0 +1,59 @@
...
+      refreshRoles();
+    }
+    else if (error) {
+      // Major failure while trying to import discord server role
+      showMessage(error.message || error.error || 'Something went wrong. Please try again', 'error');
+    }
+    else if (serverConnectFailed) {
+      showMessage('Failed to connect to Discord', 'warning');
+    }
+    // Remove the query string from the url as they are no longer needed after discord import
```
motechFR, line 51: Great! Both how you defined a meaningful variable and explained while you are doing this

motechFR (approved):

components/settings/roles/hooks/useImportDiscordRoles.ts
Devorein: Thank you 😄

PR #632, opened 2022-06-22
Email for mentioned tasks

Devorein:
1. Updated email title and subject to consider mentioned tasks
2. Show mentioned tasks along with correct data after multi-sig actions

### DIFF
--- a/background/tasks/sendNotifications/sendNotifications.ts
+++ b/background/tasks/sendNotifications/sendNotifications.ts
@@ -5,13 +5,14 @@ import type { PendingTasksProps } from 'lib/emails/templates/PendingTasks';
 import { getPendingGnosisTasks, GnosisSafeTasks } from 'lib/gnosis/gnosis.tasks';
 import log from 'lib/log';
 import * as mailer from 'lib/mailer';
+import { getMentionedTasks } from 'lib/mentions/getMentionedTasks';
 
 export async function sendUserNotifications (): Promise<number> {
 
   const notificationsToSend = await getNotifications();
 
   for (const notification of notificationsToSend) {
-    log.info('Debug: send notification to user', { userId: notification.user.id, tasks: notification.tasks.length });
+    log.info('Debug: send notification to user', { userId: notification.user.id, tasks: notification.totalTasks });
     await sendNotification(notification);
   }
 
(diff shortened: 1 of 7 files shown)


lib/emails/templates/PendingTasks.tsx
```diff
@@ -45,15 +50,44 @@ export default function PendingTasks (props: PendingTasksProps) {
   );
 }
 
-function MultisigTask ({ task }: { task: GnosisSafeTasks }) {
+const charmverseUrl = 'https://app.charmverse.io';
```
mattcasey, line 53: can you use process.env.DOMAIN for this?

lib/emails/templates/PendingTasks.tsx
```diff
@@ -45,15 +50,44 @@ export default function PendingTasks (props: PendingTasksProps) {
...
 
-  const charmUrl = 'https://app.charmverse.io/nexus';
+// https://mjml.io/try-it-live/qD1IvIVyK
```
mattcasey, line 55: this link does not match the mjml below. just remove this .. it's why we have the preview endpoint

lib/emails/templates/PendingTasks.tsx
```diff
@@ -45,15 +50,44 @@ export default function PendingTasks (props: PendingTasksProps) {
...
+            <span style={{ fontSize: 20, fontWeight: 'bold' }}>
+              ({spaceName})
+            </span>
+          </a>
+        </div>
+        <div style={{ fontWeight: 'bold', color: greyColor2 }}>
+          {text}
+        </div>
+      </MjmlText>
+      <MjmlButton align='left' padding-bottom='40px' href={`${charmverseUrl}/${spaceDomain}/${pagePath}?mentionedId=${mentionId}`}>
```
mattcasey, line 77: i believe our app is looing for "mentionId" not "mentionedId" in the URL (which is good, we should be consistent)

lib/emails/templates/PendingTasks.tsx
```diff
@@ -9,13 +9,16 @@ import { GnosisSafeTasks } from 'lib/gnosis/gnosis.tasks';
 import { shortenHex } from 'lib/utilities/strings';
 import { greyColor2 } from 'theme/colors';
 import log from 'lib/log';
```
mattcasey, line 11: The UI is a little crazy when looking at the template.. you should use actual exampls for metnions. Replace "Mention inside @b1c173…8e29 callout" with something like "@ghostpepper what do you think?" We don't even tell the user what this is right now. Please group the mentions and change layout:

**<count> Mentions**
"Hey @ghostpepper what do you..."
Page: <page> | <workspace>

"cc @ghostpepper"
Page: <page> | <workspace>

Link the title "<count> Mentions" to discussions, and each page link to the mention.

pages/api/email-templates/index.ts
```diff
@@ -15,7 +15,53 @@ const templates = {
         email: '<userEmail>',
         username: 'ghostpepper'
       },
-      tasks: [
+      totalTasks: 5,
```
mattcasey, line 18: should be 4

pages/api/email-templates/index.ts
```diff
@@ -15,7 +15,53 @@ const templates = {
...
+          username: 'Devorein',
+          avatar: 'https://s3.amazonaws.com/charm.public.test/user-content/b1c1735e-da3c-4856-bd7c-fa9b13978e29/91bd9695-3fc9-4641-9597-1bca71646f0f/6866c3f39e2382f979a99b4e9e6ec9b4.png',
+          path: 'devorein',
+          isBot: false,
+          identityType: 'Discord'
+        }
+      }, {
+        mentionId: 'da4b78aa-94d5-4587-a269-9eb586ddec76',
+        createdAt: '2022-06-20T16:33:43.370Z',
+        pageId: '07c2b5c8-32e1-4830-ba65-8d12f0d3f2b7',
```
mattcasey, line 44: it would be nicer if you had a factory function that filled in details that don't matter for actually rendering the template

background/tasks/sendNotifications/sendNotifications.ts
```diff
@@ -5,71 +5,81 @@ import type { PendingTasksProps } from 'lib/emails/templates/PendingTasks';
...
-          in: tasks.map(getTaskId)
+          in: gnosisSafeTasks.map(getGnosisSafeTaskId)
         }
       }
     });
 
-    const tasksNotSent = tasks.filter(task => !sentTasks.some(t => t.taskId === getTaskId(task)));
-    const myTasks = tasksNotSent.filter(task => Boolean(task.tasks[0].transactions[0].myAction));
+    const tasksNotSent = gnosisSafeTasks.filter(gnosisSafeTask => !sentTasks.some(t => t.taskId === getGnosisSafeTaskId(gnosisSafeTask)));
+    const myTasks = tasksNotSent.filter(gnosisSafeTask => Boolean(gnosisSafeTask.tasks[0].transactions[0].myAction));
```
mattcasey, line 63: change to "myGnosisTasks"

lib/emails/templates/PendingTasks.tsx
Devorein: You are right. I've moved it up and replaced it with the value from `process.env`

lib/emails/templates/PendingTasks.tsx
Devorein: Nice catch. Yeah I initially copied it so that we know where we used the mjml for but since we are not using it anymore I will remove it.

lib/emails/templates/PendingTasks.tsx
Devorein: Thank you 😄 That could've been a dreadful bug

pages/api/email-templates/index.ts
Devorein: Yeah I agree. Let me create that. It will be a lot easier to read and configure in the long run.

lib/emails/templates/PendingTasks.tsx
Devorein: Alright. I've made the necessary changes in the latest commit.

mattcasey (approved):

PR #633, opened 2022-06-22
Show notifications badge beside avatar on top right

Devorein:
1. Show the total number of unattended tasks you have beside your avatar on the top right
2. Used `useSWRImmutable` to fetch tasks as fetching gnosis tasks can take a long time
3. Update the tasks state right after viewing the mentioned tasks (when you navigate to the mentioned tasks tab in your nexus we mark all the mentions as "seen", now since it will impact how many tasks is shown, I had to update the tasks state)

image


Devorein:
Right now we mark the mentioned tasks as "seen/attended" whenever you move to the discussions tab, thus it decreases the number of unattended tasks you have left. That's why you see the notification count went down by 1. This was done by design. For the gnosis safes, you have to sign them on an external site so we can't count them as attended right after you've seen them

motechFR (approved):

PR #634, opened 2022-06-22
Default public pages and permission presets

motechFR:
This PR does 3 things:
- It allows space-wide default public pages.
- It brings page permissions defaults and token gates under the Permissions header.
- It enables the use of permission presets instead of individual configurations (which are still allowed)

image


AndreiMitrea (approved):
Code review

Devorein (approved):

PR #637, opened 2022-06-23
Username cleanup

AndreiMitrea:
Cleaned up some code around the username functionality.

- username is required in schema
- removed getDisplayName from codebase
- some cleanup around ENS name
- contributors use username, not address

### DIFF
--- a/pages/api/spaces/[id]/contributors/index.ts
+++ b/pages/api/spaces/[id]/contributors/index.ts
@@ -22,7 +22,7 @@ async function getContributors (req: NextApiRequest, res: NextApiResponse<Contri
   const contributors = spaceRoles.map((spaceRole): Contributor => {
     return {
       ...spaceRole.user,
-      addresses: undefined,
+      addresses: [],
       isAdmin: spaceRole.isAdmin,
       joinDate: spaceRole.createdAt.toISOString()
     } as Contributor;


components/common/UserDisplay.tsx
```diff
@@ -50,8 +49,6 @@ interface UserDisplayProps extends StyleProps {
 }
 
 function UserDisplay ({ user, linkToProfile = false, ...props }: UserDisplayProps) {
-  const ensName = useENSName(user.addresses[0]);
```
mattcasey, line 53: since we are no longer using 'addresses' you should remove it from the UserDisplayProps as well

components/common/form/InputSearchContributor.tsx
```diff
@@ -149,11 +153,10 @@ export function InputSearchContributorMultiple ({ onChange, defaultValue, ...pro
 }
 
 export function ReviewerOption ({ user, avatarSize, fontSize, fontWeight, ...props }: { fontSize?: string | number, fontWeight?: number | string, user: User, avatarSize?: 'small' | 'medium' } & HTMLAttributes<HTMLLIElement> & {component?: ElementType} & BoxProps) {
-  const ensName = useENSName(user.addresses[0]);
```
mattcasey, line 152: remove 'addresses' from the valid props of this component. Am I Missing a reason to leave the field?

pages/api/spaces/[id]/contributors/index.ts
```diff
@@ -22,9 +22,10 @@ async function getContributors (req: NextApiRequest, res: NextApiResponse<Contri
   const contributors = spaceRoles.map((spaceRole): Contributor => {
     return {
       ...spaceRole.user,
+      addresses: undefined,
```
Comment (mattcasey) [inline pages/api/spaces/[id]/contributors/index.ts:25]: This looks like it will break for clients that are using the existing frontend code which expects addresses to be an array

Comment (mattcasey) [reply pages/api/spaces/[id]/contributors/index.ts:25]: Not 100% sure but i think one way to test this could be to run the app locally on main using "NODE_ENV=production" so that it doesnt refresh, and then change the backend to your code and see if it breaks.

Comment (AndreiMitrea) [reply pages/api/spaces/[id]/contributors/index.ts:25]: I dind't see a place on the front end that still uses address from contributor. Removing the references is part of this PR.

components/common/form/InputSearchContributor.tsx
AndreiMitrea: Done, thanks!

components/common/UserDisplay.tsx
AndreiMitrea: Done, thanks!

pages/api/spaces/[id]/contributors/index.ts
Comment (AndreiMitrea) [reply pages/api/spaces/[id]/contributors/index.ts:25]: If I run on main with only the back-end change, does refresh matter?

PR #638, opened 2022-06-24
Drag and drop support for linked pages

Devorein:
I've added a new prosemirror plugin that supports drag and drop for `page` node for now. Adding support for other type of nodes can be progressively introduced, but skipped for now as it would require a bit more time to test out whether it works for all possible combinations.


AndreiMitrea (approved):
Great work. Noticed a non blocking issue where sometimes if you drop a page underneath text, it leaves an empty row in between.

Devorein:
Thank you 😄 I've managed to fix them in the latest commit.

PR #639, opened 2022-06-24
Update Lit Protocol modal

mattcasey:


.cdk/staging-stack.ts
```diff
@@ -122,7 +122,7 @@ export class CdkDeployStack extends Stack {
     *  Note: something at the end of the path is required or the Load balancer url never resolves
      */
      new CfnOutput(this, "DeploymentUrl", {
-      value: 'https://' + deploymentDomain + '/login',
+      value: 'https://' + deploymentDomain + '/',
```
mattcasey, line 125: we actually don't have a /login path anymore.. woops!

motechFR:
**Tests**

**1. Make sure no pages crash**

- [ ] Pages
- [ ] Board pages
- [ ] Board page card popup
- [ ] Settings / workspace
- [ ] Settings / contributors
- [ ] Settings / roles & permissions
- [ ] Settings / Payment methods
- [ ] Bounties / board
- [ ] Bounties / modal
- [ ] Bounties / page
- [ ] Bounties / submit
- [ ] Trash
- [ ] Nexus / home
- [ ] Nexus / identities
- [ ] Nexus / profile
- [ ] Nexus / public profile


2. Try different conditions

a. Ethereum
- [ ] Wallet
- [ ] NFT ownership
 - [ ] min. 1 token (I have 1) => managed to log in
 - [ ] min. 2 tokens (I have 1) => failed to log in
 - [ ] POAP (Awaiting confirmation for my test)

motechFR:
@mattcasey 3 general points so far:

- The generated invite link has the prod domain hardcoded (instead of sending back the staging domain). Maybe a missing environment variable?

- I tested Polygon + NFT using the charmverse launch collection with minimum of 1. I was able to enter the space from a wallet that does not have this NFT.

- We are still only returning the first token gate (that's what I saw in staging). Could we return all of them? Otherwise I think this feature wouldn't work as expected for our users

components/common/Modal/Modal.tsx
```diff
@@ -54,14 +54,15 @@ export type ModalProps = Omit<ComponentProps<typeof MuiModal>, 'children' | 'onC
   size?: ModalSize,
   children: any,
   title?: string | ReactNode,
+  noPadding?: boolean,
   onClose: () => void
 };
 
-export function Modal ({ children, size = defaultSize, title, ...props }: ModalProps) {
+export function Modal ({ children, noPadding, size = defaultSize, title, ...props }: ModalProps) {
```
motechFR, line 61: Pretty nifty for easier layout

components/common/TokenGateForm.tsx
```diff
@@ -104,10 +102,10 @@ export default function JoinSpacePage ({ onSubmit: _onSubmit }: Props) {
       });
 
     const jwt = authSig && await litClient.getSignedToken({
-      resourceId: tokenGate.resourceId as any,
       authSig,
-      chain,
-      accessControlConditions: (tokenGate.conditions as any).accessControlConditions
+      chain: (tokenGate.conditions as any).chain || 'ethereum',
+      resourceId: tokenGate.resourceId as any,
+      ...tokenGate.conditions as any
```
motechFR, line 108: This is alot of "any" types. Could we add our own manual interfaces for now based on data structure we see?

components/settings/roles/RoleSettings.tsx
```diff
@@ -17,10 +17,10 @@ import ImportDiscordRolesMenuItem from './components/ImportDiscordRolesMenuItem'
...
 import RoleRow from './components/RoleRow';
 import { useImportDiscordRoles } from './hooks/useImportDiscordRoles';
-import DefaultPagePermissions from './spacePermissions/DefaultPagePermissions';
-import PermissionConfigurationMode from './spacePermissions/PermissionConfigurationMode';
-import SpacePermissions from './spacePermissions/SpacePermissions';
-import TokenGates from './token-gates/TokenGates';
+import DefaultPagePermissions from './components/SpacePermissions/components/DefaultPagePermissions';
+import PermissionConfigurationMode from './components/SpacePermissions/components/PermissionConfigurationMode';
+import SpacePermissions from './components/SpacePermissions';
+import TokenGates from './components/TokenGates';
```
motechFR, line 23: Makes sense as a flatter reorganisation of our code

components/settings/roles/components/SpacePermissions/components/DefaultPagePermissions.tsx
```diff
@@ -1,10 +1,8 @@
 import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 import { Menu, MenuItem } from '@mui/material';
 import Grid from '@mui/material/Grid';
-import Box from '@mui/material/Box';
 import FormControlLabel from '@mui/material/FormControlLabel';
 import Switch from '@mui/material/Switch';
-import { PagePermissionLevel } from '@prisma/client';
 import Button from 'components/common/Button';
```
motechFR, line 6: Seen, will watch out to clear up unused imports in the future

components/settings/roles/components/TokenGates/TokenGates.tsx
```diff
@@ -1,38 +1,59 @@
...
+const ShareModalContainer = styled.div`
+  width: 100%;
+
+  .lsm-single-condition-select-container,
+  .lsm-condition-display,
+  .lsm-condition-container,
+  .lsm-review-conditions-group-container {
+    overflow-y: auto !important;
+  }
+  /* Remove position: absolute so we have a dynamic height */
```
motechFR, line 35: For documentation purposes, I think it could be useful to link to where these come from and what they do

https://github.com/LIT-Protocol/lit-share-modal-v3/blob/master/src/shareModal/ShareModal.css

components/settings/roles/components/TokenGates/components/TokenGateRolesSelect.tsx
```diff
@@ -62,7 +62,7 @@ export default function TokenGateRolesSelect ({ onDelete, selectedRoleIds, onCha
         disabled={!isAdmin || roles?.length === 0}
         renderValue={(roleIds) => (
           (roleIds.length === 0) ? (
-            <Button size='small' variant='text' color='secondary'>+ Add roles</Button>
+            <Button disabled size='small' variant='text' color='secondary'>+ Add roles</Button>
           ) : (
```
motechFR, line 66: What about hiding this, or adding an explanation that you should add some roles if roles.length === 0?

Also I think making it more explicit the token gate grants roles access could be handy

components/common/TokenGateForm.tsx
mattcasey: i'd rather not, we are not opinionated on the conditions. they come straight from the library

components/settings/roles/RoleSettings.tsx
mattcasey: right... in the components folder, components are the only way we organize files unless it's related to a page path

components/settings/roles/components/TokenGates/TokenGates.tsx
mattcasey: the comment would just be "removing scrollbars" which you could tell by reading the CSS.

mattcasey:
@motechFR this PR is to upgrade the lit modal, i'm going to keep scope to that for now. supporting multiple tokens would be nice but it's not a priority and can be done in another PR

PR #643, opened 2022-06-25
Skip snoozed multisig tasks in notifications count 

Devorein:
1. Skip counting snoozed multisig tasks when counting your notifications
2. After adding a snooze/removing snooze for multisig tasks we update the users state to show a live update of the notifications count


mattcasey (approved):
tested and works

PR #645, opened 2022-06-26
Better upload image ux with loading message and spinner

Devorein:
Right now if after clicking on **Choose an image** and uploading an image, the UX is extremely confusing. The user has no idea whether its uploading or not. The button is not disabled so they can upload again, or switch tabs. This PR does the following things:-
1. Show an upload message for the upload button text while the image is being uploaded
2. DIsable the tabs so that one can't switch to pasting a link
3. DIsable the upload button while its uploading
4. Show a spinner beside the upload message

image

### DIFF
--- a/components/common/ImageSelector/ImageSelector.tsx
+++ b/components/common/ImageSelector/ImageSelector.tsx
@@ -4,6 +4,7 @@ import MultiTabs from 'components/common/MultiTabs';
 import PopperPopup from 'components/common/PopperPopup';
 import { uploadToS3 } from 'lib/aws/uploadToS3Browser';
 import { ReactNode, useState } from 'react';
+import { PimpedButton } from '../Button';
 import ImageSelectorGallery from './ImageSelectorGallery';
 
 interface ImageSelectorProps {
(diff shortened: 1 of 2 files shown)


components/common/ImageSelector/ImageSelector.tsx
```diff
@@ -35,58 +37,62 @@ export default function ImageSelector ({ autoOpen = false, children, galleryImag
...
+              ...tabs,
+              [
+                'Upload',
+                <Box sx={{
+                  display: 'flex',
+                  justifyContent: 'center',
+                  width: '100%'
+                }}
+                >
+                  <PimpedButton loading={isUploading} loadingMessage='Uploading image ...' disabled={isUploading} component='label' variant='contained'>
```
mattcasey, line 52: A bit nitpicky, but please remove "..." on the button label. i think the rotating circle is enough, and we don't use "..." for any other button actions

mattcasey (approved):
minor request

components/common/ImageSelector/ImageSelector.tsx
Devorein: You are right. The three dots at the end seems redundant if we end up showing the circular loader

PR #648, opened 2022-06-27
Bounty permissions

motechFR:
This PR introduces permissions for bounties, allowing for more granular access control, and role-based bounty actions


hooks/useBounties.tsx
```diff
@@ -31,16 +33,18 @@ export const BountiesContext = createContext<Readonly<IContext>>({
...
 export function BountiesProvider ({ children }: { children: ReactNode }) {
   const [space] = useCurrentSpace();
+  const [user] = useUser();
   const [bounties, bountiesRef, setBounties] = useRefState<BountyWithDetails[]>([]);
   useEffect(() => {
     if (space) {
       setBounties([]);
       charmClient.listBounties(space.id)
         .then(_bounties => {
+          bountiesRef.current = _bounties;
```
mattcasey, line 43: this looks like a red flag. you should not have to set the current value, just call setBounties

motechFR: Good to know, thought it might help with a weird state issue I had, but all good now!

PR #649, opened 2022-06-28
Inline votes charmeditor support

Devorein:
- [ x ] Drag across a selection and click on the vote menu icon to create an inline vote around the selected text
- [ x ] UI to customize inline vote 
- [ x ] Show inline votes on the right bar alongside comments
- [ x ] Show individual inline vote when clicking on the selection


components/common/CharmEditor/components/@bangle.dev/react-menu/floating-menu.ts
```diff
@@ -4,9 +4,13 @@ import { hasComponentInSchema } from "@bangle.dev/react-menu/helper";
...
 import { PluginKey } from "prosemirror-state";
+import { markName as inlineCommentMarkName } from "../../inlineComment/inlineComment.constants"
+import { markName as inlineVoteMarkName } from "../../inlineVote/inlineVote.constants"
 
-export function toggleSubMenu(floatingMenuPluginKey: PluginKey, subMenu: 'linkSubMenu' | 'inlineCommentSubMenu'): Command {
-  let nodeName = subMenu === "inlineCommentSubMenu" ? 'inline-comment' : 'link'
+export type SubMenu = 'defaultMenu' | 'linkSubMenu' | 'inlineCommentSubMenu' | 'inlineVoteSubMenu';
+
+export function toggleSubMenu(floatingMenuPluginKey: PluginKey, subMenu: SubMenu): Command {
+  let nodeName = subMenu === "inlineCommentSubMenu" ? inlineCommentMarkName : subMenu === "inlineVoteSubMenu" ? inlineVoteMarkName : 'link'
```
AndreiMitrea, line 13: Terneray inside ternary is pretty hard to read. Can you simplify it with if/then?

components/common/CharmEditor/components/FloatingMenu.tsx
```diff
@@ -25,16 +30,15 @@ export default function FloatingMenuComponent (
...
   const { getPagePermissions, currentPageId } = usePages();
   const permissions = currentPageId ? getPagePermissions(currentPageId) : new AllowedPagePermissions();
-
+  const isUserAdmin = isAdmin();
   return (
     <FloatingMenu
       menuKey={pluginKey}
-      renderMenuType={({ type }) => {
-
+      renderMenuType={(menuType) => {
```
AndreiMitrea, line 37: Can you please add the `menuType` type here, instead of on the next line?

components/common/CharmEditor/components/FloatingMenu.tsx
```diff
@@ -51,6 +55,7 @@ export default function FloatingMenuComponent (
                 <UnderlineButton />
                 <FloatingLinkButton menuKey={pluginKey} />
                 {!inline && permissions.comment && enableComments && <InlineCommentButton enableComments menuKey={pluginKey} />}
+                {!inline && permissions.comment && isUserAdmin && enableComments && <InlineVoteButton enableVotes menuKey={pluginKey} />}
```
AndreiMitrea, line 58: I think these conditions could be extracted in a variable. For example:
`const displayInlineVoteButton = !inline && permissions.comment && isUserAdmin && enableComments`

components/common/CharmEditor/components/PageInlineVote.tsx
```diff
@@ -0,0 +1,249 @@
...
+};
+
+const StyledDiv = styled.div<{ detailed: boolean }>`
+  background-color: ${({ theme, detailed }) => detailed && theme.palette.mode !== 'light' ? theme.palette.background.default : theme.palette.background.light};
+  padding: ${({ theme }) => theme.spacing(2)};
+`;
+
+function PageInlineVoteOption (
+  { isDisabled, voteOption, voteId, checked, percentage }:
+  { voteId: string, voteOption: VoteOptions, percentage: number, checked: boolean, isDisabled: boolean }
```
AndreiMitrea, line 39: Can you please create a type for this component? Something like:
```
type PageInlineVoteOptionProps = {...}
```

components/common/CharmEditor/components/Paragraph.tsx
```diff
@@ -1,56 +1,95 @@
...
+  gap: ${({ theme }) => theme.spacing(1.5)};
   cursor: pointer;
   user-select: none;
 `;
 
 export default function Paragraph (
-  { node, children, calculateInlineComments = true, inlineCommentPluginKey }:
-  NodeViewProps & {children: ReactNode, calculateInlineComments?: boolean, inlineCommentPluginKey: PluginKey<InlineCommentPluginState>}
+  { node, children, calculateActions = true, inlineCommentPluginKey, inlineVotePluginKey }:
+  NodeViewProps & {
```
AndreiMitrea, line 29: Could you please extract a props type for this component?

lib/votes/castVote.ts
```diff
@@ -43,6 +43,9 @@ export async function castVote (choice: string, voteId: string, userId: string):
     update: {
       choice,
       updatedAt: new Date()
+    },
+    include: {
+      user: true
```
AndreiMitrea, line 48: Why is this needed? Isn't the user casting the vote the logged in user?

lib/votes/createVote.ts
```diff
@@ -62,6 +62,11 @@ export async function createVote (vote: VoteDTO): Promise<Vote> {
       }
     },
     include: {
+      userVotes: {
```
AndreiMitrea, line 65: This is for creating a vote, there are no user votes yet, so this include won't do anything.

lib/votes/getPageVotes.ts
```diff
@@ -1,18 +1,24 @@
...
+  const pageVotes = await prisma.vote.findMany({
     where: {
       pageId
     },
     include: {
       userVotes: {
         orderBy: {
           updatedAt: 'desc'
+        },
+        include: {
```
AndreiMitrea, line 15: For what are the users needed? Depending on the number of page votes, there may be a performance hit.

lib/votes/getSpaceVotes.ts
```diff
@@ -1,14 +1,21 @@
...
 
 export async function getSpaceVotes (spaceId: string): Promise<ExtendedVote[]> {
-  return prisma.vote.findMany({
+  const pageVotes = await prisma.vote.findMany({
     where: {
       spaceId
     },
     include: {
-      userVotes: true,
+      userVotes: {
```
AndreiMitrea, line 11: What is the use case for getting all the user votes with the users atached for a workspace? If there is a workspace with lots of pages, each one with lots of vote, each vote with lots of user votes, the response will be very large. Depending on the use case, either filter down to what information would actually be used, or break it down in multiple endpoint calls.

lib/votes/getVote.ts
```diff
@@ -1,14 +1,21 @@
...
-export async function getVote (id: string): Promise<Vote | null> {
-  return prisma.vote.findUnique({
+export async function getVote (id: string): Promise<ExtendedVote | null> {
+  const pageVote = await prisma.vote.findUnique({
     where: {
       id
     },
     include: {
-      userVotes: true,
+      userVotes: {
```
AndreiMitrea, line 11: This might be taxing for performance as well, depending on number of votes cast. We should filter down to what we actually need from user.

lib/votes/getSpaceVotes.ts
```diff
@@ -1,14 +1,21 @@
...
+        include: {
+          user: true
+        }
+      },
       voteOptions: true
     }
   });
+
+  return updateVotesStatus(pageVotes);
```
AndreiMitrea, line 20: This is for HTTP GET, it should be idempotent. Getting votes should not have the sideffect of update. Please consider updating the status in another place. There is also a task for creating a background job: https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=dbf0f6db-a9b5-49f3-b488-dfa67f18db5a

lib/votes/getVote.ts
```diff
@@ -1,14 +1,21 @@
...
+        include: {
+          user: true
+        }
+      },
       voteOptions: true
     }
   });
+
+  return pageVote ? (await updateVotesStatus([pageVote]))[0] : null;
```
AndreiMitrea, line 20: This is for HTTP GET, it should be idempotent. Getting vote should not have the sideffect of update. Please consider updating the status in another place. There is also a task for creating a background job: https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=dbf0f6db-a9b5-49f3-b488-dfa67f18db5a

lib/votes/interfaces.ts
```diff
@@ -22,7 +22,11 @@ export interface UserVoteDTO {
     choice: string,
 }
 
+export interface ExtendedUserVote extends UserVote {
+  user: User
```
mattcasey, line 26: we should always use the 'Contributor' interface unless it's the logged in user

pages/_app.tsx
```diff
@@ -180,6 +180,7 @@ import IconButton from '@mui/material/IconButton';
 import RefreshIcon from '@mui/icons-material/Refresh';
 
 import charmClient from 'charmClient';
+import { InlineVotesProvider } from 'hooks/useInlineVotes';
```
mattcasey, line 183: this looks like maybe we can add it outside the CharmEditor or DocumentPage, and not in _app? it's just for votes related to the document right?

pages/api/pages/[id]/votes.ts
```diff
@@ -20,25 +19,15 @@ async function getVotes (req: NextApiRequest, res: NextApiResponse<ExtendedVote[
...
+    select: {
+      spaceId: true
     }
   });
 
   if (!page) {
     throw new PageNotFoundError(pageId);
   }
 
-  const { error } = await hasAccessToSpace({
```
Comment (mattcasey) [inline pages/api/pages/[id]/votes.ts:32]: It looks like authorization was removed, but we should check that the user has access to view the page. I see computeUserPagePermissions() is used in the public page endpoint

pages/_app.tsx
mattcasey: ah you are probably using it to open the votes sidebar right?

hooks/usePageActionDisplay.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useInlineVotes } from './useInlineVotes';
+import { usePages } from './usePages';
+import { useThreads } from './useThreads';
+
+export interface IPageActionDisplayContext {
+  currentPageActionDisplay: null | 'votes' | 'comments',
+  setCurrentPageActionDisplay: React.Dispatch<React.SetStateAction<IPageActionDisplayContext['currentPageActionDisplay']>>
+}
+
+export const PageActionDisplay = createContext<IPageActionDisplayContext>({
```
mattcasey, line 14: context var names should end with "Context"

components/common/CharmEditor/components/PageInlineVote.tsx
```diff
@@ -0,0 +1,249 @@
...
+    >
+      <span>Votes</span> <Chip size='small' label={totalVotes} />
+    </Box>
+  );
+
+  const userVote = user && inlineVote.userVotes.find(_userVote => _userVote.userId === user.id);
+
+  const hasPassedDeadline = new Date(deadline) < new Date();
+
+  const relativeDate = DateTime.fromJSDate(new Date(deadline)).toRelative({ base: (DateTime.now()) });
```
mattcasey, line 106: i think you can just do DateTime.fromISO() and not have to cast the date first

components/common/CharmEditor/components/PageInlineVote.tsx
```diff
@@ -0,0 +1,249 @@
...
+      <ConfirmDeleteModal
+        title='Delete vote'
+        onClose={popupState.close}
+        open={menuState.open}
+        buttonText={`Delete ${inlineVote.title}`}
+        onConfirm={() => {
+          removeInlineVoteMark(view, inlineVote.id);
+          deleteVote(inlineVote.id);
+        }}
+        question={`Are you sure you want to delete ${inlineVote.title} vote?`}
```
mattcasey, line 222: i think this could sound weird depending on the vote title, can we just say "to delete this vote". if you like having the title, you could add it afterwards: 'delete this vote: "Should we go to Wendy's for dinner?"'

hooks/usePageActionDisplay.tsx
```diff
@@ -0,0 +1,57 @@
+import { ThreadWithCommentsAndAuthors } from 'lib/threads/interfaces';
```
mattcasey, line 1: i think this file can use the normal .ts extension

lib/inline-votes/findTotalInlineVotes.ts
```diff
@@ -0,0 +1,28 @@
...
+  const inlineVoteMarkSchema = view.state.schema.marks['inline-vote'] as MarkType;
+  const inlineVoteNodes = findChildrenByMark(node, inlineVoteMarkSchema);
+  let totalInlineVotes = 0;
+  // There is a possibility that multiple nodes can have the same threadId so use a set to capture only the unique ones
+  const voteIds: Set<string> = new Set();
+  for (const inlineVoteNode of inlineVoteNodes) {
+    // Find the inline vote mark for the node
+    const inlineVoteMark = inlineVoteNode.node.marks.find(mark => mark.type.name === inlineVoteMarkSchema.name);
+    // Only count the non-resolved votes
+    if (inlineVoteMark /* && (!inlineCommentMark.attrs.resolved) */) {
```
mattcasey, line 19: should this comment be removed?

lib/inline-votes/removeInlineVoteMark.ts
```diff
@@ -0,0 +1,30 @@
...
+      });
+    }
+  }
+
+  inlineVoteNodeWithMarks.forEach(inlineVoteNodeWithMark => {
+    const from = inlineVoteNodeWithMark.pos;
+    const to = from + inlineVoteNodeWithMark.node.nodeSize;
+    const tr = view.state.tr.removeMark(from, to, inlineVoteMarkSchema);
+    if (view.dispatch) {
+      view.dispatch(tr);
```
mattcasey, line 27: i think this should only be run once, right? so it should happen after the for loop

lib/prosemirror/highlightMarkedElement.ts
```diff
@@ -0,0 +1,49 @@
...
+  const fromNodeAfter = $from.nodeAfter;
+  const toNodeAfter = $to.nodeAfter;
+  if (!toNodeAfter) {
+    const tr = view.state.tr.setSelection(new TextSelection(view.state.doc.resolve(view.state.selection.$to.pos)));
+    view.dispatch(tr);
+    return false;
+  }
+  if (fromNodeAfter) {
+    const inlineActionMark = view.state.doc.type.schema.marks[markName].isInSet(fromNodeAfter.marks);
+    const pageActionListNode = document.querySelector(`.${className}`) as HTMLDivElement;
```
mattcasey, line 22: "NodeList" not "ListNode"

components/common/CharmEditor/components/inlineComment/inlineComment.plugins.ts
```diff
@@ -53,41 +54,13 @@ export function plugin ({ key } :{
...
-                renderSuggestionsTooltip(key, {
-                  threadIds: [threadId]
-                })(view.state, view.dispatch, view);
-              }
-            }
-          }
-          return true;
+          return highlightMarkedElement({
+            view,
+            className: 'PageThreadListBox',
```
mattcasey, line 59: class names should be lower case and hyphenated

lib/prosemirror/highlightMarkedElement.ts
mattcasey, line 22: this is actually a good case to use `id` and not class names

components/common/CharmEditor/components/Paragraph.tsx
```diff
@@ -1,56 +1,95 @@
...
+    children: ReactNode,
+    calculateActions?: boolean,
+    inlineCommentPluginKey: PluginKey<InlineCommentPluginState>,
+    inlineVotePluginKey: PluginKey<InlineVotePluginState>
+  }
 ) {
   const cardId = (new URLSearchParams(window.location.href)).get('cardId');
```
mattcasey, line 36: suggestion for this var name: `isInsideModal` or something describing the state you're checking for

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+                  setDeadline(value);
+                }
+              }}
+              onChange={(value) => {
+                if (value) {
+                  setDeadline(value);
+                }
+              }}
+              renderInput={(props) => (
+                <TextField
```
mattcasey, line 231: can u make it so that clicking this text input also opens the datepicker? i think that is what people usually expect

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+          defaultValue='default'
+          value={voteType}
+          onChange={(e) => {
+            setVoteType(e.target.value as VoteType);
+          }}
+        >
+          <FormControlLabel
+            value='default'
+            control={<Radio />}
+            label='Yes / No'
```
mattcasey, line 273: change this to Yes / No / Abstain, and then i don't think we should even show a list of options below. it's redundant.

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+                inputProps: {
+                  min: 1,
+                  max: 100,
+                  step: 1
+                }
+              }}
+            />
+          </Box>
+        </Box>
+        <RadioGroup
```
mattcasey, line 262: On this row, can you add a label? So it would look like:

**Options:**    [ ] Yes / No / Abstain     [ ] Custom

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+  const { createVote } = useInlineVotes();
+
+  useEffect(() => {
+    if (voteType === 'custom') {
+      setOptions([{
+        name: 'Option 1'
+      }, {
+        name: 'Option 2'
+      }, {
+        name: 'No change'
```
mattcasey, line 147: shouldn't this also just be "Abstain"? Might want to check with Alex

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+  return (
+    <div>
+      <FieldLabel>Options</FieldLabel>
+      {options.map((option, index) => {
+        return (
+          <ListItem sx={{ px: 0, pt: 0, display: 'flex', gap: 0.5 }}>
+            <TextField
+              error={option.name.length === 0 || Boolean(options.find((_option, _index) => _index !== index && _option.name === option.name))}
+              // Disable changing text for No change option
+              disabled={disableTextFields || index === 2}
```
mattcasey, line 87: Let the user have full control, there's no need to always force a special option

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+              value={option.name}
+              onChange={(e) => {
+                options[index] = {
+                  name: e.target.value
+                };
+                setOptions([...options]);
+              }}
+            />
+            <IconButton
+              disabled={disableDelete || options.length === 2 || (index <= 2)}
```
mattcasey, line 99: it should be index < 2. Abstain should not be required

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+              size='small'
+              onClick={() => {
+                setOptions([...options.slice(0, index), ...options.slice(index + 1)]);
```
mattcasey, line 102: splice is better for this: `arr.splice(index, 1)` to remove an item

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+      }, {
+        name: 'No'
+      }, {
+        name: 'Abstain'
+      }]);
+    }
+  }, [voteType]);
+
+  const [deadline, setDeadline] = useState(DateTime.fromMillis(Date.now()).plus({ hour: 12 }));
+  const { currentPageId } = usePages();
```
mattcasey, line 162: just noting: it really would be great if we could get currentPageId from somewhere besides usePages, this will cause the inline vote component to rerender as the user types

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -0,0 +1,292 @@
...
+}
+
+export function InlineVoteSubMenu ({ pluginKey }: { pluginKey: PluginKey }) {
+  const view = useEditorViewContext();
+  const [voteTitle, setVoteTitle] = useState('');
+  const [voteDescription, setVoteDescription] = useState('');
+  const [passThreshold, setPassThreshold] = useState<number>(50);
+  const [voteType, setVoteType] = useState<VoteType>('default');
+  const [options, setOptions] = useState<{ name: string }[]>([]);
+  const { createVote } = useInlineVotes();
```
mattcasey, line 138: i can't tell, is this component always embedded inside of CharmEditor? if so, it would be better to create a component that is only rendered when the modal is open, which consumes useInlineVotes and usePages. otherwise it will re-render as the user is typing

components/common/CharmEditor/components/inlineVote/inlineVote.specs.ts
```diff
@@ -0,0 +1,39 @@
...
+        {
+          tag: 'span.charm-inline-vote'
+        }
+      ],
+      toDOM: (mark): DOMOutputSpec => ['span', { class: `charm-inline-vote ${mark.attrs.resolved ? 'resolved' : 'active'}`, id: `inline-vote.${mark.attrs.id}` }]
+    },
+    markdown: {
+      // TODO: Fix convert to markdown
+      toMarkdown: {
+        open: '**',
```
mattcasey, line 29: i didn't know u could do this, i thought toMarkdown() had to be a function. i'd like to learn more about it though, and what needs to be fixed still?

lib/votes/updateVotesStatus.ts
```diff
@@ -0,0 +1,64 @@
...
+  const pageVoteNewStatusRecord: Record<'Passed' | 'Rejected', string[]> = {
+    Passed: [],
+    Rejected: []
+  };
+
+  const updatedPageVote = pageVotes.map(pageVote => {
+    let status = pageVote.status;
+    const { deadline, userVotes, threshold } = pageVote;
+    const totalVotes = userVotes.length;
+    const userVoteFrequencyRecord: Record<string, number> = pageVote.userVotes.reduce<Record<string, number>>((currentRecord, userVote) => {
```
mattcasey, line 14: just to keep the line short, you could remove the redundant type annotation right after the var name `: Record<string, number>` it's implied since you pass it to reduce()

components/common/CharmEditor/components/Paragraph.tsx
Devorein: Done. Thank you 😄

lib/votes/castVote.ts
Devorein: Yes, I initially added it to save some lines of code in the fe when we update the cache after casting a vote. But I can easily use to currently logged-in user to do that. Thank you. I've made the changes and confirmed that it works without returning the user.

lib/votes/createVote.ts
Devorein: It was done to stop typescript errors as `ExtendedVote` required `UserVotes` with `User` data.

lib/votes/getPageVotes.ts
Devorein: The users are needed to show the history of the votes. We are using `UserDisplay` component to show the username and the avatar, and it takes in the `user` as a prop

lib/votes/getSpaceVotes.ts
Devorein: This query will most likely be used to fetch all the votes for all the spaces the user is part of. I think joining the user along with the votes for this query might be unnecessary. I will remove it now and discuss with Alex what data we need to show on the nexus page to get a clearer picture. But the user is required for the `UserDisplay` component that is used when showing the list of votes casted

lib/votes/getPageVotes.ts
```diff
@@ -1,18 +1,24 @@
...
+          user: true
         }
       },
       voteOptions: true
     }
   });
+
+  return updateVotesStatus(pageVotes);
```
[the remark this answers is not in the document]
Devorein: Yes, I agree. I completely missed the task that's why I thought to do it in this PR. Thank you. I will remove it.

lib/votes/getSpaceVotes.ts
Devorein: Thank you. Updated.

lib/votes/getVote.ts
Devorein: Thank you. Updated.

pages/_app.tsx
Devorein: Yep, but I think I can move them a lot closer to where they are being used. For example right above the provider for Page action sidebar.

hooks/usePageActionDisplay.tsx
Devorein: Oops. Sorry for that. Thank you for catching.

components/common/CharmEditor/components/PageInlineVote.tsx
Devorein: Casting was actually done to resolve a typescript issue (otherwise we would need another much worse casting `any`). But I wasn't aware of that method.  Thank you. I will keep that in mind from now on.
image

components/common/CharmEditor/components/PageInlineVote.tsx
Devorein: Yeah, that sounds a lot better. I will go with the first one since we are already displaying the vote title on the button itself. Thank you.

hooks/usePageActionDisplay.tsx
Devorein: Well, it's a hook that's also exporting a `Context`. It's more like `usePages.tsx` rather than a module exporting only a hook.

lib/inline-votes/findTotalInlineVotes.ts
Devorein: Yes absolutely. Thank you for catching that.

lib/inline-votes/removeInlineVoteMark.ts
Devorein: Good point. I actually had the version you are talking about in stash. I did some experiment to automatically remove the completed vote marks from the editor the first time it loads, but its kind of hacky and since we already decided on a bg job for updating the status of the vote we can do the document manipulation there. But I will update it so that if in any case, we need this function, it's not dispatching multiple transactions at once.

components/common/CharmEditor/components/inlineComment/inlineComment.plugins.ts
Devorein: Updated. Thanks.

lib/prosemirror/highlightMarkedElement.ts
Devorein: Very good point. I'm not sure why I went with classes knowing its going to be unique 🤔

components/common/CharmEditor/components/Paragraph.tsx
Devorein: Yes in fact `cardId` is not used for its value. Thank you for that. I've updated it.

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Yes absolutely on it

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Nice one. Updated.

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Alright done

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Done

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Sure

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Well splice returns the removed items so we can't set those in the state

components/common/CharmEditor/components/inlineVote/inlineVote.specs.ts
Devorein: yes, I have created a [card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=efaf4859-4372-401e-9f10-0e7c829b1fef) listing the nodes that still need markdown export support. This is probably not the correct way to export votes (I copied from the heading node when creating the spec for this one that's why its there)

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: No the component is only shown when you click the vote option in the floating menu. So there is no chance it renders every time you type something in the editor.

pages/api/pages/[id]/votes.ts
Comment (Devorein) [reply pages/api/pages/[id]/votes.ts:32]: That's a good point. I will have to add it for fetching all the threads of a page as well.

pages/_app.tsx
Devorein: Update: Relocating the provider didn't fetch the votes and threads for public pages. I reverted back the changes

lib/votes/interfaces.ts
Devorein: I've removed that interface as I found a way without it.

lib/votes/createVote.ts
Devorein: Update: I've cast the return type as `ExtendedVote` and removed the include as its redundant

lib/votes/getPageVotes.ts
Devorein: Finally made the adjustments to only fetch the data required by fe. Just the `username` and `avatar`

lib/votes/getVote.ts
Devorein: Yes, I've done just that in the latest update. Thank you 😄

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
Devorein: Its been implemented after discussion.

lib/votes/updateVotesStatus.ts
Devorein: Yes thank you for catching that its been resolved.

mattcasey (approved):

components/common/CharmEditor/components/FloatingMenu.tsx
Devorein: it's not possible unfortunately as the types are coming from a component we don't control.
image

components/common/CharmEditor/components/FloatingMenu.tsx
Devorein: Thank you. I've made the changes 😄

components/common/CharmEditor/components/PageInlineVote.tsx
Devorein: Thank you. I've updated it.

PR #652, opened 2022-06-29
Vote api

AndreiMitrea:
API for voting functionality.


lib/votes/createVote.ts
```diff
@@ -0,0 +1,87 @@
...
+          id: createdBy
+        }
+      }
+    },
+    include: {
+      voteOptions: true
+    }
+  });
+
+  await prisma.voteOptions.createMany({
```
Devorein, line 76: I think we should use a transaction to create the vote and the vote options

lib/votes/createVote.ts
```diff
@@ -0,0 +1,87 @@
...
+        }
+      },
+      author: {
+        connect: {
+          id: createdBy
+        }
+      }
+    },
+    include: {
+      voteOptions: true
```
Devorein, line 72: Maybe only `select: {id: true}` as that's the only field we are using below

lib/votes/createVote.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+  });
+
+  await prisma.voteOptions.createMany({
+    data: options.map(option => ({
+      name: option.name,
+      threshold: option.threshold || DEFAULT_THRESHOLD,
```
Devorein, line 79: Maybe use `??` here rather than `||` since the user might pass `0` (an invalid threshold but it might be purposeful in some context)

lib/votes/deleteVote.ts
```diff
@@ -0,0 +1,29 @@
+import { Vote } from '@prisma/client';
+import { hasAccessToSpace } from 'lib/users/hasAccessToSpace';
+import { DataNotFoundError } from 'lib/utilities/errors';
+import { prisma } from 'db';
+import { getVote } from './getVote';
+
+export async function deleteVote (id: string, userId: string): Promise<Vote | null> {
+  const vote: Vote | null = await getVote(id);
```
Devorein, line 8: Shouldn't `getVote` return the correct type?

lib/votes/getPageVotes.ts
```diff
@@ -0,0 +1,14 @@
+import { Vote } from '@prisma/client';
+import { prisma } from 'db';
+
+export async function getPageVotes (pageId: string): Promise<Vote[]> {
```
Devorein, line 4: I believe it would be best to let prisma handle the return value or have something like `(Vote & {userVotes: UserVote[], voteOptions: VoteOption[]})[]` since the consumer would be missing out on those

lib/votes/getSpaceVotes.ts
```diff
@@ -0,0 +1,14 @@
+import { Vote } from '@prisma/client';
+import { prisma } from 'db';
+
+export async function getSpaceVotes (spaceId: string): Promise<Vote[]> {
```
Devorein, line 4: Same issue as above. Maybe have a interface inside `lib/votes/interfaces.ts` that exports `Vote` with the joined data?

lib/votes/interfaces.ts
```diff
@@ -0,0 +1,19 @@
+import { Vote } from '@prisma/client';
+
+export const DEFAULT_THRESHOLD = 50;
+
+export const VOTE_STATUS = ['InProgress', 'Passed', 'Rejected', 'Cancelled'] as const;
```
Devorein, line 5: I think we should reuse the `VoteStatus` enum?

lib/votes/updateVote.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const updatedVote = await prisma.vote.update({
+    where: {
+      id
+    },
+    data: {
+      status
+    },
+    include: {
+      voteOptions: true
```
Devorein, line 37: Shouldn't we return the `userVotes` as well? It might be available in the frontend, but it would required duct taping the existing value in the cache, rather than completely overriding it

lib/votes/getPageVotes.ts
```diff
@@ -0,0 +1,14 @@
...
+  return prisma.vote.findMany({
+    where: {
+      pageId
+    },
+    include: {
+      userVotes: true,
```
Devorein, line 10: Would it be possible to sort the `userVotes` via `updatedAt` field in descending order? It would save some burden on the fe

Devorein (changes requested):

lib/votes/getPageVotes.ts
AndreiMitrea: `Vote` is taken from `@prisma/client`, I think it includes what is on `schema.prisma`, including:
```
  userVotes     UserVote[]
  voteOptions   VoteOptions[]
  ```
  The get page votes endpoint returns the values.

File: lib/votes/updateVote.ts (hunk 8 above)
Comment (AndreiMitrea) [reply lib/votes/updateVote.ts:37]: Done, thank you!

File: lib/votes/createVote.ts (hunk 2 above)
Comment (AndreiMitrea) [reply lib/votes/createVote.ts:72]: Done, thank you!

File: lib/votes/interfaces.ts (hunk 7 above)
Comment (Devorein) [reply lib/votes/interfaces.ts:5]: Alright lets keep it like that then

File: lib/votes/createVote.ts (hunk 1 above)
Comment (AndreiMitrea) [reply lib/votes/createVote.ts]: Done, thank you.

File: lib/votes/getSpaceVotes.ts (hunk 6 above)
Comment (AndreiMitrea) [reply lib/votes/getSpaceVotes.ts]: Added `ExtendedVote` model, thank you.

File: lib/votes/getPageVotes.ts (hunk 5 above)
Comment (AndreiMitrea) [reply lib/votes/getPageVotes.ts]: Added `ExtendedVote` model, thank you.

File: lib/votes/createVote.ts (hunk 1 above)
Comment (Devorein) [reply lib/votes/createVote.ts]: Awesome 😄

File: lib/votes/getPageVotes.ts (hunk 5 above)
Comment (Devorein) [reply lib/votes/getPageVotes.ts]: Great 🥳

File: lib/votes/getSpaceVotes.ts (hunk 6 above)
Comment (Devorein) [reply lib/votes/getSpaceVotes.ts]: Yeah 🔥 This will help a lot. Thank you.

File: lib/votes/createVote.ts [hunk 10]
```diff
@@ -0,0 +1,92 @@
+
+import { prisma } from 'db';
+import { v4 as uuid } from 'uuid';
+import { Vote } from '@prisma/client';
+import { DataNotFoundError, InvalidInputError } from 'lib/utilities/errors';
+import { hasAccessToSpace } from 'lib/users/hasAccessToSpace';
+import { DEFAULT_THRESHOLD, VoteDTO, VOTE_STATUS } from './interfaces';
+import { getVote } from './getVote';
+
+export async function createVote (vote: VoteDTO): Promise<Vote | null> {
```
mattcasey, line 10: this function should always return a vote

lib/votes/deleteVote.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!vote) {
+    throw new DataNotFoundError(`There is no vote with id: ${id}`);
```
mattcasey, line 11: can we make this silently fail. there's no reason to throw an error if something doesn't exist that i asked to delete. just log a warning

lib/votes/createVote.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const { createdBy, pageId, title, description, deadline, voteOptions } = vote;
+
+  if (!pageId) {
+    throw new InvalidInputError('Please provide the id of the page where the vote is taking place.');
```
mattcasey, line 15: this validation belongs in the Next.js middleware

lib/votes/createVote.ts
```diff
@@ -0,0 +1,92 @@
...
+          connect: {
+            id: createdBy
+          }
+        }
+      },
+      select: {
+        id: true
+      }
+    }),
+    prisma.voteOptions.createMany({
```
mattcasey, line 80: you should be able to create this as part of the vote.create() method

pages/api/userVotes/index.ts
```diff
@@ -0,0 +1,26 @@
...
+import { NextApiRequest, NextApiResponse } from 'next';
+import nc from 'next-connect';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireUser)
+  .post(castVote);
+
+async function castVote (req: NextApiRequest, res: NextApiResponse<UserVote | { error: any }>) {
```
mattcasey, line 15: Please try to keep all our vote api under a single endpoint. This one can go under: /votes/[id]/cast

prisma/schema.prisma
```diff
@@ -157,19 +158,30 @@ enum VoteStatus {
   Cancelled
 }
 
+model VoteOptions {
+  name        String
+  threshold   Int
```
mattcasey, line 163: threshold should be the same for all options, please move it to Vote

mattcasey (approved):

PR #659, opened 2022-07-01
Payment icons

mattcasey:
This PR adds a default icon (the chain logo) for all tokens. I think it's better than leaving a whitespace... also mirrors what you see on Etherscan.

I also fixed a bug wehre when you select a custom token in InputSearchCrypto, we show the token address rather than the token name. I think that the code for this component could be simplified further, but saving for another day..

image

### DIFF
--- a/components/bounties/components/BountyStatusBadge.tsx
+++ b/components/bounties/components/BountyStatusBadge.tsx
@@ -14,7 +14,7 @@ import { Bounty, BountyStatus } from '@prisma/client';
 import { getChainById } from 'connectors';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
 import { usePaymentMethods } from 'hooks/usePaymentMethods';
-import { getTokenInfo } from 'lib/tokens/tokenData';
+import { getTokenAndChainInfoFromPayments } from 'lib/tokens/tokenData';
 import millify from 'millify';
 import { BOUNTY_LABELS } from 'models/Bounty';
 import Link from 'next/link';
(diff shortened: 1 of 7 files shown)


components/common/form/InputSearchCrypto.tsx
```diff
@@ -41,19 +42,17 @@ export function InputSearchCrypto ({
   const ERC20PopupState = usePopupState({ variant: 'popover', popupId: 'ERC20-popup' });
 
   useEffect(() => {
-    setInputValue(defaultValue);
```
mattcasey, line 44: @motechFR please make sure my assumptions are correct we can remove this call to setInputValue. it is what was causing us to flash the contract address briefly in the bounty editor form

connectors.ts
```diff
@@ -332,7 +332,7 @@ export function getCryptos (chainId: number): Array<string | CryptoCurrency> {
 
 }
 
-export function getChainExplorerLink (chainId: string | number, transactionOrContractId: string, endpoint: 'transaction' | 'token' = 'transaction'): string {
+export function getChainExplorerLink (chainId: string | number, transactionOrContractId: string, endpoint: 'transaction' | 'address' = 'transaction'): string {
```
motechFR, line 335: Agree, address is more straightforward than token (especially since we expect a contract address)

components/common/form/InputSearchCrypto.tsx
motechFR: @mattcasey agreed, this component was one of earliest I worked on, so probably an antipattern that came from lack of knowledge of React MUI at the time.

I've just tested the following components
- CharmEditor crypto widget
- Bounty editor form

Both of them are looking good

components/bounties/components/BountyStatusBadge.tsx
```diff
@@ -154,24 +153,15 @@ export function BountyAmount ({ bounty, truncate = false }: { bounty: Pick<Bount
...
-              tokenLogo && (
-                <Box component='span' sx={{ width: '25px', height: '25px' }}>
-                  <img
-                    height='100%'
-                    src={tokenLogo}
-                  />
-                </Box>
-              )
-            }
+                  <TokenLogo src={tokenInfo.canonicalLogo} />
```
motechFR, line 164: Nifty simplification 👍

lib/tokens/tokenData.ts
```diff
@@ -91,3 +91,46 @@ export function getTokenInfo (paymentMethods: PaymentMethod[], symbolOrAddress:
...
+  return {
+    // prefer our standard iconUrl for native tokens that may have been saved to tokenInfo
+    canonicalLogo: paymentMethod.contractAddress ? (tokenLogo || chain.iconUrl) : chain.iconUrl,
+    chain,
+    tokenName: paymentMethod.tokenName,
+    tokenSymbol: paymentMethod.tokenSymbol,
+    tokenLogo,
+    isContract: !!paymentMethod.contractAddress
+  };
+}
```
motechFR, line 136: This whole file is a good idea. Connectors.ts was getting a bit crowded!

PR #660, opened 2022-07-01
remove script api

mattcasey:
Running a next.js app just to do a migration is overkill. This is our production code, and a dev could easily leave test/dev code in here. Even just importing code from scripts could lead to unintentional consequences in production. 

I'd like to hear what issues you had with ts-node? we can talk and find a better solution


motechFR (approved):
@mattcasey agree, this would be a security risk if someone accidentally committed it with a live script

PR #661, opened 2022-07-01
redirect user to the first page that appears in the sidebar by default

mattcasey:


components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
```diff
@@ -48,10 +55,10 @@ function mapTree (items: MenuNode[], key: 'parentId', rootPageIds?: string[]): P
 
   for (i = 0; i < tempItems.length; i += 1) {
     node = tempItems[i];
-    const nodeIndex = node[key];
```
mattcasey, line 51: In mapTree, I hard-coded 'parentId' and removed the option to pass in a custom key. I think it's slightly more readable too

Devorein (approved):

PR #663, opened 2022-07-03
vote tasks in nexus page

Devorein:
[ x ] Show vote tasks tab on nexus page
[ x ] Send email for vote tasks


background/tasks/sendNotifications/sendNotifications.ts
```diff
@@ -50,32 +51,40 @@ export async function getNotifications (): Promise<PendingTasksProps[]> {
...
-    const myGnosisTasks = tasksNotSent.filter(gnosisSafeTask => Boolean(gnosisSafeTask.tasks[0].transactions[0].myAction));
+    const sentTaskIds = new Set(sentTasks.map(sentTask => sentTask.taskId));
+
+    const voteTasksNotSent = voteTasks.filter(voteTask => !sentTaskIds.has(voteTask.id));
+    const gnosisSafeTasksNotSent = gnosisSafeTasks.filter(gnosisSafeTask => !sentTaskIds.has(getGnosisSafeTaskId(gnosisSafeTask)));
+    const myGnosisTasks = gnosisSafeTasksNotSent.filter(gnosisSafeTask => Boolean(gnosisSafeTask.tasks[0].transactions[0].myAction));
 
     log.debug('Found tasks for notification', {
-      notSent: tasksNotSent.length,
+      notSent: gnosisSafeTasksNotSent.length,
```
mattcasey, line 74: include voteTasksNotSent?

components/nexus/VoteTasksList.tsx
```diff
@@ -0,0 +1,133 @@
...
+          fontSize='small'
+          sx={{
+            position: 'absolute',
+            left: -35,
+            top: '50%',
+            transform: 'translateY(-50%)'
+          }}
+        />
+      ) : null}
+      <Link target='_blank' href={voteLink}>
```
mattcasey, line 42: kind of feel that we shouldn't open a new tab, since you're still on our app

mattcasey:
I noticed a bug when changing threshold in the UI to 100 votes, it tries to save the value as a string and throws an error:
image

mattcasey:
can u put a button on the right that says "Vote now" ? Right now i'm not sure if i should click on rows or what i'm supposed to do
image

background/tasks/sendNotifications/sendNotifications.ts
Devorein: Yeah sure. I was unsure of whether it should only log gnosis tasks not sent or other ones.

components/nexus/VoteTasksList.tsx
Devorein: Hmm, you are right. I thought of doing that since they might want one tab to view the list of tasks and one to perform an action on them. But they can already do so using `ctrl + click` so its redundant

components/nexus/VoteTasksList.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }: VoteTask
+) {
+  const baseUrl = typeof window !== 'undefined' ? window.location.origin : null;
+
+  const voteLink = `${baseUrl}/${spaceDomain}/${pagePath}?voteId=${voteId}`;
+  const voteLocation = `${pageTitle || 'Untitled'} in ${spaceName}`;
+
+  return (
+    <Box>
+      <Link href={voteLink}>
```
mattcasey, line 31: this component seems to make the whole row slightly transparent on hover, which is distracting now that we have a button. rows should get darker on hover, if we do anything (following MUI table styles), but i don't think we need to have any styling. i dont really think you need to make the whole row clickable either

Devorein: Yeah good point. Since there is a CTA button we should remove the hover styling for the row. I will make the changes

Devorein: Ah I remember it now. I did that to have an easier time when creating the mock data when viewing the email locally for vote tasks, but as you said its not necessary and over optimization

mattcasey (approved):

lib/votes/getVoteTasks.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+      // No need to fetch votes that are passed deadline, those can't be voted on
+      deadline: {
+        gte: new Date()
+      }
+    },
+    orderBy: {
+      deadline: 'asc'
+    },
+    select: {
```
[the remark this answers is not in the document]
Devorein: Ah, I remember it now. I did that to have an easier time when creating the mock data when viewing the email locally for vote tasks, but as you said it's not necessary, and over-optimization

PR #664, opened 2022-07-04
move tasks badge to its own icon

mattcasey:
Preview:

image
image

I think the count number is pretty large.. but right now there's no way to know what the little dot would mean on its own. So I want to change it to a dot once we have a dropdown menu of all the tasks (a la Reddit)


components/common/PageLayout/components/Header/components/NotificationsBadge/NotificationsBadge.tsx
```diff
@@ -0,0 +1,42 @@
...
+      ? (userNotificationState.snoozedUntil && new Date(userNotificationState.snoozedUntil) > new Date() ? 0
+        : tasks.gnosis.length) : tasks.gnosis.length))
+    : 0;
+
+  return (
+    <Link href='/nexus'>
+      <IconButton size='small' sx={{ mr: 1 }}>
+        <Tooltip arrow title='Your Notifications'>
+          <Badge
+            badgeContent={2}
```
Devorein, line 25: This should be changed to `totalTasks`

Devorein (approved):

PR #667, opened 2022-07-05
Page level voting

Devorein:
[ x ] Button to create a page-level vote (subject to modification)
image
Reused the modal component when creating a vote by selecting text.


hooks/useInlineVotes.tsx
```diff
@@ -104,6 +105,20 @@ export function InlineVotesProvider ({ children }: { children: ReactNode }) {
...
       spaceId: currentSpace!.id
     });
+
+    mutateTasks((tasks) => {
+      // Add the vote to the task
+      if (tasks) {
+        tasks.votes.push({
+          ...extendedVote,
+          space: currentSpace!,
+          page: pages[currentPageId]!
```
mattcasey, line 115: it's better to add this to your check on line 111, so we don't need the "!" assertion.. its also a little safer

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -158,36 +162,53 @@ export default function Header ({ open, openSidebar, hideSidebarOnSmallScreen }:
...
-                          mr: 1
-                        }}
-                      />
-                      <ListItemText primary='View votes' />
-                    </ListItemButton>
-                  </>
+                    <>
+                      {currentSpacePermissions?.createVote && (
+                      <ListItemButton onClick={() => {
+                        setIsModalOpen(true);
```
mattcasey, line 168: should call setPageMenuOpen(false) here

components/common/CharmEditor/components/inlineVote/inlineVote.components.tsx
```diff
@@ -1,34 +1,21 @@
 import { Mark, MarkType, PluginKey, TextSelection } from '@bangle.dev/pm';
```
mattcasey, line 1: FYI these components should probably go in a folder, each with their own file

PR #668, opened 2022-07-05
Public bounties

motechFR:
Support for public bounties within a space that can be shared with the public


mattcasey (approved):
tested locally

PR #670, opened 2022-07-06
only include favorites of pages not deleted

mattcasey:
I noticed Alex was seeing the "Favorites" label in his sidebar with no pages underneath. I don't know if this was his case, but I know it is a bug - that we don't check which favorites have been deleted. Once the page is deleted after 30 days, it should be cleared, but this check helps until that happens. 

Currently what it looks like:
image

To test: add a page to your favorites and delete it. Reload the site.

### DIFF
--- a/lib/session/config.ts
+++ b/lib/session/config.ts
@@ -1,3 +1,4 @@
+import { Prisma } from '@prisma/client';
 
 declare module 'iron-session' {
   interface IronSessionData {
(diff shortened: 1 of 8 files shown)


lib/session/config.ts
```diff
@@ -15,3 +16,29 @@ export const ironOptions = {
     secure: typeof process.env.DOMAIN === 'string' && process.env.DOMAIN.includes('https')
   }
 };
+
+// a map of relationships we pull in for the logged-in user (try to keep this small)
+export const sessionUserRelations = {
```
mattcasey, line 21: i moved the relationships here since it was copy/pasted in a lot of places and was inconsistent.

Devorein: Makes sense, but maybe use type from prisma?

Devorein (approved):

lib/session/config.ts
Devorein: Make sense, but maybe use types from prisma?

PR #672, opened 2022-07-06
Vote job

AndreiMitrea:
Created background job for updating vote status.

### DIFF
--- a/background/tasks/updateVotesStatus/updateVoteStatus.ts
+++ b/background/tasks/updateVotesStatus/updateVoteStatus.ts
@@ -1,0 +1,101 @@
+import { prisma } from 'db';
+import { UserVote } from '@prisma/client';
+import { ExtendedVote, VOTE_STATUS } from 'lib/votes/interfaces';
+
+const getVotesByState = async (votes: ExtendedVote[]) => {
+
+  const passedVotes: string[] = [];
+  const rejectedVotes: string[] = [];
+
+  for (const vote of votes) {
+    if (vote.userVotes.length === 0) {
+      rejectedVotes.push(vote.id);
+      continue;
+    }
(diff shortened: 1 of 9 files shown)


background/tasks/updateVotesStatus/updateVoteStatus.ts
```diff
@@ -0,0 +1,101 @@
...
+        maxChoices = [choices[index]];
+      }
+      else if (currentIndex - index === maxCount) {
+        maxChoices.push(choices[index]);
+      }
+
+      index = currentIndex;
+    }
+
+    const maxChoicePercentage = maxCount * 100 / choices.length;
```
mattcasey, line 39: these last 10 or so lines seem like some logic that we would already have somewhere.. could we reuse it?

background/tasks/updateVotesStatus/updateVoteStatus.ts
```diff
@@ -0,0 +1,101 @@
...
+    passedVotes,
+    rejectedVotes
+  };
+};
+
+const updateVoteStatus = async () => {
+
+  const votesPassedDeadline = await prisma.vote.findMany({
+    where: {
+      status: VOTE_STATUS[0],
```
mattcasey, line 59: i think it would be more clear if you just used the actual value.. is this "InProgress"?

AndreiMitrea: Yes

background/tasks/updateVotesStatus/updateVoteStatus.ts
AndreiMitrea: Not sure.

mattcasey (approved):

PR #675, opened 2022-07-06
Lint background folder when running next lint

mattcasey:


lib/middleware/requireKeys.ts
```diff
@@ -22,12 +22,6 @@ export function requireKeys<T> (keys: Array<keyof T>, location: 'body' | 'query'
 
     for (const key of keys) {
 
-      // Null value is allowed for this key, we can proceed to check the next key
-      if (nullableKeys.indexOf(key) > -1 && toVerify[key] === null) {
```
mattcasey, line 26: @motechFR rather than fix this line, I removed the check. Since we check the value of `toVerify[key]`, the next check: `key in toVerify` is also true.

next.config.js
```diff
@@ -31,6 +31,10 @@ const config = {
   generateEtags: false,
   poweredByHeader: false,
   webpack5: true,
+  eslint: {
+    // add background to the default list of pages for eslint
+    dirs: ['pages', 'components', 'lib', 'background']
```
mattcasey, line 36: the fix

AndreiMitrea (approved):

PR #677, opened 2022-07-08
Calculations of vote state in the backend before sending to front-end

Devorein:
[ x ] calculate vote status based on deadline (in case the background job hasn't run on it)
[ x ] Don't send user votes, just the aggregated results
[ x ] add a new endpoint to get User Vote history/details


AndreiMitrea (approved):
Code review

PR #679, opened 2022-07-08
delete the new page if permissions fail after creating it

mattcasey:
Temporary fix


motechFR (approved):
Thanks for this, page permissions is on my upcoming to do list !

PR #683, opened 2022-07-11
Search inside workspace improvements

AndreiMitrea:
Made imrovements for searching inside workspace:
- highlight matched text
- hide options when not searching
- add path to result option if it is nested
- vertical scrollbar is at the search results level, not modal level
- add modal positioning

image


mattcasey (changes requested):
Request changes:
- clicking outside the search input but inside the modal clears the input for some reason
- make the "x" clear icon black, or at least use a standard blue
- hide the up/down arrow on the right. users shouldn't need/be able to toggle the list

mattcasey (approved):

PR #685, opened 2022-07-12
Update fb views in ls when referencing non nonexistent view

Devorein:
Update focalboard view record for the following cases:
[ x ] after deleting the board
[ x ] after duplicating the board
[ x ] if a non-existent view is being referenced

### DIFF
--- a/components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
+++ b/components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
@@ -3,7 +3,7 @@ import Typography from '@mui/material/Typography';
 import { useFocalboardViews } from 'hooks/useFocalboardViews';
 import useRefState from 'hooks/useRefState';
 import { Page } from 'models';
-import { useCallback, useRef, memo } from 'react';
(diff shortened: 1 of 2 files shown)


mattcasey (approved):

components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
```diff
@@ -121,6 +121,14 @@ function DraggableTreeNode ({ item, onDropAdjacent, onDropChild, pathPrefix, add
   const { expanded } = useTreeItem(item.id);
   const hideChildren = !expanded;
 
+  useEffect(() => {
+    const focalboardViewId = focalboardViewsRecord[item.id];
+    if (focalboardViewId && item.type === 'board' && !views.find(view => view.id === focalboardViewId)) {
```
mattcasey, line 126: should use view.some() when you want a boolean return value

Devorein: Right. Replacing it right now.

PR #692, opened 2022-07-13
Fixed sending mentions for pages and bounties created by user

Devorein:


jest.config.ts
```diff
@@ -30,7 +31,7 @@ export const jestConfig = {
 
   // The glob patterns Jest uses to detect test files
   testMatch: [
-    '**/lib/**/*.spec.ts'
+    path.join(__dirname, '**/lib/votes/__tests__/deleteVote.spec.ts')
```
mattcasey, line 34: was this a mistake?

mattcasey (changes requested):
i think test files were changed by accident

jest.config.ts
Devorein: yes yes, correct. Let me remove that

PR #695, opened 2022-07-13
Remember last visited workspace pages and non-workspace page

AndreiMitrea:
When accessing workspace, go to its last visited page. Go to last visited non-workspace page when accessing base url.

### DIFF
--- a/components/common/RouteGuard.tsx
+++ b/components/common/RouteGuard.tsx
@@ -3,7 +3,7 @@ import { useRouter } from 'next/router';
 import { useWeb3React } from '@web3-react/core';
 import { Web3Connection } from 'components/_app/Web3ConnectionManager';
 import { User } from '@prisma/client';
-import { useLocalStorage } from 'hooks/useLocalStorage';
+import { getKey } from 'hooks/useLocalStorage';
 import { useUser } from 'hooks/useUser';
 import { useSpaces } from 'hooks/useSpaces';
 import { isSpaceDomain } from 'lib/spaces';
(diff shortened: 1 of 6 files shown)


pages/[domain]/index.tsx
```diff
@@ -10,21 +11,27 @@ export default function RedirectToMainPage () {
   const router = useRouter();
   const [space] = useCurrentSpace();
   const { pages } = usePages();
+  const defaultPageKey: string = space?.domain ? getKey(`last-page-${space.domain}`) : '';
+  const defaultPage = defaultPageKey ? (typeof window !== 'undefined' && localStorage.getItem(defaultPageKey)) : null;
 
   useEffect(() => {
+    if (defaultPage && space && defaultPage.startsWith(`/${space.domain}/`)) {
```
Comment (mattcasey) [inline pages/[domain]/index.tsx:18]: why do we need to check that space exists or that defaultPage starts with the domain? shouldn't we be able to rely on line 14 to handle those cases?

Comment (mattcasey) [reply pages/[domain]/index.tsx:18]: we should make sure the defaultPage exists in the pages, it's possible the pages have not yet loaded for this space as well, if you're switching between spaces. we may need to either save the pageId instead of the path, or parse the path an compare the paths on each page.

components/common/RouteGuard.tsx
```diff
@@ -25,14 +26,32 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
   const isRouterLoading = !router.isReady;
   const isLoading = !isUserRequestComplete || isWalletLoading || isRouterLoading || !isSpacesLoaded;
 
+  if (typeof window !== 'undefined') {
+    const pathSegments: string[] = router.asPath.split('?')[0].split('/').filter(segment => !!segment);
+    const firstSegment: string = pathSegments[0];
```
mattcasey, line 31: I'm curious what's the reason you assert the types when defining variables? For me, it  makes the line longer and harder to read. It also doesn't mirror the style we normally follow

components/common/RouteGuard.tsx
```diff
@@ -25,14 +26,32 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
   const isRouterLoading = !router.isReady;
   const isLoading = !isUserRequestComplete || isWalletLoading || isRouterLoading || !isSpacesLoaded;
 
+  if (typeof window !== 'undefined') {
+    const pathSegments: string[] = router.asPath.split('?')[0].split('/').filter(segment => !!segment);
```
mattcasey, line 30: can you extract this logic and add some tests on it?

components/common/RouteGuard.tsx
```diff
@@ -25,14 +26,32 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
...
+    const firstSegment: string = pathSegments[0];
+    const isDomain: boolean = !!isSpaceDomain(firstSegment) || firstSegment === 'nexus';
+    const workspaceDomain = isDomain ? firstSegment : null;
+    const defaultPageKey: string = workspaceDomain ? getKey(`last-page-${workspaceDomain}`) : '';
+    const defaultWorkspaceKey: string = getKey('last-workspace');
+    const defaultPage = defaultPageKey ? localStorage.getItem(defaultPageKey) : null;
+    const defaultWorkspace = localStorage.getItem(defaultWorkspaceKey);
+
+    if (workspaceDomain && workspaceDomain !== defaultWorkspace) {
+      localStorage.setItem(defaultWorkspaceKey, workspaceDomain);
```
mattcasey, line 40: I think it would be more efficient to store the whole router.asPath, so we don't redirect the user twice

lib/spaces/index.ts
```diff
@@ -16,5 +16,5 @@ export const DOMAIN_BLACKLIST = [
 ];
 
 export function isSpaceDomain (domain: string) {
-  return domain && !DOMAIN_BLACKLIST.includes(domain);
+  return domain && !DOMAIN_BLACKLIST.includes(domain) && !domain.includes('?');
```
mattcasey, line 19: This was a great find, but I am sure we either had bugs before or will introduce new bugs with this change. Whenever we've used this method, I guarantee we just didn't consider the string might have a querystring at the end. I suggest we just strip out the querystring part in this method rather than just returning false, otherwise we should strip it out wherever this method is called.

components/common/RouteGuard.tsx
```diff
@@ -25,14 +26,32 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
...
   const isLoading = !isUserRequestComplete || isWalletLoading || isRouterLoading || !isSpacesLoaded;
 
+  if (typeof window !== 'undefined') {
+    const pathSegments: string[] = router.asPath.split('?')[0].split('/').filter(segment => !!segment);
+    const firstSegment: string = pathSegments[0];
+    const isDomain: boolean = !!isSpaceDomain(firstSegment) || firstSegment === 'nexus';
+    const workspaceDomain = isDomain ? firstSegment : null;
+    const defaultPageKey: string = workspaceDomain ? getKey(`last-page-${workspaceDomain}`) : '';
+    const defaultWorkspaceKey: string = getKey('last-workspace');
+    const defaultPage = defaultPageKey ? localStorage.getItem(defaultPageKey) : null;
```
mattcasey, line 36: It doesn't seem like you need to retrieve defaultPage or defaultWorkspace. You're retrieving it just to see if we don't need to set it again, which I think defeats the purpose of optimization. It would also reduce the complexity of this code block

mattcasey (changes requested):
see comments

pages/[domain]/index.tsx
Comment (AndreiMitrea) [reply pages/[domain]/index.tsx:18]: > why do we need to check that space exists or that defaultPage starts with the domain? shouldn't we be able to rely on line 14 to handle those conditions?

Not really

components/common/RouteGuard.tsx
AndreiMitrea: I am not consistent with this but I am used to seeing types when declaring variables. Probably from other programming languages where the types are not inferred.

mattcasey: ah ok. i'm coming at it from the other way :) I think it doesn't hurt to be explicit, at least the types are usually colored in an IDE, and I can't imagine working on our codebase w/o one.

mattcasey (approved):

PR #697, opened 2022-07-14
Using async sema to throttle discord import request

Devorein:


pages/api/discord/importRoles.ts
```diff
@@ -20,6 +21,11 @@ export interface ImportDiscordRolesPayload {
 
 export type ImportRolesResponse = { importedRoleCount: number };
 
+// requests per second = 10, timeUnit = 1sec
+const rateLimiter = RateLimit(10);
```
mattcasey, line 25: the rate limit is 50, i think 50 or 40-45 if you want to be cautious is fine.

Devorein: sure I will change that to 35 then

pages/api/discord/importRoles.ts
```diff
@@ -62,7 +68,8 @@ async function importRoles (req: NextApiRequest, res: NextApiResponse<ImportRole
       throw new Error('Guild member does not have a user property');
     }
     lastUserId = guildMember.user.id;
-    discordGuildMembersResponse = await handleDiscordResponse<DiscordGuildMember[]>(`https://discord.com/api/v8/guilds/${guildId}/members?limit=100&after=${lastUserId}`);
+    discordGuildMembersResponse = await handleDiscordResponse<DiscordGuildMember[]>(`https://discord.com/api/v8/guilds/${guildId}/members?limit=${MEMBERS_PER_REQUEST}&after=${lastUserId}`);
+    await rateLimiter();
```
mattcasey, line 72: not sure if you were copying from my code, but i think this rate limiter should go at the beginning of the loop, right?  otherwise we are waiting for no reason on the last iteration.

Devorein: yes correct. I've actually send a new commit with the change (before looking at this comment)

mattcasey (approved):

PR #699, opened 2022-07-14
Show inline action counter for headers

Devorein:


components/common/CharmEditor/components/Heading.tsx
```diff
@@ -0,0 +1,53 @@
+import { NodeViewProps, PluginKey } from '@bangle.dev/core';
```
mattcasey, line 1: I have been trying to group our prosemirror components by the name of the spec. Please put this inside a folder called `heading`.

mattcasey (approved):
Left one request for folder organization.

I just realized how this approach works with the existing 'heading' component: we're using the plugins and spec from bangle.dev, and this adds another plugin using NodeView. That's pretty clever.

components/common/CharmEditor/components/Heading.tsx
Devorein: Good point. Doing it rn.

PR #702, opened 2022-07-14
Default workspace pages

motechFR:
Populates default data for the workspace


motechFR:
Hi @mattcasey 

Good news: This functionality is now working.

We can generate a local export of the workspace pages as below screenshot using the script in scripts/exportData/exportSpacePages.ts

We can then take the path of the static data, and pass it to lib/pages/server/convertJsonPagesToPrisma.ts

This function will generate a tuple with the inputs for a Prisma transaction that will create blocks then pages.

Finally, in the POST /api/spaces endpoint, we use this convertJsonPagesToPrisma function to populate all page data for the space.

Feedback welcome on how it might be even more efficient!

Final todo: Fix up the 10 failing unit tests that all seem linked to page permissions.

image

motechFR:
When we re-re add the relationship in the future, we only need to change prisma invocations and only here

- PUT + POST pages
- lib / importFromNotion
- lib / createPublicDatabasePage

(plus re-enabling import + export functions which I commented out for now)

PR #707, opened 2022-07-17
Fixed comment marks being removed replying

Devorein:
[ x ] Relocated findChildrenByMark from `prosemirror-utils` to `lib/prosemirror`
[ x ] Added view and threads/votes as a dependency for finding inline action counter

**NOTE**: Please test it out on your end as part of the review


lib/prosemirror/findChildrenByMark.ts
```diff
@@ -0,0 +1,20 @@
+import { MarkType, Node } from '@bangle.dev/pm';
```
mattcasey, line 1: Please leave a comment with reference to the source code when copy/pasting code

mattcasey (changes requested):
We should figure out the reason why simply copying code from the library is a fix. This introduces the same method twice in our codebase, since prosemirror-utils is still a dependency, and we also lose any future improvements from the dependency.  It sounds like there is something else going on, since the functions are idempotent and you didn't have to change anything.

Devorein:
Yes please, that would be really helpful. I'm totally bamboozled as to why the copied version is working and not the original one.

lib/prosemirror/findChildrenByMark.ts
Devorein: Thank you. I completely missed that. Adding that part.

mattcasey:
Closing this, it's been fixed in https://github.com/charmverse/app.charmverse.io/pull/712

PR #708, opened 2022-07-18
Fixed popover jumping to top of page after closing

Devorein:


mattcasey (approved):
The commit message could give a bit more context. I'm not sure I noticed whatever the issue was... was it noticed in a specific popup? why does this fix work (could u reference the SO article you found, if there was one)?

Devorein:
Yes absolutely. So this is a fix for [this](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=ff261464-b62c-4e62-8a14-5890088825fb) card. I spent a lot of time previously trying to find a solution, as I assumed it was a prosemirror bug, and gave up eventually. But I gave it another shot and this time I just searched for the bug and a [MUI github issue](https://github.com/mui/material-ui/issues/10072) thread popped up. Used that prop and that was the solution thankfully.

PR #709, opened 2022-07-18
Fix parent page call in the duplicatePage method

motechFR:
Fixes a call we forgot to revert when removing parentPage relationship

### DIFF
--- a/lib/pages/server/duplicatePage.ts
+++ b/lib/pages/server/duplicatePage.ts
@@ -21,11 +21,7 @@ export async function duplicatePage (pageId: string, userId: string, parentId?: 
   const transactions: PrismaPromise<any>[] = [prisma.page.create({
     data: {
       id: createdPageId,
-      parentPage: parentIdToAssign ? {
-        connect: {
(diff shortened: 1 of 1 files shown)


mattcasey (approved):

lib/pages/server/duplicatePage.ts
```diff
@@ -21,11 +21,7 @@ export async function duplicatePage (pageId: string, userId: string, parentId?:
   const transactions: PrismaPromise<any>[] = [prisma.page.create({
     data: {
       id: createdPageId,
-      parentPage: parentIdToAssign ? {
```
mattcasey, line 24: interesting, i'm surprised that Typescript didn't catch this

mattcasey: Did some research.. Typescript doesn't warn when there are extra properties, at least without using some interesting use of generics: https://stackoverflow.com/questions/49580725/is-it-possible-to-restrict-typescript-object-to-contain-only-properties-defined

PR #710, opened 2022-07-18
Don't display user uuid on card when no username

AndreiMitrea:
Don't display the user uuid on a card's user property.


components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
```diff
@@ -51,7 +51,7 @@ const UserProperty = (props: Props): JSX.Element => {
     const workspaceUsersById = useAppSelector<{[key:string]: IUser}>(getWorkspaceUsers)
 
     if (props.readonly) {
-        return (<div className='UserProperty octo-propertyvalue readonly'>{workspaceUsersById[props.value]?.username || props.value}</div>)
+        return (<div className='UserProperty octo-propertyvalue readonly'>{workspaceUsersById[props.value]?.username || ''}</div>)
```
mattcasey, line 54: instead could you show "Not found" or something like "(missing name)" instead of nothing?

PR #711, opened 2022-07-18
Cascade board permissions down to the cards

motechFR:
This PR ensures that whenever we update permissions for a board, they get cascaded downwards


.vscode/settings.json
```diff
@@ -20,6 +20,8 @@
   "editor.codeActionsOnSave": {
     "source.organizeImports": false
   },
+  "editor.detectIndentation": false,
+  "editor.tabSize": 2,
```
mattcasey, line 24: this doesn't actually do anything right? since we aren't using tabs for spacing?

pages/api/permissions/index.ts
```diff
@@ -44,6 +45,12 @@ async function addPagePermission (req: NextApiRequest, res: NextApiResponse<IPag
     throw new ActionNotPermittedError('You cannot make page public.');
   }
 
+  const { type: pageType } = await prisma.page.findUnique({
+    where: {
+      id: pageId
+    }
+  }) as Page;
```
mattcasey, line 52: you should handle the case if page comes back null, or else the deconstruction will throw an error

pages/api/permissions/index.ts
```diff
@@ -53,15 +60,27 @@ async function addPagePermission (req: NextApiRequest, res: NextApiResponse<IPag
...
+  }
+  // Old behaviour where we setup permissions after a page permission is added
+  else {
+    const permissionsAfter = await prisma.pagePermission.count({
+      where: {
+        pageId
+      }
+    });
+
+    // TODO - This could be an async job, but there is a risk of the UI being out of sync
```
mattcasey, line 80: this doesn't look like a TODO. it is, technically async, and it's safer to wait for all async ops to complete before returning a response

pages/api/permissions/index.ts
```diff
@@ -53,15 +60,27 @@ async function addPagePermission (req: NextApiRequest, res: NextApiResponse<IPag
...
+    await prisma.$transaction([
+      args.updateManyArgs ? prisma.pagePermission.updateMany(args.updateManyArgs) : null,
+      args.createManyArgs ? prisma.pagePermission.createMany(args.createManyArgs) : null
+    ].filter(a => a !== null) as PrismaPromise<any>[]);
 
-  // TODO - This could be an async job, but there is a risk of the UI being out of sync
-  if (permissionsAfter > permissionsBefore) {
-    await setupPermissionsAfterPagePermissionAdded(createdPermission.id);
+  }
+  // Old behaviour where we setup permissions after a page permission is added
```
mattcasey, line 72: this isn't technically 'old' behavior right?

lib/permissions/pages/triggers/board-page-permission-updated.ts
```diff
@@ -0,0 +1,124 @@
...
+import { IPageWithPermissions } from 'lib/pages/server';
+import { DataNotFoundError } from 'lib/utilities/errors';
+import { BoardPagePermissionUpdated } from '../interfaces';
+
+/**
+ * Generates the prisma operations for the board and its descendants
+ * @param param0
+ * @returns
+ */
+export async function boardPagePermissionUpdated ({ boardId, permissionId }: BoardPagePermissionUpdated):
```
mattcasey, line 13: this seems like a good candidate for a test or two

pages/api/permissions/index.ts
```diff
@@ -53,15 +60,27 @@ async function addPagePermission (req: NextApiRequest, res: NextApiResponse<IPag
...
-      pageId
-    }
-  });
+  if (pageType === 'board') {
+    const args = await boardPagePermissionUpdated({ boardId: pageId, permissionId: createdPermission.id });
+
+    await prisma.$transaction([
+      args.updateManyArgs ? prisma.pagePermission.updateMany(args.updateManyArgs) : null,
+      args.createManyArgs ? prisma.pagePermission.createMany(args.createManyArgs) : null
+    ].filter(a => a !== null) as PrismaPromise<any>[]);
```
mattcasey, line 69: you can use `isTruthy` from lib/utilities/react and i don't think you'll need to add a type assertion here

lib/permissions/pages/triggers/board-page-permission-updated.ts
```diff
@@ -0,0 +1,124 @@
+import { PagePermission, Prisma } from '@prisma/client';
```
mattcasey, line 1: filenames should be camelcase. It looks like the permissions folder has a mix of hyphen and camelCase. Same principle as our UI: please try to be consistent

mattcasey (changes requested):
See my comments... I mostly care about cleaning up comments and the destructuring on the result from a db query

.vscode/settings.json
motechFR: I'm personally using tabs for spacing. Our codebase has a natural tab=2, but my VSCode was acting a bit weird since I reinstalled it.

I plugged this in for code consistency. It won't change anything, only ensuring that tabs still continue to look like what we have so far.

pages/api/permissions/index.ts
motechFR: Agreed, at the time I made a note here as I wasn't sure how long permission cascading would take.

Since we are working on eliminating race conditions, and a more performant assignment process, this TODO is now irrelevant

pages/api/permissions/index.ts
motechFR: You're right, this is an inaccurate statement. Will update accordingly

lib/permissions/pages/triggers/board-page-permission-updated.ts
```diff
@@ -0,0 +1,124 @@
...
+        pageId: childPage.id,
+        [targetGroupKey]: targetGroupValue,
+        permissionLevel: targetPermission.permissionLevel,
+        inheritedFromPermission: targetPermissionId
+      });
+    }
+
+  });
+
+  // // Find inheritance references to drop
```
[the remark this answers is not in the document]
motechFR: Good spot, thanks, deleted

pages/api/permissions/index.ts
motechFR: Good spot, thanks! All sorted

mattcasey (approved):

PR #712, opened 2022-07-19
Use decorations for inline comments

mattcasey:
This PR has some cleanup/reorganization stuff in it so I've added comments to the important part. I didn't realize the fix would also be so complicated (using decorations)! The reason we want to use decorations is that editing the DOM won't trigger updates via prosemirror.

This works:
- create a decoration for each comment that triggers the comment popup and group by heading or paragraph

To do:
- find a way to update the counter when comment count is updated
- Possible other cases? I removed some de-duping logic and i don't show anything unless it actually has an 'id' in the attribute. I also am not checking if the thread was resolved already...


components/common/CharmEditor/components/InlineActionCounter.tsx
```diff
@@ -30,75 +30,73 @@ const InlineActionCountContainer = styled.span`
...
-  children: ReactNode,
   calculateActions?: boolean,
   inlineCommentPluginKey: PluginKey<InlineCommentPluginState>,
   inlineVotePluginKey: PluginKey<InlineVotePluginState>
   node: NodeViewProps['node']
 }
 
-export default function InlineActionCounter (
-  { node, children, calculateActions = true, inlineCommentPluginKey, inlineVotePluginKey }: InlineActionCounterProps
+function InlineActionCounter (
```
mattcasey, line 39: we probably still need this component but only for votes

components/common/CharmEditor/components/inlineComment/inlineComment.plugins.tsx
```diff
@@ -0,0 +1,147 @@
...
+        getReferenceElement: referenceElement(key, (state) => {
+          const { selection } = state;
+          return {
+            end: selection.to,
+            start: selection.from
+          };
+        })
+      }
+    }),
+    new Plugin({
```
mattcasey, line 88: this is a new plugin which adds comments as decorations

lib/prosemirror/highlightMarkedElement.ts
```diff
@@ -19,31 +19,39 @@ export function highlightMarkedElement ({
...
-        // If the page thread list isn't open, then we need to show the inline thread component
-        renderSuggestionsTooltip(key, {
-          ids: [actionId]
-        })(view.state, view.dispatch, view);
+    return highlightElement({ id: actionId, view, elementId, markName, key, prefix });
+  }
+  return false;
+}
+
+export function highlightElement ({ id, key, prefix, elementId, view }:
```
mattcasey, line 28: i extracted this logic to a new function so i could call it when i already have the node and don't need to look at the selection

components/common/CharmEditor/components/FloatingMenu.tsx
```diff
@@ -121,13 +119,6 @@ export function floatingMenuPlugin ({ key, readOnly, enableComments = true }:{ke
         return null;
       }
 
-      // If we are inside an inline comment
-      if (hasComponentInSchema(state, inlineCommentMarkName)) {
```
mattcasey, line 125: @Devorein i don't think we were really relying on this code so i deleted it because i found that when i clicked on text with an "orphan" mark, it would open the new comment popup instead of the normal editing tools, and this was the culprit. I'm not sure if we should also remove the following code for inline votes, but i left it alone

theme/@bangle.dev/styles.scss
```diff
@@ -341,7 +337,10 @@ ol li {
   display: flex;
   flex-direction: row;
   position: relative;
-  left: -1.5rem;
+}
+
+.bangle-editor li[data-bangle-is-todo] > span:first-child {
```
mattcasey, line 342: made some slight CSS changes to todo list, because the negative left margin was also affecting the alignment of the decorations

components/common/CharmEditor/components/FloatingMenu.tsx
Devorein: Yeah, I think the new behavior is better. Thank you for bringing this up.

Devorein:
I tested it out in the staging environment and everything worked as expected except for the following few:-
1. Un-resolving comment doesn't bring back the mark
2. Comment and vote counter are on top of one another
3. Clicking on a comment from sidebar does scroll to the correct paragraph but it doesn't highlight it

PR #715, opened 2022-07-20
add public api endpoint to retrieve bounties

mattcasey:
This endpoint was requested by Gateway. It is only temporarily needed by them.. eventually the idea is for us to call their API to mint NFTs instead of them calling us.


motechFR (approved):
@mattcasey made some small fixes

- Updated Open API doc to explain how to get and use the API key
Add this handly doc for future reference https://spec.openapis.org/oas/v3.0.3#openapi-object

- Fixed endpoint documentation to make sure it specifies an array is returned

- Added 1 integration test to make sure the API returns what we expect

PR #716, opened 2022-07-20
add scripts to generate and test an api key

mattcasey:
just another way to generate keys. it was easier to just call the api directly since i have access to the production db. eventually we'll have a formal way to do it :P


motechFR (approved):
Simple but effective!

PR #717, opened 2022-07-20
Mobile improvements: signup and navigation

mattcasey:
In this PR, I was focusing on the top navigation and general layout for mobile.
- Signup: centered our logo and added margin on either side of forms
- Navigation: smaller font, hide breadcrumbs and some icons on mobile
- Page layout: got rid of extra left margin and made margins more narrow
- Fixed some inconsistent typography inside some Focalboard components

In another PR, I want to make a 'mobile' sidebar similar to Notion, which includes links to alerts, profile, etc. so they don't have to always be visible in the right corner


Devorein (approved):

Devorein:
A few things I noticed while reviewing
1. If I'm on the nexus page I am not able to see the right workspaces strip, this means there is no way to get back to my workspace other than using the back button
2. On a smaller screen may be put the share and favorite under the page context menu (...)? (The profile & notification would go inside the left sidebar as you mentioned in the PR)
3. The Change cover and Remove button are not positioned correctly in the mobile view. They get hidden underneath the page icon.
4. No ways to switch between database views in mobile view. The user can only add new views

PR #719, opened 2022-07-21
Create bounty page

AndreiMitrea:
When creating a bounty, also create an associated bounty page.


prisma/schema.prisma
```diff
@@ -289,6 +291,7 @@ model Bounty {
   space             Space              @relation(fields: [spaceId], references: [id], onDelete: Cascade)
   applications      Application[]
   permissions       BountyPermission[]
+  page              Page?
```
Devorein, line 294: We are already using linkedTaskId to link a bounty to a page, so I'm not sure how this will behave when we assign a bounty to a page, for example, if a task card is assigned a bounty which page should it use the assigned one or the bounty page one?

Devorein (approved):

pages/api/pages/[id]/index.ts
```diff
@@ -78,7 +87,28 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
+
+  if (isBountyPage) {
+    [pageWithPermission] = await prisma.$transaction([
+      prisma.page.update(pageUpdateData),
+      prisma.bounty.update({
+        where: {
+          id: page.bountyId || undefined
+        },
+        data: {
+          title: req.body.title,
```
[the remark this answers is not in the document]
Comment (mattcasey) [reply pages/api/pages/[id]/index.ts:102]: That's correct. Please do not keep them in sync, this PR needs to make it so we never read from bounties

prisma/schema.prisma
mattcasey: The end result should be that we no longer use linkedTaskId on bounties, since there will always be a page. But we need to make sure we do that and get rid of linkedTaskId. @AndreiMitrea do you think it would make sense to remove it in this PR?

scripts/createBountyPages.ts
```diff
@@ -0,0 +1,43 @@
+import { Bounty, PageType, Prisma } from 'prisma/prisma-client';
+import log from 'lib/log';
+import { prisma } from '../db';
+
+(async () => {
+  const bounties: Array<Partial<Bounty>> = await prisma.bounty.findMany();
```
mattcasey, line 6: always use `Partial<Bounty>[]` notation for arrays

prisma/schema.prisma
AndreiMitrea: @mattcasey Sure

pages/api/pages/[id]/index.ts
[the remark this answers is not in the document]
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:102]: I've removed the code to keep the bounty in sync with the page in the latest commits

Devorein (approved):

PR #721, opened 2022-07-21
Update to Prisma 4.1

mattcasey:
The schema file changes were fully automated, following the guide here: https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4


pages/api/v1/databases/[id]/search.ts
```diff
@@ -110,12 +110,10 @@ async function searchDatabase (req: NextApiRequest, res: NextApiResponse) {
   };
 
   if (searchQuery.query?.title) {
-    prismaQueryContent.Page = {
-      every: {
```
Comment (mattcasey) [inline pages/api/v1/databases/[id]/search.ts:114]: Typescript said 'every' is not valid, from the types I can see prisma is now expecting a normal "PageWhereInput" object

Comment (motechFR) [reply pages/api/v1/databases/[id]/search.ts:114]: Having worked deeply with their documentation recently, I can confirm there's a few inconsistencies here and there.

prisma/schema.prisma
```diff
@@ -1,84 +1,55 @@
-// This is your Prisma schema file,
-// learn more about it in the docs: https://pris.ly/d/prisma-schema
-
 generator client {
-  provider        = "prisma-client-js"
-  binaryTargets   = ["linux-musl", "darwin", "native"]
-  previewFeatures = ["filterJson"]
```
mattcasey, line 7: filterJson is no longer preview 🎉

motechFR (approved):

PR #722, opened 2022-07-22
Return header if identity type is not discord

Devorein:


AndreiMitrea (approved):

pages/api/discord/disconnect.ts
```diff
@@ -45,7 +45,7 @@ async function disconnectDiscord (req: NextApiRequest, res: NextApiResponse) {
 
   // If identity type is not Discord
   if (user.identityType !== IDENTITY_TYPES[1]) {
-    res.status(200).json({ success: 'ok' });
+    return res.status(200).json({ success: 'ok' });
```
mattcasey, line 48: i assume this is so we stop execution so we don't return two responses.. nice  catch!

mattcasey (approved):

pages/api/discord/disconnect.ts
Devorein: Yep. Thanks to the discord error message.

PR #733, opened 2022-07-25
hide comment and vote sidebars on mobile

mattcasey:
- completely remove components on mobile to help optimize performance
- also hide components when browser is resized using CSS


Devorein (approved):

Devorein:
Approved but there seems to be a small style issue for the right sidebar
image

At the moment there are no backgrounds for the Votes or the Comments header
image

PR #734, opened 2022-07-26
fix extra space on focalboard in mobile view

mattcasey:
On mobile, you can scroll the focalboard view left and right  even though it's just whitespace. I figured out it was the title input being 500+ pixels wide was stretching the page. I don't know why it is that wide by default, but setting it to 100% fixes the problem. I found that normally it has a "100% width" when it's a child of CardDetail, so I figure this is a safe update.

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/viewTitle.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/viewTitle.tsx
@@ -24,7 +24,7 @@ type Props = {
     setPage: (page: Partial<Page>) => void
 }
 
-const ViewTitle = React.memo((props: Props) => {
+const ViewTitle = (props: Props) => {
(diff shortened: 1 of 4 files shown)


components/common/BoardEditor/focalboard/src/components/viewTitle.tsx
```diff
@@ -131,6 +131,6 @@ const ViewTitle = React.memo((props: Props) => {
             }
         </div>
     )
-})
+}
 
-export default ViewTitle
+export default React.memo(ViewTitle)
```
mattcasey, line 136: not related to layout, but using memo this way allows there to be the component name 'ViewTitle' in any error stacktraces

Devorein (approved):

PR #735, opened 2022-07-26
Mobile: Allow toggling the sidebar on Nexus pages

mattcasey:


adapters/http/fetch.server.ts
```diff
@@ -0,0 +1,6 @@
+import fetch, { RequestInit } from 'node-fetch';
+import { transformResponse } from './fetch';
+
+export default function fetchWrapper<T> (resource: string, init?: RequestInit): Promise<T> {
```
mattcasey, line 4: added this helper for making api calls on the backend

lib/deepdao/getAggregatedData.ts
```diff
@@ -19,15 +21,24 @@ export async function getAggregatedData (userPath: string): Promise<DeepDaoAggre
     throw new DataNotFoundError();
   }
 
-  const participationScores = user.addresses.length !== 0 ? await Promise.all(user.addresses.map(address => getParticipationScore(address))) : [];
+  const participationScores = (await Promise.all(
```
mattcasey, line 24: graceful handling if no api key exists. the api response without a key is { message: 'ForbiddenAccess' } and blowing up further down below

Devorein (approved):

PR #737, opened 2022-07-26
Multi token gates

motechFR:
- Auto verify gates a user is eligible for

- Simplify token gate logic so that it takes a spaceDomain instead of also containing that form


components/common/TokenGate/TokenGateForm.tsx
```diff
@@ -0,0 +1,188 @@
+import Grid from '@mui/material/Grid';
```
mattcasey, line 1: Please structure this folder:
TokenGateForm/
TokenGateForm/index.tsx // exports TokenGateForm.tsx
TokenGateForm/TokenGateForm.tsx
TokenGateForm/TokenGateOptions.tsx // this could also be put inside a components folder, but just one or two children can be an exception

components/common/TokenGate/TokenGateForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { checkAndSignAuthMessage } from 'lit-js-sdk';
+import { useEffect, useState } from 'react';
+import TokenGateOption from './TokenGateOption';
+
+interface Props {
+  onSubmit: (values: Space) => void;
+  spaceDomain: string
+}
+
+export default function JoinSpacePage ({ onSubmit: _onSubmit, spaceDomain }: Props) {
```
mattcasey, line 27: rename to "TokenGateForm"

pages/join.tsx
```diff
@@ -46,10 +56,62 @@ export default function CreateSpace () {
...
+          fullWidth
+          value={spaceDomain}
+          helperText={userInputStatus}
+          InputProps={{
+            endAdornment: isTruthy(spaceInfo) && <CheckIcon color='success' />
+          }}
+        />
+
+        {
+          isTruthy(spaceInfo) && (
```
mattcasey, line 110: FYI isTruthy is not really needed here - Typescript knows that spaceInfo is not null and will strip it out in the following code for you. I only added that utility for when we map arrays, because Typescript isn't smart enough to strip the null/falsey values of union types on its own:
```
const maybes = [true, false] as const;
// doesn't compile:
const truths: true[] = maybes.filter(Boolean);
// compiles:
const truthsOk: true[] = maybes.filter(isTruthy);
```
const

components/common/TokenGate/TokenGateOption.tsx
```diff
@@ -0,0 +1,57 @@
...
+import Grid from '@mui/material/Grid';
+import Typography from '@mui/material/Typography';
+import { useWeb3React } from '@web3-react/core';
+import { TokenGateWithRoles } from 'lib/token-gates/interfaces';
+import { humanizeAccessControlConditions } from 'lit-js-sdk';
+import { useEffect, useState } from 'react';
+
+interface Props {
+  tokenGate: TokenGateWithRoles,
+  select: (tokenGate: TokenGateWithRoles) => void,
```
mattcasey, line 13: I think calling this `onSelect` would make it a little more obvious that it's an event handler

mattcasey (changes requested):
Code looks good, just a few requests! But after reading this, I realize you made it so that the user selects which token gate they want. I was thinking we would simply verify all token gates and apply all applicable roles at once, but this is definitely an improvement. We can check with Alex if it's good for now.

My only concern with the current approach is the green check mark looks like a verification, and it's not obvious that I have to click a gate to select it. At first, I was going to suggest we use a RadioGroup but MUI has another component that looks even better called ToggleButton: https://mui.com/material-ui/react-toggle-button/. What do you think?

Also, we could update the instructions when there's more than one gate: "To join, please select the condition which you want to verify"

components/common/TokenGate/TokenGateForm.tsx
motechFR: Gotcha, makes sense as it allows for flat export hierarchy from components/common even if a specific component has a few sub-components.

pages/join.tsx
motechFR: Understood, sounds like I went a bit overboard on isTruthy usage, will remember that specific use case for next times (so objects don't need it)

components/common/TokenGate/TokenGateOption.tsx
motechFR: Done, can we add this to our UI design principles doc under a code styling header?

components/common/TokenGateForm/TokenGateForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+          setTokenGates(null);
+          setIsLoading(false);
+        });
+    }
+  }, [spaceDomain]);
+
+  async function generateAuthSignature () {
+    // Reset the current state
+    setTokenGateResult(null);
+    window.localStorage.removeItem('lit-auth-signature');
```
mattcasey, line 64: I would like to remove this... it's only to protect if the lit js sdk breaks its compatibility with the current authSig right? I can't see that happening any time soon, at least until we update our lit js sdk. And we don't have the same defensive logic for admins when they create token gates or test them.

mattcasey (approved):

components/common/TokenGate/TokenGateOption.tsx
mattcasey: we can. but i think this is so small, its weird to sit on its own. if you look at any event handlers in JS, they all start with the word "on". i dont expect someone to be looking this up in our docs

PR #739, opened 2022-07-27
Support drag and drop for all components

mattcasey:


Devorein:
It seems to be working for the most part. But throws an error at times
image
I think the best solution would be to now wrap ReactDnD provider around the whole component tree, just around those components that requires it
this seems to be a lead https://github.com/react-dnd/react-dnd/issues/1534 but there is no official solution for it. The closest one was to have atleast chrome 78 which I definitely have.

Devorein (approved):

PR #724, opened 2022-07-22
Updated bounty ui

Devorein:
[ x ] Create a bounty
[ x ] Be able to edit bounty from the new UI
[ x ] "View mode" display for bounties from the card view
[ x ] Reviewer Mode: view list of applications
[ x ] Create form to apply to a bounty and update submission

PR #743, opened 2022-07-28
Add a new column to determine if submissions are locked

mattcasey:
This PR depends on https://github.com/charmverse/app.charmverse.io/pull/724.  I can either just merge it into that one, or we could temporarily hide the option to lock submissions in the new UI and then merge this to main after the new bounty ui is deployed. and have it re-enable the lock in the UI. Let me know what you prefer @Devorein


lib/applications/actions/__tests__/approveApplication.spec.ts
```diff
@@ -61,9 +60,7 @@ describe('approveApplication', () => {
     const bounty = await createBounty({
       title: 'My bounty',
       createdBy: adminUser.id,
-      spaceId: adminSpace.id,
-      // Random person
-      reviewer: null
```
mattcasey, line 66: removing fields that are no longer valid input

lib/applications/actions/approveApplication.ts
```diff
@@ -1,12 +1,10 @@
 import { Application } from '@prisma/client';
-import { DataNotFoundError, LimitReachedError, UnauthorisedActionError, UndesirableOperationError } from 'lib/utilities/errors';
-import { hasAccessToSpace } from 'lib/users/hasAccessToSpace';
+import { DataNotFoundError, LimitReachedError } from 'lib/utilities/errors';
 import { prisma } from 'db';
-import { getBounty } from 'lib/bounties';
+import { getBountyOrThrow } from 'lib/bounties';
```
mattcasey, line 4: i also replaced "getBounty" with "getBountyOrThrow" to save a bunch of code. it turns out we always want to throw a DataNotFoundError... but it's nice to have both methods available i guess

__integration-tests__/server/pages/api/bounties/[id]/close-submissions.spec.ts
```diff
@@ -81,13 +75,11 @@ describe('POST /api/bounties/{submissionId}/close-submissions - close a bounty t
...
 
-    const result = (await request(baseUrl)
+    await request(baseUrl)
       .post(`/api/bounties/${bounty.id}/close-submissions`)
       .set('Cookie', adminCookie)
       .send({})
-      .expect(200)).body as BountyWithDetails;
-
-    expect(result.status).toBe('inProgress');
+      .expect(200);
```
Comment (mattcasey) [inline __integration-tests__/server/pages/api/bounties/[id]/close-submissions.spec.ts:82]: @motechFR one question here: the status returned is now "open" and not "inProgress". I removed the expectation from this specific tests since it's just testing that an admin can take action. A better place to test status could be the previous test, but is the status supposed to change if it's locked?

Devorein (approved):

lib/applications/actions/approveApplication.ts
Devorein: For a second I thought this was a prisma method as 4.x added `findOrThrow`

PR #748, opened 2022-07-29
Hide headers when no bounties exist

mattcasey:


motechFR:
@mattcasey Works well, good idea breaking those components up!

One small feedback, I think we should reduce vertical spacing (at least in case of empty bounty list)

image

image

motechFR (approved):

mattcasey:
> One small feedback, I think we should reduce vertical spacing (at least in case of empty bounty list)

Thanks, actually the video was slightly lower than the column headers. I like them roughly the same height so that when you add your first bounty, the top of the page content doesn't jump

PR #749, opened 2022-07-29
Fix visibility of public bounties

motechFR:
Fixes visibility of public bounties by passing bounty page data to use pages hook

Fixes public pages to show an error if the public space does not exist

### DIFF
--- a/components/share/PublicBountyList.tsx
+++ b/components/share/PublicBountyList.tsx
@@ -11,8 +11,10 @@ import TokenGateForm from 'components/common/TokenGateForm';
 import { Web3Connection } from 'components/_app/Web3ConnectionManager';
 import { useContributors } from 'hooks/useContributors';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import { usePages } from 'hooks/usePages';
 import { useSnackbar } from 'hooks/useSnackbar';
 import { useUser } from 'hooks/useUser';
+import { IPageWithPermissions } from 'lib/pages';
 import { usePopupState } from 'material-ui-popup-state/hooks';
 import { BountyWithDetails } from 'models';
 import { useRouter } from 'next/router';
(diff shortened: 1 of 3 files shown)


components/share/PublicPage.tsx
```diff
@@ -48,8 +48,9 @@ export default function PublicPage () {
   const pageIdOrPath = router.query.pageId instanceof Array ? router.query.pageId.join('/') : router.query.pageId as string;
   const dispatch = useAppDispatch();
   const { pages, currentPageId, setCurrentPageId } = usePages();
+  const [loadingSpace, setLoadingSpace] = useState(true);
   const [currentSpace] = useCurrentSpace();
-  const [, setSpaces] = useSpaces();
+  const [_, setSpaces] = useSpaces();
```
mattcasey, line 53: minor, but this is a warning by the linter since we never use `_`

motechFR: Oops, added as test, forgot to remove. Doing so now

components/share/PublicPage.tsx
```diff
@@ -61,7 +62,11 @@ export default function PublicPage () {
       const spaceDomain = (router.query.pageId as string[])[0];
       charmClient.getPublicSpaceInfo(spaceDomain).then(space => {
         setSpaces([space]);
-      });
+        setLoadingSpace(false);
+      })
+        .catch(() => {
+          setLoadingSpace(false);
```
mattcasey, line 68: this loading var needs to be set after loading regular public pages too, right now they never appear to load

mattcasey: Also, this is a good case for `.finally()`, vs adding it to .then() and .catch()

components/share/PublicBountyList.tsx
```diff
@@ -66,6 +68,21 @@ export default function PublicBountyList () {
     if (space) {
       charmClient.listBounties(space.id, true)
         .then(_bounties => {
+
+          const pagesToSet: Record<string, IPageWithPermissions> = _bounties.reduce((pageMap, bounty) => {
```
mattcasey, line 72: this approach doesn't work, because usePages() will automatically refresh itself every 5 minutes

PR #755, opened 2022-07-29
Add token gate to new bounty v3 UI

mattcasey:
This PR generally moves the token gate logic and modal into the BountyApplicantForm, similar to what you recommended Mo. BountySignupButton could almost just be a drop-in button anywhere we need a signup via token gate, there's nothing really specific to bounties there... but for now i didn't want to generalize until we need it


mattcasey:
Preemptive merging since this is a bug fix, will have reviewed on Monday

PR #757, opened 2022-08-01
Bounty actions inside card options menu

Devorein:


mattcasey (changes requested):
something got messed up with the merge conflicts:
image

components/[pageId]/DocumentPage/components/BountyProperties.tsx
```diff
@@ -84,6 +85,8 @@ export default function BountyProperties (props: {children: ReactNode, readOnly?
   const assignedRoleSubmitters = permissions?.bountyPermissions?.submitter?.filter(p => p.group === 'role').map(p => p.id as string) ?? [];
   const selectedReviewerUsers = permissions?.bountyPermissions?.reviewer?.filter(p => p.group === 'user').map(p => p.id as string) ?? [];
   const selectedReviewerRoles = permissions?.bountyPermissions?.reviewer?.filter(p => p.group === 'role').map(p => p.id as string) ?? [];
+  const router = useRouter();
+  const isSharedPage = router.pathname.startsWith('/share');
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties.tsx:89]: this logic exists in DocumentPage already, you should just pass it in as a prop

components/[pageId]/DocumentPage/components/BountyProperties.tsx
```diff
@@ -371,18 +374,23 @@ export default function BountyProperties (props: {children: ReactNode, readOnly?
...
+        </>
       )}
       <Stack flexDirection='row' justifyContent='space-between' gap={2} alignItems='center'>
         {
           (permissions && !permissions.userPermissions?.review) ? (
-            <BountyReviewers
-              bounty={bounty}
-              permissions={permissions}
-            />
+            <Stack gap={1} mb={1} width='100%'>
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties.tsx:387]: i feel like <Divider /> and no <Stack /> would be simpler to do here. then just add some vertical margin using `sx`

components/bounties/components/ApplicationEditorForm.tsx
```diff
@@ -0,0 +1,159 @@
+import { yupResolver } from '@hookform/resolvers/yup';
```
mattcasey, line 1: We already have a copy of this as a sub-component inside components/bounties/components/BountyApplicantForm. (I wanted to try nesting some of these because there are too many components in one folder)

components/bounties/components/BountiesKanbanView.tsx
```diff
@@ -7,19 +7,27 @@ import { useRouter } from 'next/router';
...
+  refreshBounty?: (bountyId: string) => void
 }
 
-export default function BountiesKanbanView ({ bounties }: Omit<Props, 'publicMode'>) {
-
-  const [activeBountyPage, setBountyPage] = useState<Page | null>(null);
-  const { pages } = usePages();
+export default function BountiesKanbanView ({ bounties, refreshBounty }: Omit<Props, 'publicMode'>) {
+  const [bountyPage, setBountyPage] = useState<Page | null>(null);
+  const [selectedBounty, setSelectedBounty] = useState<BountyWithDetails | null>(null);
```
mattcasey, line 28: I was going to make a point that you should use the same adjective (whether it's "active" or "selected") for both Page and Bounty vars here, but they actually should be stored in the same state. Doesn't seem like there's a need to de-couple them. I think it would be more clear this way, to show that they should be set together:
```
const [activePage, setActivePage] = useState<{ page: Page, bounty: BountyWithDetails } | null>(null);

```

mattcasey (changes requested):
see comments

components/bounties/components/BountiesKanbanView.tsx
```diff
@@ -70,21 +88,70 @@ export default function BountiesKanbanView ({ bounties }: Omit<Props, 'publicMod
...
             ))}
           </div>
         ))}
       </div>
 
-      {activeBountyPage && <PageDialog page={activeBountyPage} onClose={closePopup} />}
+      {bountyPage && selectedBounty && (
+      <PageDialog
+        toolsMenu={(
+          <List dense>
```
mattcasey, line 109: 1) no need to optionally pass in toolsMenu. We always want the menu to exist when viewing a page.
2) please reuse the same component from cardDialog.tsx (or possibly even better, have cardDialog import it from PageDialog.. I could see PageDialog become a defacto global page modal)

components/[pageId]/DocumentPage/components/BountyProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/BountyProperties.tsx:89]: Ow good catch. Changing it rn.

components/bounties/components/ApplicationEditorForm.tsx
Devorein: Ok got another duplicated component `SubmissionEditorForm`. Removed that one as well.

components/bounties/components/BountiesKanbanView.tsx
Devorein: Updated. I agree they both are used in a similar way and when one gets set the other one does as well so no real need to decouple them. Replaced state with activePage as well

components/bounties/components/BountiesKanbanView.tsx
Devorein: Gotcha. One reason I made `toolsMenu` options is because when you are creating a bounty for the first time no tools menu is shown on the page dialog. And I refactored the code a bit to do exactly as you said, `cardDialog` now uses `PageDialog` a defacto global page modal component

mattcasey (approved):

PR #762, opened 2022-08-02
Lazy-load the CharmEditor

mattcasey:


Devorein:
image
Creating a bounty throws this error

components/common/IntlProvider.tsx
```diff
@@ -0,0 +1,24 @@
+import { ReactNode, useEffect } from 'react';
+import { IntlProvider } from 'react-intl';
+import { useAppDispatch, useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
+import { getMessages } from 'components/common/BoardEditor/focalboard/src/i18n';
+import { fetchLanguage, getLanguage } from 'components/common/BoardEditor/focalboard/src/store/language';
+
+export default function ReactDndProvider ({ children }: { children: ReactNode }) {
```
Devorein, line 7: Not sure why its called ReactDndProvider 🤔

Devorein (changes requested):

Devorein:
Tested all the major operations of the platform to make sure nothing broke. Thankfully caught one. Other than that everything looks and works great. We can merge the PR once we handle that case

Devorein:
I've fixed both of the issues in the latest commits. I will do one more round of QA to make sure everything is a-okay

Devorein (approved):

components/common/IntlProvider.tsx
mattcasey: mistake!

PR #764, opened 2022-08-02
Prompt user for unsaved changes

Devorein:
A hook called `useNavigationLocked` to prompt user to save unsaved changes during three actions
1. Clicking on an internal link
2. Reloading the page
3. Going back/forward in browser

The changes have been implemented in these two places but adding it for other places would be a breeze now
1. Workspace settings
2. Inline page comment

Issues:
1. We can't make use of our custom confirm dialog unfortunately
2. Sometimes you need to click on Ok inside the window.confirm dialog twice
3. Reloading the page first shows regular non submitted form reload dialog then ours

References: 
1. [Gist](https://gist.github.com/Tymek/df2021b77fcea20cabaef46bbee8b001)
2. [Issue#2694](https://github.com/vercel/next.js/issues/2694)
3. [Discussion#12348](https://github.com/vercel/next.js/discussions/12348)


mattcasey (changes requested):
it didnt work when i tested with comment sidebar:
https://www.loom.com/share/0433e0d745064ed8a96e7d83cc889678

This current approach does not actually stop the page from navigating. Also, clicking "OK" should not take any action, it should simply leave me where i am, so no "onConfirm" callback should be necessary.

I also could see we'd want to prevent a modal from closing or the sidebar.. I can close my comments sidebar and the message disappears again.

It does help prevent me from refreshing the page on accident, though. I would like to see if that's enoguh for this feature for alex

components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
```diff
@@ -56,23 +58,29 @@ export default function ConnectSnapshot () {
...
+      const spaceWithDomain = await charmClient.updateSnapshotConnection(space?.id as any, formValues);
+      setSpace(spaceWithDomain);
+    }
+    catch (err) {
+      setFormError(err as any);
+    }
+    setTouched(false);
   }
 
+  usePreventReload(touched);
```
mattcasey, line 71: hooks are best to be called at the top of a method

mattcasey (approved):

components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
Devorein: From now on we've decided to follow a format for declaring react artifacts inside a functional component

PR #765, opened 2022-08-02
Inline proposal votes

motechFR:
This PR extends the vote model to provide context on whether this vote is inline or proposal level

It also provides visual differentiation for proposal level votes and inline votes


components/nexus/VoteTasksList.tsx
```diff
@@ -146,6 +161,8 @@ export function VoteTasksListRow (
 
 export function VoteTasksList ({ error, tasks, mutateTasks }: VoteTasksListProps) {
 
+  const { pages } = usePages();
```
mattcasey, line 164: looks like you can remove this?

components/votes/VotesPage.tsx
```diff
@@ -70,7 +74,7 @@ export default function VotesPage () {
           </Box>
         </Grid>
       </Grid>
-      <VotesTable votes={data ? sortedVotes : undefined} mutateVotes={mutate} />
+      <VotesTable votes={sortedVotes} mutateVotes={mutateVotes} />
```
mattcasey, line 77: we should pass in 'undefined' until there is data, so that it shows it is loading. that, or add an 'isLoading' prop?

prisma/schema.prisma
```diff
@@ -253,6 +258,7 @@ model Vote {
   spaceId     String        @db.Uuid
   threshold   Int
   type        VoteType      @default(SingleChoice)
+  context     VoteContext?   @default(inline)
```
mattcasey, line 261: i think you can change this to mandatory since it has a default

components/votes/components/VoteIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import Box from '@mui/material/Box';
+import TaskIcon from '@mui/icons-material/TaskOutlined';
+import MuiVoteIcon from '@mui/icons-material/HowToVoteOutlined';
+import { Vote, VoteStatus } from '@prisma/client';
+
+type Props = Pick<Vote, 'context'> & {status: 'Draft' | VoteStatus}
+
+export default function VoteIcon ({ context, status }: Props) {
+
+  const proposalIconColor = status === 'InProgress' ? 'primary' : 'secondary';
```
mattcasey, line 10: for now i recommend not coloring the icons, since status is already visible. or maybe have a prop to trigger the color on hover for the sidebar of pages, and it should work for all statuses and types

components/votes/VotesPage.tsx
mattcasey: I was going to also ask can we check that pages are loaded before showing the votes? It jumps for me since pages load after the votes (this was probably happening before as well).
It looks like we need to export a new 'isLoading' prop from usePages, but that could be useful elsewhere. I can take care of that in another PR if you want

mattcasey (changes requested):
Few minor requests, overall works well!

mattcasey (approved):

components/nexus/VoteTasksList.tsx
motechFR: Thanks, done!

components/votes/VotesPage.tsx
motechFR: Hey @mattcasey I handled that in this [commit](https://github.com/charmverse/app.charmverse.io/pull/765/commits/8ad4842641ad897d2c6438cc8fa38e89e5f42115)

Pretty simple approach for now, but should avoid the initial flashing. (I considered Object.keys(pages).length === 0 to be a sign that the pages are loading.

prisma/schema.prisma
motechFR: Done 👍

PR #767, opened 2022-08-03
fix saving votes from inside a card

mattcasey:
I noticed I couldn't save a new vote when looking at a card view inside of focalboard. It's because the pageId was 'null' inside the hook - it doesn't seem to get upated when the URL changes.

I also moved the call to useVotes for `createVote` into the modal component because all the parents were having to do the same logic.

### DIFF
--- a/components/votes/components/CreateVoteModal.tsx
+++ b/components/votes/components/CreateVoteModal.tsx
@@ -5,10 +5,11 @@ import { DateTimePicker } from '@mui/x-date-pickers/DateTimePicker';
 import Button from 'components/common/Button';
 import FieldLabel from 'components/common/form/FieldLabel';
 import Modal from 'components/common/Modal';
-import { VoteDTO, ExtendedVote } from 'lib/votes/interfaces';
+import { ExtendedVote } from 'lib/votes/interfaces';
 import { DateTime } from 'luxon';
 import { Dispatch, SetStateAction, useEffect, useState } from 'react';
 import { usePages } from 'hooks/usePages';
+import { useVotes } from 'hooks/useVotes';
 import AddCircle from '@mui/icons-material/AddCircle';
 import DeleteIcon from '@mui/icons-material/Delete';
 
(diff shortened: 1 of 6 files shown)


hooks/useVotes.tsx
```diff
@@ -87,7 +87,6 @@ export function VotesProvider ({ children }: { children: ReactNode }) {
     const extendedVote = await charmClient.createVote({
       ...votePayload,
       createdBy: user.id,
-      pageId: cardId ?? currentPageId,
```
mattcasey, line 90: The other place we use cardId in this hook is to tell if we should skip loading votes on line 38. Do you know @Devorein what was the context for that logic? Why can't we load votes for any page?

components/votes/components/CreateVoteModal.tsx
```diff
@@ -78,20 +79,21 @@ function InlineVoteOptions (
 
 interface CreateVoteModalProps {
   onClose?: () => void;
-  createVote: (votePayload: Omit<VoteDTO, 'createdBy' | 'spaceId'>) => Promise<ExtendedVote>;
-  postCreateVote?: (vote: ExtendedVote) => void;
+  onCreateVote?: (vote: ExtendedVote) => void;
```
mattcasey, line 82: trying to standardize on callback names: they should start with the word 'on'. and it is usually implied that the action already took place

Devorein: Good idea. I will keep that in mind from now 'on' when naming props

hooks/useVotes.tsx
Devorein: I can't exactly remember the reasoning behind it but I think it had something to do with the right sidebar showing inside card view which we decided on not showing. As for the global context, that would require a separate PR for the refactor I believe, updating currentPageId with the cardId if any card page is opened, right now it points to the board page.

Devorein (approved):

hooks/useVotes.tsx
Devorein: I can't exactly remember the reasoning behind it but I think it had something to do with the right sidebar showing inside card view which we decided on not showing. As for the global context, that would require a separate PR for the refactor I believe, updating currentPageId with the cardId if any card page is opened, right now it points to the board page.

mattcasey: Oh, ok.. i wonder if we could control that from the card view itself, though, even if we had to pass a prop down thru to DocumentPage. we don't need the cardId, just the fact that the sidebar shouldn't be viewed

PR #768, opened 2022-08-03
Use CharmEditor for comments

mattcasey:
Stores the JSON inside of 'block.fields.content', and the raw text is still 'block.title'. I checked that this is backwards-compatible, I'm not sure if we even need to do a migration?

Other notes:
- I fixed up the UI a bit for avatars and made font size smaller (I Like that Notion does it)
- I removed some code from focalboard to grab the current user that doesn't work but wasn't being used
- removed draft-js


Devorein (approved):

Devorein:
A small issue cropped up during QA. But for some reason, it's not reproducible in local environment but is present in the staging
Here is what's shown in my local env
image
And this is staging
image

PR #771, opened 2022-08-03
Disabled state for batch bounty payment button

Devorein:


components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -123,20 +123,24 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
...
+            sx={{ ml: 1 }}
+            variant='outlined'
+            color='secondary'
+            disabled={isDisabled}
+          >
+            Batch Payment ({selectedApplicationIds.length})
+          </Button>
+        </div>
+      </Tooltip>
+      {!isDisabled && (
```
mattcasey, line 143: is there a reason you added this here? I think it's better to just have the Modal in place rather than add/remove it to the DOM whenever this state changes

components/bounties/components/BountySubmissionsTable.tsx
```diff
@@ -238,7 +238,6 @@ export default function BountySubmissionsTable ({ bounty, permissions }: Props)
 
   return (
     <>
-      {permissions.userPermissions.review && (
```
mattcasey, line 241: i added this condition because there's no reason someone should need to see application count unless they are also viewing more than just their application

mattcasey: Also it doesn't really matter, since we'll be completely hiding the table when Mo makes his update :)

components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -123,20 +123,24 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
...
     <>
-      <Button
-        {...bindTrigger(popupState)}
-        sx={{ ml: 1 }}
-        variant='outlined'
-        color='secondary'
-      >
-        Batch Payment ({selectedApplicationIds.length})
-      </Button>
+      <Tooltip arrow placement='top' title={isDisabled ? 'Please ensure that your wallet is connected and on the right chain.' : ''}>
```
mattcasey, line 130: `Batch payment requires at least one Completed bounty on the ${getChainById(gnosisSafeChainId)?.chainName} network`

mattcasey (changes requested):
please change the tooltip on the button, other stuff is not as important

components/bounties/components/MultiPaymentModal.tsx
Devorein: Yeah correct its unnecessary optimization

components/bounties/components/MultiPaymentModal.tsx
Devorein: Updated. Thank you.

components/bounties/components/BountySubmissionsTable.tsx
Devorein: Right, but I asked Alex about this and he said that it should always be visible.

mattcasey (approved):

PR #774, opened 2022-08-03
Add notifications for comments on bounties and other FB comment blocks

mattcasey:
image


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -168,7 +168,7 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false }: Documen
                     <hr />
                     <CommentsList
                       comments={comments}
-                      rootId={card?.rootId ?? page.spaceId}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:171]: At first I wanted to use rootId, but it didn't work because we were using spaceId. I don't think we rely on this being spaceId, but it also should not be a spaceId.. rootId should point to a root piece of content (a page or a block).

lib/mentions/__tests__/getMentionedTasks.spec.ts
```diff
@@ -136,30 +147,37 @@ beforeAll(async () => {
...
 
+    function expectSome <T> (arr: T[], condition: (item: T) => boolean) {
+      expect(arr.some(condition)).toBeTruthy();
+    }
+
     expect(marked.length).toBe(1);
-    expect(unmarked.length).toBe(5);
-    expect(marked[0].mentionId).toBe(mentionIds[0]);
-    expect(marked[0].pageId).toBe(page1.id);
+    expect(unmarked.length).toBe(6);
```
mattcasey, line 160: i rewrote these expectations since we can't guarantee the order of results

lib/mentions/getMentionedTasks.ts
```diff
@@ -66,134 +81,160 @@ export async function getMentionedTasks (userId: string): Promise<MentionedTasks
...
 
   // Get the marked mention task ids (all the discussion type tasks that exist in the db)
-  const markedMentionTaskIds = new Set(markedMentionTasks.map(markedMentionTask => markedMentionTask.taskId));
+  const notifiedTaskIds = new Set(notifications.map(notification => notification.taskId));
 
-  // Get all the pages of all the spaces this user is part of
-  const pages = await prisma.page.findMany({
+  const context: GetMentionsInput = { userId, username, spaceRecord, spaceIds };
+
+  const { mentions, mentionedUserIds } = await Promise.all([
```
mattcasey, line 88: refactored this to separate out the query/transformation logic

lib/mentions/getMentionedTasks.ts
```diff
@@ -213,37 +254,67 @@ export async function getMentionedTasks (userId: string): Promise<MentionedTasks
...
+            spaceId: page.spaceId,
+            spaceDomain: spaceRecord[page.spaceId].domain,
+            pagePath: page.path,
+            spaceName: spaceRecord[page.spaceId].name,
+            userId: mention.createdBy,
+            pageTitle: page.title,
+            text: mention.text,
+            bountyTitle: page.title,
+            commentId: null,
+            type: page.bountyId ? 'bounty' : 'page'
```
mattcasey, line 309: i removed the extra query on bounties since we retrieve them now thru pages

lib/mentions/__tests__/getMentionedTasks.spec.ts
Devorein: Yes, you are absolutely right. I will keep that in mind from now on.

Devorein (changes requested):
1. Clicking on a discussion card (for bounty page) to the wrong page. It directs to `/bounties/{bountyId}` not `bounties?bountyId={bountyId}`
2. mentions inside bounty page content is marked as `page` but it should be `bounty`

lib/mentions/getMentionedTasks.ts
Devorein: @mattcasey This should be used in all cases, not only for this one. Every comment belongs to a page (either a block or a regular page), but it could either be bounty or a page

mattcasey:
> Clicking on a discussion card (for bounty page) to the wrong page. It directs to /bounties/{bountyId} not bounties?bountyId={bountyId}

I believe we're just sending you to the page path, which I am fine with. We have discussed a separate story to not navigate at all

mattcasey:
> All of these can be done for both page.type of bounty and page so we should use the page.bountyId to decide whether its a bounty or not, right now for 1 and 2 we explicitly mark them as page
Actually I might look at the code, i don't see a real reason to differentiate between these, except maybe if we show an icon in the

lib/mentions/getMentionedTasks.ts
mattcasey: not sure what "this" you're talking about, lets chat in the standup

Devorein:
We should also remove the `commentId` & `mentionId` from the url when someone gets redirected to the page from the nexus discussion tasks

PR #776, opened 2022-08-04
Fetch individual discord user when importing roles

Devorein:
Some considerations
1. Fetching all users that have their discord account connected to charmverse for a specific space could cause some bottleneck if there are lots of them. Should we be careful about that or is that something we can skip for now?
2. How to handle the case when failing to fetch an individual user? When the import is done should there be a message that says failed to import/connect `X` users?
3. Right now as soon as you authorize your server the import starts and you are redirected to where you came from (settings/roles), and we show a loader underneath the `Roles` section indicating that roles are being imported. Should that be more prominent via a modal that says `Discord roles import is in progress. Please wait...`?

### DIFF
--- a/pages/api/discord/importRoles.ts
+++ b/pages/api/discord/importRoles.ts
@@ -7,6 +7,7 @@ import { handleDiscordResponse } from 'lib/discord/handleDiscordResponse';
 import { findOrCreateRoles } from 'lib/roles/createRoles';
 import { assignRolesFromDiscord, DiscordGuildMember } from 'lib/discord/assignRoles';
 import { DiscordServerRole } from 'lib/discord/interface';
+import { isTruthy } from 'lib/utilities/types';
 
 const handler = nc({
   onError,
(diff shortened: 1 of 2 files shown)


pages/api/discord/importRoles.ts
```diff
@@ -55,34 +52,41 @@ async function importRoles (req: NextApiRequest, res: NextApiResponse<ImportRole
...
-  else {
-    const rolesRecord = await findOrCreateRoles(discordServerRoles, spaceId, req.session.user.id);
-    // Remove the roles imported from guild.xyz
-    for (const roleId of Object.keys(rolesRecord)) {
-      const role = rolesRecord[roleId];
-      if (role?.sourceId && role.source === 'guild_xyz') {
-        delete rolesRecord[roleId];
-      }
+  // TODO: How to handle error fetching a single user
+  const discordGuildMemberResponses = await Promise.all(discordConnectedMembers.map(discordConnectedMember => (handleDiscordResponse<DiscordGuildMember>(`https://discord.com/api/v8/guilds/${guildId}/members/${discordConnectedMember.discordUser!.discordId}`) as Promise<{status: 'success', data: DiscordGuildMember}>)));
```
mattcasey, line 77: please make this line shorter

Devorein: Good idea. I will do it rn.

motechFR: Same comment as @mattcasey 

Also, we may have quite a few Discord connected members. We need a way to rate limit our requests

motechFR (changes requested):
@Devorein you just need to handle the failure case on individual requests by catching them and returning a value indicatibg this call failed

pages/api/discord/importRoles.ts
Devorein: Correct. Its been fixed in the latest commit

pages/api/discord/importRoles.ts
```diff
@@ -55,34 +53,60 @@ async function importRoles (req: NextApiRequest, res: NextApiResponse<ImportRole
...
-    // Remove the roles imported from guild.xyz
-    for (const roleId of Object.keys(rolesRecord)) {
-      const role = rolesRecord[roleId];
-      if (role?.sourceId && role.source === 'guild_xyz') {
-        delete rolesRecord[roleId];
-      }
+  const discordGuildMemberResponses = await Promise.all(
+    discordConnectedMembers.map(
+      discordConnectedMember => (
+        new Promise<DiscordGuildMember | null>((resolve) => {
```
mattcasey, line 80: there's no need to wrap this in a Promise constructor, just return the promise

mattcasey (approved):

PR #780, opened 2022-08-04
Use CharmEditor to submit comments

mattcasey:
image
 
 I re-used the UI from the existing comment list, i think it helps a bit since this ends up being a comment anyway


Devorein:
Found two issues
1. No content after the mention gets added to the comment content.
2. Can't create any new comments after sending the first now
I will look for a fix and push them

Devorein (approved):

motechFR (approved):
Tested and approved

PR #786, opened 2022-08-08
Fixed public board card page not showing

Devorein:


mattcasey:
was fixed in https://github.com/charmverse/app.charmverse.io/pull/787

PR #789, opened 2022-08-09
Confirm card conversion to bounty

mattcasey:
Rather than try to pass around an empty bounty, I added a new 'draftBounty' to the state inside useBounties hook which is used to populate the form inside BountyProperties.

See loom: https://www.loom.com/share/ca32b3ae9b7b4373a453fc3991e60645

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -186,7 +186,7 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false }: Documen
                     pageUpdatedBy={page.updatedBy}
                   />
                 )}
-                {(draftBounty || (page.bountyId && bountyPermissions)) && (
+                {(draftBounty || page.bountyId) && (
                   <BountyProperties
                     bountyId={page.bountyId}
                     pageId={page.id}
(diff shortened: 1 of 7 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -177,26 +175,27 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false }: Documen
...
+                    card={card}
+                    cards={cards}
+                    activeView={activeView}
+                    views={boardViews}
+                    readonly={cannotEdit}
+                    pageUpdatedAt={page.updatedAt.toString()}
+                    pageUpdatedBy={page.updatedBy}
+                  />
                 )}
-                {bounty && bountyPermissions && (
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:191]: I removed the check for bountyPermissions, we're basically just using it to show the values of Reviewer, but it's better to show the rest of the form while it loads

components/common/form/InputSearchBlockchain.tsx
```diff
@@ -53,6 +53,7 @@ export default function InputSearchBlockchain ({
       onChange={(_, _value: IChainDetails) => {
         if (_value?.chainId) {
           onChange(_value.chainId);
+          setValue(_value);
```
mattcasey, line 56: this is a fix for a bug that's in production - you can't change the blockchain when adding a new payment method

Devorein (approved):

Devorein (approved):

PR #791, opened 2022-08-10
Add inline database component

mattcasey:
This PR allows users to embed databases inside prosemirror docs


Devorein:
Can't seem to create a new board from the left sidebar. This is the error that is shown
image

Devorein (changes requested):
Looks really good and is almost fully functional apart from the bug mentioned above

mattcasey:
@Devorein i just moved ReactDNDProvider to the _app. I believe it's always attaching a listener to window anyway. This fixed drag and drop for kanban boards and table columns for me. I couldn't reproduce your error but maybe it is fixed as well.

Devorein (approved):

PR #792, opened 2022-08-10
Resolve threads from removed document text 

Devorein:


charmClient.ts
```diff
@@ -253,6 +253,10 @@ class CharmClient {
     return http.POST<{importedRolesCount: number}>('/api/guild-xyz/importRoles', payload);
   }
 
+  resolveMultipleThreads (payload: ResolveMultipleThreads) {
```
mattcasey, line 256: please put this next to the other calls to api/threads

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -345,17 +347,30 @@ function CharmEditor (
   const _isEmpty = checkForEmpty(content);
   const [isEmpty, setIsEmpty] = useState(_isEmpty);
   const [currentUser] = useUser();
-  const onContentChangeDebounced = onContentChange ? debounce((view: EditorView) => {
+  const onContentChangeDebounced = onContentChange ? debounce((view: EditorView, prevDoc?: Node<any>) => {
```
mattcasey, line 350: isn't prevDoc always there?

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -90,7 +92,7 @@ export function charmEditorPlugins (
       pageId?: string | null,
       userId?: string | null,
       readOnly?: boolean,
-      onContentChange?: (view: EditorView) => void,
+      onContentChange?: (view: EditorView, prevDoc: Node<any>) => void,
```
mattcasey, line 95: I think `EditorState['doc']` would be a slightly more accurate/informative type to use instead of `Node<any>`

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -345,17 +347,30 @@ function CharmEditor (
...
     const doc = view.state.doc.toJSON() as PageContent;
     const rawText = view.state.doc.textContent as string;
+    if (prevDoc && pageId) {
+      const deletedThreadIds = extractDeletedThreadIds(
+        specRegistry.schema,
+        view.state.doc,
+        prevDoc
+      );
+      if (deletedThreadIds.length) {
+        charmClient.resolveMultipleThreads({
```
mattcasey, line 360: it would be good to add a .catch() here and log the error

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -345,17 +347,30 @@ function CharmEditor (
...
+          threadIds: deletedThreadIds,
+          pageId
+        });
+      }
+    }
     onContentChange({ doc, rawText });
-  }, 100) : undefined;
+  }, 1000) : undefined;
```
mattcasey, line 367: we should keep this at 100 please. i don't think we should extend the delay for calling onContentChange

lib/inline-comments/extractDeletedThreadIds.ts
```diff
@@ -0,0 +1,34 @@
...
+          threadIds.add(mark.attrs.id);
+        }
+      });
+    }
+  });
+
+  return threadIds;
+}
+
+export function extractDeletedThreadIds (schema: Schema, curDoc: Node, prevDoc: Node) {
```
mattcasey, line 20: good candidate for a test or two

pages/api/threads/resolve.ts
```diff
@@ -0,0 +1,48 @@
...
+import nc from 'next-connect';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser)
+  .post(requireKeys<ResolveMultipleThreads>(['threadIds', 'pageId'], 'body'), resolveThreads);
+
+async function resolveThreads (req: NextApiRequest, res: NextApiResponse) {
+
+  const { threadIds, pageId } = req.body as {
```
mattcasey, line 17: should be `as ResolveMultipleThreads`?

lib/threads/interfaces.ts
```diff
@@ -27,3 +27,8 @@ export interface ThreadCreate {
 export interface ThreadWithCommentsAndAuthors extends Thread {
   comments: CommentWithUser[]
 }
+
+export interface ResolveMultipleThreads {
```
mattcasey, line 31: ResolveMultipleThreads sounds like an action.. i think a better name for interface could be `MultipleThreadsInput`

mattcasey (changes requested):
This is a good size PR and I just had a few comments. When testing, I noticed that the comments' state doesn't update in the sidebar until I refresh the page or collapse and then hide comment sidebar. That should be fixed before we go to production

charmClient.ts
Devorein: Updated in the latest commit

components/common/CharmEditor/CharmEditor.tsx
Devorein: Yes it's provided but we are calling `onContentChangeDebounced` when we have stopped resizing (for example image or video), by calling `onResizeStop`. So I decided not to pass the `prevDoc` there and which would bypass auto thread resolving

components/common/CharmEditor/CharmEditor.tsx
Devorein: Updated and yeah good point. I usually hover over types and then just copy paste but this is a bit more abstract.

components/common/CharmEditor/CharmEditor.tsx
Devorein: Nice catch, I've updated it to log the error

components/common/CharmEditor/CharmEditor.tsx
Devorein: Oh I think I first added 1000 ms to test it out then completely forgot to remove it. 1000 ms delay is very noticeable. Sorry for that. I've reverted back the 1000ms delay for content change to 100ms and added a 1000ms denounced call to resolve the threads

lib/inline-comments/extractDeletedThreadIds.ts
Devorein: Yep gonna write them right now

Devorein: Done. The most difficult part of the setup but this is a good step towards writing tests for our prosemirror stuffs

pages/api/threads/resolve.ts
Devorein: Correct. Wrote it over there first, then realized its gonna be needed in a lot of other places so might as well write the type for it, but forgot to replace it back when destructuring the payload

lib/threads/interfaces.ts
Devorein: Should we come up with a convention on how to write payload types? Maybe `ResolveMultipleThreadsPayload` or `ResolveMultipleThreadsInput`? Because I can see that we are writing the payload types very close to the function name in quite a few other places.

mattcasey: either of those would be good. Since this is so coupled to the request i think it belongs in the api folder, honestly

mattcasey (approved):

PR #793, opened 2022-08-10
Remove hack for loading data into Redux

mattcasey:
I realized that for loading blocks into Redux/focalboard, we were still using reverse-engineering the URL to determine the workspace. I added that hack because originally when implementing focalboard, I didn't have context available. This is hard to follow and adds some delay for the first page load.

Once clients have a chance to be updated (24 hours or so), I'll remove the backend code that was being used.


components/common/BoardEditor/focalboard/src/store/initialLoad.ts
```diff
@@ -10,24 +10,16 @@ import { RootState } from './index';
...
-  async () => {
+  async ({ spaceId }: { spaceId: string }) => {
 
-    const [workspace, workspaceUsers, blocks, userWorkspaces] = await Promise.all([
-      charmClient.getWorkspace(),
-      charmClient.getWorkspaceUsers(),
-      charmClient.getAllBlocks(),
-      charmClient.getUserWorkspaces()
+    const [workspaceUsers, blocks] = await Promise.all([
+      charmClient.getWorkspaceUsers(spaceId),
```
mattcasey, line 16: I removed some context we are not even using, AFAICT. Not sure if we can rely on Typescript to catch any cases that might still be using it, though.

components/settings/contributors/InviteLinks/InviteLinks.tsx
```diff
@@ -52,7 +52,7 @@ export default function InviteLinkList ({ isAdmin, spaceId }: { isAdmin: boolean
     <>
       <Legend>
         Invite Links
-        {isAdmin && <Button variant='outlined' sx={{ float: 'right' }} onClick={open}>Create a link</Button>}
+        {isAdmin && <Button variant='outlined' sx={{ float: 'right' }} onClick={open}>Createa a link</Button>}
```
msadura, line 55: typo

```suggestion
        {isAdmin && <Button variant='outlined' sx={{ float: 'right' }} onClick={open}>Create a link</Button>}
```

pages/_app.tsx
```diff
@@ -246,9 +243,9 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
         <ThemeProvider theme={theme}>
           <Web3ReactProvider getLibrary={getLibrary}>
             <Web3ConnectionManager>
-              <ReduxProvider store={store}>
-                <IntlProvider>
-                  <DataProviders>
+              <DataProviders>
+                <FocalBoardProvider>
```
msadura, line 247: tiny but really nice cleanup - ReduxProvider with store from focalboard was a bit confusing when you looked into this file 👍

PR #794, opened 2022-08-11
Feat/avatar nft

msadura:
- picking default NFT avatar while creating new user
- loading user's NFTs 
- showing nft gallery where user can pick one as avatar
- a bit of refactor in a few places

https://user-images.githubusercontent.com/5198394/184178830-e0e45d45-babf-485c-aa84-8938901ec862.mp4


prisma/schema.prisma
```diff
@@ -271,6 +271,9 @@ model User {
   updatedAt         DateTime               @default(now())
   addresses         String[]
   avatar            String?
+  avatarContract    String?
+  avatarTokenId     String?
+  avatarTokenChain  Int?
```
mattcasey, line 276: Hm, this could also be `avatarContractChain`. What about `avatarNetwork` or `avatarBlockchain`? or change `avatarContract` to `avatarTokenContract` so we're consistent.

pages/api/profile/avatar.ts
```diff
@@ -0,0 +1,74 @@
...
+
+handler
+  .put(updateAvatar);
+
+async function updateAvatar (req: NextApiRequest, res: NextApiResponse<LoggedInUser | {error: string}>) {
+  const { avatar, avatarTokenId, avatarContract } = req.body as UserAvatar;
+  const { id: userId } = req.session.user;
+
+  if (!avatar || !avatarContract || !avatarTokenId) {
+    res.status(400).json({ error: 'Invalid avatar data' });
```
mattcasey, line 26: we're in the process of replacing this pattern with throwing exceptions. Can you change this to:

```
import { InvalidInputError } from 'lib/utilities/errors';
...
throw new InvalidInputError(`Invalid avatar data`);
```

components/common/HexagonalAvatarMask.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export function HexagonalAvatarMask () {
+  return (
+    <svg height='0' viewBox='0 0 200 188' width='0'>
+      <defs>
+        <clipPath clipPathUnits='objectBoundingBox' id='hexagon-avatar' transform='scale(0.005 0.005319148936170213)'>
```
mattcasey, line 7: at first i was going to warn about the 'id' prop and that it needs to be unique, but i saw we just need one instance of this to reference. 'id' could be a prop that gets passed in?

components/profile/PublicProfile.tsx
```diff
@@ -1,5 +1,5 @@
 import { Divider, Grid, Stack } from '@mui/material';
-import UserDetails, { UserDetailsProps } from './components/UserDetails';
+import UserDetails, { UserDetailsProps } from './components/UserDetails/UserDetails';
```
mattcasey, line 2: a pattern we usually implement is to add an `index.tsx` file that re-exports the default component and optionally the other exports as well. This makes the imports look cleaner:
```
import UserDetails from './UserDetails';
export default UserDetails;
export * from './UserDetails';
```

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
+import { UserAvatar } from 'lib/users/interfaces';
+import Dialog from '@mui/material/Dialog';
```
mattcasey, line 3: FYI we have a transformation step from Next that maps  `import { Dialog } from '@mui/material'` to the tree-shaking-friendly format. I find it's easier to add/remove components this way, also IDE's will automatically create the imports this way.

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -0,0 +1,72 @@
...
+import DialogTitle from '@mui/material/DialogTitle';
+import { useMyNfts } from 'hooks/useMyNfts';
+import Skeleton from '@mui/material/Skeleton';
+import Stack from '@mui/material/Stack';
+
+import Grid from '@mui/material/Grid';
+
+import { CircularProgress, DialogContent, Typography } from '@mui/material';
+import { NftGalleryItem } from 'components/profile/components/NftAvatarGallery/NftGalleryItem';
```
mattcasey, line 12: this should be a relative import:
`import NftGalleryItem from './NftGalleryItem';`

And each component file/folder should export the component by default.

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -0,0 +1,72 @@
...
+            {nfts?.length ? nfts?.map(nft => (
+              <Grid key={`${nft.contract}-${nft.tokenId}`} item xs={6} sm={3} sx={{ minWidth: 110 }}>
+                <NftGalleryItem nft={nft} onClick={() => onSelect?.(nft)} isSelected={getIsSelected(nft)} />
+              </Grid>
+            )) : <Box sx={{ p: 4, textAlign: 'center' }}><Typography>You do not own any NFTs</Typography></Box>}
+          </Grid>
+        )}
+
+        {isSaving && (
+        <Box sx={{ position: 'absolute', top: 0, right: 0, left: 0, bottom: 0, display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
```
mattcasey, line 65: this would be a good candidate for styled components

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </Grid>
+            ))}
+          </Grid>
+        ) : (
+          <Grid container spacing={1}>
+            {nfts?.length ? nfts?.map(nft => (
+              <Grid key={`${nft.contract}-${nft.tokenId}`} item xs={6} sm={3} sx={{ minWidth: 110 }}>
+                <NftGalleryItem nft={nft} onClick={() => onSelect?.(nft)} isSelected={getIsSelected(nft)} />
+              </Grid>
+            )) : <Box sx={{ p: 4, textAlign: 'center' }}><Typography>You do not own any NFTs</Typography></Box>}
```
mattcasey, line 60: I think you could make more use of MUI's props (sx is slightly slower for rendering than props, apparently):
```
<Box p={4}><Typography align='center'>...</Typography></Box>
```

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -0,0 +1,72 @@
...
+                <Stack spacing={1}>
+                  <Skeleton variant='rectangular' width={100} height={80} />
+                  <Skeleton variant='text' sx={{ fontSize: '1rem' }} />
+                </Stack>
+              </Grid>
+            ))}
+          </Grid>
+        ) : (
+          <Grid container spacing={1}>
+            {nfts?.length ? nfts?.map(nft => (
```
mattcasey, line 56: i think it's unecessary to check for length when mapping or looping over arrays because it's basically a no-op, and one less if/else condition to read

components/profile/components/NftAvatarGallery/NftGalleryItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const onAvatarClick = () => {
+    if (!isSelected) {
+      onClick();
+    }
+  };
+
+  return (
+    <Card
+      sx={{ p: 1, cursor: isSelected ? 'auto' : 'pointer', position: 'relative', height: '100%' }}
+      onMouseEnter={() => setIsHovered(true)}
```
mattcasey, line 33: i like the attention to detail :)

components/profile/components/UserDetails/userDetails.hook.ts
```diff
@@ -0,0 +1,41 @@
+import { useState } from 'react';
```
mattcasey, line 1: For organization, please create a folder 'hooks' and put useUserDetails as a file inside the folder

components/profile/components/UserDetails/UserDetails.tsx
```diff
@@ -61,31 +63,19 @@ function UserDetails ({ readOnly, user, updateUser }: UserDetailsProps) {
     setTimeout(() => setIsDiscordUsernameCopied(false), 1000);
   };
 
-  const handleUserUpdate = async (data: Partial<User>) => {
-    const updatedUser = await charmClient.updateUser(data);
-    if (updateUser) {
-      updateUser(updatedUser);
-    }
-  };
+  const { handleUserUpdate, handleNftAvatarUpdate, isSaving } = useUserDetails({ readOnly, user, updateUser });
```
mattcasey, line 66: ya i think this is a good cleanup

components/settings/workspace/LargeAvatar.tsx
```diff
@@ -64,62 +72,118 @@ const getIcons = (editIcon: ReactNode, deleteIcon: ReactNode, avatar: string | n
...
 
 export default function LargeAvatar (props: LargeAvatarProps) {
-  const { name, image, updateImage, variant, editable } = props;
+  const { name, image, updateAvatar, variant, editable, canSetNft, isSaving, updateImage, isNft } = props;
   const inputFile = useRef<HTMLInputElement>(null);
+  const editIconRef = useRef(null);
+  const [menuAnchorEl, setMenuAnchorEl] = useState<null | HTMLElement>(null);
+  const [isGalleryVisible, setIsGalleryVisible] = useState(false);
+
+  const onUploadClick = () => {
```
mattcasey, line 81: `function onUploadClick() { }` is our preferred format to define functions

components/settings/workspace/LargeAvatar.tsx
```diff
@@ -64,62 +72,118 @@ const getIcons = (editIcon: ReactNode, deleteIcon: ReactNode, avatar: string | n
...
+        }}
+      />
+      <StyledAvatarWithIcons
+        avatar={image}
+        name={name}
+        variant={variant}
+        icons={icons}
+        isNft={isNft}
+      />
+      {canSetNft && (
```
mattcasey, line 176: this almost feels like it should just be a new component, say `NFTAvatar` that wraps LargeAvatar. I'm ok with this for now though

hooks/useMyNfts.ts
```diff
@@ -0,0 +1,12 @@
+import charmClient from 'charmClient';
+import { useUser } from 'hooks/useUser';
+import useSWR from 'swr';
+
+export const useMyNfts = () => {
+  const [user] = useUser();
+  const { data, error: serverError } = useSWR('/nfts/list', () => charmClient.nft.list(user?.addresses || []));
```
mattcasey, line 7: I think it would be better to just use the addresses on the server-side session. We don't need to expose a free API service for checking anyone's NFTs. Also, I'm not sure there's an actual bug but if we call this before the user is loaded, we'd never see your NFT's because '/nfts/list' is a static key. you would want to do something like `useSWR(user ? '/nfts/list' : null, ...)` or `useSWR(`/nfts/list/{user?.id}`, ...)`

lib/blockchain/provider/alchemy.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const getOwners = async (contractAddress: string, tokenId: string, chainId: SupportedChainId = 1) => {
+  const url = `${getAlchemyBaseUrl(chainId)}/getOwnersForToken?contractAddress=${contractAddress}&tokenId=${tokenId}`;
+  const res = await GET<{ owners: string[] }>(url);
+
+  return res.owners;
+};
+
+export const alchemyApi = { getNfts, getNft, getOwners };
```
mattcasey, line 83: i'd prefer we just use, as needed:
```
import * as alchemyApi from 'lib/blockhain/provider/alchemy
```

lib/blockchain/provider/types.ts
```diff
@@ -0,0 +1,26 @@
+type NftMedia = {
```
mattcasey, line 1: please use 'interfaces.ts' instead of 'types.ts', again just for consistency sake

lib/blockchain/provider/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  bytes: number
+  format: string
+  gateway: string
+  raw: string
+  thumbnail: string
```
mattcasey, line 6: we've been adding semi-colons to end of each line for types and interfaces. It bugs me that TS allows comma, semi-colon, nothing... i wonder if we could add a lint rule for this?

lib/users/createUser.ts
```diff
@@ -19,12 +21,15 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
   }
   else {
     const ens: string | null = await getENSName(address);
+    const userId = uuid();
+    const avatar = await getDefaultAvatar(address, userId);
```
mattcasey, line 25: please add a try/catch around this. i realize you added one for the call to S3 but not the call to alchemy, i think it's best to put it here since creating a user profile is high priority. (you could say the same about getENSName, but at least it does gracefully handle the error already)

lib/users/getDefaultAvatar.ts
```diff
@@ -0,0 +1,39 @@
...
+      return {
+        avatar,
+        avatarContract: nft.contract,
+        avatarTokenId: nft.tokenId,
+        avatarTokenChain: nft.chainId
+      };
+    }
+    // eslint-disable-next-line no-empty
+    catch (e) {
+      log.debug('Failed to upload default NFT avatar', e);
```
mattcasey, line 33: log.warn would be better

lib/users/interfaces.ts
```diff
@@ -0,0 +1,7 @@
+
+export type UserAvatar = {
+  avatar: string | null
+  avatarContract: string | null;
+  avatarTokenId: string | null;
+   avatarTokenChain: number | null
```
mattcasey, line 6: extra whitespace, and missing semi-colon

mattcasey: btw i like that you added a ffile here. We have some stuff in `models` folder that i'd like to get rid of, since most of our models now come from Prisma, and we've started the practice of putting interfaces inside of `lib`

components/common/HexagonalAvatarMask.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export function HexagonalAvatarMask () {
```
mattcasey, line 3: please make it a default export

pages/api/nft/list.ts
```diff
@@ -0,0 +1,29 @@
...
+    userAddresses = userAddresses.split(',').map(a => a.trim());
+  }
+
+  const nfts = await alchemyApi.getNfts(userAddresses || [], chainId);
+  const mappedNfts = nfts.map(nft => mapNftFromAlchemy(nft, chainId));
+
+  res.status(200).json(mappedNfts);
+}
+
+export default withSessionRoute(handler);
```
mattcasey, line 29: you only need to wrap a handler with withSessionRoute() if you access `res.session`. i left another comment that we could use this to check `res.session.user.addresses` instead of passing it from the front-end

pages/api/profile/avatar.ts
```diff
@@ -0,0 +1,74 @@
...
+import { mapNftFromAlchemy } from 'lib/nft/utilities/mapNftFromAlchemy';
+import { getUserProfile } from 'lib/users/getUser';
+import { getFilenameWithExtension } from 'lib/utilities/getFilenameWithExtension';
+import { UserAvatar } from 'lib/users/interfaces';
+import { withSessionRoute } from 'lib/session/withSession';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .put(updateAvatar);
```
mattcasey, line 19: please add `requireUser` whenever accessing req.session.user. There's lots of examples

mattcasey (changes requested):
see comments

prisma/schema.prisma
msadura: lets go with `avatarChain` to keep it short

pages/api/profile/avatar.ts
msadura: yup, I really like it 👍

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
msadura: oh this is actually great! Yeah makes imports much cleaner

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
msadura: kk, I left Box to properly center text, but used styled props

components/settings/workspace/LargeAvatar.tsx
msadura: ok good to know. Personally I prefer arrow functions because they look also cleaner when you wrap them in hooks like useMemo / useCallback

components/settings/workspace/LargeAvatar.tsx
msadura: yeah, probably a good idea, especially that orginal "LargeAvatar" is placed in workspace while NFT avatar will be used for user profile

hooks/useMyNfts.ts
msadura: good call, I updated it to `useSWR(user && `/nfts/list/${user.id}` so it should fetch unly when there is an user + refetch when it change

lib/blockchain/provider/types.ts
msadura: I believe that if you use `;` for line endings in code then we should keep them for types / interfaces as well for consistency. I am pretty sure there is lint rule for that, I wil lcheck it out

pages/api/nft/list.ts
msadura: Thx for the explanation, it is all clear now :)

Initially I did not use `session.user.addresses` due to note in user session interface:

```
  interface IronSessionData {
    // this data is only useful for authentication - it is not kept up-to-date!
    user: { id: string, addresses: string[] };
  }
 ```

but I guess user addresses is not something that changes often so I added them to type. Let me know if that will be fine.

msadura:
I think I was able to address all of the mentioned comments. Thank you for such detailed insights and explanations.

A couple of notes:
- you mentioned formatting types / interfaces - there is eslint rule for that: https://typescript-eslint.io/rules/member-delimiter-style/ and according to it:
> Semicolon style (default, preferred in TypeScript).

I tried to add it but it generated ~1k errors. Ofc lint can easily fix them for us, but I do not want to mix it with this pr to avoid massive conflicts. I think it would be the best to gather a few ideas related to linting / formatting code, add rules (and possibly prettier) and re-format everything at once :)

- I added default exports to components, but this is something to reconsider - default exports are rather considered to be a bad practise, example article: https://blog.piotrnalepa.pl/2020/06/26/default-exports-vs-named-exports/
TLDR:
- with "named" imports you cannot mess names so it is easier to understand what function / component you are looking at
- better support for IDE imports

pages/api/nft/list.ts
mattcasey: oh, this just made me remember that we don't keep the user object updated very well on the session, sorry about that. I'd like to change this to just capture userId, just haven't had time to QA. we can easily use the user id to retrieve the profile, though

mattcasey: This works for now.

mattcasey:
You made some good points about named exports. I'd like to keep consistent for now but am open to revisiting this as well.

mattcasey (approved):

PR #796, opened 2022-08-11
Highlight permission inconsistencies

motechFR:
Shows editors within the bounty form if submitters or reviewers cannot view a page

image


pages/api/pages/[id]/restrict-permissions.ts
```diff
@@ -0,0 +1,34 @@
...
+  const computed = await computeUserPagePermissions({
+    pageId,
+    userId: req.session?.user?.id
+  });
+
+  if (computed.grant_permissions !== true) {
+    throw new UnauthorisedActionError('You do not have permission to restrict permissions on this page');
+  }
+
+  const updatedPage = await lockToBountyCreator({ pageId });
```
Comment (mattcasey) [inline pages/api/pages/[id]/restrict-permissions.ts:29]: i see what you did here... it will only ever apply to the bounty creator. that addresses my concern about letting anyone 'restrict editing'

mattcasey (approved):
Looks good, I tested once as a reviewer and once as an applicant.

PR #803, opened 2022-08-12
fix filename parsing when uploading to s3

mattcasey:
we need to make sure we take the _last_ part of the string, if a filename has periods inside of it the updated logic would fail to extract the extension


lib/utilities/__tests__/encodeFilename.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { encodeFilename } from '../encodeFilename';
+
+describe('amount formatting', () => {
```
msadura, line 3:
```suggestion
describe('encoding filename', () => {
```

msadura (approved):

lib/utilities/__tests__/encodeFilename.spec.ts
mattcasey: thanks! I've never seen a commit suggestion, i made a commit on my own tho

PR #805, opened 2022-08-15
Feat/avatar popup

msadura:
Show avatar select popup when new user account is created

https://user-images.githubusercontent.com/5198394/184589397-c9e834d9-40ec-435a-ad11-1f05f068ea5e.mp4

Additional changes:
- moved common global components from _app.tsx
- added 2 avatar sizes: xl and 2xl to avoid creating a custom styled components for them
- added helper hooks for file pickers and uploading selected file to s3


components/_app/GlobalComponents.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+export default function GlobalComponents () {
+  return (
+    <>
+      <HexagonalAvatarMask id='hexagon-avatar' />
+      <SetAvatarPopup />
+      <Snackbar />
+    </>
+  );
+}
```
motechFR, line 14: I think this refactoring is a bit excessive as we only use these in one place.

components/_app/GlobalComponents.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Snackbar } from '@mui/material';
+import HexagonalAvatarMask from 'components/common/HexagonalAvatarMask';
+import SetAvatarPopup from 'components/profile/components/SetAvatarPopup/SetAvatarPopup';
+
+export default function GlobalComponents () {
+  return (
+    <>
+      <HexagonalAvatarMask id='hexagon-avatar' />
+      <SetAvatarPopup />
```
motechFR, line 10: Is the intention to have this AvatarPopup across the app?

Otherwise we could relocate this to the actual join page

components/common/Avatar.tsx
```diff
@@ -41,8 +73,8 @@ export type InitialAvatarProps = {
   avatar: string | null | undefined;
   className?: string;
   name?: string;
-  variant?: 'circular' | 'rounded' | 'square';
-  size?: 'xSmall' | 'small' | 'medium' | 'large';
+  variant?: AvatarVariant;
+  size?: AvatarSize;
```
motechFR, line 77: Good cleanup 👍

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -41,42 +40,44 @@ export default function NftAvatarGallery ({ onSelect, isVisible, onClose, isSavi
...
+          ))}
+        </Grid>
+      ) : (
+        <Grid container spacing={1}>
+          {nfts?.length ? nfts?.map(nft => (
+            <Grid key={`${nft.contract}-${nft.tokenId}`} item xs={6} sm={3} sx={{ minWidth: 110 }}>
+              <NftGalleryItem nft={nft} onClick={() => onSelect?.(nft)} isSelected={getIsSelected(nft)} />
+            </Grid>
+          )) : (
+            renderEmpty()
```
motechFR, line 71: When rendering components, for consistency it's best to use the following style

<RenderEmpty />

You'll just want to capitalise the function name and make empty message a prop of this component

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -41,42 +40,44 @@ export default function NftAvatarGallery ({ onSelect, isVisible, onClose, isSavi
...
-              <Box p={4} textAlign='center' flex={1}>
-                <Typography>You do not own any NFTs</Typography>
-              </Box>
-            )}
-          </Grid>
-        )}
+  return (
+    <Box position='relative'>
+      {isLoading ? (
+        <Grid container spacing={1}>
```
motechFR, line 54: Suggestion; Since you have the same Grid container, you could use it once and then use the ternary expression to define its content

components/profile/components/UserDetails/hooks/useUpdateProfileAvatar.ts
```diff
@@ -0,0 +1,39 @@
+import { UserAvatar } from 'lib/users/interfaces';
+import { useCallback, useState } from 'react';
+import { useUser } from 'hooks/useUser';
+import charmClient from 'charmClient';
+
+const isAvatarObject = (
+  avatar: string | UserAvatar
+): avatar is UserAvatar => typeof avatar === 'object';
```
motechFR, line 8: I'm not familiar with "is" notation. Would you mind explaining what it does?

I think this could be clearer as a simple boolean return (unless I missed something)

components/profile/components/UserDetails/hooks/useUpdateProfileAvatar.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    setIsSaving(true);
+    try {
+      const updatedUser = await charmClient.profile.setAvatar(updatedAvatar);
+      setUser(updatedUser);
+
+      return updatedUser;
+    }
+    catch (e) {
+      return user;
```
motechFR, line 30: I think we should alert the user that the avatar upload failed. Otherwise they'll be surprised to see it didn't change

You can use the useSnackbar hook's showMessage function for this

hooks/useS3UploadInput.ts
```diff
@@ -0,0 +1,11 @@
...
+import { uploadToS3 } from 'lib/aws/uploadToS3Browser';
+
+export const useS3UploadInput = (onFileUpload: (url: string) => void) => {
+  async function uploadFile (file: File) {
+    const { url } = await uploadToS3(file);
+    onFileUpload(url);
+  }
+
+  return useFilePicker(uploadFile);
+};
```
motechFR, line 11: Could you explain a bit more the logic behind this utility please, and how it works inside SetAvatarPopup for example?

hooks/useFilePicker.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    const file = event.target.files?.[0];
+    if (file) {
+      onFile(file);
+    }
+
+    event.target.value = '';
+  };
+
+  return { inputRef, openFilePicker, onFileChange };
```
motechFR, line 22: General consistency, our hooks should consumable by components.

It seems useFilePicker is only used inside useS3UploadInput, so I'd recommend either having a subfolder, or defining the useFilePicker inside the S3 upload input file (since it's the only consumer)

hooks/useUser.tsx
```diff
@@ -36,7 +38,11 @@ export function UserProvider ({ children }: { children: ReactNode }) {
     }
   }, [account]);
 
-  const value = useMemo(() => ({ user, setUser, isLoaded, setIsLoaded }) as IContext, [user, isLoaded]);
+  const updateUser = useCallback((updatedUser: Partial<LoggedInUser>) => {
```
motechFR, line 41: Makes sense having this, nice one

motechFR:
@msadura Finished reviewing your PR. Pleasure to read and really appreciating your attention to detail and emphasis on clean code.

Looking forward to reading your feedback on my comments so we can get this approved and merged

components/_app/GlobalComponents.tsx
msadura: I like to have clean files, easy to read. `_app.tsx` has 300+ lines of code already, so it felt like a good idea to take a part of it to external file. Yeah, I know I extracted 2 lines of code and added 14, but we will have a place to add global components in the future.
Overall `_app.tsx` holds quite a lot of code and logic so I wouldn't say any refactor of it is excessive.

components/_app/GlobalComponents.tsx
msadura: yup, I talked with @mattcasey and seems like there are a few "entry points" for new user so it is safer to have this popup available globally

components/profile/components/UserDetails/hooks/useUpdateProfileAvatar.ts
msadura: It is so called "custom type guardian". When you have union type ( `string | UserAvatar`) you can use it to narrow down the type of variable. It is not about returning boolean, but rather to verify the type of `avatar` variable.
Actually I saw it somewhere in CV codebase so I decided to go with it. It is really useful technique anyway: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

hooks/useFilePicker.ts
msadura: >  hooks should consumable by components.

I am not sure what is the reason for this, but cannot entirely agree on that :D Hooks can be used by components, but at the same time they can be used by other hooks, for better composition and separation of concerns.

In this poarticular case - I noticed that every time file picker input looks pretty much the same with repeated logic, for example:

```
                    onChange={async (e) => {
                      const firstFile = e.target.files?.[0];
                      if (firstFile) {
                        const { url } = await uploadToS3(firstFile);
                        onPdfSelect(url);
                      }
                    }}
                    ```

this does not look clean because code is being duplicated + declaring this sort of logic directly in jsx part is just harded  to read.

Because of that I extracted this kind of logic to 2 reusable hooks:
1) useFilePicker - this handles just a part of logic responsible for picking a file 
2) useS3UploadInput - this uses 1st hook and uploads selected file to s3 with a callback

why 2 separate hooks? Because they do separate things :) you can pick a file and do not need to upload it to s3, at the same time you can upload file to s3 form a different source than file input.

Generally I am a fan of properly structured code that can be reusable. Additionally I believe it is better to separate a bit components to a part which handles logic, and a part of "presentation" - where you structure you jsx elements and style them

File: scripts/sendNotifications.ts [hunk 11]
```diff
@@ -1,7 +1,8 @@
 
 import { getNotifications, sendUserNotifications } from 'background/tasks/sendNotifications/sendNotifications';
+import log from 'lib/log';
 
 (async () => {
   const r = await sendUserNotifications();
-  console.log('result', r);
+  log.info('result', r);
```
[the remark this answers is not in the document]
msadura: eslint was throwing an error about console.log 😅

lib/users/createUser.ts
```diff
@@ -21,15 +19,12 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
   }
   else {
     const ens: string | null = await getENSName(address);
-    const userId = uuid();
-    const avatar = await getDefaultAvatar(address, userId);
```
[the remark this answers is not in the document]
msadura: I added default avatar in my previous PR. There was no way for new user to select it manually so we picked just recent NFT avatar.
Now we will offer a popup where user can select his avatar while creating an account, so it does not make sense to set default at the same time

hooks/useS3UploadInput.ts
msadura: Basically, I noticed that every time "pick a file + uplaod to s3" logic is being copy-pasted to many different inputs. Purpose of this hook is to abstract it from `<input>` and expose set of props that can be simply connected to input to make it work. As an example I used it for `LargeAvatar` and `SatAvatarPopup`.

For example:
Screenshot 2022-08-27 at 14 33 05

msadura:
@motechFR  thank you for the review! I think I addressed all the updates / comments :)

components/_app/GlobalComponents.tsx
motechFR: Makes sense

components/profile/components/UserDetails/hooks/useUpdateProfileAvatar.ts
motechFR: Thanks for sharing! Very interesting technique indeed, will use this in future when opportunity presents itself

hooks/useS3UploadInput.ts
motechFR: Oh, thanks for sharing!

hooks/useFilePicker.ts
motechFR: Fair point, I like your point about the division between upload logic and input binding, makes sense and +1 for clean, reusable code

motechFR:
Hey @msadura 

Just noticed 2 user experience bugs

1. I can't upload an image. S3 works perfectly, but the update gets sent to the API, which rejects because the endpoint requires an avatar contract and token id

PS: Appreciated you using our convention for throwing API errors!

2. The error thrown by showMessage doesn't actually show in the UI. Could it be the GlobalComponents overlay has a higher z-index?

Otherwise code looks good, just need to fix these 2 functional items

msadura:
@motechFR thank you for extensive testing! Actually I have a problem with uploading images to s3 on my local env and I couldn't test this scenario. I will fix it anyway.

motechFR (approved):
@msadura Tested S3 upload and worked just fine. NFT selection works too.

It seems the snackbar still isn't showing, were you able to get it to work?

We'll want visual feedback that an error during upload occurred.

Otherwise looks good, approving PR, thanks for the super constructive conversation

components/_app/GlobalComponents.tsx
mattcasey: One other thing to consider is whether this component does a lot of rendering, which it doesn't seem like it does. For example, anything that consumes 'usePages' will be re-rendered as the user is typing, currently.

components/profile/components/UserDetails/hooks/useUpdateProfileAvatar.ts
mattcasey: fyi we use this in the 'isTruthy' method in lib/utilities/react :)

scripts/sendNotifications.ts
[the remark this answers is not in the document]
mattcasey: i hate that it does this, been wondering if we can exclude eslint from the scripts folder?

msadura:
Thanks again @motechFR ! Just in case I increased z-index of snackbar, but it should be working and show error message. Unless it is an issue on specific browser 🤔 
Overall I synced branch with main, resolved conflicts and looks like everything should be fine and ready 💪

scripts/sendNotifications.ts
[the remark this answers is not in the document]
msadura: good call @mattcasey - I added `/scripts` to `.eslintignore`

msadura:
I synced this branch with `mainz once again and resolved conflicts. Additionally I renamed avatar variants to be inlined with naming convention from main (`xLarge`, `2xLarge` instead of `xl`, `2xl`).

PR #813, opened 2022-08-18
Auto-upload the embedded image

motechFR:
Converts embedded base64 images to S3 uploads

### DIFF
--- a/components/common/CharmEditor/components/ResizableImage.tsx
+++ b/components/common/CharmEditor/components/ResizableImage.tsx
@@ -1,13 +1,17 @@
+import Alert from '@mui/material/Alert';
 import { NodeViewProps, Plugin, RawSpecs } from '@bangle.dev/core';
 import { EditorState, EditorView, Node, Slice, Transaction } from '@bangle.dev/pm';
+import { useEditorViewContext } from '@bangle.dev/react';
 import { useTheme } from '@emotion/react';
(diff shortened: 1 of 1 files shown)


motechFR:
Testing complete. This approach will detect existing base64 images and upload them to S3

components/common/CharmEditor/components/ResizableImage.tsx
```diff
@@ -163,15 +186,49 @@ function ResizableImage ({ readOnly, onResizeStop, node, updateAttrs, selected }
...
+    for (let i = 0; i < bufferLength; i += chunkSize) {
+      const chunk = fileContent.slice(i, i + chunkSize);
+      bufferChunks.push(chunk);
+    }
+
+    const file: File = new File(bufferChunks, fileName, { type: `image/${fileExtension}` });
+
+    uploadToS3(file)
+      .then(({ url }) => {
+        node.attrs.src = url;
```
Devorein, line 215: We need to update the attribute by calling `updateAttrs` and passing in the new src value

Devorein (changes requested):

Devorein:
I just tried to copy and paste a large base64 encoded image and the app froze😶
Actually just as you said its happening locally only but not on staging environment

Devorein (approved):

PR #815, opened 2022-08-19
Server tree error

motechFR:


Devorein (approved):

components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
```diff
@@ -170,16 +172,16 @@ function PageNavigation ({
       id: droppedItem.id,
       index, // send it to the end
       parentId
-    });
+    }).then(((updatedPage) => {
+      setPages(_pages => ({
```
mattcasey, line 176: it was by design that we didn't wait to show the page update, did you test this on an account where the query is slow?

pages/api/pages/[id]/index.ts
```diff
@@ -69,6 +71,23 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
     throw new NotFoundError();
   }
 
+  const hasNewParentPage = typeof updateContent.parentId === 'string';
+
+  if (hasNewParentPage) {
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:76]: this adds a slow query to the update route. we should also check that the parentId has changed, since we have the current parentId, or perhaps it's time to even create a new endpoint so it doesn't slow down updates when the user is just typing or changing the title

components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
motechFR: Yes, agreed this may introduce a latency. I tested this by going into staging, adding quite a few pages then running the request. Since I have a higher latency based on my location, I did notice a small UI delay.

You are correct there may be better ways to handle this.

What I noticed, however was that I was able to reposition a page to a child and have this persisted in database, which was the concern behind this PR

I think the much better solution, will be
- Check for the parent id condition in UI and cancel the drop if not valid
- Reset back as it was so we don't await the update and have the UI issues

mattcasey (changes requested):
please see my comments, and add more context here for this fix, since it looks like a hack. the error will be confusing to the user

components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
motechFR: @mattcasey Updated this code section to add the "don't move parent into child" logic into the UI

pages/api/pages/[id]/index.ts
```diff
@@ -69,6 +71,23 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
+    const { flatChildren } = await resolvePageTree({
+      pageId,
+      flattenChildren: true
+    });
+
+    const newParentId = updateContent.parentId as string;
+
+    if (newParentId === pageId || flatChildren.some(p => p.id === newParentId)) {
+      throw new UndesirableOperationError(`You cannot reposition a page to be a child of ${newParentId === pageId ? 'itself' : 'one of its child pages'}`);
```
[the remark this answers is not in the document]
Comment (motechFR) [reply pages/api/pages/[id]/index.ts:85]: These concepts are unrelated. This code section only ensures we aren't making a page a child of one of its existing children

PR #817, opened 2022-08-21
Revamped user profile nexus page with deepdao integration

Devorein:
[Loom Recording](https://www.loom.com/share/091af29895584fc6b08a34deeda5c141)

### DIFF
--- a/components/profile/components/DeepDaoData.tsx
+++ b/components/profile/components/DeepDaoData.tsx
@@ -174,13 +174,41 @@ function DeepDaoOrganizationRow ({ organization }: DeepDaoOrganizationRowProps) 
           <Stack gap={2}>
             {
                 (currentTask === 'vote' ? votes : proposals).map((event, eventNumber) => (
-                  <Stack key={event.id} flexDirection='row' justifyContent='space-between'>
-                    <Stack flexDirection='row' gap={1} alignItems='center'>
-                      {event.verdict ? <ThumbUpIcon color='success' fontSize='small' /> : <ThumbDownIcon color='error' fontSize='small' />}
+                  <Stack key={event.id} flexDirection='row' gap={1}>
+                    <Stack
+                      flexDirection='row'
+                      gap={1}
+                      alignItems='center'
+                      sx={{
+                        alignSelf: 'flex-start'
+                      }}
+                    >
+                      {event.verdict ? (
+                        <ThumbUpIcon
+                          color='success'
+                          fontSize='small'
+                        />
+                      ) : <ThumbDownIcon color='error' fontSize='small' />}
                       <Typography fontWeight={500}>{eventNumber + 1}.</Typography>
-                      <Typography>{event.title}</Typography>
                     </Stack>
-                    <Typography variant='subtitle1'>{showDateWithMonthAndYear(event.createdAt, true)}</Typography>
+                    <Stack
+                      gap={0.5}
+                      sx={{
+                        flexGrow: 1,
+                        flexDirection: {
+                          sm: 'column',
+                          md: 'row'
+                        },
+                        alignItems: 'flex-start'
+                      }}
+                    >
+                      <Typography sx={{
+                        flexGrow: 1
+                      }}
+                      >{event.title}
+                      </Typography>
+                      <Typography variant='subtitle2' textAlign={{ sm: 'left', md: 'right' }} minWidth={100}>{showDateWithMonthAndYear(event.createdAt, true)}</Typography>
+                    </Stack>
                   </Stack>
                 ))
               }
--- a/components/profile/components/DeepDaoData.tsx
+++ b/components/profile/components/DeepDaoData.tsx
@@ -10,6 +10,8 @@ import ForumIcon from '@mui/icons-material/Forum';
 import { useState } from 'react';
 import ThumbUpIcon from '@mui/icons-material/ThumbUp';
 import ThumbDownIcon from '@mui/icons-material/ThumbDown';
+import { GetPoapsResponse } from 'lib/poap';
+import { ExtendedPoap } from 'models';
 import { UserDetailsProps } from './UserDetails';
 
 export function AggregatedDataItem ({ value, label }: { value: number, label: string }) {
@@ -154,7 +156,14 @@ function DeepDaoOrganizationRow ({ votes, proposals, organization, earliestEvent
 
 }
 
-export function AggregatedData ({ user }: Pick<UserDetailsProps, 'user'>) {
+type OrganizationDetails = DeepDaoOrganization & {
+  proposals: DeepDaoProposal[],
+  votes: DeepDaoVote[],
+  oldestEventDate: string,
+  latestEventDate: string
+}
+
+export function DeepDaoData ({ user, poapData }: Pick<UserDetailsProps, 'user'> & {poapData: GetPoapsResponse | undefined}) {
   const { data, isValidating } = useSWRImmutable(user ? `userAggregatedData/${user.id}` : null, () => {
     return charmClient.getAggregatedData(user.id);
   });
@@ -171,12 +180,28 @@ export function AggregatedData ({ user }: Pick<UserDetailsProps, 'user'>) {
     return null;
   }
 
-  const organizationsRecord: Record<string, DeepDaoOrganization & {proposals: DeepDaoProposal[], votes: DeepDaoVote[]}> = data.organizations
+  const poaps: (Partial<ExtendedPoap> & {createdAt: string})[] = [];
+
+  const hiddenPoapIds: Set<string> = new Set();
+  const visiblePoapIds: Set<string> = new Set();
+
+  poapData?.hiddenPoaps.forEach(poap => {
+    poaps.push({ ...poap, createdAt: poap.created as string });
+    hiddenPoapIds.add(poap.id as string);
+  });
+  poapData?.visiblePoaps.forEach(poap => {
+    poaps.push({ ...poap, createdAt: poap.created as string });
+    visiblePoapIds.add(poap.id as string);
+  });
+
+  const organizationsRecord: Record<string, OrganizationDetails> = data.organizations
     .reduce((acc, org) => ({ ...acc,
       [org.organizationId]: {
         ...org,
         proposals: [],
-        votes: []
+        votes: [],
+        oldestEventDate: '',
+        latestEventDate: ''
       } }), {});
 
   data.proposals.forEach(proposal => {
@@ -188,16 +213,11 @@ export function AggregatedData ({ user }: Pick<UserDetailsProps, 'user'>) {
   });
 
   // Sort the proposals and votes based on their created at date and attach organization data with it
-  const events = [...data.proposals, ...data.votes]
-    .sort((eventA, eventB) => eventA.createdAt > eventB.createdAt ? 1 : -1)
-    .map(event => ({ ...event, organization: organizationsRecord[event.organizationId] }));
+  const events = [...data.proposals.map(proposal => ({ type: 'proposal', ...proposal })), ...data.votes.map(vote => ({ type: 'vote', ...vote }))]
+    .sort((eventA, eventB) => eventA.createdAt > eventB.createdAt ? 1 : -1);
 
   const organizationIdSet: Set<string> = new Set();
   const sortedOrganizationIds: string[] = [];
-  const organizationEventDates: Record<string, {
-    oldest: string
-    latest: string
-  }> = {};
 
   events.forEach(event => {
     if (!organizationIdSet.has(event.organizationId)) {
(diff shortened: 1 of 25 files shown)


pages/api/profile/poaps.ts
```diff
@@ -41,9 +41,10 @@ async function getUserPoaps (req: NextApiRequest, res: NextApiResponse<GetPoapsR
     throw new InvalidStateError('User not found');
   }
 
-  const poaps: Partial<ExtendedPoap>[] = await getPOAPs(user.addresses);
-  const hiddenPoaps = poaps.filter((poap: Partial<ExtendedPoap>) => hiddenPoapIDs.find((tokenId :string) => poap.tokenId === tokenId));
-  const visiblePoaps = poaps.filter((poap: Partial<ExtendedPoap>) => !hiddenPoapIDs.find((tokenId :string) => poap.tokenId === tokenId));
+  const poaps: ExtendedPoap[] = await getPOAPs(user.addresses);
+
+  const hiddenPoaps = poaps.filter((poap: ExtendedPoap) => hiddenPoapIDs.find((tokenId :string) => poap.tokenId === tokenId));
```
msadura, line 46:
```suggestion
  const hiddenPoaps = poaps.filter((poap: ExtendedPoap) => hiddenPoapIDs.includes(poap.tokenId));
```

pages/api/profile/poaps.ts
```diff
@@ -41,9 +41,10 @@ async function getUserPoaps (req: NextApiRequest, res: NextApiResponse<GetPoapsR
...
+  const visiblePoaps = poaps.filter((poap: ExtendedPoap) => !hiddenPoapIDs.find((tokenId :string) => poap.tokenId === tokenId));
```
msadura, line 47:
```suggestion
  const visiblePoaps = poaps.filter((poap: ExtendedPoap) => !hiddenPoapIDs.includes(poap.tokenId));
```

pages/api/nft/list.ts
```diff
@@ -13,9 +13,9 @@ handler
   .get(getNfts);
 
 async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {error: string}>) {
+  // address = '0x155b6485305ccab44ef7da58ac886c62ce105cf9'
```
msadura, line 16: remove comment

components/profile/components/UserCollectives.tsx
```diff
@@ -0,0 +1,153 @@
...
+      type: 'nft',
+      date: nft.timeLastUpdated,
+      id: nft.tokenId,
+      image: nft.image,
+      title: nft.title,
+      link: `https://app.poap.xyz/token/${nft.tokenId}`
+    });
+  });
+
+  collectives = collectives.sort((collectiveA, collectiveB) => collectiveA.date > collectiveB.date ? -1 : 1);
```
msadura, line 99: according to types, dates are strings so sorting will not work properly here.
PS. Array.sor

```
collectives.sort((collectiveA, collectiveB) => new Date(collectiveB.date) - new Date(collectiveA.date));
```
should to the job.
PS. you do not need to reassign collectives variable, because .sort mutates the orginal array anyway

pages/api/profile/poaps.ts
Devorein: Thanx, that was unnecessarily complicated code 😛. Updated

pages/api/profile/poaps.ts
Devorein: Good catch. Updated 😄

pages/api/nft/list.ts
Devorein: We left a similar comment inside `lib/deepdao/client.ts` to help us test our poaps/nfts and deepdao data

components/profile/components/UserCollectives.tsx
Devorein: It's actually possible to compare date strings using comparison operators (we are using the same pattern in a lot of places), but I agree it's not intuitive that that's the case. I've updated to cast it to date, but the above snippet throws the typescript error `ts(2363)`. Good catch on the sort one, I always thought that it returned a new array and didn't sort in place. Updated

components/profile/components/DeepDaoData.tsx
```diff
@@ -0,0 +1,163 @@
...
+  if (!data) {
+    return null;
+  }
+
+  const organizationsRecord: Record<
+    string,
+    OrganizationDetails | undefined
+  > = data.organizations
+    .reduce((acc, org) => ({
+      ...acc,
```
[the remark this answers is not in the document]
Devorein: We seem to be following this pattern in a lot of places, but I've updated this portion of the code to attach the value to `acc[org.organizationId] = org` rather than spreading

components/profile/components/DeepDaoData.tsx
```diff
@@ -0,0 +1,163 @@
...
+        ...org,
+        oldestEventDate: '',
+        latestEventDate: '',
+        proposals: [],
+        votes: []
+      }
+    }), {});
+
+  // Sort the proposals and votes based on their created at date and attach organization data with it
+  const events = [...data.proposals.map(proposal => ({ type: 'proposal', ...proposal })), ...data.votes.map(vote => ({ type: 'vote', ...vote }))];
```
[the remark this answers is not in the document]
motechFR: Hey @msadura great comments! Agree with what you're saying.

@Devorein Can you extract these data manipulation methods to a helper in the client-accessible area of the deepdao lib?

components/profile/components/DeepDaoData.tsx
```diff
@@ -0,0 +1,165 @@
...
+        organization.latestEventDate = event.createdAt;
+      }
+      else if (organization.latestEventDate < event.createdAt) {
+        organization.latestEventDate = event.createdAt;
+      }
+    }
+  });
+
+  const sortedOrganizations = Object.values(organizationsRecord).filter(isTruthy)
+    .sort((orgA, orgB) => orgA.latestEventDate > orgB.latestEventDate ? -1 : 1)
```
motechFR, line 110: Small optimisation, it's better to filter before sorting. This way you only sort on data you intend to keep

components/profile/components/DeepDaoOrganizationRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  latestEventDate: string
+}
+
+interface DeepDaoOrganizationRowProps {
+  organization: OrganizationDetails
+}
+
+export default function DeepDaoOrganizationRow ({ organization }: DeepDaoOrganizationRowProps) {
+  const [isCollapsed, setIsCollapsed] = useState(true);
+  const [currentTask, setCurrentTask] = useState<'vote' | 'proposal'>('vote');
```
motechFR, line 39: Could be better renamed as currentTab?

components/profile/components/DeepDaoOrganizationRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  const theme = useTheme();
+
+  const proposals: DeepDaoEvent[] = organization.proposals
```
motechFR, line 42: Question; I see lots of votes, but empty proposals each time, why is this?

pages/api/nft/list.ts
motechFR: I added a comment here to explain why we have this to fellow dev team members

motechFR (approved):
This PR does what it says, some small comments but looks good to merge

components/profile/components/UserCollectives.tsx
```diff
@@ -0,0 +1,153 @@
...
+import { useMyNfts } from 'hooks/useMyNfts';
+import { GetPoapsResponse } from 'lib/poap';
+import { showDateWithMonthAndYear } from 'lib/utilities/dates';
+import { usePopupState } from 'material-ui-popup-state/hooks';
+import { ExtendedPoap } from 'models';
+import { KeyedMutator } from 'swr';
+import ManagePOAPModal from './ManagePOAPModal';
+import { isPublicUser, UserDetailsProps } from './UserDetails';
+
+interface Collective {
```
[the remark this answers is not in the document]
Devorein: No it's not used anywhere else. We are just using it as component props. So I think we can keep it where its located at the moment

components/profile/components/DeepDaoOrganizationRow.tsx
Devorein: It depends on the wallet address. You can visit deepdao, navigate to the people section and sort by proposals to get the user wallet address with the highest amount of proposals. https://deepdao.io/people

components/profile/components/DeepDaoData.tsx
```diff
@@ -0,0 +1,165 @@
+import { Box, Chip, CircularProgress, Divider, Grid, Paper, Stack, Typography } from '@mui/material';
```
mattcasey, line 1: this component is not only DeepDao (we include our workspaces as DAOs for example), AggregatedData was a better name for it. Please rename to something more general to avoid confusion

theme/index.ts
```diff
@@ -107,6 +107,22 @@ export const createThemeLightSensitive = (mode: PaletteMode) => {
         fontWeight: 700
       },
       h2: {
+        fontSize: '1.75rem',
```
mattcasey, line 110: where did these values come from? Did you check all the places we're using h* tags to make sure this is correct? There is no requirement we need to use h tags in the profile layout

pages/api/nft/list.ts
```diff
@@ -13,9 +13,10 @@ handler
   .get(getNfts);
 
 async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {error: string}>) {
+  // Use this address for testing as it has alot of on-chain data
+  // req.session.user.addresses = ['0x155b6485305ccab44ef7da58ac886c62ce105cf9'];
```
mattcasey, line 17: Please note this field is not updated. We should query the information instead of relying on session data

components/profile/components/DeepDaoData.tsx
Devorein: Thats a good point. Updated

pages/api/nft/list.ts
Devorein: Thank you 😄

lib/poap/interfaces.ts
```diff
@@ -1,8 +1,8 @@
 import { ExtendedPoap } from 'models';
 
 export interface GetPoapsResponse {
-    visiblePoaps: Array<Partial<ExtendedPoap>>;
-    hiddenPoaps: Array<Partial<ExtendedPoap>>;
+    visiblePoaps: Array<ExtendedPoap>;
```
mattcasey, line 4: please clean up so we are not using `Array` but `[]`

components/profile/components/DeepDaoData.tsx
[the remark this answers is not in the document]
Devorein: Merging the proposal and votes into 1 array was to attach the type discriminator for easier event detection, and to reduce the number of lines to detect the earliest and latest event date. But I agree, it should be extracted to a separate lib method for reusability and better readability. Updated 😄

mattcasey (changes requested):
I'm not seeing my NFT when looking at the public view:
image

components/profile/components/DeepDaoData.tsx
Devorein: Good call. Reverting it back to `AggregatedData` rn.

lib/poap/interfaces.ts
Devorein: Done. But there are still a lot of places we are using `Array` rather than `[]` 🤔

pages/api/nft/list.ts
Devorein: Updated api endpoint to take in `userId` so that a user's nfts can be viewed publicly

theme/index.ts
Devorein: We are actually not using `h2-4` in any place. So I assumed that we are not using h5 and h6 but we are. I've reverted back changes from theme.

pages/api/nft/list/[userId].ts
```diff
@@ -1,21 +1,35 @@
...
 const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
 
 handler
-  .use(requireUser)
   .get(getNfts);
 
 async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {error: string}>) {
-  const { addresses } = req.session.user;
-  const chainId = 1;
+  // address = '0x155b6485305ccab44ef7da58ac886c62ce105cf9'
```
Comment (mattcasey) [inline pages/api/nft/list/[userId].ts:16]: i think we can delete this now? or can u bring back the comment of what this is for

pages/api/nft/list/[userId].ts
```diff
@@ -1,21 +1,35 @@
...
+    }
+  });
+
+  if (!user) {
+    throw new InvalidStateError('User not found');
+  }
 
+  const { addresses } = user;
+  const chainId = 1;
   const nfts = await alchemyApi.getNfts(addresses || [], chainId);
```
Comment (mattcasey) [inline pages/api/nft/list/[userId].ts:33]: we don't need `|| []` here anymore

hooks/useMyNfts.ts
```diff
@@ -2,9 +2,10 @@ import charmClient from 'charmClient';
 import { useUser } from 'hooks/useUser';
 import useSWR from 'swr';
 
-export const useMyNfts = () => {
+export const useMyNfts = (userId?: string) => {
   const { user } = useUser();
```
mattcasey, line 6: it seems like we could just remove useUser and pass the userID in instead

hooks/useMyNfts.ts
```diff
@@ -2,9 +2,10 @@ import charmClient from 'charmClient';
...
-  const { data, error: serverError } = useSWR(user && `/nfts/list/${user.id}`, () => charmClient.nft.list());
+  userId = user?.id || userId;
```
mattcasey, line 7: userId should take precendence (but see my note about using useUser in the first place)

pages/api/nft/list/[userId].ts
Comment (Devorein) [reply pages/api/nft/list/[userId].ts:33]: Oops my bad. Removed 😛

hooks/useMyNfts.ts
Devorein: Updated to only use `userId`

hooks/useMyNfts.ts
Devorein: Ahh silly mistake. Yes updated the hook to take in the userId rather than using the current user

components/profile/components/DeepDaoOrganizationRow.tsx
Devorein: Agree

mattcasey (approved):

PR #819, opened 2022-08-22
Use contributors hook for Focalboard properties

mattcasey:
This fixes a "bug" where contributor list was not being updated. The useContributors hook uses SWR so it will periodically refresh. I also used our standard UserDisplay component, so now we support user avatars.

We could almost get rid of workspaceusers in focalboard, except it's still used for sorting by user. This would have the same issue as card properties, but going to leave that alone for now until we figure out how to have blocks update autmoatically


Devorein (approved):

components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
```diff
@@ -27,47 +23,36 @@ const selectStyles = {
...
-    </div>
-  );
-}
-
-function UserProperty (props: Props): JSX.Element {
-  const workspaceUsers = useAppSelector<IUser[]>(getWorkspaceUsersList);
-  const workspaceUsersById = useAppSelector<{[key:string]: IUser}>(getWorkspaceUsers);
+function UserProperty (props: Props): JSX.Element | null {
+  const [contributors] = useContributors();
+  const contributorMap = contributors.reduce<Record<string, Contributor>>((acc, contributor) => {
```
msadura, line 28: as far as I understand this map is to easily get contributor by id. In such case I would consider 2 solutions:

1) Move this map to useContributor context and memoize it there - it can be then reused in other places if needed without creating it over again

2) Probably even more widely used solution - expose getter function from useContributors hook like `getContributor(id: string)` which would basically do:

```
const getContributor =  useCallback(() => contributors.find(c => c.id === id), [contributors])
```

That would be nice long-term solution for cases when you need to get contributor by id, but at the same time it would rather require refactoring useContributors to return object with methods  rather than array

PR #821, opened 2022-08-22
Validate token gate has at least 1 condition before creation

motechFR:


pages/api/token-gates/index.ts
```diff
@@ -26,6 +28,13 @@ async function saveTokenGate (req: NextApiRequest, res: NextApiResponse) {
     throw error;
   }
 
+  // Make sure token gate has at least 1 condition.
+  const hasValidCondition = (req.body.conditions?.unifiedAccessControlConditions as AccessControlCondition[])?.some(c => isTruthy(c.chain));
```
mattcasey, line 32: no need to use isTruthy here since you're not trying to reduce the type in an array. `.filter(Boolean)` is fine too

mattcasey (approved):

pages/api/token-gates/index.ts
motechFR: I like the semanticness of isTruthy, but fair enough for using Boolean when it's very deterministic as in this case

PR #823, opened 2022-08-23
update lit modal

mattcasey:
this version include a fix that disallows users to save empty conditions by selecting 'multiple conditions'

### DIFF
--- a/components/settings/roles/components/TokenGates/components/TokenGateRolesSelect.tsx
+++ b/components/settings/roles/components/TokenGates/components/TokenGateRolesSelect.tsx
@@ -1,5 +1,6 @@
-import { Box, Chip, FormControl, MenuItem, Select } from '@mui/material';
+import { Box, Chip, FormControl, MenuItem, Select, Tooltip, Typography } from '@mui/material';
 import { SelectChangeEvent } from '@mui/material/Select';
+import { InfoOutlined as InfoOutlinedIcon } from '@mui/icons-material';
 import { useMemo } from 'react';
 import styled from '@emotion/styled';
 import Button from 'components/common/Button';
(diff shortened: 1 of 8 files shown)


components/settings/roles/components/TokenGates/components/TokenGatesTable.tsx
```diff
@@ -109,7 +109,7 @@ export default function TokenGatesTable ({ isAdmin, onDelete, tokenGates }: Prop
             <TableCell sx={{ px: 0 }}>Description</TableCell>
             <TableCell>
               <Tooltip arrow placement='top' title='Automatically assign these roles to new users'>
-                <span>Included Roles</span>
+                <span>Assigned Roles</span>
```
mattcasey, line 112: We use the verbs "Assign", "Include" and "Add", but I think only one is necessary and makes it feel more deliberate.

components/settings/roles/components/TokenGates/components/TokenGateRolesSelect.tsx
```diff
@@ -50,6 +51,19 @@ export default function TokenGateRolesSelect ({ onDelete, selectedRoleIds, onCha
 
   const isAdmin = useIsAdmin();
 
+  if (roles?.length === 0) {
+    return (
+      <Box display='flex' justifyContent='center'>
```
mattcasey, line 56: instead of just a disabled button, replace the input with a tooltip:
image

mattcasey:
While working on this, I did some clean-up of the UI to make things look more consistent:
image

motechFR (approved):
Tested and approved, new Lit Modal is blocking me from creating a multiple condition gate with 0 conditions

PR #824, opened 2022-08-23
fix share button not appearing on database pages

mattcasey:
this was a little rabbit hole. I did some clean-up too:
- Fixed a bug: Share button UI doesn't currently update at all when you toggle public pages
- The Header now keeps track of its own page Id, referred to as 'basePageId', instead of currentPageId which we should probably remove eventually from the global hook
- Replace logic to display "..." menu by checking page type rather than the URL path
- reorganized the ShareBountyButton and ShareToWeb components (I just moved them into the components folder)


motechFR (approved):
@mattcasey looks good!

I made 1 small change, deleted the refreshPage call since the parent is already doing this.

Rename parent refreshPermissions to refreshPageAndPermissions to be clearer on its intent

mattcasey:
thanks @motechFR ! this must have been broken since i did the refactor... :/

PR #831, opened 2022-08-23
Inline database full width

mattcasey:
@Devorein please test in Chrome. I tested this in Firefox, both on an inline database that needed scrollbar and one that didn't. It seems to work pretty well, if there is a little delay while resizing:
image

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -17,6 +17,7 @@ import { Page, PageContent } from 'models';
 import dynamic from 'next/dynamic';
 import { useRouter } from 'next/router';
 import { memo, useCallback, useEffect, useState } from 'react';
+import { useElementSize } from 'usehooks-ts';
 import BountyProperties from './components/BountyProperties';
 import CreateVoteBox from './components/CreateVoteBox';
 import PageBanner from './components/PageBanner';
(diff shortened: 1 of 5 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -121,93 +124,97 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false }: Documen
...
-          <CharmEditor
-            key={page.id}
-            content={page.content as PageContent}
-            onContentChange={updatePageContent}
-            readOnly={cannotEdit}
-            pageActionDisplay={!insideModal ? currentPageActionDisplay : null}
-            pageId={page.id}
-            disablePageSpecificFeatures={isSharedPage}
-            enableVoting={true}
+        <div ref={containerRef}>
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:140]: i had to add a special div to check its width so that we know the available area minus the scrollbar, if one exists

components/common/CharmEditor/components/inlineDatabase/components/InlineDatabase.tsx
```diff
@@ -41,8 +41,9 @@ const StylesContainer = styled.div`
     margin: 0 -24px;
     padding-left: 24px;
     ${({ theme }) => theme.breakpoints.up('md')} {
-      margin: 0 -80px;
-      padding-left: 80px;
+      --side-margin: ${({ containerWidth }) => `calc((${containerWidth}px - 100%) / 2)`};
```
mattcasey, line 44: We need two widths: the one of the container and one of the charmeditor/inline database. I used 100% inside of calc() instead of using more JS for the latter. Seems to work!

Devorein (approved):

PR #834, opened 2022-08-24
bring back top-left icon for Nexus

mattcasey:
This brings back the original idea for navigation for Nexus, but includes the notifcation badge on top of the user's avatar.

I also removed the top-right chain/username. We put that there because it's familiar in Web3, but for our app it doesn't really serve a practical purpose other than a link to your profile. 

image


motechFR (approved):
Looks great! Approved with small tweaks

PR #837, opened 2022-08-24
Initialise playwright config

motechFR:
Note: Currently having issues installing dependencies

https://github.com/microsoft/playwright/issues/11122

If having issues with **sudo npx playwright install-deps**, we can use **sudo apt-get install gstreamer1.0-libav** to install the required browsers


package.json
```diff
@@ -34,12 +35,14 @@
     "db-tool:test": "dotenv -e .env.test.local -- npx prisma studio",
     "test:setup-db": "./scripts/configure-db.sh",
     "test:browser": "npm run test-setup && jest --config='./jest.config-client.ts'",
+    "test:browser:e2e": "dotenv -e .env.test.local -- npx playwright test __e2e__/public-page.spec.ts --timeout=60000",
```
mattcasey, line 38: The command grammar is <action>:<test_suite>:<env>. Also, this tests more than just the browser  - the server too! In this case, I think we should just come up with an alternative to "browser" like "test:e2e"

package.json
```diff
@@ -17,6 +17,7 @@
     "migrate:staging": "npx dotenv -e .env.staging -- prisma migrate deploy",
     "server:dev": "next dev",
     "debug": "NODE_OPTIONS='--inspect' next",
+    "debug:playwright": "dotenv -e .env.test.local -- npx playwright test __e2e__/public-page.spec.ts --headed --timeout=0",
```
mattcasey, line 20: maybe "debug:e2e"?

tests-examples/demo-todo-app.spec.ts
```diff
@@ -0,0 +1,398 @@
+import { test, expect, type Page } from '@playwright/test';
```
mattcasey, line 1: was it intentional to leave this file here? I think it would be better to link to the API docs, and usually our own tests will become better examples.

components/common/Page/PageDialog.tsx
```diff
@@ -158,8 +160,11 @@ export default function PageDialog (props: Props) {
           )}
           onClose={onClose}
         >
-          {page && <DocumentPage insideModal page={page} setPage={setPage} readOnly={props.readOnly} />}
+          <div data-test='page-dialog'>
```
mattcasey, line 163: We should try not to add extra HTML elements for data-test attribute, it can mess with the CSS of surrounding code. Can you move this to the Dialog or DocumentPage component instead?

__e2e__/public-page.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  await copyButton.click({ force: true });
+
+  const clipboardContent = await page.evaluate(async () => {
+    return navigator.clipboard.readText();
+  });
+
+  expect(clipboardContent.match(shareUrl)).not.toBe(null);
+
+  // Part B - Visit this page as a non logged in user
+  const publicContext = await browser.newContext({});
```
mattcasey, line 67: a new user should be a separate test

__e2e__/public-page.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  const targetRedirectPath = `${baseUrl}/${domain}/${boardPage?.path}`;
+
+  await page.goto(targetRedirectPath);
+
+  // Act ----------------------
+  // Part A - Prepare the page as a logged in user
+  // 1. Make sure the board page exists and cards are visible
+  await page.waitForResponse(/\/api\/blocks/);
+
+  await expect(page.locator('data-test=kanban-card').first()).toBeVisible();
```
mattcasey, line 35: this expectation seems too specific for a test about public pages

__e2e__/public-page.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+  const userContext = await browser.newContext({ permissions: ['clipboard-read', 'clipboard-write'] });
+  const page = await userContext.newPage();
+
+  const { user: profile, space, pages } = await createUserAndSpace({ browserPage: page });
+
+  const domain = space.domain;
+
+  const boardPage = pages.find(p => p.type === 'board' && p.title.match(/tasks/i) !== null);
+
+  await page.goto(`${baseUrl}`);
```
mattcasey, line 24: this goto() seems unecessary based on line 28

__e2e__/record-custom.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { chromium } from '@playwright/test';
```
mattcasey, line 1: please use camelCase for files

__e2e__/public-page.spec.ts
```diff
@@ -0,0 +1,111 @@
+import { Browser, chromium, expect, test } from '@playwright/test';
```
mattcasey, line 1: use camelCase for file names

testing/playwright.ts
```diff
@@ -0,0 +1,70 @@
...
+  walletAddress?: string}): Promise<LoggedInUser> {
+  return browserPage.request.post(`${baseUrl}/api/profile`, {
+    data: {
+      address: walletAddress
+    }
+  }).then(res => res.json());
+}
+
+export async function createSpace ({ browserPage, createdBy }: {browserPage: BrowserPage} & Pick<Space, 'createdBy'> & Partial<Pick<Space, 'permissionConfigurationMode'>>): Promise<Space> {
+  return browserPage.request.post(`${baseUrl}/api/spaces`, {
```
mattcasey, line 18: Did you try using our standard test fixtures from testing/setupDatabase.ts instead of calling the api? Calling the API thru the browser is unnecessary overhead. We don't need to test that the API can be called, and if we wanted to test that it works then we should have tests on the API itself.

testing/playwright.ts
```diff
@@ -0,0 +1,70 @@
+import { Space } from '@prisma/client';
```
mattcasey, line 1: this file could just go into __e2e__ since that's the only place it will get used, and i would rename it to e2eUtils or something. At first i assumed it was some playwright config

mattcasey (changes requested):
i added "M:" (for "Must") to the beginning of my comments where i'd like a response at least. The idea is a suggestion from that article about using Must/Could/Should/Would.

package.json
motechFR: Gotcha,

Updated local test to test:e2e, and the ci version to test:e2e:ci

package.json
motechFR: Agreed, let's keep our script names library agnostic

tests-examples/demo-todo-app.spec.ts
motechFR: Thanks for the spot, we definitely don't need this folder.

We have a working example that uses our own product now

components/common/Page/PageDialog.tsx
motechFR: @mattcasey Just finished re-testing with alternative approach.

If you see the images below, you can see the selector was dropped. Not sure why this is.

image
image


--

I agree with you, we don't want to have to edit production code just to implement the test.

As a solution, I've updated the dialog selector to rely on the .Dialog classname of the dialog div

__e2e__/public-page.spec.ts
motechFR: Thanks for the push for us to find an even better way!

The key question to achieve this is, how can we run the tests serially, so previous tests can setup data for latter tests?

I found this specific method we can use to define a serially executed test suite (without losing the benefit of parallel execution of all test suites)
https://playwright.dev/docs/api/class-test#test-describe-serial

The tests are now split in 2

__e2e__/public-page.spec.ts
motechFR: I was running into an issue where the modal would collapse before it could get interacted with properly.

My solution was to wait for the page's board to finish rendering, and only then interact with the share modal

__e2e__/record-custom.spec.ts
motechFR: I've added some documentation inline about this page's purpose.

It allows us to spin up a logged in space and start recording our interactions, generating various playwright code assertions we can copy paste into a test.

testing/playwright.ts
motechFR: There are 2 aspects that led to this approach of passing calls through the browser:

1. Our setupDatabase imports prisma. Using these methods in the browser causes a compilation error

2. I did go with direct API calls, but then had to handle the returned cookies. With the current approach, the browser calls create user endpoint, and handles the setting of the returned cookie.

testing/playwright.ts
motechFR: Makes sense, I've relocated the page

motechFR:
@mattcasey Thanks for all the feedback. All implemented and works in CI too.

For running the tests, you should be able to use **npm run test:e2e** and it'll run automagically.

You can copy the command content and add "--headed" option so you actually see what's going on in the browser

mattcasey (approved):
works now, nice job

PR #839, opened 2022-08-24
Tt3 deploy docker compose

yiwork:


package.json
```diff
@@ -14,7 +14,7 @@
     "cron:dev": "dotenv -e .env.local -- ts-node ./background/cron.ts",
     "migrate:test": "dotenv -e .env.test.local -- npx prisma migrate deploy",
     "migrate:prod": "npx dotenv -e .env.staging -- prisma migrate deploy",
-    "migrate:staging": "npx dotenv -e .env.staging -- prisma migrate deploy",
+    "migrate:staging": "prisma migrate deploy",
```
mattcasey, line 17: removed it from wrong place

Dockerfile
```diff
@@ -14,4 +14,20 @@ ENV PORT 3000
 ENV NEXT_TELEMETRY_DISABLED 1
 
 EXPOSE 3000
-CMD ["sh", "-c", "exec npm run $DOCKER_NPM_COMMAND"]
+CMD ["npm", "run", "start:test"]
```
mattcasey, line 17: i think this would break production

yiwork: Yes, but refer to docker-compose.yml line #7 `command: [ "npm", "run", "${NPM_SCRIPT}" ]`

At docker-compose run time I override the built-in command inside this dockerfile with the one in docker-compose. 

Note that what I'm leveraging is the shell environment variable that's present when running docker-compose (i.e. the env variable that's interpreted *outside* the docker container).  

I _still_ advise against using variables inside the Dockerfile itself b/c that would be variables to be interpreted *inside* the docker container, which doesn't get interpreted until you wrap said command with `sh -c` (which is a command running inside a shell inside a docker container).  This would make the shell process number 1, preventing system signals from getting to the program that you are running. But if you must use `sh -c` in the `CMD` you should definitely use `exec` along with it. 

Basically: 
`CMD ["sh", "-c", "npm", "run", "$NPM_SCRIPT"]` => BAD
`CMD ["sh", "-c", "exec", "npm", "run", "$NPM_SCRIPT"]` => BETTER
`CMD ["npm", "run", "start:staging"]` in Dockerfile then use docker-compose to override `CMD ["npm", "run", "$SCRIPT_VAR_FROM_DOCKER_COMPOSE"]` => BEST

mattcasey: @yiwork but we still need this for production

PR #840, opened 2022-08-25
move source selection to focalboard

mattcasey:
this moves all the view source selection to the centerPanel component, splits up the 'board' as the inline board and teh 'board' based on whichever view is active

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -113,6 +113,14 @@
         ]
       }
     ],
+    "react/jsx-indent": [
+      2,
+      2,
+      {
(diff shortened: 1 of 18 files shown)


.eslintrc.json
```diff
@@ -113,6 +113,14 @@
         ]
       }
     ],
+    "react/jsx-indent": [
```
mattcasey, line 116: this is intentional... it's bugging me lol. so now:
```
// right:
{booleanCheck && (
  <IconButton />
)}

// wrong:
{booleanCheck && (
<IconButton />
)}
```

PR #844, opened 2022-08-26
Display inline database in public pages

Devorein:


components/common/BoardEditor/focalboard/src/components/SourceSelection.tsx
```diff
@@ -10,6 +10,7 @@ import ArrowBackIosIcon from '@mui/icons-material/ArrowBackIos';
 import { usePages } from 'hooks/usePages';
 import PagesList from 'components/common/CharmEditor/components/PageList';
 import { isTruthy } from 'lib/utilities/types';
+import { useRouter } from 'next/router';
```
mattcasey, line 13: can be removed?

components/common/BoardEditor/focalboard/src/components/SourceSelection.tsx
```diff
@@ -66,13 +67,14 @@ export default function SourceSelection (props: Props) {
...
                 sx={{ fontSize: 'inherit', textDecoration: 'underline' }}
+                disabled={props.readOnly}
               >
                 Select a data source
               </Button>to continue
             </Typography>
           </Stack>
         </Box>
-        <Collapse in={Boolean(sidebarState)} orientation='horizontal'>
+        <Collapse in={props.readOnly === true ? false : Boolean(sidebarState)} orientation='horizontal'>
```
mattcasey, line 77: sidebarState is the only thing that should determine if this is open or not. i think you could just change the default value based on readOnly?:

```
  const [sidebarState, setSidebarState] = useState<SidebarState>(props.readOnly ? null : 'select-source');
```

mattcasey (approved):

PR #852, opened 2022-08-28
Hide user nft in nexus profile

Devorein:


components/profile/components/ManageProfileItemModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+      metadata: {
+        imageURL: nft.image ?? nft.imageThumb,
+        title: nft.title,
+        date: nft.timeLastUpdated
+      },
+      type: 'nft'
+    };
+  });
+
+  const handleHideProfileItem = (profileItemId: string) => {
```
msadura, line 148: Honestly, logic in this component looks a bit nasty.

1) On each update you run a lot of logic which feel unnecessary.
2) For some reason you keep visible / hidden items as separate arrays, just to merge them into single array
3) 2-step updates of hide / show look kinda weird

I think issue in this logic feel like wrong data model. Why do you keep visible / hidden items in 2 separate arrays insteal of 1 with some sort of  `isHidden: boolean` flag?

components/profile/components/ManageProfileItemModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+      hiddenProfileItems: hiddenProfileItemIds.filter(profileItemId => !existingHiddenProfileItemIds.includes(profileItemId))
+        .map(hiddenProfileItemId => profileItemsRecord[hiddenProfileItemId]),
+      shownProfileItems: shownProfileItemIds.filter(profileItemId => !existingVisibleProfileItemIds.includes(profileItemId))
+        .map(shownProfileItemId => profileItemsRecord[shownProfileItemId])
+    });
+    setTabIndex(0);
+    save();
+  };
+
+  const shownProfileItems = shownProfileItemIds.map(profileItemId => profileItemsRecord[profileItemId]);
```
msadura, line 177: Honestly `i think this logic needs a bit of rethinking / different data modeling. There is so many maps / merges over the same data that is feels not really optimal.

pages/api/nft/list/[userId].ts
```diff
@@ -30,10 +42,16 @@ async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {e
...
   const chainId = 1;
-  const nfts = await alchemyApi.getNfts(addresses, chainId);
+  const nfts = await alchemyApi.getNfts(['0x155b6485305ccab44ef7da58ac886c62ce105cf9'], chainId);
+
   const mappedNfts = nfts.map(nft => mapNftFromAlchemy(nft, chainId));
+  const hiddenNfts = mappedNfts.filter((nft) => hiddenNftIds.includes(nft.tokenId));
+  const visibleNfts = mappedNfts.filter((nft) => !hiddenNftIds.includes(nft.tokenId));
 
-  res.status(200).json(mappedNfts);
+  res.status(200).json({
```
Comment (msadura) [inline pages/api/nft/list/[userId].ts:51]: Honestly I do not think this sort of data structure makes sense.
1) You add quite a lot of logic to split hidden/ visible items  on the server
2) Then on client you add even more logic to merge or handle this data on the frontend.

Once again, I believe this data structure is not optimal and it would be better to rethink it a bit to properly fit the purpose of use. I am not sure how do you handle this sor of issues as a team - any brainstorm sessions? CC: @mattcasey

lib/nft/getNfts.ts
```diff
@@ -0,0 +1,9 @@
+import * as alchemyApi from 'lib/blockchain/provider/alchemy';
+import { mapNftFromAlchemy } from 'lib/nft/utilities/mapNftFromAlchemy';
+
+export async function getNFTs (addresses: string[]) {
```
msadura, line 4: THis logic is copied from `pages/api/nft/list/[userId].ts` - can you reuse it there instead of just copying?

pages/api/nft/list/[userId].ts
```diff
@@ -30,10 +42,16 @@ async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {e
 
   const { addresses } = user;
   const chainId = 1;
-  const nfts = await alchemyApi.getNfts(addresses, chainId);
+  const nfts = await alchemyApi.getNfts(['0x155b6485305ccab44ef7da58ac886c62ce105cf9'], chainId);
```
Comment (msadura) [inline pages/api/nft/list/[userId].ts:45]: looks like hardcoded temp data for tests

msadura (changes requested):

components/profile/components/ManageProfileItemModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+    visiblePoaps: ExtendedPoap[];
+    hiddenPoaps: ExtendedPoap[];
+    hiddenNfts: NftData[];
+    visibleNfts: NftData[];
+    save: () => void,
+    close: () => void,
+    isOpen: boolean,
+};
+
+const TabPanel = styled(Box)`
```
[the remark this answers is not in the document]
mattcasey: Agreed, and he could just use the width attribute: `<Box width='100%' py={2} />`

lib/charmClient/interface.ts
```diff
@@ -0,0 +1,6 @@
+import { NftData } from 'lib/nft/types';
+
+export interface GetNftsResponse {
```
mattcasey, line 3: i don't think we need yet another interface file, this would also get huge quickly since it's not namespaced. we should just put the interface in the nftAPi file.

lib/profileItem/interfaces.ts
```diff
@@ -0,0 +1,6 @@
+import { ProfileItem } from '@prisma/client';
+
+export interface UpdateProfileItemRequest {
```
mattcasey, line 3: this should go inside the api endpoint file

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -59,7 +60,7 @@ export default function NftAvatarGallery ({ onSelect, isVisible, onClose, isSavi
           </Grid>
         ) : (
           <Grid container spacing={1}>
-            {nfts?.length ? nfts?.map(nft => (
```
mattcasey, line 62: there's no need to check for a length before mapping. its just more to read, we dont really save any performance

pages/api/nft/list/[userId].ts
```diff
@@ -1,20 +1,32 @@
...
 
 handler
   .get(getNfts);
 
-async function getNfts (req: NextApiRequest, res: NextApiResponse<NftData[] | {error: string}>) {
+async function getNfts (req: NextApiRequest, res: NextApiResponse<GetNftsResponse | {error: string}>) {
   // address = '0x155b6485305ccab44ef7da58ac886c62ce105cf9'
   const userId = req.query.userId as string;
+
+  const hiddenNftIds: Array<string> = (await prisma.profileItem.findMany({
```
Comment (mattcasey) [inline pages/api/nft/list/[userId].ts:19]: dont use Array<> notation

pages/api/nft/list/[userId].ts
Comment (mattcasey) [reply pages/api/nft/list/[userId].ts:51]: In response to your point, I agree we should try to have logic in one place if at all possible. I generally like to put as much logic as possible in the backend, until it has to start handling different logic for more than one 'view'. And there is one big exception: when it's to merge data sources, since the client always has context like users, workspace, info, etc and it allows us to have one source of truth.

Regarding brainstorm sessions, we don't have a standard format, but we do talk a lot ad hoc. I think Safwan and Mo pair fairly often, and sometimes we go through our code at the end of the daily standup (leading sometimes to an hour-long meeting but no one is required to stay for deep dives). A quick 5 minutes call between you too might also be sufficient. When using Discord, default to having discussions in a public channel (usually #team) so that everyone benefits and is aware of any decisions (so try not to use DM so much).

One final note @msadura: when giving feedback I like to keep this article in mind: https://allthecode.co/blog/post/how-to-get-better-at-code-reviews. We don't use the "S/W/C/M:" prefixes they suggest (maybe we should!) but I'll make sure to point out when something is so bad I think it *must* be fixed before merging. Non-optimal code usually falls under "Should" for me, because I want us to prioritize speed over perfection, and our code changes often.

scripts/updateUserPoap.ts
```diff
@@ -0,0 +1,22 @@
+import log from 'lib/log';
+import { prisma } from '../db';
+
+(async () => {
+  const poaps = await prisma.poap.findMany({});
```
mattcasey, line 5: this was not part of the required scope but thanks for taking care of it now

components/profile/PublicProfile.tsx
```diff
@@ -14,12 +15,18 @@ export default function PublicProfile (props: UserDetailsProps) {
...
+      : charmClient.nft.list(props.user.id);
+  });
+
   return (
     <Stack spacing={2}>
       <UserDetails {...props} />
       <Divider />
       <AggregatedData user={props.user} />
-      <UserCollectives user={props.user} mutatePoaps={mutatePoaps} poapData={poapData} />
+      <UserCollectives user={props.user} mutatePoaps={mutatePoaps} poapData={poapData} mutateNfts={mutateNfts} nftData={nftData} />
```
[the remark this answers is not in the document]
mattcasey: I agree, in fact i'd say we only need to call the endpoint when the user opens the modal

pages/api/profile/items.ts
```diff
@@ -0,0 +1,61 @@
...
+          type: profileItem.type
+        }
+      })));
+    }
+    catch (err) {
+      log.error('Error updating profile items', err);
+    }
+  }
+
+  return res.status(200).json({});
```
mattcasey, line 58: just use .end() instead of json({}), an empty object gives the appearance something is being returned or could be

pages/api/profile/items.ts
```diff
@@ -0,0 +1,61 @@
...
+          id: profileItem.id,
+          userId: req.session.user.id,
+          metadata: profileItem.metadata === null ? undefined : profileItem.metadata,
+          isHidden: true,
+          type: profileItem.type
+        }
+      })));
+    }
+    catch (err) {
+      log.error('Error updating profile items', err);
```
mattcasey, line 54: why is this try/catch necessary? if it's actually an error, then it should bubble up otherwise you should be using log.warn

pages/api/profile/items.ts
```diff
@@ -0,0 +1,61 @@
...
+import { NextApiRequest, NextApiResponse } from 'next';
+import nc from 'next-connect';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireUser)
+  .put(updateUserProfileItems);
+
+async function updateUserProfileItems (req: NextApiRequest, res: NextApiResponse<any | {error: string}>) {
```
mattcasey, line 16: this whole flow is an anti-pattern right now: why are we not just saving the user actions as they click to hide/show an individual item? but i know it was like that before, so it's a bit beyond scope. i will make a task to remove the "save button"

pages/api/nft/list/[userId].ts
Comment (msadura) [reply pages/api/nft/list/[userId].ts:51]: Actually what I want to point here is not that this code must be fixed, but from long-term perspective we can improve our approach for some problems to actually speed things up. From what I see in visible / hidden items processing here is adding complexity, where it does not feel needed. 
Current flow iterates over data on the server side to split them, and then they are merged back on the frontend to be processed + update loop in multiple setState's feel overcomplicated.

> Non-optimal code usually falls under "Should" for me, because I want us to prioritize speed over perfection, and our code changes often.

I totally agree with that, but I think lesson can be still learned for the future. Actually I am not talking about making code perfect, but talking about speed and potential code changes:
- having more logic needs more time to actually write the code
- having more complex code takes more time to understand and review
- talking about code changes - complex data structures are actually harder to change. For example - what if you would like to add some sort of filtering or grouping items by different category that visible / hidden - it would be hardly possible without rewriting solution proposed here.

TLDR: My key point here is not to make code perfect, but rather make it simple. For me also means actually better speed and maintainability so this is somehow aligned with overall team goals :) I still need to learn alot about CV codebase, but I will be happy to help with solving this sort of issues as simple  and fast as possible 👍

components/profile/PublicProfile.tsx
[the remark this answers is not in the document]
Devorein: I believe we either wanted to make `PublicProfile` the central component to fetch all the data related to that individual user, or other components did make use of the poaps and nfts before. Either way, since it's not being used by any other components I will transfer the code to fetch the nfts and poaps inside `UserCollectives`  (also rename the component to ProfileItem)

components/profile/components/ManageProfileItemModal.tsx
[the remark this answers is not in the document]
Devorein: Turns out we didn't even need the styled component, removing `width=100%` didn't change anything. I replaced all `TabPanels` with regular MUI `Box`

components/profile/components/ManageProfileItemModal.tsx
```diff
@@ -0,0 +1,238 @@
...
+    border-radius: 10%;
+    display: none;
+
+    & > svg {
+      color: white;
+      cursor: pointer;
+    }
+`;
+
+function ProfileItemsTab ({ profileItems, onClick, tab }: { tab: 'visible' | 'hidden', onClick: (profileItemId: string) => void, profileItems: Pick<ProfileItem, 'id' | 'metadata' | 'type'>[]}) {
```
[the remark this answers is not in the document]
Devorein: I'm not sure whether we are following that approach everywhere. We do have a lot of cases where we have multiple components defined inside a single file, maybe have a discussion on when to have a single component inside a file

components/profile/components/ManageProfileItemModal.tsx
Devorein: I agree that its a bit complex, but the lack of comments doesn't help either. The previous logic was actually a lot harder to follow (even with just poaps). I think we need to update the data that is being send from the backend first before refactoring on the frontend

lib/nft/getNfts.ts
Devorein: Yes absolutely. Thank you for catching that 😬

pages/api/nft/list/[userId].ts
Comment (Devorein) [reply pages/api/nft/list/[userId].ts]: Removed thanks 🙏🏼

lib/charmClient/interface.ts
Devorein: Yes, I agree. I've relocated it inside `lib/nft/interfaces`

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
Devorein: The check for length is to handle the case when there is no nfts since its a ternary operator

pages/api/profile/items.ts
Devorein: Good point. I've updated another place where we are using `.json({})` to just `.end()`

pages/api/profile/items.ts
Devorein: Updated

pages/api/profile/items.ts
Devorein: Yeah I can create another PR off of this branch to handle that use case, the whole code related to hiding/showing nft/poaps can get a lot more readable if we just use the row-based action pattern and get rid of grouped nfts/poaps (by using a isHidden field)

components/profile/components/ManageProfileItemModal.tsx
Devorein: This component has been completely removed. We decided on showing a visibility toggle icon beside each dao/nft/poap row to make the code a lot easier to maintain, and frankly for a better UX

components/profile/components/ManageProfileItemModal.tsx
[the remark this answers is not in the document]
mattcasey: We don't have a hard requirement to have a file per component, but it is a good idea to split out children if they are very complicated

pages/api/nft/list/[userId].ts
Comment (mattcasey) [reply pages/api/nft/list/[userId].ts]: > My key point here is not to make code perfect, but rather make it simple. For me also means actually better speed and maintainability so this is somehow aligned with overall team goals
I agree with that too, and I prefer to let the reviewee make the ultimate judgement call when possible, as it's also an opportunity to learn.

I also like that you mention 'lessons to learn for the future.' Even if the reviewee doesn't make the change now, they can at least keep the feedback in mind for next time or if they revisit this code.

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
```diff
@@ -59,7 +60,7 @@ export default function NftAvatarGallery ({ onSelect, isVisible, onClose, isSavi
...
+            {allNfts.length ? allNfts.map(nft => (
```
mattcasey: please change this to:
```
  {allNfts.map(nft => (
```

The check for length is unecessary, i'm not sure we can add a lint for something like that

lib/profileItem/interfaces.ts
mattcasey: > Moving it inside the API endpoint file would cause us to unintentionally import client code.
You mean for the client code to import backend code? Because it is a type, you can just use `import type`. Having the interfaces close to the function is more important than build errors.

mattcasey: I added a "Types" section to our coding style guide: https://app.charmverse.io/charmverse/page-4558717445075624.

components/profile/components/NftAvatarGallery/NftAvatarGallery.tsx
mattcasey: sorry, my mistake!

Devorein: Since we are using a ternary operator and handling empty state, we need to check the nft length first

components/profile/components/ManageProfileItemModal.tsx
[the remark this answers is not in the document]
msadura: Yeah I agree that sometimes you can have more than one "internal" component if needed. In this case we are talking about really large file (300+ lines) so it would be much easier to read and follow.

msadura (approved):
Thx for all the updates! I am approving to not block this, but lets give final voice to @mattcasey :)

mattcasey (approved):
Approved with two UI requests:
- make the eye icon on the far right, vertically centered (same area as the community rows). we always put actions at the end of the row, and it makes it easier to disable multiple rows quickly
- make the fade-out even more dramatic, like 0.25 opacity
- suggestion: pass in a simple function like "toggleVisibility" callback to the ProfileItems component, so that the logic to mutate state is all handled in PublicPage. See how simple CommunityRow is, we just call updateDaoProfileItem(community) (that function actually toggles visibility, maybe we could re-use it)

PR #853, opened 2022-08-29
Cleanup: co-locate apis for charmClient, create bounty api, and remove dead code

mattcasey:
This PR creates a new namespace on charmClient: `charmClient.bounties.*`.

(Going forward, we should use namespaced methods on CharmClient, because the current single file has grown too large).


__integration-tests__/server/pages/api/bounties/pool.spec.ts
```diff
@@ -1,227 +0,0 @@
-/* eslint-disable @typescript-eslint/no-unused-vars */
```
mattcasey, line 1: I removed this code because we're no longer using it

msadura (approved):
awesome cleanup!

PR #856, opened 2022-08-29
Fetch charmverse spaces for nexus profile

Devorein:
image
The vote without the title is for a proposal page since there is no title for those sort of votes

### DIFF
--- a/lib/deepdao/getAggregatedData.ts
+++ b/lib/deepdao/getAggregatedData.ts
@@ -5,7 +5,7 @@ import { getSpacesOfUser } from 'lib/spaces/getSpacesOfUser';
 import { DataNotFoundError } from 'lib/utilities/errors';
 import { isUUID } from 'lib/utilities/strings';
 import { isTruthy } from 'lib/utilities/types';
-import { getParticipationScore, getProfile } from './client';
+import { getProfile } from './client';
 import { DeepDaoAggregateData, DeepDaoOrganization, DeepDaoProfile, DeepDaoVote } from './interfaces';
 
 export async function getAggregatedData (userPath: string, apiToken?: string): Promise<DeepDaoAggregateData> {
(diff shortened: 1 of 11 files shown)


lib/deepdao/getAggregatedData.ts
```diff
@@ -37,18 +37,68 @@ export async function getAggregatedData (userPath: string, apiToken?: string): P
...
+      voteOptions: true,
+      threshold: true,
+      type: true,
+      status: true
+    }
+  });
+
   return {
-    daos: workspacesCount + participationScores.reduce((acc, cur) => acc + cur.data.daos, 0),
+    daos: userWorkspaces.length + participationScores.reduce((acc, cur) => acc + cur.data.daos, 0),
```
mattcasey, line 79: why are we not just using the length of organizations? is there a difference between dao and organizations from the data? i'm not sure we make that very clear in our UI

lib/deepdao/getAggregatedData.ts
```diff
@@ -37,18 +37,68 @@ export async function getAggregatedData (userPath: string, apiToken?: string): P
...
+        ...userWorkspaces.map(userWorkspace => ({
+          organizationId: userWorkspace.id,
+          name: userWorkspace.name
+        } as DeepDaoOrganization))
+      ],
     proposals: profiles.reduce<DeepDaoProfile['proposals']>((proposals, profile) => ([...proposals, ...profile.data.proposals]), []),
-    votes: profiles.reduce<DeepDaoProfile['votes']>((votes, profile) => ([...votes, ...profile.data.votes]), []),
+    votes: [
+      // Deepdao votes
+      ...profiles.reduce<DeepDaoProfile['votes']>((votes, profile) => ([...votes, ...profile.data.votes]), []),
```
mattcasey, line 92: I think this could be written in a simpler way, and not using spread operator as Marek mentioned:
```
Object.values(profiles).map(profile => profile.data.votes)
```

mattcasey (approved):

lib/deepdao/getAggregatedData.ts
Devorein: I haven't seen any discrepancies between the organization.length and daos so far yet. Not sure why deepdao decided to differentiate between them.

PR #859, opened 2022-08-30
some cleanup for hide nfts PR

mattcasey:
- renamed 'organizations' to 'communities'
- split out profile logic from deepdao client/api types into lib/profile folder

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -124,7 +124,7 @@ enum SpaceOperation {
 }
 
 enum ProfileItemType {
-  dao
+  community
   nft
   poap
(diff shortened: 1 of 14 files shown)


prisma/schema.prisma
```diff
@@ -124,7 +124,7 @@ enum SpaceOperation {
 }
 
 enum ProfileItemType {
-  dao
+  community
```
mattcasey, line 127: highlighting that i'm changing the type enum, it might require you to delete your data or in staging if we've already run the migration

Devorein (approved):

PR #861, opened 2022-08-30
Create and configure proposal reviewers and authors

Devorein:


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -23,6 +23,7 @@ import CreateVoteBox from './components/CreateVoteBox';
 import PageBanner from './components/PageBanner';
 import PageDeleteBanner from './components/PageDeleteBanner';
 import PageHeader from './components/PageHeader';
+import { ProposalProperties } from './components/ProposalProperties/ProposalProperties';
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:26]: needs an index.tsx file so you don't have to import the nested component

prisma/schema.prisma
```diff
@@ -311,6 +322,8 @@ model User {
   userVotes         UserVote[]
   initiatedVotes    Vote[]
   profileItems      ProfileItem[]
+  proposalAuthors   ProposalAuthor[]
```
mattcasey, line 325: authoredProposals?

prisma/schema.prisma
```diff
@@ -311,6 +322,8 @@ model User {
...
+  proposalReviewers ProposalReviewer[]
```
mattcasey, line 326: authoredReviewers? (note: on User model, all the 'reviewers' are THIS user so i think the focus should be on the proposals that are being linked)

prisma/schema.prisma
```diff
@@ -628,6 +641,33 @@ model Comment {
   user      User      @relation(fields: [userId], references: [id])
 }
 
+model Proposal {
+  id        String             @id @default(uuid()) @db.Uuid
+  createdBy String             @db.Uuid
+  spaceId   String             @db.Uuid
```
mattcasey, line 647: i think this should cascade delete if the space is deleted

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -0,0 +1,77 @@
...
+            // excludedIds={[...selectedReviewerUsers, ...selectedReviewerRoles]}
+              sx={{
+                width: '100%'
+              }}
+            />
+          </div>
+        </div>
+      </Stack>
+      <Stack flexDirection='row' justifyContent='space-between' gap={2} alignItems='center'>
+        <div
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:47]: always use Box for inline styles. all the ones you'er using are actually props:
`<Box display='flex' height='fit-content' flexGrow={1} />`

pages/api/pages/index.ts
```diff
@@ -55,7 +57,36 @@ async function createPage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
-  const page = await prisma.page.create({ data: typedPageCreationData });
+  let page: Page;
+
+  if (typedPageCreationData.type === 'proposal') {
+    const proposalId = v4();
+    typedPageCreationData.proposalId = proposalId;
+    // Using a transaction to ensure both the proposal and page gets created together
+    const [createdPage] = await prisma.$transaction([
+      prisma.page.create({ data: typedPageCreationData }),
+      createProposal({
```
mattcasey, line 68: at the moment, i think we should just be calling prisma.proposal.create. unless there's extra business logic required, etc. it's more confusing in this example because we use prisma.page directly above.

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:26]: Updated

prisma/schema.prisma
Devorein: That's a lot better term. Thanks.

prisma/schema.prisma
Devorein: Updated

prisma/schema.prisma
Devorein: Thats right. I also forgot to add a relation with the space model. Updated.

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:47]: I actually copied pretty much the whole component from our `BountyProperties.tsx` component. I will make the updates in both places.

pages/api/pages/index.ts
Devorein: Yes good point. I just transferred all the logic to create a page along with the proposal inside the lib function. It will be easier to test it out

lib/proposal/interface.ts
```diff
@@ -0,0 +1,11 @@
+import { Proposal, ProposalAuthor, ProposalReviewer, ProposalReviewerGroup } from '@prisma/client';
+
+export interface UpdateProposalRequest {
```
mattcasey, line 3: this is only used in charmClient, it could just go in that file

lib/proposal/interface.ts
```diff
@@ -0,0 +1,11 @@
...
+  authors: string[]
+  reviewers: {
+    group: ProposalReviewerGroup
+    id: string
+  }[]
+}
+
+export type ProposalWithUsers = Proposal & {authors: ProposalAuthor[], reviewers: ProposalReviewer[]}
```
mattcasey, line 11: you could still use interface here with `interface ProposalWithUsers extends Proposal { ... } `

prisma/schema.prisma
```diff
@@ -628,6 +649,40 @@ model Comment {
...
+  status    ProposalStatus
+  authors   ProposalAuthor[]
+  reviewers ProposalReviewer[]
```
mattcasey, line 658: @Devorein i thoguht during our call today we agreed to break this out into tables for reviewerUsers and reviewerRoles? can you explain here why you decided to keep them in the same table

prisma/schema.prisma
```diff
@@ -628,6 +649,40 @@ model Comment {
...
+
+model ProposalReviewer {
+  id         String                @id @default(uuid()) @db.Uuid
+  proposalId String                @db.Uuid
+  proposal   Proposal              @relation(fields: [proposalId], references: [id], onDelete: Cascade)
+  roleId     String?               @db.Uuid
+  userId     String?               @db.Uuid
+  role       Role?                 @relation(fields: [roleId], references: [id], onDelete: Cascade)
+  reviewer   User?                 @relation(fields: [userId], references: [id], onDelete: Cascade)
+  group      ProposalReviewerGroup
```
mattcasey, line 680: if this is a discriminator, please name it to 'type'

pages/api/pages/index.ts
```diff
@@ -55,7 +56,18 @@ async function createPage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
 
-  const page = await prisma.page.create({ data: typedPageCreationData });
+  let page: Page;
+
+  if (typedPageCreationData.type === 'proposal') {
+    page = await createProposal({
+      pageCreateInput: typedPageCreationData,
+      spaceId,
+      userId
+    }) as PageWithProposal;
```
mattcasey, line 66: Please remove this type assertion and handle the 'null' case instead. It actually looks like the types for createProposal may be wrong?

pages/api/pages/[id]/index.ts
```diff
@@ -109,6 +109,12 @@ async function updatePage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
         include: {
           sourcePermission: true
         }
+      },
+      proposal: {
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:113]: is this JOIN really necessary?

lib/pages/interfaces.ts
```diff
@@ -2,9 +2,11 @@ import { Block, Page, PagePermission, Space } from '@prisma/client';
 import { Board } from 'lib/focalboard/board';
 import { BoardView } from 'lib/focalboard/boardView';
 import { Card } from 'lib/focalboard/card';
+import { ProposalWithUsers } from 'lib/proposal/interface';
 
 export interface IPageWithPermissions extends Page {
   permissions: (PagePermission & {sourcePermission: PagePermission | null}) []
+  proposal?: ProposalWithUsers | null
```
mattcasey, line 9: we need to be passing around less info, not more, in our pages hook. please just query the proposal data you need when looking at a page, sa we do with bounties

mattcasey (changes requested):
I know this may be painful, but we shouldn't attach the proposal to the global collection of pages. Loading all the content into usePages is already very slow right now and we know we need to reduce it, this moves us in the wrong direction. The data should be loaded when you load a page, similar to how we load the bounties, or we could have a global 'useProposals' hook as well

lib/pages/server/getAccessiblePages.ts
```diff
@@ -113,6 +113,12 @@ export function generateAccessiblePagesQuery ({ spaceId, userId, archived }: Pag
         include: {
           sourcePermission: true
         }
+      },
+      proposal: {
+        include: {
+          authors: true,
```
mattcasey, line 119: remember this data actually can be joined on the frontend, or else be explicit about which user fields you need so we dont pull in private data

mattcasey (commented):
Changes:
1) the usePages hook should only be returning metadata and not content, and this moves us further away from that goal. None of the proposal content is needed in a global sense from what i can tell. Eventually we should just have DocumentPage or ProposalsPage load the extra content they need. You might consider a global useProposals() hook, as @motechFR recommended. But if it's global, it still means we're loading data we don't need, so I'm not sure that will make the job any easier.
2) use user data from the front-end, or be explicit about which fields you need when querying for reviewers and authors so that we don't expose user wallet addressess

lib/proposal/interface.ts
Devorein: Relocated the type from `lib/proposals/interface.ts` to be inside `charmClient/apis/proposalApis.ts`. Also updated the `updateProposal` lib method to make use of the interface

lib/proposal/interface.ts
Devorein: Right. I've gotten used to writing `type` everywhere. Updated.

pages/api/pages/index.ts
Devorein: Thank you. I just saw it (the types were wrong indeed). Updated.

lib/pages/server/getAccessiblePages.ts
Devorein: This itself doesn't pull in private data, just the `userId` and the `proposalId`, no other information as we are not joining with the `user` table

lib/pages/interfaces.ts
Devorein: Yes, you are right. I've updated to not retrieve the proposal from the page. There is a separate endpoint to fetch an individual proposal and used that

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts]: Correct. I've updated the code to fetch only a single proposal rather than join the proposal with the page.

prisma/schema.prisma
Devorein: We have decided to go with the current approach and use two optional columns to store the id of the reviewer

prisma/schema.prisma
Devorein: We have decided to remove the discriminator and do the validation that only either userId or roleId is set in a single record.

mattcasey, line 680: @Devorein please remove this field

mattcasey (approved):

prisma/schema.prisma
Devorein: Yes updated. It was unfortunately tangled in a lot of places so had to do a lot of QA to make sure nothing was broken

PR #862, opened 2022-08-30
clean up nft organization and fix unique ids

mattcasey:
- create a unique id for poap and NFT, since tokenId can sometimes be the same (it could start at 1, for instance and just be a counter)
- i merged nft and poap lib folders under 'blockchain'. The lib folder is pretty big and it's easier to maintain with less folders
- show counts of votes and bounties on an organization and hide empty tabs


mattcasey:
pre-emptively merging to fix the bug

PR #868, opened 2022-09-01
fix creating nested databases

mattcasey:
Woops... we were using different ids for the page than the board/cards. Need to figure out how to fix for boards that have been created now..


lib/pages/addPage.ts
```diff
@@ -22,7 +22,7 @@ interface AddPageResponse {
 
 export async function addPage ({ createdBy, spaceId, shouldCreateDefaultBoardData, ...page }: NewPageInput): Promise<AddPageResponse> {
 
-  const pageId = v4();
+  const pageId = page?.id || v4();
```
mattcasey, line 25: fix here... we were passing in id when calling addPage() inside of useNestedPage()

Devorein (approved):

PR #869, opened 2022-09-01
Show CharmVerse bounties on personal nexus

mattcasey:


lib/applications/__tests__/getCompletedApplicationsOfUser.spec.ts
```diff
@@ -1,42 +0,0 @@
-import { User, Space } from '@prisma/client';
```
mattcasey, line 1: sorry i'm deleting all these tests.. but at the end of the day, i dont think they were testing much beyond that the query was written correctly. i ended up using a broader query and filtering in getAggregatedData

charmClient/charmClient.ts
```diff
@@ -593,8 +593,8 @@ class CharmClient {
     return http.POST('/api/tasks/mark', tasks);
   }
 
-  getAggregatedData (userPath: string) {
-    return http.GET<AggregatedProfileData>(`/api/public/profile/${userPath}/aggregate`);
+  getAggregatedData (userId: string) {
+    return http.GET<AggregatedProfileData>(`/api/public/profile/${userId}/aggregate`);
```
mattcasey, line 597: i checked and we never pass in a user path, so making it more strict to only accept a userId

Devorein: Make sense.

Devorein (approved):

PR #871, opened 2022-09-01
Prevent server errors due to invalid input

mattcasey:
- Fixes two errors on the server due to invalid input
- Display an error when someone is trying to use a domain that already exists:
image

### DIFF
--- a/pages/[domain]/settings/workspace.tsx
+++ b/pages/[domain]/settings/workspace.tsx
@@ -1,6 +1,6 @@
 import { yupResolver } from '@hookform/resolvers/yup';
 import LaunchIcon from '@mui/icons-material/LaunchOutlined';
-import { Box, Typography } from '@mui/material';
+import { Box, FormHelperText, Typography } from '@mui/material';
 import Grid from '@mui/material/Grid';
 import TextField from '@mui/material/TextField';
 import charmClient from 'charmClient';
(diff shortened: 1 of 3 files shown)


pages/[domain]/settings/workspace.tsx
```diff
@@ -51,18 +52,24 @@ export default function WorkspaceSettings () {
...
     const newDomain = space.domain !== values.domain;
-    const updatedSpace = await charmClient.updateSpace({ ...space, ...values });
-    if (newDomain) {
-      window.location.href = router.asPath.replace(space.domain, values.domain);
-    }
-    else {
-      setSpace(updatedSpace);
-    }
-    reset(updatedSpace);
+    charmClient.updateSpace({ ...space, ...values })
```
Comment (msadura) [inline pages/[domain]/settings/workspace.tsx:60]: question out of curiosity - is using ".then" chaining over async / await here just a preference?

Comment (motechFR) [reply pages/[domain]/settings/workspace.tsx:60]: @msadura I think (although don't know) the aim was to have some error handling here which we didn't have before, and using promise.then.catch was syntactically simpler than await + try / catch block (especially since this is the end of the function)

pages/api/session/login.ts
```diff
@@ -1,15 +1,17 @@
...
 import { withSessionRoute } from 'lib/session/withSession';
 import { LoggedInUser } from 'models';
 import { updateGuildRolesForUser } from 'lib/guild-xyz/server/updateGuildRolesForUser';
 import { sessionUserRelations } from 'lib/session/config';
 
 const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
 
-handler.post(login);
+handler
+  .use(requireKeys(['address'], 'body'))
```
motechFR, line 13: Is this endpoint used for Discord login?

motechFR (approved):
Tested & approved, nifty cleanup! Small question on the login endpoint, otherwise good to go

pages/[domain]/settings/workspace.tsx
Comment (mattcasey) [reply pages/[domain]/settings/workspace.tsx:60]: We usually default to await, but I actually thought this was a good use case to use .then().catch() because i didn't want to try/catch the entire block where we set the window.location, etc. and i didn't want to have to instantiate null variables first.

pages/api/session/login.ts
mattcasey: Good question. That's acutally handled in api/discord/callback. And you can see this code would fail anyway, the error is because we were trying to query address by "undefined"

PR #873, opened 2022-09-02
create global page dialog

mattcasey:
This PR introduces a new hook `usePageDialog()` to reduce boilerplate when showing the PageDialog (popup view of pages) on a page. This could be useful on the nexus or tasks page, for instance:
```
import { usePageDialog } from 'components/PageDialog/hooks/usePageDialog';

function MyComp () {
  const { showPage } = usePageDialog();
  
  onClickPage (pageId: string) {
     showPage({ pageId });
  }
}
```

or you can even just pass a pageId into the new PageLink component. It will have an Href for people that want to use right-click to open the link in a new tab, but will show the modal if you click on it:
```
  <PageLink pageId={page.id} href={...} />
```

This PR doesn't add the popup to any new locations, and one place I didn't change yet was 'CenterPanel' in focalboard code however it seems like we could replace that one too.

(This cleanup came out of some work on personal nexus, but realized I need to also handle pages the user can't access, and didn't want to go down that rabbit hole yet)


components/common/BoardEditor/focalboard/src/components/cardDialog.tsx
```diff
@@ -73,52 +66,17 @@ const CardDialog = (props: Props): JSX.Element | null => {
     }
   }, []);
 
-  const handleDeleteCard = async () => {
-    if (!card) {
-      Utils.assertFailure()
-      return
-    }
-    // TelemetryClient.trackEvent(TelemetryCategory, TelemetryActions.DeleteCard, {board: props.board.id, view: props.activeView.id, card: card.id})
-    await mutator.deleteBlock(card, 'delete card')
```
mattcasey, line 82: i checked and it appears that mutator.deleteBlock() ends up calling `modifyChildPages()` on the back-end just like charmClient.deletePage(), so it should be fine not to have this custom method

mattcasey: for later.. if we moved the 'Convert to bounty' button to PageDialog component, I think we could get rid of cardDialog completely

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
```diff
@@ -142,7 +142,7 @@ export default function BountyProperties (props: {
 
   useEffect(() => {
     refreshSubmissions();
-  }, [currentBounty]);
+  }, [currentBounty?.id]);
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:145]: fixed a bug here, we were loading applicants many times which I think delayed it from showing up

.eslintrc.json
```diff
@@ -86,6 +86,7 @@
         "allowForLoopAfterthoughts": true
       }
     ],
+    "no-restricted-exports": "off",
```
mattcasey, line 89: allows us to do `export { default } from 'Component'`

components/common/Link.tsx
```diff
@@ -24,27 +25,51 @@ const StyledMuiLink = styled(MuiLink)`
...
   );
 }
+
+interface PageLinkProps extends Props {
+  bountyId?: string;
+  pageId?: string;
+}
+
+// use this link component to display a page inside a modal
+export function PageLink ({ bountyId, pageId, ...props }: PageLinkProps) {
```
mattcasey, line 61: i'm not using this yet, wasn't sure if i should just take it out for now

components/common/PageDialog/PageDialogGlobal.tsx
```diff
@@ -0,0 +1,49 @@
+import { useEffect, useState } from 'react';
+import charmClient from 'charmClient';
+import type { IPageWithPermissions } from 'lib/pages/interfaces';
+import { useBounties } from 'hooks/useBounties';
+import log from 'lib/log';
+import PageDialog from './PageDialog';
+import { usePageDialog } from './hooks/usePageDialog';
+
+// a wrapper of page dialog that uses usePageDialogHook
+export default function PageDialogGlobal () {
```
mattcasey, line 10: I suppose this 'global' version of PageDialog could just be combined with the original component if we replace PageDialog in centerPanel...

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
Comment (msadura) [reply components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:145]: I haven't seen the rest of the logic, but shouldn't it be:

```
useEffect(() => {
    if (currentBounty?.id) {
        refreshSubmissions();
    }
  }, [currentBounty?.id]);
 ```
 
 unless you want to call `refreshSubmissions` even if there is no id

msadura (approved):
I think the ida of this global component is quite handy. My only concern is that `interface PageDialogContext ` will grow in time and might get harder to use (we might need more props for more specific cases). I wouldn't worry about it too much for now anyway.

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
Comment (mattcasey) [reply components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:145]: there's a check in refreshSubmissions but i'm going to add the check you suggested too

PR #874, opened 2022-09-02
Publish a draft proposal

Devorein:


components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -154,6 +165,62 @@ export default function ProposalProperties ({ proposalId, readOnly }: ProposalPr
...
+                </Typography>
+              </>
+            )}
+          </Box>
+        </MenuItem>
+
+        {(proposal.status === 'private_draft' || proposal.status === 'draft') && (
+          <MenuItem
+            disabled={!isProposalAuthor}
+            onClick={async () => {
```
Comment (msadura) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:197]: this logic is pretty much duplicated, so out could create a finction with common logic:

```
async function updateStatus (newStatus)  {
              await charmClient.proposals.updateStatus(newStatus);
              refreshProposal();
              proposalMenuState.close();
            }
            ```

and you could pass updates status value to it

File: components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx [hunk 2]
```diff
@@ -154,6 +165,62 @@ export default function ProposalProperties ({ proposalId, readOnly }: ProposalPr
         my: 2
       }}
       />
+      <Menu {...bindMenu(proposalMenuState)}>
+        <MenuItem
+          disabled={!isProposalAuthor}
+          onClick={async () => {
+            await charmClient.proposals.updateStatus(proposal.id, proposal.status === 'private_draft' ? 'draft' : proposal.status === 'draft' ? 'private_draft' : 'draft');
```
Comment (msadura) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:172]: this nested ternary is really hard to understand, but if I get it correctly it will do the same as `status === 'draft' ? 'private_draft' : 'draft'`

motechFR:
**Product improvements**
- Allow moving discussion stage proposal back to private draft and draft

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -32,6 +37,7 @@ export default function ProposalProperties ({ proposalId, readOnly }: ProposalPr
   const canUpdateAuthors = status === 'draft' || status === 'private_draft' || status === 'discussion';
   const canUpdateReviewers = status === 'draft' || status === 'private_draft';
   const reviewerOptionsRecord: Record<string, ({group: 'role'} & ListSpaceRolesResponse) | ({group: 'user'} & Contributor)> = {};
+  const isProposalAuthor = (user && proposal.authors.map(author => author.userId).includes(user.id));
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:40]: Simpler way

```typescript
const isProposalAuthor = (user && proposal.authors.some(author => author.id === user.id))
```

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:172]: Agree with @msadura it would be better to extract the logic of which status update should take place upwards in the scope or as a separate helper function

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -154,6 +165,62 @@ export default function ProposalProperties ({ proposalId, readOnly }: ProposalPr
...
+              ) : (
+                <>
+                  <ArrowBackIcon fontSize='small' />
+                  <Typography>
+                    Move to private draft
+                  </Typography>
+                </>
+              )}
+            </Box>
+          </MenuItem>
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:221]: Simpler approach, have a component that takes  a proposal and emits the correct options

lib/proposal/updateProposalStatus.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const proposalStatusTransitionRecord: Record<ProposalStatus, ProposalStatus[]> = {
+  private_draft: ['draft', 'discussion'],
+  draft: ['private_draft', 'discussion'],
+  discussion: ['draft', 'review'],
+  review: ['reviewed'],
+  reviewed: ['vote_active'],
+  vote_active: ['vote_closed'],
+  vote_closed: []
+};
```
motechFR, line 13: You could import this to the front end for deciding which menu options to provide users with.

That way your UI and back-end available next stages will always be consistent

pages/api/proposals/[id]/status.ts
```diff
@@ -0,0 +1,49 @@
...
+      authors: true,
+      reviewers: true
+    }
+  });
+
+  if (!proposal) {
+    throw new NotFoundError();
+  }
+
+  const isCurrentUserProposalAuthor = proposal.authors.some(author => author.userId === userId);
```
Comment (motechFR) [inline pages/api/proposals/[id]/status.ts:34]: This is a very good permissions MVP for a specific scenario, no extra complexity.

motechFR (changes requested):
Main improvements are unifying the logic for next / previous available stages

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:197]: Right. I've refactored the whole code to make use of a state transition record. The logic is a lot simpler now and relies on a single source of truth.

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:172]: Thank you. I've updated the code to rely on a single source of truth using a state transition record. The logic is a lot simpler and easier to understand.

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:40]: Yes one loop is always better than two 😛

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:221]: Just updated the code to rely on a status transition record. The code is much simpler now.

lib/proposal/updateProposalStatus.ts
Devorein: Yes. Moved it to a separate file to be importable from both client and server.

pages/api/proposals/[id]/status.ts
Comment (Devorein) [reply pages/api/proposals/[id]/status.ts:34]: Awesome. Made sure to add this permission logic to the fe as well. The inputs are disabled if you are not an author. Basic primitive simple permission for now.

motechFR (approved):

PR #877, opened 2022-09-02
Automagic proposal page permissions

motechFR:
This PR auto-updates proposal permissions based on the current proposal stage, and who is a reviewer author.

It also adds support so that proposal editors can still toggle the page as public, but not modify underlying permissions


components/common/PageLayout/components/Header/components/ShareButton/components/ShareToWeb.tsx
```diff
@@ -135,6 +140,22 @@ export default function ShareToWeb ({ pageId, pagePermissions, refreshPermission
         )
       }
 
+      {
```
msadura, line 143: as far as I can see only message changes in those laerts so I woud do something like this:

```
 <Alert severity='info'>
            {getShareAlertMessage()}
          </Alert>
```

and ofc `getShareAlertMessage` would return proper string based on conditions

lib/pages/utils.ts
```diff
@@ -1,4 +1,12 @@
+import { PageType } from '@prisma/client';
 
 export function getPagePath () {
   return `page-${Math.random().toString().replace('0.', '')}`;
 }
+
+export function canReceiveManualPermissionUpdates ({ pageType }: {pageType: PageType}): boolean {
+  if (pageType === 'card_template' || pageType === 'proposal') {
```
msadura, line 8: `return pageType === 'card_template' || pageType === 'proposal' ? false : true`

or even simpler version:

`return pageType !== 'card_template' && pageType !== 'proposal' `

components/common/PageLayout/components/Header/components/ShareButton/components/ShareToWeb.tsx
motechFR: Good spot, thanks!

lib/pages/utils.ts
motechFR: Makes sense from optimisation viewpoint, in this case I prefer to keep the verbose version as is, simply for readability, and future extensibility.

motechFR:
Currently battling an issue with playwright error

Below article shares the exact cause I'm facing

https://decembersoft.com/posts/error-selector-creators-expect-all-input-selectors-to-be-functions/

__integration-tests__/server/pages/api/permissions/index.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+    });
+
+    await request(baseUrl)
+      .delete('/api/permissions')
+      .set('Cookie', userCookie)
+      .send({
+        permissionId: permissionToDelete.id
+      })
+      .expect(401);
+  });
```
mattcasey, line 411: very thorough test cases...  they all seem relevant :)

lib/proposal/interface.ts
```diff
@@ -1,6 +1,17 @@
...
+  group: Extract<AssignablePermissionGroups, 'role' | 'user'>
+  id: string
+}
 
 export interface ProposalWithUsers extends Proposal {
   authors: ProposalAuthor[],
   reviewers: ProposalReviewer[]
 }
+
+export interface UpdateProposalRequest {
```
mattcasey, line 14: why was this moved here? it seems very specific to the method 'updateProposal'. As we discussed in Discord, our default is to keep interfaces close to the code, and not in a 'catch all' file

lib/proposal/interface.ts
```diff
@@ -1,6 +1,17 @@
...
+  authors: string[]
```
mattcasey, line 15: needs semi-colon at the end of these lines

lib/proposal/interface.ts
mattcasey: FWIW, placement of a single TS interface is not a big deal but I'm making this comment so that we're aligned in general.

lib/proposal/syncProposalPermissions.ts
```diff
@@ -0,0 +1,308 @@
...
+  const [page, proposal] = await Promise.all([
+    prisma.page.findUnique({
+      where: {
+        proposalId
+      },
+      include: {
+        permissions: true
+      }
+    }),
+    prisma.proposal.findUnique({
```
mattcasey, line 68: not a big deal i guess, but couldn't you gather the page and proposal in one go using prisma?
```
const { proposal, ...page } = await prisma.page.findUnique({
  where: {
     proposalId,
  },
  include: {
    proposal: {
      include: {
         authors: true,
         reviewers: true
      }
  }
})
```

lib/proposal/syncProposalPermissions.ts
```diff
@@ -0,0 +1,308 @@
...
+      permissions: {
+        include: {
+          sourcePermission: true
+        }
+      }
+    }
+  });
+
+  if (children.length > 0) {
+    children = (await resolvePageTree({
```
mattcasey, line 100: What do you think of this alternative way to extract the flatChildren property:
```
({ flatChildren: children } = await resolvePageTree({
   ...
}))
```
I think it's a little better because the assignment of flatChildren stays on the left side of the `=` sign, but I don't feel strongly about either approach

lib/proposal/syncProposalPermissions.ts
```diff
@@ -0,0 +1,308 @@
...
+
+  if (reviewerPermissionSetting !== null) {
+    proposal.reviewers.forEach(reviewer => {
+
+      const assignedAuthor = reviewer.userId ? createProposalPermissionArgs.find(permissionArgs => {
+        return permissionArgs.data.user?.connect?.id === reviewer.userId;
+      }) : undefined;
+
+      // If author is also a reviewer, only assign a permission if this is higher
+      if (assignedAuthor && comparePermissionLevels({
```
mattcasey, line 153: suggestion: compute comparePermissionLevels() on a line before this conditional, it's hard to tell you are comparing it's output to 'more' in the current format

lib/proposal/syncProposalPermissions.ts
```diff
@@ -0,0 +1,308 @@
...
+        base: assignedAuthor.data.permissionLevel as PagePermissionLevel, comparison: reviewerPermissionSetting }) === 'more'
+      ) {
+        assignedAuthor.data.permissionLevel = reviewerPermissionSetting;
+      }
+      else if (!assignedAuthor) {
+        const newId = v4();
```
mattcasey, line 159: why do you create uuid's? Postgres will do that for you automatically

pages/api/pages/index.ts
```diff
@@ -72,7 +73,15 @@ async function createPage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
-    const pageWithPermissions = await setupPermissionsAfterPageCreated(page.id);
+    const proposalIdForPermissions = page.type === 'proposal' ? page.id : (page.parentId ? (await resolvePageTree({
+      pageId: page.id
+      // includeDeletedPages: true
+    })).parents.find(p => p.type === 'proposal')?.id : undefined);
+
+    await (proposalIdForPermissions ? syncProposalPermissions({ proposalId: proposalIdForPermissions as string })
+      : setupPermissionsAfterPageCreated(page.id));
+
+    const pageWithPermissions = await getPage(page.id) as IPageWithPermissions;
```
mattcasey, line 84: please avoid using `as <interface>` just to strip out the possibility of null. It's better to handle the null case instead

mattcasey (approved):
You are a permissions fiend.

I left some comments/suggestions about best practices but overall looks good

Devorein:
Tested it out in staging and all the cases worked out fine. Approved.

Devorein (approved):

lib/proposal/interface.ts
motechFR: In general, I think it's best if we define interfaces on the server-side to have a single source of truth.

We might have multiple clients in the future (whether UI, apps or bots) which will be talking to the server, so I think that's the most relevant place to have definitions.

To go with this convention of inlining definitions, I've moved this interface to updateProposal method

lib/proposal/syncProposalPermissions.ts
motechFR: Thanks for the spot!

lib/proposal/syncProposalPermissions.ts
motechFR: My Typescript linter is throwing some red on this approach, let's look at this on a next chat

pages/api/pages/index.ts
motechFR: Gotcha, updated

motechFR:
Thanks for the feedback everyone!

PR #879, opened 2022-09-05
Show status for proposals in table

Devorein:


components/proposals/components/ProposalsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+import GridHeader from 'components/common/Grid/GridHeader';
+import LoadingComponent from 'components/common/LoadingComponent';
+import PageDialog from 'components/common/Page/PageDialog';
+import useTasks from 'components/nexus/hooks/useTasks';
+import { usePages } from 'hooks/usePages';
+import { IPageWithPermissions } from 'lib/pages';
+import { ProposalWithUsers } from 'lib/proposal/interface';
+import { humanFriendlyDate, toMonthDate } from 'lib/utilities/dates';
+import { useCallback, useState } from 'react';
+import { ProposalStatusChip } from '../../../components/[pageId]/DocumentPage/components/ProposalProperties/ProposalStatusBadge';
```
mattcasey, line 15: It would be better to put shared components inside the components/proposals folder (so move ProposalStatusBadge to here)

components/proposals/components/ProposalsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+                deleteProposal={deleteProposal}
+                editProposal={(proposalId) => openPage(proposalId)}
+                proposal={proposal}
+              />
+            </Grid>
+          </GridContainer>
+        ) : null;
+      })}
+      {activePage && (
+        <PageDialog
```
mattcasey, line 108: replace with `usePageDialog`

lib/proposal/getProposalsBySpace.ts
```diff
@@ -0,0 +1,37 @@
...
+                spaceId,
+                userId,
+                isAdmin: true
+              }
+            }
+          }
+        }]
+      }
+    },
+    include: {
```
mattcasey, line 32: this should be joined with contributors on the frontend

pages/[domain]/proposals/index.tsx
```diff
@@ -0,0 +1,15 @@
+import ProposalsPage from 'components/proposals/ProposalsPage';
+import getPageLayout from 'components/common/PageLayout/getLayout';
+import { setTitle } from 'hooks/usePageTitle';
+
+export default function VotesPage () {
```
Comment (mattcasey) [inline pages/[domain]/proposals/index.tsx:5]: fix "VotesPage"

components/proposals/ProposalsPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+import NewProposalButton from 'components/votes/components/NewProposalButton';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import useSWR from 'swr';
+import ProposalsTable from './components/ProposalsTable';
+
+export default function ProposalsPage () {
+  const [currentSpace] = useCurrentSpace();
+
+  const { data, mutate: mutateProposals } = useSWR(() => `proposals/${currentSpace?.id}`, () => currentSpace ? charmClient.proposals.getProposalsBySpace(currentSpace.id) : [], {
+    fallbackData: undefined
```
mattcasey, line 14: i don't think this does anything since you're passing in undefined?

components/proposals/ProposalsPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { CenteredPageContent } from 'components/common/PageLayout/components/PageContent';
+import NewProposalButton from 'components/votes/components/NewProposalButton';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import useSWR from 'swr';
+import ProposalsTable from './components/ProposalsTable';
+
+export default function ProposalsPage () {
+  const [currentSpace] = useCurrentSpace();
+
+  const { data, mutate: mutateProposals } = useSWR(() => `proposals/${currentSpace?.id}`, () => currentSpace ? charmClient.proposals.getProposalsBySpace(currentSpace.id) : [], {
```
mattcasey, line 13: it's better to return `null` in the first input to useSWR, until currentSpace exists, so that the 2nd function doesn't even get called

mattcasey (changes requested):
Before this PR is merged, please remove the aggregated user data from ProposalWithUsers - u may want to get rid of this interface completely. we should just be using the hook 'useContributors' to retrieve user data

components/proposals/components/ProposalsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+  }, [pages]);
+
+  function closePage () {
+    setActivePage(null);
+  }
+
+  async function deleteProposal (proposalId: string) {
+    const page = pages[proposalId];
+    if (page) {
+      await charmClient.archivePage(page.id);
```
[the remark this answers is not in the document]
Devorein: Updated.

components/proposals/components/ProposalsTable.tsx
Devorein: Wow. Its so much easier now 🤯 Updated 🥳

pages/[domain]/proposals/index.tsx
Comment (Devorein) [reply pages/[domain]/proposals/index.tsx:5]: Oops. Thank you for that. Updated.

components/proposals/ProposalsPage.tsx
```diff
@@ -0,0 +1,50 @@
...
+    fallbackData: undefined
+  });
+
+  // For now, consider that empty pages list means we are loading pages
```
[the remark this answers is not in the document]
Devorein: The comment was actually there before this PR 👀

components/proposals/ProposalsPage.tsx
Devorein: I copied most of the code from the `VotesPage` component, it was also used there. Removed it from both places as swr by default returns `undefined`

components/proposals/ProposalsPage.tsx
Devorein: Updated both in `VotesPage` and `ProposalsPage`

Devorein:
Right now we are just fetching the `userId` (which is a must) and the `proposalId` for authors and additional `roleId` for reviewers. No data about the author and reviewer is being fetched other than that and we join the data on the frontend using the id.

lib/proposal/getProposalsBySpace.ts
Devorein: It is joined on the frontend with the contributors based on the userId and roleId, we are not getting any other user data (apart from the id) when joining on the backend

Devorein:
I've added back the filter and sort options for the proposals table. Completely forgot I removed it when copying code from the Votes table. For the sort only Created at is added as an option (we could add more new options later so kept it for now) and as for the filter options its all the available status of the proposal along with 'all' to include proposals of any status

mattcasey (approved):

PR #884, opened 2022-09-07
make it easier to use discord on alternate local ip addresses

mattcasey:
for testing locally, this allows you to use your laptop's IP address instead of localhost to connect thru Discord

### DIFF
--- a/lib/discord/loginByDiscord.ts
+++ b/lib/discord/loginByDiscord.ts
@@ -9,7 +9,7 @@ import log from 'lib/log';
 
 export default async function loginByDiscord ({ code, hostName }: { code: string, hostName?: string }) {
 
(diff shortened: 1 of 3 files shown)


lib/discord/loginByDiscord.ts
```diff
@@ -9,7 +9,7 @@ import log from 'lib/log';
 
 export default async function loginByDiscord ({ code, hostName }: { code: string, hostName?: string }) {
 
-  const domain = hostName?.startsWith('localhost') ? `http://${hostName}` : `https://${hostName}`;
+  const domain = process.env.NODE_ENV === 'development' ? `http://${hostName}` : `https://${hostName}`;
```
mattcasey, line 12: I believe that NODE_ENV is set to development by Next.js by default

Devorein (approved):

PR #885, opened 2022-09-07
bug fix: make it possible to scroll sidebar on mobile

mattcasey:
Simple fix to only enable drag-and-drop at certain angles on mobile


lib/browser.ts
```diff
@@ -1,5 +1,5 @@
 
-export function isMobile (): boolean {
+export function isTouchScreen (): boolean {
```
mattcasey, line 2: I renamed this to be more specific

Devorein (approved):
Tested it out on all draggable views and everything works great.

PR #888, opened 2022-09-07
Open vote for reviewed proposal

Devorein:


components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -17,17 +17,28 @@ import { proposalStatusTransitionRecord, proposalStatusTransitionPermission, PRO
...
+import { IPageWithPermissions } from 'lib/pages';
+import { useState } from 'react';
+import HowToVoteOutlinedIcon from '@mui/icons-material/HowToVoteOutlined';
+import CreateVoteModal from 'components/votes/components/CreateVoteModal';
 import { ProposalStatusChip } from '../../../../proposals/components/ProposalStatusBadge';
 
 interface ProposalPropertiesProps {
   proposalId: string,
   readOnly?: boolean
+  page: IPageWithPermissions
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:30]: since page is a requirement i'd probably just grab proposalId from it instead of passing in separately

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
```diff
@@ -15,7 +15,7 @@ import { usePopupState } from 'material-ui-popup-state/hooks';
 import { useEffect, useState } from 'react';
 import PublishingForm from './PublishingForm';
 
-export default function PublishToSnapshot ({ page }: {page: Page}) {
+export default function PublishToSnapshot ({ page, disabled = false, button = true }: {button?: boolean, disabled?: boolean, page: Page}) {
```
mattcasey, line 18: Please simplify this:
- the component already uses 'usePages()' hook. This means we can just pass a pageId instead, will also simplify the props for ProposalPropeties (just pass a pageId in instead of the whole page. This will help us when we want to stop passing entire page content around).
- instead of passing in a 'button' prop to determine how to wrap the element, just wrap the element at the call sites in Header.tsx and ProposalProperties with the correct component.

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
```diff
@@ -53,52 +53,62 @@ export default function PublishToSnapshot ({ page }: {page: Page}) {
...
-    </ListItemButton>
+        </Link>
+      )
+    }
+    </>
+  );
 
+  return (
+    button ? (
+      <ListItemButton disabled={disabled}>
```
mattcasey, line 105: actually the link needs to wrap the ListItem or else you can't click the sides of the component. perhaps you could pass in ListItemButton or MenuItem as a 'component' prop?

mattcasey (changes requested):
can't click the sides of the 'publish to snapshot' menu item anymore because the link is not completely wrapping  the element

components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/ProposalProperties/ProposalProperties.tsx:30]: Good point. Updated.

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
Devorein: Updated. Thanks.

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
Devorein: The only prop PublishToSnapshot now takes is the `pageId`. I've wrapped the component with the appropriate components where it's used. It seems like it's not working on production as well.
Updates: I've tried to pass in the component to use as wrapper, but now the modal can't be closed :( I've added a card to handle this, as its getting quite frustrating trying to fix this stupid bug

motechFR:
Tested with Safwan, this is good to go!

PR #890, opened 2022-09-07
Show active votes and polls in the Nexus

Devorein:


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -204,7 +200,7 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false, parentPro
                       pageUpdatedBy={page.updatedBy}
                     />
                   )}
-                  {proposalId && <ProposalProperties pageId={proposalId} proposalId={proposalId as string} readOnly={readonlyProposalProperties} />}
+                  {page.type === 'proposal' && page.proposalId && <ProposalProperties pageId={page.proposalId} proposalId={page.proposalId} readOnly={readOnly} />}
                   {(draftBounty || page.bountyId) && (
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:204]: Please revert this change.

We find a proposalId in the current page or parents, so that we can display the proposal properties across all child pages

motechFR:
Outstanding product questions:
- Only give you most relevant notification per stage?
- Change labels of actions?

components/[pageId]/DocumentPage/components/PageTemplateBanner.tsx
```diff
@@ -21,15 +20,12 @@ const StyledPageTemplateBanner = styled(Box)<{card?: boolean}>`
...
+export function PageTemplateBanner ({ page }: {page: IPageWithPermissions}) {
   const [space] = useCurrentSpace();
-  const { pages } = usePages();
-
   const theme = useTheme();
-  const page = pages[pageId] as IPageWithPermissions;
 
   const isShowingCard = page.type.match('card') !== null;
-  const board = isShowingCard ? pages[page.parentId as string] as IPageWithPermissions : undefined;
+  const board = isShowingCard ? page : undefined;
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/PageTemplateBanner.tsx:28]: The intention of this is to provide the board to the card page template, so we can navigate you to the board from the card.

Can you please fix this?

components/nexus/ProposalTasksList.tsx
```diff
@@ -0,0 +1,147 @@
...
+};
+
+/**
+ * Page only needs to be provided for proposal type proposals
+ */
+export function ProposalTasksListRow (
+  props: {proposalTask: ProposalTask}
+) {
+
+  const {
```
motechFR, line 25: W/C: I've seen this pattern creeping into our code. We usually destructure directly in the function definition

components/nexus/TasksPage.tsx
```diff
@@ -49,13 +51,16 @@ const TASK_TABS = [
...
+  { icon: <TaskOutlinedIcon />, label: 'Proposal', type: 'proposal' }
 ] as const;
 
+type TaskType = (typeof TASK_TABS)[number]['type'];
+
 export default function TasksPage () {
   const router = useRouter();
   const { user } = useUser();
-  const [currentTask, setCurrentTask] = useState(router.query?.task ?? 'multisig');
+  const [currentTask, setCurrentTask] = useState<TaskType>((router.query?.task ?? 'multisig') as TaskType);
```
motechFR, line 63: W: This variable is used for selecting the current group, not an atomic task.

It would be better for code readability to rename it to something along the "group" idea.

components/nexus/TasksPage.tsx
```diff
@@ -131,7 +137,7 @@ export default function TasksPage () {
           />
         ))}
       </Tabs>
-      {currentTask === 'multisig' ? <GnosisTasksList error={error} mutateTasks={mutateTasks} tasks={tasks} /> : currentTask === 'discussion' ? <MentionedTasksList mutateTasks={mutateTasks} error={error} tasks={tasks} /> : currentTask === 'vote' ? <VoteTasksList mutateTasks={mutateTasks} error={error} tasks={tasks} /> : null}
+      {currentTask === 'multisig' ? <GnosisTasksList error={error} mutateTasks={mutateTasks} tasks={tasks} /> : currentTask === 'discussion' ? <MentionedTasksList mutateTasks={mutateTasks} error={error} tasks={tasks} /> : currentTask === 'vote' ? <VoteTasksList mutateTasks={mutateTasks} error={error} tasks={tasks} /> : currentTask === 'proposal' ? <ProposalTasksList error={error} tasks={tasks} /> : null}
```
motechFR, line 140: S: It's a bit hard to follow what's happening with this nested ternary. Could you please break this up?

Maybe with a format like

{
 currentTask === 'multisig' && <Component1>
}

{
 currentTask === 'proposal' && <Component2>
}

jest.config.ts
```diff
@@ -32,7 +33,7 @@ export const jestConfig = {
 
   // The glob patterns Jest uses to detect test files
   testMatch: [
-    '**/lib/**/*.spec.ts'
+    path.join(__dirname, 'lib/proposal/__tests__/getProposalTasks.spec.ts')
```
motechFR, line 36: M: Please drop this change, as otherwise our CI would go from 300+ tests to maybe 5 ;)

lib/proposal/getProposalTasks.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+      ]
+    },
+    include: {
+      page: true,
+      space: true
+    }
+  });
+
+  return [
```
motechFR, line 120: W: I think this could be written in simpler fashion

Step 1:
- Merge your prisma queries into a single query with an or statement (If you'd like, also add a small comment on each or object explaining why)

Step 2:
- Map your results using extractProposal data and calculate extractProposalData second argument as part of the mapping

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:204]: Thank you for catching that. Reverted back the changes.

components/[pageId]/DocumentPage/components/PageTemplateBanner.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/PageTemplateBanner.tsx:28]: Yes fixed.

components/nexus/ProposalTasksList.tsx
Devorein: Agreed. Most of the time I skip it if there are multi-level restructuring, but I will update it here anyways

components/nexus/TasksPage.tsx
Devorein: Good one. Currently, the code is quite difficult to read. This will certainly improve the readability.

jest.config.ts
Devorein: Yeah clearly a violation of our guidelines.

lib/proposal/getProposalTasks.ts
Devorein: Yep updated the code to use a single query to fetch all the proposals and then convert them to proposal task by assigning an action.

lib/proposal/interface.ts
```diff
@@ -11,3 +11,12 @@ export interface ProposalWithUsers extends Proposal {
   reviewers: ProposalReviewer[]
 }
 
+export interface ProposalTask {
```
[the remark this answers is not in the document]
Devorein: Updated thank you.

motechFR:
We need to fix a final bug:

If I click an action in Nexus (ie. Review a "in review" proposal), go to the proposal, take that action, come back to nexus, the task is not replaced by the new task

PR #893, opened 2022-09-07
install playwright dependencies as part of cached dependencies step

mattcasey:
One of my builds timed out waiting for playwright to install. But we should  be able to cache it

### DIFF
--- a/.github/actions/install/action.yml
+++ b/.github/actions/install/action.yml
@@ -36,7 +36,7 @@ runs:
       shell: bash
       if: steps.cache_playwright.outputs.cache-hit != 'true'
       run: |
-        npx playwright install-deps chromium chrome
+        npx playwright install chromium chrome --with-deps
 
     - name: Cache dependencies
       id: cache_node_modules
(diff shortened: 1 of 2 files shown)


.github/actions/install/action.yml
```diff
@@ -33,4 +33,5 @@ runs:
       shell: bash
       if: steps.cache_node_modules.outputs.cache-hit != 'true'
       run: |
-        npm ci --no-audit --no-fund
\ No newline at end of file
+        npm ci --no-audit --no-fund
+        npx playwright install --with-deps
```
mattcasey, line 37: trying this command based on the official example: https://playwright.dev/docs/ci-intro

motechFR:
Looks like this worked!

I went on the CI action and re-ran it. The install deps took 30 seconds instead of 5 minutes which means it cached properly

Screenshot from 2022-09-07 22-19-18
Screenshot from 2022-09-07 22-15-31

motechFR:
@mattcasey As an interim option, I've split the playwright install into 2 steps,

- The npm install in the cached section
- The linux deps install just before browser test exection

As per your input, also narrowed it down to only Chromium browser, hopefully this works!

motechFR:
@mattcasey Seems like the split approach is a non starter.

We'll need to cache the system dependencies as part of our build. Will require some searching to see where they get saved

image

motechFR (approved):

PR #896, opened 2022-09-08
fix loading of new workspace

mattcasey:
This took me hours to figure out, basically when you create a new workspace from the sidebar, it would update the 'spaces' context, which would trigger an update inside both usePages and the EditorPage component. EditorPage calls 'setPages' from usePages() which overrides the API result from the call to charmClient inside of usePages().

I forget why we need to load the page here, I don't think we always need it, but it looks like it's only intended to load the first time the component is mounted, so I just check that the pageId is not equal to 'currentPageId'.

### DIFF
--- a/components/[pageId]/EditorPage/EditorPage.tsx
+++ b/components/[pageId]/EditorPage/EditorPage.tsx
@@ -12,7 +12,7 @@ import BoardPage from '../BoardPage';
 import DocumentPage from '../DocumentPage';
 
 export default function EditorPage ({ pageId }: { pageId: string }) {
-  const { setIsEditing, pages, setCurrentPageId, setPages, getPagePermissions } = usePages();
+  const { setIsEditing, pages, currentPageId, setCurrentPageId, setPages, getPagePermissions } = usePages();
   const [, setTitleState] = usePageTitle();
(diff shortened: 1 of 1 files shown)


Devorein (approved):

Devorein (approved):
Shouldn't we add `currentPageId` inside the deps array for `useEffect`

components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -25,7 +25,7 @@ export default function EditorPage ({ pageId }: { pageId: string }) {
   useEffect(() => {
     async function main () {
       setIsAccessDenied(false);
-      if (pageId && pagesLoaded && space) {
+      if (pageId && pagesLoaded && space && pageId !== currentPageId) {
```
Comment (Devorein) [inline components/[pageId]/EditorPage/EditorPage.tsx:28]: We could probably also check if the `currentPageId` is empty, as this is the place where it gets updated. But this worked just as fine.

Devorein (approved):

components/[pageId]/EditorPage/EditorPage.tsx
Comment (mattcasey) [reply components/[pageId]/EditorPage/EditorPage.tsx:28]: i think if you are navigating between pages you might need to refresh the data

PR #898, opened 2022-09-08
Fix API error that was causing public pages error

motechFR:
Simple fix, we couldn't load the public page as the endpoint was erroring out on empty content


motechFR:
Closing as handled by 899

PR #900, opened 2022-09-09
save the channel that notifications are received

mattcasey:
should help a bit with QA and possibly product analytics


Devorein:
We need to run a migration right? At the moment we only mark the mention type notifications as being viewed from the webapp, the others are done via email.

Devorein (approved):

PR #902, opened 2022-09-09
Disable new proposals button if user cannot create pages

motechFR:
Usability fixes for permissions
- Disable create proposal button if new page creation is disabled

- TODO: Don't allow admin override in editing proposal content (They can still delete a proposal and toggle it public)


mattcasey (approved):
nice find! thanks.. are we checking permissions for new bounty as well?

motechFR:
@mattcasey yep, we have the established workflow of "create bounty" or "suggest bounty"

PR #905, opened 2022-09-10
Fix comment and vote options not visible on floating menu

Devorein:


components/[pageId]/BoardPage/BoardPage.tsx
```diff
@@ -27,14 +28,18 @@ interface Props {
   setPage: (p: Partial<Page>) => void;
 }
 
-export default function BoardPage ({ page, setPage, readOnly }: Props) {
+export default function BoardPage ({ page, setPage, readOnly = false }: Props) {
   const router = useRouter();
   const board = useAppSelector(getCurrentBoard);
+  const { getPagePermissions } = usePages();
```
Comment (mattcasey) [inline components/[pageId]/BoardPage/BoardPage.tsx:34]: please don't add usePages here, it is triggered as the user types

PR #907, opened 2022-09-10
Show public bounty toggle

mattcasey:
Please double-check my work.. I had to remove some checks for the current page's permissions. mabe it was just copypasta from the other share button?


motechFR (approved):
@mattcasey approved!

Nice spot on the copy paste delete, always love a good negative diff PR!

PR #908, opened 2022-09-11
fix popup modal for public pages

mattcasey:
I needed to add the 'PageDialogProvider' around the public page component. The methods from `usePageDialog()` just silently fail without it (the methods values are no-ops). We could just put it inside _app so it's always there, but I didn't because it relies on Comments and Votes providers which are added in PageLayout.tsx, and I'd have to move those into _app as well.

### DIFF
--- a/pages/api/pages/[id]/index.ts
+++ b/pages/api/pages/[id]/index.ts
@@ -38,8 +38,31 @@ async function getPageRoute (req: NextApiRequest, res: NextApiResponse<IPageWith
     userId
   });
 
-  if (permissions.read !== true) {
+  if (permissions.read !== true && page.type !== 'bounty') {
     throw new ActionNotPermittedError('You do not have permission to view this page');
+  }
+  else if (permissions.read !== true && page.type === 'bounty') {
+    const bounty = await prisma.bounty.findUnique({
+      where: {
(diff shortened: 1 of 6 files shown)


motechFR (approved):
@mattcasey approved

We still have the issue of showing the apply now button & triggering the workflow. Will address in separate PR

I also added an integration test to cover the API fix I made

pages/api/pages/[id]/index.ts
```diff
@@ -38,9 +38,32 @@ async function getPageRoute (req: NextApiRequest, res: NextApiResponse<IPageWith
...
+        space: true
+      }
+    });
+
+    if (!bounty) {
+      throw new NotFoundError(`Bounty with id ${page.bountyId} not found`);
+    }
+
+    // We surface space-accessible bounties to the public if the space has enabled the public bounty board option
+    if (bounty.space.publicBountyBoard && bounty.permissions.some(p => p.spaceId === bounty.spaceId)) {
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:60]: @motechFR I noticed that this change is needed, but i can also still hide a bounty from public view (by updating its individual Share permissions) - how is that? Why do we need a special case here for bounty pages?

PR #912, opened 2022-09-13
enable viewing public bounty pages on their own

mattcasey:
This is a temporary fix until we can change permissions properly

### DIFF
--- a/pages/api/public/pages/[...pageId]/index.ts
+++ b/pages/api/public/pages/[...pageId]/index.ts
@@ -151,13 +151,15 @@ async function getPublicPage (req: NextApiRequest, res: NextApiResponse<PublicPa
     throw new NotFoundError('Space domain is required');
   }
 
-  const space = await prisma.space.findUnique({ where: { domain: spaceDomain } });
+  const space = await prisma.space.findUnique({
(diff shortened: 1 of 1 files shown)


pages/api/public/pages/[...pageId]/index.ts
```diff
@@ -147,19 +147,26 @@ async function getPublicPage (req: NextApiRequest, res: NextApiResponse<PublicPa
...
+    throw new NotFoundError('Space domain is required');
+  }
+
+  const space = await prisma.space.findUnique({ where: { domain: spaceDomain } });
+
+  if (!space) {
+    throw new NotFoundError('Space not found');
+  }
+
+  if (pagePath) {
```
Comment (motechFR) [inline pages/api/public/pages/[...pageId]/index.ts:160]: For safety, could we say if no page path and no page?

motechFR (approved):

Devorein (approved):

pages/api/public/pages/[...pageId]/index.ts
Comment (mattcasey) [reply pages/api/public/pages/[...pageId]/index.ts:160]: We haven't queried the page yet. It is never defined at this point

Comment (mattcasey) [reply pages/api/public/pages/[...pageId]/index.ts]: done

PR #913, opened 2022-09-13
turning off datadog agent using compose profile

yiwork:
adding COMPOSE_PROFILE env variable and specified profiles of which service to run when running a specific profile. This should allow us to easily turn on/off datadog agent


yiwork:
checkout screenshot: https://share.getcloudapp.com/d5unWPJR. No more datadog container

mattcasey (approved):

PR #914, opened 2022-09-13
Fix public bounty visibility

motechFR:


motechFR:
@mattcasey All done and workflow restored.

I am adding an e2e test, to make sure the bounty will show up. However, cannot complete it as playwright refuses to execute (see bug below).

It's to do with our import tree. Added some resources on ESM.

A simple way to solve this would be to have client-side helper methods which we call through the browser, or alternatively solve our import issues.

https://github.com/microsoft/playwright/issues/14371
https://playwright.dev/docs/test-typescript#typescript-with-esm

image

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
```diff
@@ -50,6 +52,13 @@ export default function BountyProperties (props: {
   const { user } = useUser();
   const { setPages, pages } = usePages();
 
+  const router = useRouter();
+
+  const isPublic = router.asPath.split('/')[1] === 'share';
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:57]: this is already handled in the PublicPage component. We should try to avoid having child components be smart about the URL being used

components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx
```diff
@@ -60,7 +88,7 @@ export default function BountySignupButton () {
             : (
               <TokenGateForm
                 onSuccess={() => {
-                  window.location.reload();
+                  router.push(`/${space?.domain}/${bountyPage.path}`);
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx:91]: are you sure this works? I think there's usually some things we need to do like refresh user permissions, which reloading the page handled

hooks/usePages.tsx
```diff
@@ -61,15 +61,7 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
...
-    const isPublicBountiesPage = (router.route === '/share/[...pageId]') && (router.query.pageId?.[1] === 'bounties');
-    if (isPublicBountiesPage) {
-      // retrieve the pages we need for public bounties
-      return charmClient.bounties.listBounties(currentSpace.id, true)
-        .then(bounties => bounties.map(bounty => bounty.page).filter(isTruthy));
-    }
-    else {
-      return charmClient.getPages(currentSpace.id);
-    }
+    return charmClient.getPages(currentSpace.id);
```
mattcasey, line 64: this is a nice benefit/cleanup

mattcasey (approved):
approved with one comment

Devorein:
It seems like we are duplicating a lot of code from `testing/setupDatabase.ts` inside `__e2e__/utilities.ts`. Is there any reason for now using the functions from `setupDatabase`?

Devorein:
Small ui issue on public bounty cards
image
Mark Complete should be disabled just like delete

Devorein (approved):

components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx:91]: @mattcasey Based on your feedback, I'm going with a simpler option, which is to perform a full app load by using window.location.href to make sure everything is up to date

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:57]: Thanks for this feedback @mattcasey 

There is also the nuance of why something should be read only, and whether we want to show all components.

I'd prefer to do this as part of a /router/share cleanup as this doesn't affect functionality for now

PR #915, opened 2022-09-13
Build the image then deploy EBStalk using the image

yiwork:


mattcasey (approved):
looks ok for staging, i think production still needs to be updated? (you can use the github action to deploy a docker image instead of zip file)

PR #916, opened 2022-09-14
Support multiple chains for Personal NFT

Devorein:


pages/api/profile/avatar.ts
```diff
@@ -21,21 +21,20 @@ handler
...
   let avatarUrl = avatar || '';
-  const chainId = 1;
-  const isNftAvatar = avatarContract && avatarTokenId;
+  const isNftAvatar = avatarContract && avatarTokenId && avatarChain;
 
   // Provided NFT data
   if (isNftAvatar) {
     const user = await getUserProfile('id', req.session.user.id);
-    const owners = await alchemyApi.getOwners(avatarContract, avatarTokenId, chainId);
+    const owners = await alchemyApi.getOwners(avatarContract, avatarTokenId, avatarChain as alchemyApi.SupportedChainId);
```
mattcasey, line 37: The type of avatarChain should be defined on UserAvatar, and we shouldn't need to typecast here since we do it on line 24

mattcasey (changes requested):
make sure we are not supporting any testnet networks

pages/api/profile/avatar.ts
Devorein: With that I had to change the type in a lot of function parameters in `lib/blockchain/nfts`, but now they are more strictly typed

pages/api/nft/list/[userId].ts
```diff
@@ -40,7 +42,9 @@ async function getNFTsMiddleware (req: NextApiRequest, res: NextApiResponse<NftD
   }
 
   const { addresses } = user;
-  const mappedNfts = await getNFTs(addresses);
+  const mappedNfts = (await Promise.all(
+    RPCList.filter(rpc => SupportedChainIds.includes(rpc.chainId as any)).map(rpc => getNFTs(addresses, rpc.chainId as SupportedChainId))
```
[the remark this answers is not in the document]
Comment (Devorein) [reply pages/api/nft/list/[userId].ts:46]: yep went with a constant array of supported mainnet chain ids.

mattcasey (approved):

PR #917, opened 2022-09-14
Dispatch notifications

Devorein:


lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -0,0 +1,154 @@
+import { Page, ProposalStatus, Space, WorkspaceEvent } from '@prisma/client';
+import { prisma } from 'db';
+import { ProposalWithUsers } from './interface';
+
+export type ProposalTask = {
+  id: string
+  action: 'discuss' |
```
motechFR, line 7: These action types should be extracted to their own type

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export async function getProposalTasksFromWorkspaceEvents (userId: string, workspaceEvents: WorkspaceEvent[]) {
+  const proposalTasks: ProposalTask[] = [];
+
+  // Sort the events based on their created date to help in deduping
+  workspaceEvents.sort((we1, we2) => we1.createdAt > we2.createdAt ? 1 : -1);
+
+  const proposals = await prisma.proposal.findMany({
+    where: {
+      id: {
```
motechFR, line 27: For better resilience, target proposal.pageId

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -0,0 +1,154 @@
...
+        in: workspaceEvents.map(workspaceEvent => workspaceEvent.pageId)
+      }
+    },
+    include: {
+      authors: true,
+      reviewers: true,
+      space: true,
+      page: true
```
motechFR, line 35: S:You should narrow down the page with select, to optimise to load only the required data

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -0,0 +1,154 @@
...
+  const spaceIds = spaceRoles.map(spaceRole => spaceRole.spaceId);
+  // Get all the roleId assigned to this user
+  const roleIds = spaceRoles.map(spaceRole => spaceRole.spaceRoleToRole[0].role.id);
+
+  const proposalsRecord: Record<string, ProposalWithUsers & {
+    space: Space
+    page: Page | null
+  }> = {};
+
+  proposals.forEach(proposal => {
```
motechFR, line 74: C: You could use Object.reduce to directly instantiate this object

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -0,0 +1,154 @@
...
+  // A single proposal might trigger multiple workspace events
+  // We want to register them as a single event
+  // This set keeps track of the proposals encountered
+  const visitedProposals: Set<string> = new Set();
+
+  for (const workspaceEvent of workspaceEvents) {
+    const proposal = proposalsRecord[workspaceEvent.pageId];
+
+    if (!visitedProposals.has(proposal.id)) {
+      const { newStatus } = workspaceEvent.meta as {
```
motechFR, line 87: W: It would be better to have a standard definition for each WorkspaceEvent

For example


```typescript

type ProposalStatusChangeMetaData
type ProposalStatusChangeWorkspaceEvent = WorkspaceEvent & {type: 'proposal_status_change} & {meta: ProposalStatusChangeMetaData}

```

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Agreed. Updated.

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Updated. Thank you. This is a great improvement.

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Updated thanks.

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Updated. Thank you.

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Agreed. Updated.

PR #919, opened 2022-09-15
New proposal status not refreshing

Devorein:


mattcasey (changes requested):
can you describe the case the proposal status doesn't refresh? we should really try to avoid needing to mutate SWR config directly. It looks like we already call mutateProposals() when the popup disappears.

mattcasey (approved):

PR #922, opened 2022-09-15
Copy workspace

motechFR:
Copy all pages and board pages from one space to another


motechFR:
@mattcasey Tagging you since you feedback was instrumental on this v2 :)

components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
```diff
@@ -12,7 +12,7 @@ import { useTreeItem } from '@mui/lab/TreeItem';
 import PageTreeItem from './PageTreeItem';
 import BoardViewTreeItem from './BoardViewTreeItem';
 
-export type MenuNode = Pick<Page, 'id' | 'title' | 'icon' | 'index' | 'parentId' | 'path' | 'type' | 'createdAt' | 'deletedAt'> & { isEmptyContent?: boolean };
+export type MenuNode = Pick<Page, 'id' | 'title' | 'spaceId' | 'icon' | 'index' | 'parentId' | 'path' | 'type' | 'createdAt' | 'deletedAt'> & { isEmptyContent?: boolean };
```
msadura, line 15: I would consider if it is worth to use `Pick` in such cases or just write a new type - at first glance you will do not know what are the types for `'id' | 'title' | 'spaceId' | 'icon' | 'index' | 'parentId' | 'path' | 'type' | 'createdAt' | 'deletedAt'`. Nothing important, but just to rethink approach in the future

lib/pages/interfaces.ts
```diff
@@ -42,7 +42,7 @@ export interface PublicPageResponse {
 // Generic type A is optional, we can mount additional properties on basic node definitions
 
 // eslint-disable-next-line @typescript-eslint/ban-types
-export type PageNode<A = {}> = Pick<Page, 'id' | 'type' | 'parentId' | 'index' | 'createdAt' | 'deletedAt'> & A
+export type PageNode<A = {}> = Pick<Page, 'id' | 'type' | 'parentId' | 'index' | 'createdAt' | 'deletedAt'> & Partial<Pick<Page, 'id' | 'type' | 'parentId' | 'boardId' | 'cardId'>> & A
```
msadura, line 45: hmm this got a bit complicated :D I am not sure if I understand what it should do, but shouldn't it be:

`export type PageNode<A = {}> = Pick<Page, 'id' | 'type' | 'parentId' | 'index' | 'createdAt' | 'deletedAt'> & Partial<Pick<Page, 'boardId' | 'cardId'>> & A`

I mean will union of `Pick` and `Partial<Pick` on the same properties do anything?

components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
motechFR: Pick makes more sense here, as we are deriving the node type directly from the page.

This ensures a single source of truth.

lib/pages/interfaces.ts
motechFR: Copy / paste fail, thanks for spotting, fixed

msadura (approved):

PR #924, opened 2022-09-16
Proposal tasks fixes

Devorein:


lib/proposal/__tests__/getProposalTasks.spec.ts
```diff
@@ -1,11 +1,43 @@
...
+      proposalStatus: 'private_draft',
+      spaceId: space.id,
+      authors: [user.id],
+      reviewers: [],
+      userId: user.id
+    });
+
+    const proposalTasks = await getProposalTasks(user.id);
+
+    expect(proposalTasks).toEqual(expect.arrayContaining([
```
msadura, line 32: I think it would be good to make sure that archived proposal is not returned in the array, something like:

```
    expect(proposalTasks).not.toEqual(expect.arrayContaining([
      expect.objectContaining({
        id: archivedProposal.id,
      })
    ]));
    ```

+ you need to store archived proposal data in variable

**Review by msadura — COMMENTED** [review]
Really nice stuff with testing flows!

**Comment by motechFR** [comment]
Small fix needed: Remove task from nexus once I've voted. Currently this requires a reload

File: components/nexus/ProposalTasksList.tsx [hunk 2]
```diff
@@ -69,7 +70,8 @@ export function ProposalTasksListRow (
               display='flex'
               sx={{
                 color: {
-                  xs: 'inherit'
+                  xs: 'inherit',
+                  md: theme.palette.link
```
mattcasey, line 74: What is the reason to change the color based on screen size? That seems like an anti-pattern. Also, I think you can just use 'palette.link'. No need to import the theme

components/nexus/ProposalTasksList.tsx
```diff
@@ -1,3 +1,4 @@
+import { useTheme } from '@emotion/react';
 import TaskOutlinedIcon from '@mui/icons-material/TaskOutlined';
 import { Alert, Button, Card, Grid, Link, Typography } from '@mui/material';
```
mattcasey, line 3: Please *always* use the 'common' Button component. It fixes one issue I ran into where "In Review" ends up on two lines

mattcasey: And also always use the common Link component, not the one from MUI

components/proposals/components/ProposalActionsMenu.tsx
```diff
@@ -46,7 +47,8 @@ export default function ProposalActionsMenu ({ deleteProposal, editProposal, pro
         {deleteProposal && (
           <MenuItem
             dense
-            onClick={() => {
+            onClick={async () => {
+              await refetchTasks();
```
mattcasey, line 51: You should call refetchTasks after you delete the proposal, and make sure it doesn't block the UI

components/nexus/ProposalTasksList.tsx
```diff
@@ -32,7 +33,7 @@ export function ProposalTasksListRow (
 ) {
   const proposalLink = `/${spaceDomain}/${pagePath}`;
   const proposalLocation = `${pageTitle || 'Untitled'} in ${spaceName}`;
```
mattcasey, line 35: remove the redundant page title

components/nexus/ProposalTasksList.tsx
Devorein: Updated to just be 'inherit' at all times.

components/nexus/ProposalTasksList.tsx
Devorein: Updated. Thank you. It's also responsible for client-side routing so as not to reload the whole page after routing.

components/proposals/components/ProposalActionsMenu.tsx
Devorein: Updated. I've also called refetchTasks when you cast a vote to remove the task

components/nexus/ProposalTasksList.tsx
Devorein: Updated

PR #925, opened 2022-09-16
Remove yarn.lock

msadura:
I habitually used `yarn install` instead of `npm install` and it created unnecessary `yarn.lock` file


mattcasey (approved):

mattcasey:
we can add yarn.lock to .gitignore if it's helpful

msadura:
> we can add yarn.lock to .gitignore if it's helpful

I think it is better not to use it since we decided to go with npm (actually it installed a few wrong version compared to npm's lock)

PR #926, opened 2022-09-16
Feat/analytics plausible

msadura:
Base boilerplate for plausible analytics. Turned out there is official guide for adding them to nextjs so I used it to keep things simple: https://plausible.io/docs/nextjs-integration

TODO: I need to double check how plausible works, but we will most likely need an account for our domain: https://plausible.io/#pricing

NOTE: Code will keep plausible turned off until we set up domain, so it should be safe to merge.


msadura:
We will not use plausible for now so I am closing this PR

PR #928, opened 2022-09-16
Delete proposal after deleting page

Devorein:


pages/api/pages/[id]/index.ts
```diff
@@ -142,6 +142,13 @@ async function deletePage (req: NextApiRequest, res: NextApiResponse<ModifyChild
 
   const modifiedChildPageIds = await modifyChildPages(pageId, userId, 'delete');
 
+  // Only top level page can be proposal page
+  await prisma.proposal.delete({
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:146]: please move this into modifyChildPages, where we delete other related docs like bounties and blocks

lib/proposal/getProposalTasks.ts
```diff
@@ -14,16 +15,16 @@ export interface ProposalTask {
 export function extractProposalData (proposal: Proposal & {
   space: Space,
   page: Page | null
-}, action: ProposalTask['action']): ProposalTask {
-  return {
+}, action: ProposalTask['action']): ProposalTask | null {
+  return proposal.page ? {
```
mattcasey, line 19: note for the future: I think we should probably always be querying by _pages_ first and not proposals

lib/pages/server/deleteArchivedPages.ts
```diff
@@ -2,6 +2,17 @@ import { prisma } from 'db';
 import { DateTime } from 'luxon';
 
 export async function deleteArchivedPages (maxDay: number) {
+  await prisma.proposal.deleteMany({
```
mattcasey, line 5: please export the count of deleted proposals and add it to the metrics inside of task.ts like we do for other documents

mattcasey, line 5: i think we may need to do the same for bounties... what do u think?

pages/api/pages/[id]/index.ts
Comment (Devorein) [reply pages/api/pages/[id]/index.ts:146]: Updated. Thanks.

lib/pages/modifyChildPages.ts
```diff
@@ -6,6 +6,13 @@ export type ChildModificationAction = 'delete' | 'restore' | 'archive'
 
 export async function modifyChildPages (parentId: string, userId: string, action: ChildModificationAction) {
 
+  // Only top level page can be proposal page
+  await prisma.proposal.delete({
+    where: {
+      id: parentId
```
mattcasey, line 12: even if the pageId happens to match the id, we don't have any guarantees. can you change this to delete based on the page?

PR #930, opened 2022-09-16
Use username as default profile path

msadura:
https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=25ac8182-5056-4b58-9902-8b2b06068146

Username as default user profile path (when available)

### DIFF
--- a/lib/users/createUser.ts
+++ b/lib/users/createUser.ts
@@ -27,7 +27,7 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
       data: {
         addresses: [address],
         identityType: IDENTITY_TYPES[0],
-        username: ens || shortenHex(address),
+        username,
         path: isUserPathAvailable ? username : null
       },
       include: sessionUserRelations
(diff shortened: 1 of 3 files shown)


lib/users/createUser.ts
```diff
@@ -19,12 +20,15 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
...
     const ens: string | null = await getENSName(address);
+    const username = ens || shortenHex(address);
+    const isUserPathAvailable = await isProfilePathAvailable(username);
 
     const newUser = await prisma.user.create({
       data: {
         addresses: [address],
         identityType: IDENTITY_TYPES[0],
-        username: ens || shortenHex(address)
+        username: ens || shortenHex(address),
```
mattcasey, line 30: it would be better to just use 'username' here instead of duplicating the logic from line 23

mattcasey (approved):

PR #936, opened 2022-09-19
Expand bounty kanban cards to fit title

Devorein:


msadura:
Keep in ming fixing tests before merging :)

PR #938, opened 2022-09-19
Add bounty button moved under properties

msadura:
https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=9264fe9f-6b23-40de-b0ec-1d4354f3d0be

Screenshot 2022-09-19 at 15 42 45

and  a bit of refactor


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -19,6 +19,7 @@ import dynamic from 'next/dynamic';
 import { useRouter } from 'next/router';
 import { memo, useCallback, useEffect, useState } from 'react';
 import { useElementSize } from 'usehooks-ts';
+import CardActions from 'components/common/BoardEditor/focalboard/src/components/cardDetail/CardActions/CardActions';
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:22]: we always add an index.tsx file which would export CardActions by default, so that the import doesn't have to have the name twice. (this is another thing I'm open to revisiting but I'd like to be consistent for now)

components/common/BoardEditor/focalboard/src/components/cardDetail/CardActions/CardActions.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <Box display="flex">
+        <div className='octo-propertyname add-property'>
+          <MenuWrapper>
+            <Button>
+              <FormattedMessage
+                id='CardDetail.add-property'
+                defaultMessage='+ Add a property'
+              />
+            </Button>
+            <Menu position='bottom-start'>
```
mattcasey, line 50: What is the benefit we get out of combining these two buttons in one component or problem you were trying to solve? I think one downside is that the focalboard properties-related logic is now split up. It's nice to keep tyhat somewhat together since it's using focalboard hooks and other components (which we'll refactor as time goes, hopefully but there's no timeline).  Product-wise, I don't see a reason to group all card actions in one place either. The logical grouping is that "add a property" is next to the list of properties and "convert to bounty" now makes the bounty properties appear.

components/common/BoardEditor/focalboard/src/components/cardDetail/CardActions/CardActions.tsx
```diff
@@ -0,0 +1,71 @@
...
+                  };
+                  const templateId = await mutator.insertPropertyTemplate(board, activeView!, -1, template);
+                  setNewTemplateId(templateId);
+                }}
+              />
+            </Menu>
+          </MenuWrapper>
+        </div>
+
+        <AddBountyAction readonly={readonly} cardId={pageId} />
```
mattcasey, line 68: Please make sure this falls on its own line.

components/common/BoardEditor/focalboard/src/widgets/buttons/button.tsx
```diff
@@ -18,16 +18,16 @@ type Props = {
     rightIcon?: boolean
 }
 
-function Button (props: Props): JSX.Element {
+function Button ({ size = 'small', ...props }: Props): JSX.Element {
```
mattcasey, line 21: Request: whenever you have to do anywork with buttons in focalboard, you might as well replace with our standard one first: components/common/Button. We should replace them all eventually

mattcasey (approved):
Tested locally. Personally, I think the 'Add a bounty' button should fall below "Add a property" and I left a comment about organization but nothing blocking and otherwise works well :)

components/common/BoardEditor/focalboard/src/components/cardDetail/CardActions/CardActions.tsx
msadura: My initial thinkink was that "add a bounty" button did not suit card properties component. With UI update you mentioned (`add a bounty `under `add a property`) it is no longer an issue so I can revert this component.

components/common/BoardEditor/focalboard/src/widgets/buttons/button.tsx
msadura: good to know thx!

msadura:
Updated design

Screenshot 2022-09-20 at 08 14 59

mattcasey (approved):

PR #939, opened 2022-09-19
Setup ESlint rule for preferring import type

motechFR:
This will simplify our dependency trees and builds by ensuring simple interface imports don't expand a dependency tree


msadura:
TBH  I did not know consistent type imports are this helpful. If they are then maybe we should make this error rather than warn?
BTW Can lint fix that wth `--fix` option? If it can then it would be nice to run it and add updated files to PR

msadura (approved):
Awesome stuff 💪

PR #940, opened 2022-09-19
Replace dots with hyphen in default user path

msadura:
Instead of using `...` from username, default path will have `-`

(request on discord)


mattcasey (approved):
thanks for entertaining my nitpicks :)

PR #942, opened 2022-09-20
Feat/mixpanel

msadura:
Draft of mixpanel implementation.  I wrapped raw mixpanel instance with our own `track` method which has string event typing - it will guard events and their payload.
Additionally I tried to split event types by category / feature to avoid one gigantic file with all the events.

I still need to handle mixpanel user handling, but let me know what you guys think about this approach / code organisation. @motechFR @mattcasey @Devorein


lib/analytics/mixpanel/interfaces/ProposalEvents.ts
```diff
@@ -0,0 +1,6 @@
+
+export type ProposalEvents= {
+  proposalCreated: { id: string; workspaceId: string; userId: string; }
+  proposalStageCreated: { id: string; workspaceId: string; userId: string; stageName: string }
+}
+
```
motechFR, line 6: Alternative style suggestion

_Btw I think as a team we prefer interfaces (let's discuss), but go for type if you need it_

Also for readability of the object typings, newlining and proper indentation are appreciated

Reasons for below:
- It's nicer to know what the actual shape of the event is

```typescript
export interface ProposalCreatedEvent  {
  id: string; 
 workspaceId: string;
 userId: string;
}

export interface ProposalStageChangedEvent {
  id: string;
  workspaceId: string;
  userId: string;
  stageName: string
}

export interface ProposalEvents {
 proposalCreated: ProposalCreatedEvent
 proposalStageChanged: ProposalStageChangedEvent
}

/* Ensures definition of Proposal event names happens locally here*/

export type ProposalEventType = keyof ProposalEvents

```

lib/analytics/mixpanel/server.ts
```diff
@@ -0,0 +1,16 @@
...
+  mixpanelInstance = Mixpanel.init(process.env.MIXPANEL_API_KEY as string);
+}
+
+function track<T extends MixpanelEventName> (eventName: T, params: MixpanelEvent[T]) {
+  mixpanelInstance?.track(eventName, params);
+}
+
+export const mixpanel = {
+  track
+};
```
motechFR, line 16: Why are we exporting an obj with a single prop? Why not the function directly?

lib/users/createUser.ts
```diff
@@ -34,6 +35,8 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
       include: sessionUserRelations
     });
 
+    mixpanel.track('userCreated', { userId: newUser.id, isWallet: true, isDiscord: false });
```
motechFR, line 38: Awesome

msadura:
This pr should be ready to re-review :)

Types for event are updated, introduced a few base types to reuse.

Added events related to user and proposals

lib/metrics/mixpanel/interfaces/UserEvents.ts
```diff
@@ -0,0 +1,24 @@
...
+  roles?: string[];
+  result: 'pass' | 'fail'
+}
+
+export interface SpaceJoined extends SpaceEvent {
+  source: 'invite_link' | 'public_bounty_token_gate' | 'token_gate'
+}
+
+export interface UserEvents {
+  UserCreated: UserCreatedEvent;
```
mattcasey, line 18: I think these should be user_created, etc.

lib/token-gates/applyTokenGates.ts
```diff
@@ -123,6 +131,8 @@ export async function applyTokenGates ({ spaceId, userId, tokens, commit }: Toke
       });
     }));
 
+    updateTrackUserProfileById(userId);
```
mattcasey, line 134: Can you make sure to handle possible exceptions? either here using .catch() or internally in the method

lib/metrics/mixpanel/server.ts
```diff
@@ -0,0 +1,33 @@
...
+export function trackUserAction<T extends MixpanelEventName> (eventName: T, params: MixpanelEvent[T]) {
+  const { userId, ...restParams } = params;
+
+  // map userId prop to distinct_id required by mixpanel to recognize the user
+  const mixpanelTrackParams: MixpanelTrackBase = {
+    distinct_id: userId,
+    ...restParams
+  };
+
+  mixpanelInstance?.track(eventName, mixpanelTrackParams);
```
mattcasey, line 20: we probably need to handle any exceptions here right? i would suggest using async/await for this method, even tho we don't await on it on the call sites. it's good to expose promises as high as possible imo

mattcasey (changes requested):
looks good, just want to change the event names to match what's in the doc: https://app.charmverse.io/charmverse/page-516372129559042 (UserCreated -> Sign up, for example)

lib/token-gates/applyTokenGates.ts
msadura: good call with exceptions!

lib/metrics/mixpanel/interfaces/UserEvents.ts
msadura: event names in code will be `event_name` and they will be converted to `Event Name` when send to mixpanel

msadura:
@motechFR @mattcasey pr udpated:
- changes event names
- added try / catch in mixpanel methods - might not be needed, but will take care of unhandled rejections

msadura:
Additional updated:
- event data props will also be human readable
- updated logic to verify if user connected via wallet
- updated logic of joining space via public bounty to reflect that in event

lib/metrics/mixpanel/server.ts
```diff
@@ -0,0 +1,69 @@
+import { capitalize } from 'lodash';
```
mattcasey, line 1: i think mixpanel.ts would be a better filename... we are not planning on having a browser implementation at all

lib/metrics/mixpanel/server.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+
+export function updateTrackUserProfile (user: LoggedInUser) {
+  const profile = {
+    $created: user.createdAt,
+    $name: user.username,
+    'Is Connected to Discord': !!user.discordUser,
+    'Is Connected via Wallet': !!user.wallets.length,
+    'Workspaces Joined': user.spaceRoles.map(sr => sr.spaceId)
```
mattcasey, line 39: use workspace name?

lib/metrics/mixpanel/server.ts
```diff
@@ -0,0 +1,69 @@
...
+import Mixpanel from 'mixpanel';
+
+import { isDev } from 'config/constants';
+import type { LoggedInUser } from 'models';
+
+import type { MixpanelEvent, MixpanelEventName, MixpanelTrackBase } from './interfaces/index';
+
+let mixpanelInstance: Mixpanel.Mixpanel | null = null;
+
+if (process.env.MIXPANEL_API_KEY && isDev) {
```
mattcasey, line 11: i dont think we need to check 'isDev' here. it should be enough to just use the api key if it's set

lib/metrics/mixpanel/trackPageAction.ts
```diff
@@ -0,0 +1,10 @@
+import { prisma } from 'db';
+import type { PageEventNames } from 'lib/metrics/mixpanel/interfaces/PageEvents';
+import { trackUserAction } from 'lib/metrics/mixpanel/server';
+
+export async function trackPageAction (eventName: PageEventNames, userId: string, pageId: string) {
```
mattcasey, line 5: please use object notation, at least for the attributes. i could see someone easily mixing userId and pageId since the'yre both strings

lib/metrics/mixpanel/updateTrackUserProfileById.ts
```diff
@@ -0,0 +1,13 @@
...
+import { getUserProfile } from 'lib/users/getUser';
+
+export async function updateTrackUserProfileById (userId: string) {
+  try {
+    const user = await getUserProfile('id', userId);
+
+    updateTrackUserProfile(user);
+  }
+  catch (e) {
+    // Failed to update tracking profile
```
mattcasey, line 11: Always log something in the error path, log.warn would be good to use here because it won't trigger alerts but we can use  it for debugging

lib/users/createUser.ts
```diff
@@ -41,6 +42,9 @@ export async function createUserFromWallet (address: string): Promise<LoggedInUs
       include: sessionUserRelations
     });
 
+    updateTrackUserProfile(newUser);
+    trackUserAction('sign_up', { userId: newUser.id, identityType: 'Wallet' });
```
mattcasey, line 46: Do we need to wait for the user profile to be added before sending a user event?

lib/votes/castVote.ts
```diff
@@ -34,5 +36,9 @@ export async function castVote (choice: string, vote: Vote & { voteOptions: Vote
     }
   });
 
+  if (vote.context === VoteContext.proposal) {
+    trackUserAction('new_vote_created', { userId, resourceId: voteId, spaceId: vote.spaceId, platform: 'charmverse' });
```
mattcasey, line 40: i don't think this was chris's intention, "userVote" in our data model is just a user casting their choice on a poll or proposal. i think he meant to track new polls. i asked on the doc

pages/api/session/login.ts
```diff
@@ -43,6 +45,9 @@ async function login (req: NextApiRequest, res: NextApiResponse<LoggedInUser | {
   await updateGuildRolesForUser(user.wallets.map(w => w.address), user.spaceRoles);
   await req.session.save();
 
+  updateTrackUserProfile(user);
```
mattcasey, line 48: why do we need to update user profile on sign in? (what properties could have changed?)

lib/metrics/mixpanel/interfaces/index.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export interface MixpanelUserProfile {
+  $created: Date;
+  $name: string;
+  discordConnected: boolean;
+  walletConnected: boolean;
+  spaces?: string[];
+}
+
+export type MixpanelEvent = UserEvents & ProposalEvents & PageEvents
```
mattcasey, line 33: to keep consistent naming convention, 'MixpanelEvent'  should be 'MixpanelEvents'. I think "MixPanelEventMap" would be a bit better

lib/metrics/mixpanel/interfaces/PageEvents.ts
```diff
@@ -0,0 +1,14 @@
+import type { ResourceEvent, SpaceEvent } from 'lib/metrics/mixpanel/interfaces';
```
mattcasey, line 1: this "works" but it's a recursive dependency. Please fix... 'interfaces/index' imports PageEvents, but PageEvents imports index.ts. Looks like ResourceEvent and SpaceEvent should be their own file

lib/metrics/mixpanel/interfaces/ProposalEvents.ts
```diff
@@ -0,0 +1,22 @@
+import type { ProposalStatus } from '@prisma/client';
+
+import type { SpaceEvent } from 'lib/metrics/mixpanel/interfaces';
```
mattcasey, line 3: same comment as PageEvents... keep dependencies single-direction. I think it's very confusing to follow even if it works in typescript

.ebextensions/00_env_vars.config
```diff
@@ -11,6 +11,7 @@ option_settings:
         DB_USER: "{{pull:secretsmanager:/io.cv.app/stg/db:SecretString:username}}"
         DD_API_KEY: "{{pull:secretsmanager:/io.cv.app/shared/datadog:SecretString:dd_api_key}}"
         DEEP_DAO_BASE_URL: "https://api.deepdao.io"
+        MIXPANEL_API_KEY: "{{pull:secretsmanager:/io.cv.app/stg/mixpanel:SecretString:mixpanel_api_key}}"
```
mattcasey, line 14: @msadura i added mixpanel api key and updated your branch with main. We now have an api key for staging (the previous one) and one for production that hasn't been used yet

mattcasey:
@msadura i know this story's been going on for a while but seems like it's worth adding groups from the git-go: that way we dont need to send 'workspaceName' as part of the event, we can probably just use the workspace id as the group key: https://developer.mixpanel.com/docs/nodejs#sending-group-identifiers-with-events

lib/users/createUser.ts
msadura: We do not need to wait and we actually don't wait - both actions are async and to not wait for each other

pages/api/session/login.ts
msadura: It was in case user profile was not existing yet. But if we will migrate all users to mixpanel we can skip this `updateTrackUserProfile` call

pages/api/pages/[id]/details.ts
```diff
@@ -34,6 +35,8 @@ async function getPageDetailsHandler (req: NextApiRequest, res: NextApiResponse<
     throw new ActionNotPermittedError('You do not have permission to view this page');
   }
 
+  trackPageAction('page_load', userId, pageId);
```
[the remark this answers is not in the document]
Comment (msadura) [reply pages/api/pages/[id]/details.ts:38]: it should be loaded for the first time you open a page and then cached so it should not spam this more. I will double check it tho. If this logic will not be sufficient then we will need `/track` endpoint and simply call it when user navigates to as page.

msadura:
@mattcasey I adressed you comments, thx for review!

spaceId will be send as group indentifier to all mixpanel events (except for login / signup events, because we basically do not have space information there).

2 things to do (I will let you judge if in this PR or separate):
- I need to verify if mixpanel will accept `Space Id[]` in user profile to connect it to proepr groups (docs mention having multiple groups per user is possible)
- I need to add updating group profile - group profile will contain space information (I think name for starters?), so I need a logic to create / update group profile when new space is created / updated

mattcasey (approved):
really nice work!

PR #944, opened 2022-09-20
Stepper for Proposals

Devorein:


mattcasey (approved):

mattcasey:
@Devorein I think a better UX would be to load the component IMMEDIATELY if the page.type === 'proposal' and then update the state once the data comes in. That way you don't have a jumping layout when it first loads

Devorein:
I think it's best if we do that on a separate card, as we can't use page.type === 'proposal' for nested pages of a proposal. Plus, this should also include the proposal properties, not just the status steppers. I've added a  separate card for it.

PR #946, opened 2022-09-21
Lasting proposals & marking proposals in nexus

Devorein:


background/tasks/sendNotifications/sendNotifications.ts
```diff
@@ -72,7 +72,7 @@ export async function getNotifications (): Promise<PendingTasksProps[]> {
           in: [
             ...gnosisSafeTasks.map(getGnosisSafeTaskId),
             ...voteTasks.map(voteTask => voteTask.id),
-            ...workspaceEvents.map(workspaceEvent => workspaceEvent.id)
+            ...workspaceEvents.map(workspaceEvent => `${workspaceEvent.id}.${user.id}`)
```
mattcasey, line 75: do we have the same issue with votes?

components/nexus/ProposalTasksList.tsx
```diff
@@ -110,11 +114,36 @@ export function ProposalTasksListRow (
...
+              unmarked: []
+            }
+          } : undefined;
+        }, {
+          revalidate: false
+        });
+      }
+    }
+    main();
+  }, [tasks, user]);
```
mattcasey, line 146: we only want to run this if user.id changes

lib/proposal/getProposalTasks.ts
```diff
@@ -27,145 +30,142 @@ export function extractProposalData (proposal: Proposal & {
...
+        id: workspaceEvent?.id ?? proposal.id,
+        pagePath: (proposal.page as Page).path,
+        pageTitle: (proposal.page as Page).title,
+        spaceDomain: proposal.space.domain,
+        spaceName: proposal.space.name,
+        status: proposal.status,
+        action
+      };
+
+      if (workspaceEvent && !userNotificationIds.has(`${workspaceEvent.id}.${userId}`)) {
```
mattcasey, line 158: as mentioned above... the formatting of the id should really only be handled in one place, or in a utility function. that would allow us to see where it gets used at least, and we make sure it's all consistent

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
```diff
@@ -120,7 +120,7 @@ export async function getProposalTasksFromWorkspaceEvents (userId: string, works
       }));
 
       const proposalTask: Omit<ProposalTask, 'action'> = {
-        id: workspaceEvent.id,
+        id: `${workspaceEvent.id}.${userId}`,
```
mattcasey, line 123: +1 can we do this inside getProposalTasks as well?

mattcasey (changes requested):
TL;DR: can you reduce the amount of places we format the unique id for proposal tasks. But Also, isn't this an issue for votes? And did you consider just fixing the constraint on the db?

background/tasks/sendNotifications/sendNotifications.ts
Devorein: I've removed the userId when creating the workspace event id, it turns out we had two unique directives for our UserNotification prisma schema. Removing that fixed the issue.

lib/proposal/getProposalTasks.ts
```diff
@@ -27,145 +30,142 @@ export function extractProposalData (proposal: Proposal & {
...
+      const action = getProposalAction(
+        {
+          currentStatus: proposal.status,
+          isAuthor,
+          isReviewer
+        }
+      );
+
+      const proposalTask = {
+        id: workspaceEvent?.id ?? proposal.id,
```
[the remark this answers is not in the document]
Devorein: Updated code to use a uuid, rather than relying on proposal if when workspace event doesn't exist. Normally workspace event does exist (unless its a newly created proposal), but now we will create a new workspace event anytime a proposal is created.

lib/proposal/getProposalTasksFromWorkspaceEvents.ts
Devorein: Removed userId when generating proposalTask id

prisma/schema.prisma
```diff
@@ -139,6 +139,7 @@ enum ProposalStatus {
 
 enum WorkspaceEventType {
   proposal_status_change
+  proposal_create
```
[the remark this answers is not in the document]
Devorein: Yep I will remove the migration and update the code dependant upon it.

mattcasey (approved):

PR #947, opened 2022-09-21
Suggestion mode

mattcasey:


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -79,8 +83,8 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false, parentPro
     }
   }, [page.bountyId]);
 
-  const cannotEdit = readOnly || !pagePermissions?.edit_content;
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:82]: For readOnly, we check page permissions in both parents of DocumentPAge already, no need to check again

components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -50,10 +54,15 @@ export const BangleEditor = React.forwardRef<
...
-      pmViewOpts
+      pmViewOpts,
+      enableSuggestions
     });
     const [editor, setEditor] = useState<CoreBangleEditor>();
     const nodeViews = useNodeViews(renderRef);
+    const { user } = useUser();
+
+    // set current
+    editorViewPayloadRef.current.enableSuggestions = enableSuggestions;
```
mattcasey, line 65: use a ref so that inside dispatchTransaction, we get the *real* value and not just the first value

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -153,29 +155,32 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false, parentPro
...
-          {page.headerImage && <PageBanner headerImage={page.headerImage} readOnly={cannotEdit} setPage={setPage} />}
+          {/* temporary? disable editing of page meta data when in suggestion mode */}
+          {page.headerImage && <PageBanner headerImage={page.headerImage} readOnly={readOnly || enableSuggestingMode} setPage={setPage} />}
           <Container
             top={pageTop}
             fullWidth={page.fullWidth ?? false}
           >
             <CharmEditor
-              key={page.id}
+              key={page.id + editMode}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:165]: refresh the entire view when editMode changes

components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -60,8 +63,26 @@ export default function EditorPage ({ pageId }: { pageId: string }) {
 
   }, [pageId, pagesLoaded, space, user]);
 
+  // set page attributes of the primary charm editor
+  useEffect(() => {
+    if (!editMode) {
+      if (currentPagePermissions?.edit_content) {
```
Comment (msadura) [inline components/[pageId]/EditorPage/EditorPage.tsx:69]: setPageProps({ permissions: currentPagePermissions, editMode: currentPagePermissions?.edit_content ? 'editing' : 'viewing' });

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -79,8 +83,8 @@ function DocumentPage ({ page, setPage, insideModal, readOnly = false, parentPro
     }
   }, [page.bountyId]);
 
-  const cannotEdit = readOnly || !pagePermissions?.edit_content;
   const cannotComment = readOnly || !pagePermissions?.comment;
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:86]: W: Just noticed you moved the logic for deciding readOnly outside this component. Since we use it in a couple places, the original idea was to pass readOnly from external component if that component use case requires readOnly (such as publicPage) and have aDRY approach to consuming page permissions

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:167]: Nice one, re-rendering the charmEditor based on readOnly changes has always been one of those things that catches us

motechFR:
**Tests**

- [X] Add a suggestion and delete underlying text in edit mode
Expected: Suggestion box disappears automatically

- [X] Accept a suggestion
Expected: Suggestion loses special marking and becomes "normal" page content viewable in "view" and "edit" mode

- [X] Hide suggestions in view mode
Expected: Content without any suggestions appears

- [X] Share page to public
Expected: Shows same content as edit/view mode without any suggestions

- [ ] Comment a suggestion
Expected: Comment shows up with suggestion content
Issue: Highlighted part of suggestion for comment disappears in suggestion box. But, if we accept suggestion; it gets accepted

Expected: Suggestion icon + Comment icon in same line
Issue: Only 1 icon shows

Screenshot from 2022-09-26 17-51-48
Screenshot from 2022-09-26 17-54-03


- [ ] Create inline poll for a suggestion
Same expects and issues as for the comment (This time, we only see the poll icon)
Screenshot from 2022-09-26 18-02-28
Screenshot from 2022-09-26 18-03-00

- [ ]  Suggest embedded widgets
Expected: Unknown if we should support this

Issue: Tried adding an inline database, it doesn't come through as  suggestion.

Issue: Added crypto price. It created 2 of them; which still shows unconfigured in the edit mode.

Suggestion: In the spirit of suggested edits; should we drastically restrict functionality in suggest mode? Specifically
- Text
- Formatting

Screenshot from 2022-09-26 18-09-22

Devorein:
- [ ] Removing editor state toggle button for board pages
- [ ] Still showing a wrong suggestion creation time (It was created a few seconds ago)
imagei
- [ ] Changing the editor state shifts the scrollbar to the top (I think this is a general issue with modals and menus)
Other than that, everything is working great 🥳 . This is a marvellous feat of human engineering. I don't have any other way to explain this ...

PR #948, opened 2022-09-21
Adding healthcheck endpoint and also fixed datadog metric submission

yiwork:


docker-compose.yml
```diff
@@ -13,12 +13,35 @@ services:
...
+    volumes:
+      - $PWD:/app
+    env_file:
+      - ".env"
+    depends_on:
+      - localpostgres
+    profiles:
+      - dev
+
+  localpostgres:
```
mattcasey, line 36: so we now have it set up to run locally? neat - I'd like to add the commands to our README. would you mind? or tell me what would the commands be to start up docker

pages/api/health/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import log from 'lib/log';
+import { gauge, count } from 'lib/metrics';
+
+export default function handler (req: NextApiRequest, res: NextApiResponse) {
+  res.status(200).json({ health: 'ok' });
+}
+
+// need to add a router to this.
+export function testDDGauge (req: NextApiRequest, res: NextApiResponse) {
+  gauge('.health.dd.gauge', 99);
```
mattcasey, line 11: i dont think you need the starting "." right?

mattcasey (approved):
approved with one question

pages/api/health/index.ts
mattcasey: I noticed it looks right in datadog, maybe they do something special with the prefix? Do you think we always need to have a "."

docker-compose.yml
yiwork: We can have it setup to run locally if we chose to. Though I haven't actually ran it to see if it worked. Can work on that next.

pages/api/health/index.ts
yiwork: Good question, don't know. I'll remove it and see what happens

PR #949, opened 2022-09-21
disable token gate test if litClient is not initialized

mattcasey:
Loom of current user experience: https://www.loom.com/share/98b4541d752446149aafb1bd7c44ad09

### DIFF
--- a/components/settings/roles/components/TokenGates/components/TokenGatesTable.tsx
+++ b/components/settings/roles/components/TokenGates/components/TokenGatesTable.tsx
@@ -73,10 +73,12 @@ export default function TokenGatesTable ({ isAdmin, onDelete, tokenGates }: Prop
   async function testConnect (tokenGate: TokenGate) {
     setTestResult({ status: 'loading' });
     try {
-
+      if (!litClient) {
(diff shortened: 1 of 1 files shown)


components/settings/roles/components/TokenGates/components/TokenGatesTable.tsx
```diff
@@ -73,10 +73,12 @@ export default function TokenGatesTable ({ isAdmin, onDelete, tokenGates }: Prop
   async function testConnect (tokenGate: TokenGate) {
     setTestResult({ status: 'loading' });
     try {
-
+      if (!litClient) {
```
mattcasey, line 76: this case should never happen but i feel like it's better than using the "!" operator

Devorein: It will happen if the network gets disconnected during the connection phase. I've encountered it a bunch of time in my local environment

mattcasey: oh, i just meant that this function won't be called because i short-circuit on the CHip now :)

Devorein (approved):

PR #951, opened 2022-09-21
Proposals template

motechFR:
Admins can manage workspace proposal templates

Users can create a proposal if they have the space / create page permission


lib/templates/proposals/createProposalTemplate.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const proposalId = v4();
+
+  return prisma.page.create({
+    data: {
+      id: proposalId,
+      path: getPagePath(),
+      content: pageContent?.content,
+      contentText: pageContent?.contentText ?? '',
+      title: pageContent?.title ?? 'Untitled',
```
mattcasey, line 30: FYI we never hard-code Untitled, it's handled in the UI

PR #953, opened 2022-09-21
Fix tests for Proposal notifications  PR

mattcasey:
I fixed the tests, but also simplified the createProposal() method by removing the Prisma interface from the library function. I think it's a kind of awkward shape when you're just trying to pass in properties:
For example, now you can just pass `{ createdBy: userId }` instead of `{ author: { connect: { id: userId } } }`

I also removed the extra workspace event type. I don't think it really warrants that, the _use case_ is exactly the same as proposal_status_change. It's just the first step in the flow.


pages/api/pages/index.ts
```diff
@@ -46,39 +46,22 @@ async function createPage (req: NextApiRequest, res: NextApiResponse<IPageWithPe
...
-  if (typedPageCreationData.type === 'proposal') {
-    page = await createProposal({
-      pageCreateInput: typedPageCreationData,
+  if (pageCreationData.type === 'proposal') {
+    ({ page } = await createProposal({
+      ...pageCreationData,
       spaceId,
-      userId
-    });
-    await prisma.workspaceEvent.create({
```
mattcasey, line 71: i moved this to createProposal, it seems like we will always want it to exist

lib/proposal/createProposal.ts
```diff
@@ -1,52 +1,48 @@
...
-          authors: {
-            create: {
-              author: {
-                connect: {
-                  id: userId
-                }
-              }
-            }
+  const [proposal, page, workspaceEvent] = await prisma.$transaction([
+    prisma.proposal.create({
```
mattcasey, line 14: i split out the creation of the proposal because Prisma doesn't let you combine the plain column values for foreign keys and also the 'connect'/'create' syntax for some reason

lib/proposal/createProposal.ts
```diff
@@ -1,52 +1,48 @@
...
+        spaceId,
+        status: 'private_draft',
+        authors: {
+          create: {
+            userId: createdBy
           }
         }
       }
-    },
-    include: {
```
mattcasey, line 41: we were only using this for asserting results on the test.

prisma/migrations/20220921220303_remove_workspace_event_type/migration.sql
```diff
@@ -0,0 +1,14 @@
+/*
```
mattcasey, line 1: this is pretty ugly migration! if you agree w/my change, please just remove the original migration that added the new type

Devorein: Yep I think this is a lot cleaner. I will remove my migration and yours as well.

pages/api/pages/index.ts
Devorein: Right thank you for that.

PR #957, opened 2022-09-22
Feat/filtering by category

msadura:
Proposal category in proposals table + filtering by category

Screenshot 2022-09-22 at 11 42 19
Screenshot 2022-09-22 at 11 13 25


msadura:
+ Added creating default categories to the scope of this pr

lib/proposal/getDefaultCategories.ts
```diff
@@ -0,0 +1,15 @@
...
+const titles = [
+  'Operational Budget',
+  'Grants',
+  'Investment',
+  'Elections',
+  'Governance',
+  'Other'
+];
+
+export function getDefaultCategories (spaceId: string): Omit<ProposalCategory, 'id'>[] {
```
motechFR, line 13: I would rename this to generateDefaultCategoriesInput since it will be consumed by Prisma

PR #960, opened 2022-09-22
Mobile view support for proposal page stepper

Devorein:


components/proposals/components/ProposalStepper.tsx
```diff
@@ -101,11 +107,38 @@ export default function ProposalStepper (
               </Stack>
             </Grid>
             {statusIndex !== proposalStatuses.length - 1 && (
-              <Grid item md={12 / 13}>
+              <Grid
+                item
+                md={12 / 13}
+                sx={{
+                  '&.MuiGrid-root.MuiGrid-item': {
```
msadura, line 114: this feels hacky. Why do you need to style refering specific classes?

Devorein: Yeah, ikr. Its due to a specific class added to the proposal and bounty properties to make them look like our focalboard properties. So the sx style for Divider was being overridden by `.focalboard hr` and I had to use the class itself to increase the specificity. I will give it another shot and see if its possible without using these verbose classes

mattcasey: or just delete the rule about focalboard hr. I'm not really aware of us needing an hr tag in the layout, and if it is we could replace it with Divider

PR #968, opened 2022-09-22
Script to create default categories

msadura:
Script to update all existing spaces with default categories. It will only update spaces without any category.

Additionally small prisma schema cleanup.

### DIFF
--- a/scripts/createDefaultProposalCategories.ts
+++ b/scripts/createDefaultProposalCategories.ts
@@ -1,0 +1,30 @@
+import { ProposalStatus } from '@prisma/client';
+import { prisma } from 'db';
+import { generateDefaultCategoriesInput } from 'lib/proposal/generateDefaultCategoriesInput';
+
+async function migrateProposal () {
+  const spacesWithoutCategories = await prisma.space.findMany({
+    where: {
+      proposalCategory: {
+          none: {} // Space has no categories
+      }
+    },
+    select: {
+      id: true
+    }
+  });
+
+  console.log('🔥 Count of spaces without categories:', spacesWithoutCategories.length);
+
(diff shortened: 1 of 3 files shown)


scripts/createDefaultProposalCategories.ts
```diff
@@ -0,0 +1,30 @@
+import { ProposalStatus } from '@prisma/client';
+import { prisma } from 'db';
+import { generateDefaultCategoriesInput } from 'lib/proposal/generateDefaultCategoriesInput';
+
+async function migrateProposal () {
+  const spacesWithoutCategories = await prisma.space.findMany({
+    where: {
+      proposalCategory: {
+          none: {} // Space has no categories
```
motechFR, line 9: Nice, hadn't used this before

Turns out prisma supports this natively and simply with syntax you used
https://www.prisma.io/docs/guides/database/troubleshooting-orm/help-articles/finding-entities-based-on-relation#solution

scripts/createDefaultProposalCategories.ts
```diff
@@ -0,0 +1,30 @@
...
+      }
+    },
+    select: {
+      id: true
+    }
+  });
+
+  console.log('🔥 Count of spaces without categories:', spacesWithoutCategories.length);
+
+  await Promise.all(spacesWithoutCategories.map(({ id }) => createCategoriesForSpace(id)))
```
motechFR, line 19: Could you wrap this inside a transaction? Will ensure we don't have a half successful migration

msadura: I can but I wasn't sure if we want it - if for some reason script fails, you can safely re-run it because it takes only spaces without categories.
Anyway I think I can add transaction as it is general good practise

msadura: done.

mattcasey (approved):

PR #969, opened 2022-09-22
Admin overrides

motechFR:
Admin can now update proposal properties and move it across stages


mattcasey (approved):

Devorein (approved):

Devorein:
I fixed a small UI bug where we show the admin is able to change the proposal status to any stage (blue circle).
image

PR #974, opened 2022-09-23
Feat/dev title

msadura:
I realised I often use a few tabs with CV open in dev and prod. Not sure if you like this, but small tweak will help to recognize which one is dev one easily

Screenshot 2022-09-23 at 11 34 25

@motechFR @mattcasey @Devorein

### DIFF
--- a/config/constants.ts
+++ b/config/constants.ts
@@ -1,0 +1,1 @@
+export const isDev = process.env.NODE_ENV === 'development';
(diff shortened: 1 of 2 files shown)


motechFR:
Very good idea!

Devorein:
+1 to that. I get confused a lot as well on which one is development and which one is production.

mattcasey (approved):
nice :)

PR #975, opened 2022-09-23
Rationalise permissions

motechFR:
This PR moves the use of space permission "createVote" to be used for the creation of proposals

It adds a "create poll" ability at the page permission level so that inline poll creation depends on your access level to that page

For proposals, only proposal editors can create polls


msadura (approved):
Awesome work around tests! 🥇

PR #979, opened 2022-09-24
Speed up jest

motechFR:
Pass maxWorkers flag as simple solution to accelerate jest


mattcasey:
This... looks like it halved the time to run tests? Comparing the latest test run on main, basic tests took 5min/30s and server tests 3 minutes: https://github.com/charmverse/app.charmverse.io/actions/runs/3113476587/jobs/5048154863.

For this PR it says 2 minutes and 1min/30 seconds!

motechFR:
@mattcasey All works, this should also work on a computer with extra params. Worked fine with me.

After that first win, I went back and decided to see if we could take this further.

Finally found the magic CI config so that we can prioritise running tests that have been affected by the current

mattcasey:
> Finally found the magic CI config so that we can prioritise running (Unit + Integration) tests that have been affected by the current PR
It's normal to always run all tests in CI. How would Jest know when there's a link between code and tests that is outside of the git commit change. If I changed code in an api endpoint, how does Jest know which api test to run?

lib/bounties/createBounty.ts
```diff
@@ -25,6 +25,9 @@ export async function createBounty ({
   permissions
 }: BountyCreationData) {
 
+  // Test commit to trigger CI
+  const test = 1;
```
msadura, line 29: you can do: `git commit –allow-empty -m "Reuild or any message"` - it will create an empty commit which you can push to retrigger CI without any changes

mattcasey: neat

motechFR: @msadura Yep know that trick, here in this PR, what I'm testing is a bit different.

I want jest to magically run only a subset of tests which have themselves (or stuff they import) been affected in the current PR.

Checkout
.github/actions/install/action.yml
.github/workflows/test_and_deploy.yml

package.json (l.41)

To see how it all comes together where we assess the current PR tree against the head commit of main branch

motechFR:
> 

@mattcasey Agree we want to run all tests in CI. All I'm trying to do, is get Jest to run what might be affected first before running the full suite

In the screenshot below, you can see I've added an extra test step, which runs with the changedSince setup.

Of course, we then run unit tests and integration tests. Totally agree we want to have the full suite run in CI! I'm just trying to speed up even further the moment where a developer discovers if they've broken something or not.

Current issue: Getting jest to not re run tests in steps 2 and 3 that were already run in step 1

Screenshot from 2022-09-26 20-19-39

PR #980, opened 2022-09-26
Override broken react deps for npm

msadura:
This overrides react deps for all packages and will avoid issue with npm install:

Screenshot 2022-09-26 at 10 20 42


mattcasey:
This is an amazing solution, thanks! I thought we'd need to fork the dependency

mattcasey:
FYI, it's fine to just delete and re-generate the package-lock.json

mattcasey (approved):

PR #981, opened 2022-09-26
Eslint rule for types separator

msadura:
This PR introduces unifying of types and interfaces syntax and adds `;` as required separator and fixes all the files.

If you are afraid of potential conflicts in PR, here is how to deal with that easily:

- merge main (which will have new rule)
- accept your changes for all the files
- run `npm run lint --fix` - it will reformat your code
- you are all set :)


.eslintrc.json
```diff
@@ -36,6 +36,7 @@
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-inferrable-types": "off",
     "@typescript-eslint/no-empty-interface": "off",
+    "@typescript-eslint/member-delimiter-style": "error",
```
motechFR, line 39: Nice, we definitely needed standards on this

motechFR:
@msadura Do we have the rule for unifying types and interfaces yet? Couldnt see that in eslint file

msadura:
> we have the rule for unifying types and interfaces yet? Couldnt see that in eslint file

I am not sure if we will be able to get any rule for that. Feels like continuous manual work 😅 but I check if there is any solution

mattcasey:
@motechFR @msadura what do you guys think of allowing commas for single-line types? I don't know why, it feels more natural in that case to read and the rule allows for multiline vs single

mattcasey:
There's actually a rule for type vs interface:
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-definitions.md

msadura:
> @motechFR @msadura @Devorein what do you guys think of using commas for single-line types? That's what I've gotten used to, and the rule allows for multiline vs single. We didn't discuss 'commas' vs 'semi-colons' in our tech chat, but I would also be fine just using _commas_ for type props everywhere too, if eslint can correct it all for us.

works for me :)

msadura:
updates:
- fixed conflicts
- updated single-line types to use `,` instead of `;`
- added `@typescript-eslint/object-curly-spacing` rule to enforce space in curly brackets for both obejct and types `{ asd: string }`  intead of `{asd:1}`

asd for "types vs interfaces" I would not add the rune in this pr to avoid massive changes at once

@mattcasey @motechFR

mattcasey (approved):

PR #982, opened 2022-09-26
Support multiple multi-sig wallets on the same chain

Devorein:


components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -48,17 +51,27 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
     () => getSafesForAddress({ signer: signer!, chainId: chainId!, address: account! })
   );
 
-  // use first multisig wallet
-  const multiSigWallet = (safeInfos || [])[0] as SafeData | undefined;
-  const gnosisSafeAddress = multiSigWallet?.address;
-  const gnosisSafeChainId = multiSigWallet?.chainId;
+  const safeDataRecord = useMemo(() => {
```
mattcasey, line 54: this is an unecessary use of useMemo, and also it's watching the wrong variable

components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -142,10 +157,68 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
...
+              <Box justifyContent='space-between' gap={2} alignItems='center'>
+                <div
+                  className='octo-propertyrow'
+                  style={{
+                    height: 'fit-content',
+                    display: 'flex',
+                    alignItems: 'center'
+                  }}
+                >
+                  <div className='octo-propertyname octo-propertyname--readonly'>
```
mattcasey, line 190: Please do *not*  use focalboard classes  anywhere outside of focalboard. You shoudl use standard MUI components for field labels

components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -35,6 +36,8 @@ interface TransactionWithMetadata extends MetaTransactionData, Pick<Bounty, 'rew
 
 export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDetails[]}) {
   const [isLoading, setIsLoading] = useState(false);
+  const [gnosisSafeData, setGnosisSafeData] = useState<SafeData | undefined>(undefined);
```
mattcasey, line 39: please set the first wallet by default

components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -201,15 +274,15 @@ export default function MultiPaymentModal ({ bounties }: {bounties: BountyWithDe
...
-            <MultiPaymentButton
-              chainId={gnosisSafeChainId}
-              safeAddress={gnosisSafeAddress}
-              transactions={selectedApplicationIds.map(selectedApplicationId => applicationTransactionRecord[selectedApplicationId])}
-              onSuccess={onPaymentSuccess}
-              isLoading={isLoading}
-            />
-
-            <Typography color='secondary' variant='caption'>Safe address: {shortenHex(gnosisSafeAddress)}</Typography>
+            {(gnosisSafeChainId && gnosisSafeAddress) ? (
```
mattcasey, line 277: we should always show this button. it includes a tooltip asking the user to connect their gnosis safe if it's not connected

mattcasey (changes requested):
see comments

components/bounties/components/MultiPaymentModal.tsx
Devorein: Updated.

components/bounties/components/MultiPaymentModal.tsx
Devorein: We are doing the same for bounty and proposal properties

components/bounties/components/MultiPaymentModal.tsx
Devorein: Updated.

components/bounties/components/MultiPaymentModal.tsx
Devorein: Updated. Previously, we used the same logic to show the multi-payment button. I've removed the extra disabled button.

PR #983, opened 2022-09-26
Seed pages

motechFR:
This pull request allows us to generate an arbitrary number of pages with actual page content inside a space.

We can also generate X amount of boards each containing X amount of cards.

This will come in very handy for performance testing.

To use this, run one of the 3 scripts in scripts/seedTestPages.ts

Nifty feature, you can pass a "nestedPercent" parameter to the seedTestPages fn so that X percent of pages have a parent page.

Screenshot from 2022-09-26 16-27-34

### DIFF
--- a/testing/generate-page-stub.ts
+++ b/testing/generate-page-stub.ts
@@ -1569,13 +1569,15 @@ export function pageContentStub (): Pick<Prisma.PageCreateInput, 'content' |'con
 /**
  * A page stub with a significant amount of content
  */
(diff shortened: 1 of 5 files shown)


msadura (approved):
This is just AMAZING!

testing/generate-page-stub.ts
```diff
@@ -0,0 +1,1605 @@
+import type { Prisma } from '@prisma/client';
```
mattcasey, line 1: @motechFR please use camelcase

PR #988, opened 2022-09-26
adding different stage builds for the app. 

yiwork:
added prisma to production as that is needed to build prisma db files


package.json
```diff
@@ -15,9 +15,9 @@
     "migrate:test": "dotenv -e .env.test.local -- npx prisma migrate deploy",
     "migrate:prod": "prisma migrate deploy",
     "migrate:staging": "npx dotenv -e .env.staging -- prisma migrate deploy",
-    "server:dev": "next dev",
     "debug:e2e": "dotenv -e .env.test.local -- npx playwright test __e2e__/public-page.spec.ts --headed --timeout=0",
     "notifications:debug": "ts-node ./scripts/readNotifications.ts",
+    "server:dev": "npx prisma migrate dev && next dev",
```
mattcasey, line 20: This command is aliased as 'npm start'... I prefer we don't call prisma every time we use it

PR #989, opened 2022-09-26
Accelerate jest further

motechFR:
Some further attempts to speed up tests


msadura (approved):

mattcasey (approved):

mattcasey:
I merged main into this branch and re-install node_modules, but now the tests are failing due to a weird error: [Jest worker encountered 4 child process exceptions, exceeding retry limit](https://stackoverflow.com/questions/70007464/jest-worker-encountered-4-child-process-exceptions-exceeding-retry-limit). It may be just a side effect of some test not passing? But if so, that means this change may make debugging tests much harder...

motechFR:
@mattcasey Im going to close out this PR for now as our tests are running pretty fast anyways, we can always reinvestigate later

PR #997, opened 2022-09-26
update to node 18

mattcasey:
Now's a good time to do it, since Node 16 will go into maintenance this year: https://github.com/nodejs/release#release-schedule

Node v18 announcement with 'whats new': https://nodejs.org/en/blog/announcements/v18-release-announce/


mattcasey:
DNS including 'localhost' now maps to the IPv6 address, not IPv4. Hence some integration tests were failing because the server doesn't spuport ipv6

msadura (approved):

PR #1000, opened 2022-09-27
Remove some types from models (deprecated) folder

mattcasey:
The models folder isn't really used anymore, and we usually put shared types inside of lib/<domain>/interfaces.ts.
This PR removes types for bounties and tasks from the folder.


motechFR (approved):
Looks good!

msadura (approved):

PR #1001, opened 2022-09-27
Feat/pages refactor

msadura:
Step 0 of reworking how pages work. This actually does not bring any optimisations yet, but in my opinion logic of updating / mutating pages via `setPages` all over the place was a bit overcomplicated. 2 palces still left with `usePages` as I didn't see clear way on removing it.

Additionally I removed explicit `pages` state in `usePages` - `swr` cache is a nice way of storing and updating the data vis mutations without need of having them duplicated in local state.

In final approach I would remove state mutators from `usePages` to external hook, but I didn't want to go too deep in single PR

@mattcasey @motechFR @Devorein let me know if you agree with proposed solution :)


hooks/usePages.tsx
```diff
@@ -154,12 +172,36 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
...
+    mutate(pagesData => {
+      if (pagesData) {
+        const udpatedData = { ...pagesData };
+        pageIds.forEach(id => delete udpatedData[id]);
+        return udpatedData;
+      }
+    }, { revalidate });
+  }, [mutate]);
+
+  const updatePage = useCallback(async (updates: PageUpdates, revalidateCache = false) => {
```
msadura, line 195: this will call api + mutate state

hooks/usePages.tsx
```diff
@@ -154,12 +172,36 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     }
   }
 
+  const mutatePage = useCallback((page: PageUpdates, revalidate = false) => {
```
msadura, line 175: this util can take page or partial page and will mutate local state - if page is now in cache it will get upserted

hooks/usePages.tsx
```diff
@@ -154,12 +172,36 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
...
+    mutate(pagesData => {
+      const currentPageData = pagesData?.[page.id];
+      if (pagesData) {
+        const updatedData: IPageWithPermissions = currentPageData ? { ...currentPageData, ...page } : page as IPageWithPermissions;
+        return { ...pagesData, [page.id]: updatedData };
+      }
+    }, { revalidate });
+  }, [mutate]);
+
+  const mutatePagesRemove = useCallback((pageIds: string[], revalidate = false) => {
```
msadura, line 185: helper for removing pages from local state - it can take array of strings for updating many pages in one call

msadura:
Overall changes:
- usePages() will now store pages data without `content` and `contentText` fields - this intriduces `PageMeta` type 
- new `usePageDetails` hook will fetch page details and take care of caching / udpates. In the future we might want to move more properties (like permissions)
- page content updates will mutate only `usePageDetails` data, not the general pages context, leading to faster udpates
- introduced content loader for first page details load
- introduced `hasContent` column in page - to keep track of content existence in PageMeta
- introduced `fallbackPreviewUrl` for cards - to be able to show card gallery preview without content field

Devorein (approved):
I tested out pretty much everything I could think found. Found no major issues, but a couple of places where we can make a few updates.

1. If I create a new page and add some content, it doesn't change the content icon on the left sidebar. Maybe we can update the page's hasContent field in the cache
2. I think making the charmEditor readOnly when there is no pageDetails inside DocumentPage would remove the sudden layout shift due to the circular loader. I've tested it out and its a better ux. First we would need to change the key of CharmEditor to be `key={page.id + editMode + !!pageDetails}`, then update readOnly prop to be `readOnly={readOnly || !pageDetails}` and we are done!
3. I see a lot of spaceId=undefined, when fetching the details of a page even when I'm inside a workspace. We can either remove it completely, as the data can be obtained from page.spaceId or pass it when there is one
4. type `PageUpdates` and `PageDetailsUpdates` have an extra `{id: string}` in their definition. The issue is that we are directly passing that inside `updatePage` to update the page, without first extracting the id. So a malicious user can update the id of a page 😬

Other than that its working like a charm. Amazing work🥳 Can't wait to see this in production

msadura:
1. `hasContent` cache is being updated, but indeed sidebar is not updated. I see this issue in prod app already, but I will see if it is an easy fix.
2. I will try your solution. I was trying to avoid rendering charm editor without need but if it will avoig page jump I will go with it. 
3. and 4. are the same way as they were in the past. 3. will be easy update so I will add it. as for 4. I will see, if it is a wuick update I will add it, otherwise I would do it in separate PR

charmClient/apis/pagesApi.ts
```diff
@@ -0,0 +1,26 @@
+import type { Page, ProfileItem } from '@prisma/client';
```
motechFR, line 1: Good idea! Liking this progressive addition of domain specific APIs and uncluttering of CharmClient

lib/pages/interfaces.ts
```diff
@@ -89,6 +89,17 @@ export type TargetPageTreeWithFlatChildren<T extends PageNode = PageNode> = {
   flatChildren: PageNodeWithChildren<T>[];
 }
 
-export type PageWithProposal = (Page & { proposal: ProposalWithUsers | null })
+export type PageMeta = Omit<IPageWithPermissions, 'content' | 'contentText'>
```
motechFR, line 92: W: Small suggestion; add a comment to explain this is a page without any content that we can use for page lists

lib/pages/interfaces.ts
```diff
@@ -89,6 +89,17 @@ export type TargetPageTreeWithFlatChildren<T extends PageNode = PageNode> = {
...
 
-export type PagesMap<P extends IPageWithPermissions | PageNode = IPageWithPermissions> = Record<string, P | undefined>;
+export type PageDetails = {
+  id: string;
+  content: Prisma.JsonValue | null;
```
motechFR, line 96: S: It's best to avoid leaky abstractions. In this case, since the interface is imported client-side, I'd avoid any reference to prisma. We could replace this by any or JSON depending on what you prefer.

lib/pages/server/getAccessiblePages.ts
```diff
@@ -22,6 +25,52 @@ export function includePagePermissions (): Prisma.PageInclude & {
   };
 }
 
+function selectPageFields () {
+  const permissions = {
```
motechFR, line 29: The includePagePermissions utility on l.13 of this file can generate this for you :)

hooks/usePageDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const updatedDetails: PageDetails = {
+      id: pageId,
+      content: updatedPage.content,
+      contentText: updatedPage.contentText
+    };
+
+    const hasContent = !checkIsContentEmpty(updatedPage.content as PageContent);
+
+    // Update pages context data only when hasContent value changed
+    mutate<PagesMap<IPageWithPermissions>>(`pages/${currentSpace?.id}`, pages => {
```
motechFR, line 34: W: It wasn't initially clear how this related to usePages.

I think it would be good to have a function that generates the cache key and is exported from usePages

This would make it clear that we are mutating a cache built there

lib/pages/addPage.ts
```diff
@@ -78,8 +78,8 @@ export async function addPage ({ createdBy, spaceId, shouldCreateDefaultBoardDat
     }
   }
 
-  await mutate(`pages/${spaceId}`, (pages: Page[]) => {
-    return [...pages, newPage];
+  await mutate(`pages/${spaceId}`, (pages: Record<string, Page>) => {
```
motechFR, line 81: 3rd instance of using this cache key, definitely worthy of an explicit function that produces it and will make the dependency clear :)

hooks/usePageDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const currentPageData = pages?.[pageId];
+      if (currentPageData && currentPageData.hasContent !== hasContent) {
```
motechFR, line 36: Shouldn't this also be responsible for initially setting the page inside the usePages SWR cache? I think it would be cleaner so we don't directly mutate a page in usePages

msadura: but this way we keen usePages cache separated, and we update it only when needed (when hasContent chganges). Thanks to that, we do not rerender whole usePages context on every page update. I will post video to demonstrate the difference

components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -33,9 +34,9 @@ export default function EditorPage ({ pageId }: { pageId: string }) {
       setIsAccessDenied(false);
       if (pageId && pagesLoaded && space) {
         try {
-          const page = await charmClient.getPage(pageId, space.id);
+          const page = await charmClient.pages.getPage(pageId, space.id);
           if (page) {
-            setPages((_pages) => ({ ..._pages, [page.id]: page }));
+            mutatePage(page);
```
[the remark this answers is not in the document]
Comment (msadura) [reply components/[pageId]/EditorPage/EditorPage.tsx:39]: yeah exactly, I used it initially, but then I relized decoupling them wil llead to less rerender.

lib/pages/addPage.ts
msadura: sounds like a good idea

hooks/usePageDetails.tsx
msadura: yeah I will do that, and I would even add it to a separate hooks to not mess with usePages too much

lib/pages/server/getAccessiblePages.ts
msadura: haha true thx

mattcasey:
image

This is a screenshot of the test data. We need to make sure that the images are properly attributed to each gallery card, and I guess we'll need a migration script to add them to existing pages

prisma/schema.prisma
```diff
@@ -245,22 +246,23 @@ model Page {
...
+  updatedAt          DateTime  @default(now())
+  updatedBy          String    @db.Uuid
+  spaceId            String    @db.Uuid
+  parentId           String
+  rootId             String    @db.Uuid
+  schema             Int
+  type               String
+  title              String
+  fields             Json
+  fallbackPreviewUrl String?
```
mattcasey, line 262: this should go into 'fields', that is where all the content for blocks are stored, especially since it's specific to a certain type of block

components/common/BoardEditor/focalboard/src/mutator.ts
```diff
@@ -1,4 +1,5 @@
 import charmClient from 'charmClient';
+import { PageMeta } from 'lib/pages';
```
mattcasey, line 2: should use type here, lib/pages exports a lot of stuff including bangle.dev library

scripts/updatePageHasContent.ts
```diff
@@ -0,0 +1,26 @@
...
+      id: true,
+      content: true
+    }
+  });
+
+  const pagesWithContent = pages.filter(p => !checkIsContentEmpty(p.content as PageContent))
+
+  console.log('🔥 Count of pages with content:', pagesWithContent.length);
+
+  await prisma.$transaction(pagesWithContent.map(({ id }) => prisma.page.update({
```
mattcasey, line 17: rather than doing thousands of queries in one go, i think it would be faster and safer to do one query? `{ where: { id: { in: pagesWithContent.map(p => p.id) } }`

lib/pages/server/getPageDetails.ts
```diff
@@ -0,0 +1,31 @@
+import type { Prisma } from '@prisma/client';
+import { prisma } from 'db';
+import { validate } from 'uuid';
+import type { PageDetails } from '../interfaces';
+
+export async function getPageDetails (pageIdOrPath: string, spaceId?: string): Promise<PageDetails | null> {
```
mattcasey, line 6: suggestion: rename to getPageContent?

msadura: tbh I called it `details` because I had in mind that we might port more stuff there (you mentioned permissions as an example)

hooks/usePageDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+import type { IPageWithPermissions, PageDetails, PageDetailsUpdates, PagesMap } from 'lib/pages';
+import { checkIsContentEmpty } from 'lib/pages/checkIsContentEmpty';
+import debounce from 'lodash/debounce';
+import type { PageContent } from 'models';
+import { useCallback, useEffect, useMemo } from 'react';
+import useSWR, { mutate } from 'swr';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+
+export function usePageDetails (pageIdOrPath: string, spaceId?: string) {
+  const { data: pageDetails, error, mutate: mutateDetails } = useSWR(pageIdOrPath ? [`pages/details/${pageIdOrPath}`, spaceId] : null, () => charmClient.pages.getPageDetails(pageIdOrPath, spaceId));
```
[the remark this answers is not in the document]
msadura: ah you might be right

prisma/schema.prisma
msadura: moved to Page under `galleryImg` column

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -501,13 +503,15 @@ function CharmEditor (
...
       }}
       placeholderComponent={(
-        <Placeholder
-          sx={{
-            // This fixes the placeholder and cursor not being aligned
-            top: -34
-          }}
-          show={isEmpty && !readOnly}
-        />
+        placeholder || (
```
msadura, line 506: I didn't see clear way of adding loader when pageDetails are loading, so I pass it as custom placeholder

msadura:
@Devorein thx for noted issues, everything got cleared out. Loading of content looks much better with your approach, thx!

@motechFR  @mattcasey  I think I also addressed most of things you mentioned.

`fallbackPreviewImg` was renamed to `galleryImg` and moved to `Page` column. I check need to double check if it is always properly filled when new page is created.

I also tested update scripts locally, they work as expected. I just hope it will be possible to run them on prod db with more records

prisma/schema.prisma
```diff
@@ -216,6 +216,8 @@ model Page {
   title              String
   content            Json?
   contentText        String
+  galleryImg         String?
```
mattcasey, line 219: can this be `galleryImg`? we have a table called `headerImage` on the same database

msadura: ok done

PR #1002, opened 2022-09-27
Fix comment menu

motechFR:
Fixing the menu so that user can't delete another user's comment

New logic:
- page.comment only needed to create a comment, resolve or delete threqd
- Editing or deleting a comment only possible on your own. Needs to be your own comment


msadura:
what kind of magic is that 😆 

Screenshot 2022-09-27 at 13 08 09

motechFR:
> what kind of magic is that laughing (0 files changed)
> 
> Screenshot 2022-09-27 at 13 08 09

Magic of the merge commit, I was surprised too!

msadura (approved):

PR #1003, opened 2022-09-27
Added link to the kanbanCard | Only usable on right click

valentinludu:
Added a link to wrap the kanbanCard, but only usable if you write click and open in new tab


msadura (changes requested):
Linking part works properly, that is nice 💪 

Unfortunately, original card look and behavior changes a bit:
- title is not in primary color instead of white (most likely due to href styling), which is undesired
- when you hover property row, property tooltip appears even I do not hover proerty itself (see video)

https://user-images.githubusercontent.com/5198394/192509306-448fff3c-2a98-41e9-ab08-63ac4bea33ce.mov

valentinludu:
@msadura
Fixed the issues
Fixed the pipeline

msadura:
Yeah looks and works fine now :) One thing that I would improve - `components/common/PageDialog/PageDialog.tsx` uses pretty much the same logic for getting page url. Would be nice to extract that logic to a hook and reuse it in both places

msadura (approved):

mattcasey (approved):

PR #1004, opened 2022-09-27
Add Publish to Snapshot in vote stage

Devorein:


components/votes/components/CreateVoteModal.tsx
```diff
@@ -253,17 +258,27 @@ export default function CreateVoteModal ({ open = true, onClose, onCreateVote, i
...
+            disabled={disabledSave}
+          >
+            Create
+          </Button>
+          or
+          {proposal?.status === 'reviewed' && (
+            <Tooltip title={!isProposalAuthor ? 'Only proposal authors can publish to snapshot' : ''}>
+              <div>
+                <Button disabled={!isProposalAuthor}>
+                  <PublishToSnapshot typography pageId={proposal.id} />
```
msadura, line 276: This is incorrectly composed. You added a `typography` prop which is a bit misleading, just to change the component which should be displayed inside of a button. Additionally this component will be rendered inside of a button with its own click handler which creates a bit of a mess.

Proper solution would be to add optional `renderContent` prop to PublishToSnapshot which would then pick onClick and label value and render a button.

Other, even easier would be to extract PublishToSnapshot's logicto hook and reuse it for for a button and in orignal component

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
```diff
@@ -13,7 +14,7 @@ import { usePopupState } from 'material-ui-popup-state/hooks';
 import { useEffect, useState } from 'react';
 import PublishingForm from './PublishingForm';
 
-export default function PublishToSnapshot ({ pageId }: {pageId: string}) {
+export default function PublishToSnapshot ({ pageId, typography = false }: { typography?: boolean, pageId: string}) {
```
msadura, line 17: This `typography` prop is really misleading and tell nothing about its purpose. What is the idea behind it?

components/votes/components/CreateVoteModal.tsx
Devorein: It would be difficult to use a hook as the component provides its own icons and has its own UI logic. But I can see the use case of using `renderContent` prop that can provide `onClick` and `text` content to the caller and its up to the caller to use which ever component to use

components/common/PageLayout/components/Header/components/Snapshot/PublishToSnapshot.tsx
Devorein: The component returns a react fragment to make it usable in any context. Its currently used in two places. One for the top-level actions you can take for a page by clicking on [...], another on inside a vote modal. For the actions one we want it to act like a menu item. but for votes, it should be a button, so their presentation and ux are different. At first, I used `useTypography`, but then I remembered react might mistake it for a hook 😛 but then I went with more ambiguous `typography` 🤦🏼 This was the easiest fix. Based on that prop we either use regular typography to render or menu/list items. Below is how it looks right now with the prop, the 2nd one without it. Not too much of a difference but its noticeable
image
image

msadura (approved):
Awesome refactor with renderContent! GJ!

PR #1006, opened 2022-09-27
Align kanban card title

valentinludu:
Checking if there are elements below or not in order to add that margin-bottom


components/common/BoardEditor/focalboard/src/components/kanban/kanbanCard.scss
```diff
@@ -67,11 +68,14 @@
     flex: 1 1 auto;
     font-weight: 600;
     line-height: 20px;
-    margin-bottom: 8px;
```
mattcasey, line 70: I noticed that there's 2 px of padding on top of 'octo-titletext' which makes the vertical margins uneven:
image
Maybe just have a default margin-bottom of 2px?

mattcasey (changes requested):
took a little too much spacing off the bottom

mattcasey (approved):

PR #1008, opened 2022-09-27
Use tx inside interactive prisma transactions

motechFR:
This ensures interactive transactions use the transaction client instead or Prisma directly

TODO: We should look at passing this to methods that have prisma internally and are consumed by transactions


Devorein:
I just search for `prisma.$transcation(async` in vscode and there seems to be a few other places where we are using `prisma` rather than `tx`. Even without considering the prisma usage by internal libraries, we should at least replace all the surface level ones.

Devorein (approved):

PR #1010, opened 2022-09-27
Use arrows and enter key in quick find

valentinludu:
What I did:
- use only one mouse hover. Before there were 2 types of hover.
- clicking anywhere inside the row of a tile will get you the desired page. Before only clicking on the left half of the container worked
- The use of down/up keys
- Added the option for the user to navigate with the down key and hit enter to go to that page
- For the above changes I am sending the user with the help of the onChange prop instead of using the link that was before. The only downside is that the user does not see the link on hover. If needed i can add it back with a preventDefault() -> I just didn't want to add a link that does not do anything.


components/common/PageLayout/components/SearchInWorkspaceModal.tsx
```diff
@@ -177,9 +174,10 @@ function SearchInWorkspaceModal (props: SearchInWorkspaceModalProps) {
                   <Stack>
                     <StyledTypographyPage>
                       {
-                        parts.map((part: { text: string; highlight: boolean; }, _index: number) => {
+                        parts.map((part: { text: string; highlight: boolean; }) => {
                           return (
                             <span
+                              key={part.text}
```
mattcasey, line 180: 'text' is not unique, pages can have the same title

mattcasey (changes requested):
i really like that you figured out how to fix pagination natively. i would have just stared by listening to keyboard events, etc :)

However i notice a couple issues: https://www.loom.com/share/6fc25fff298c4f1082136b411732c10a
1) up/down makes me jump all over the place sometimes. looking at the html (see below) I wonder if it has to do with the id's being out of order as well?
image

2) pages appear more than once, I discovered this bug existed before your PR but it might be part of the issue with #1.

components/common/PageLayout/components/SearchInWorkspaceModal.tsx
mattcasey, line 180: at first i thought that 'text' was the full page title, now i realize its just part of the mapped text. but i still think it could be duplicate, if any of the segments happen to be the same?

mattcasey (approved):

PR #1020, opened 2022-09-28
Fix login redirect to workspace

mattcasey:
When logging in thru Metamask, we dump you into the /signup page. I tried reallly hard to just 'fix' the current logic on the login page (pages/index.tsx), but the problem is that the wallet address appears way before the user in useUser() hook is updated. I even tried logging the user in, but it conflicted with RouteGuard logging the user in as well. So the easiest fix right now was to check on the /signup page if the user has a workspace to go to. This also will have the least affect on @motechFR 's PR to authenticate login.

Added playwright tests for login and signup using Metamask too.

### DIFF
--- a/__e2e__/login.spec.ts
+++ b/__e2e__/login.spec.ts
@@ -29,7 +29,7 @@ test('login - allows user to login and see their workspace', async () => {
   await page.goto(baseUrl);
 
   // should redirect to workspace
-  await page.waitForURL(`**/${space.domain}/${docPage.path}`);
-  await page.locator('text=[Your DAO]sdfdfddf Home').first().waitFor();
+  await page.waitForURL(`*/${space.domain}/${docPage.path}`);
+  await page.locator('text=[Your DAO] Home').first().waitFor();
 
 });
--- a/__e2e__/login.spec.ts
+++ b/__e2e__/login.spec.ts
@@ -1,34 +1,18 @@
 import { expect, test } from '@playwright/test';
 import { v4 } from 'uuid';
-import { baseUrl, mockWeb3 } from './utilities';
+import { baseUrl } from './utilities';
 
 test('login - sets a cookie inside the user browser', async ({ page }) => {
 
   // Arrange
+  const walletAddress = v4();
 
-  await mockWeb3(page, () => {
-
-    const walletAddress = '0xd73b04b0e696b0945283defa3eee453814758f1a';
-
-    // @ts-ignore - added by mockWeb3
-    Web3Mock.mock({
-      blockchain: 'ethereum',
-      accounts: {
-        return: [walletAddress]
-      }
-    });
+  // Act
+  await page.request.post(`${baseUrl}/api/profile`, {
+    data: {
+      address: walletAddress
+    }
   });
-
-  await page.goto('http://localhost:3000/');
-  await page.locator('text=Connect Wallet').click();
-  await page.locator('text=Connect Wallet').click();
-  await page.locator('text=Connected').waitFor();
-
-  // await page.request.post(`${baseUrl}/api/profile`, {
-  //   data: {
-  //     address: walletAddress
-  //   }
-  // });
 
   const cookies = await page.context().cookies();
 
--- a/__e2e__/login.spec.ts
+++ b/__e2e__/login.spec.ts
@@ -12,7 +12,7 @@ test('login - allows user to login and see their workspace', async () => {
 
   const sandbox = await browser.newContext();
   const page = await sandbox.newPage();
-  const { space, walletAddress } = await generateUserAndSpace();
+  const { space, page: docPage, walletAddress } = await generateUserAndSpace();
 
   await mockWeb3(page, { walletAddress }, context => {
 
@@ -29,7 +29,7 @@ test('login - allows user to login and see their workspace', async () => {
   await page.goto(baseUrl);
 
   // should redirect to workspace
-  await page.waitForURL(`**/${space.domain}`);
+  await page.waitForURL(`**/${space.domain}/${docPage.path}`);
   await page.locator('text=[Your DAO]sdfdfddf Home').first().waitFor();
 
 });
--- a/__e2e__/login.spec.ts
+++ b/__e2e__/login.spec.ts
@@ -28,8 +28,8 @@ test('login - allows user to login and see their workspace', async () => {
 
   await page.goto(baseUrl);
 
-  // should redirect to workspace
-  await page.waitForURL(`*/${space.domain}/${docPage.path}`);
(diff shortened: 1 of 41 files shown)


components/signup/SignupForm.tsx
```diff
@@ -43,14 +44,6 @@ export default function SignupPageContent () {
   const router = useRouter();
   const { user } = useUser();
 
-  async function createWorkspace () {
-    router.push('/createWorkspace');
```
mattcasey, line 47: Side cleanup: replaced onClick with href attributes. I don't know why we used onClick, so something's probably going to break of course.. but I have a test on it that shows it's working :)

pages/api/blocks/index.ts
```diff
@@ -2,7 +2,7 @@
 import type { Block, Prisma } from '@prisma/client';
 import { prisma } from 'db';
 import { InvalidStateError, NotFoundError, onError, onNoMatch, requireUser } from 'lib/middleware';
-import { getPagePath } from 'lib/pages';
+import { getPagePath } from 'lib/pages/utils';
```
mattcasey, line 5: Adding '/utils' to the path fixes the compile errors caused by importing bangle.dev (an esm module, I didn't want to go down the rabbit hole of trying to make it work)

__e2e__/login.spec.ts
```diff
@@ -1,22 +1,35 @@
...
   });
 
-  const cookies = await page.context().cookies();
+  await page.goto(baseUrl);
 
-  // Assert
-  expect(cookies.some(cookie => cookie.name === 'charm.sessionId')).toBe(true);
+  // should auto redirect to workspace
+  await page.waitForURL(`**/${space.domain}/${docPage.path}`);
+  await page.locator(`text=${docPage.title}`).first().waitFor();
```
motechFR, line 33: I like this method, nicer than previous isVisible I was doing

__e2e__/signup.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  await page.waitForURL('**/signup');
+  await page.locator('data-test=close-profile-avatar').click();
+  await page.locator('data-test=goto-create-workspace').click();
+
+  await page.waitForURL('**/createWorkspace');
+  await page.locator('data-test=workspace-domain-input').fill(uniqueDomain);
+  await page.locator('data-test=create-workspace').click();
+
+  await page.waitForURL(`**/${uniqueDomain}`);
+  await page.locator('text=[Your DAO] Home').first().waitFor();
```
motechFR, line 46: Looks good to me

__e2e__/utils.ts
```diff
@@ -154,23 +158,21 @@ export async function generateBounty ({ content = undefined, contentText = '', s
...
   const user = await createUserFromWallet(walletAddress);
 
   const existingSpaceId = user.spaceRoles?.[0]?.spaceId;
 
-  let space = null;
+  let space: Space;
 
   if (existingSpaceId) {
-    space = await prisma.space.findUnique({ where: { id: user.spaceRoles?.[0]?.spaceId }, include: { apiToken: true, spaceRoles: true } });
+    space = await prisma.space.findUniqueOrThrow({ where: { id: user.spaceRoles?.[0]?.spaceId }, include: { apiToken: true, spaceRoles: true } });
```
motechFR, line 173: Good spot

components/signup/SignupForm.tsx
```diff
@@ -65,24 +58,28 @@ export default function SignupPageContent () {
...
         <Grid item xs sx={{ height: '100%', margin: 'auto' }}>
-          <AdventureCard onClick={createWorkspace}>
-            <ImageContainer>
-              <Image src={rocketImage} />
-            </ImageContainer>
-            <PrimaryButton size='large'>
-              Create a new workspace
-            </PrimaryButton>
-          </AdventureCard>
+          <Link href='/createWorkspace'>
```
motechFR, line 61: Alot simpler, not sure at which point simple redirect became an onClick, good pattern to watch out for in future code reviews

lib/notion/importFromWorkspace.ts
```diff
@@ -11,7 +11,7 @@ import { createBoardView } from 'lib/focalboard/boardView';
 import { createCard } from 'lib/focalboard/card';
 import { MAX_IMAGE_WIDTH, MIN_IMAGE_WIDTH } from 'lib/image/constants';
 import log from 'lib/log';
-import { getPagePath } from 'lib/pages';
+import { getPagePath } from 'lib/pages/utils';
```
motechFR, line 14: What do you think of moving this function to lib/utilities/string since it's really a random page domain generator?

Would also avoid any future lib/pages accidental imports

pages/index.tsx
```diff
@@ -34,10 +35,11 @@ export default function LoginPage () {
       router.push('/nexus');
     }
     else if (spaces.length > 0) {
-      const isValidDefaultWorkspace = !!defaultWorkspace && spaces.some(space => defaultWorkspace.startsWith(`/${space.domain}`));
-      router.push(isValidDefaultWorkspace ? defaultWorkspace : `/${spaces[0].domain}`);
+      router.push(getDefaultWorkspaceUrl(spaces));
     }
     else {
+      // Note that a user logging in will be redirected to /signup, because the 'user' and 'spaces' are loaded async after the wallet address appears.
+      // TODO: Find a way to connect the state between hooks (wallet address and loaded user)
```
motechFR, line 42: I will see if this makes sense to address in the wallet PR Im working on

__e2e__/utils.ts
```diff
@@ -63,18 +66,19 @@ export async function getPages ({ browserPage, spaceId }: { browserPage: Browser
  */
 export async function createUserAndSpace ({
   browserPage,
-  walletAddress = v4(),
+  walletAddress = Wallet.createRandom().address,
   permissionConfigurationMode = 'collaborative'
 }: {
   browserPage: BrowserPage;
   walletAddress?: string;
-} & Partial<Pick<Space, 'permissionConfigurationMode'>>): Promise<{ user: LoggedInUser, space: Space, pages: IPageWithPermissions[] }> {
```
motechFR, line 71: W: I noticed you dropped the return type, I think it's better to have these as explicit so that future changes know what the return type should be

motechFR (approved):

lib/notion/importFromWorkspace.ts
mattcasey: Ya I don't see a strong argument either way. Am going to save that for future cleanup tho

__e2e__/utils.ts
mattcasey: I took it out because I usually see the return type by hovering over the method where it's called. Don't have a strong opinion so I'll add it back

PR #1021, opened 2022-09-28
Add signatures for wallet ID

motechFR:
In progress PR to require signatures and use them to login user & create token gate

- [X] Detect user with an active account but no authorisation signature
- [X] Sign a message
- [X] Store message in lit-compliant format

- [x] Clean UI that explains to user why they are being asked to sign
_Current prototype: A page that takes responsibility for calling the sign method_

- [x] Require wallet auth in registration w/ wallet
- [x] Require wallet auth in login w/ wallet

TBC
- [x] Find a way to connect the state between hooks (wallet address and loaded user)
_Comes from PR #1020 _

Stretch
- [x] Create token gate reusing our authsig
- [x] Verify yourself in a token gate reusing our authsig
- [x] Authenticate to a public bounty


mattcasey:
Please fix the layout of login:
image

components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx
```diff
@@ -103,3 +103,7 @@ export function BountySignupButton ({ bountyPage }: Props) {
     </>
   );
 }
+function lowerCaseEquals (arg0: string, arg1: string) {
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx:106]: i think u meant to delete this?

components/common/RouteGuard.tsx
```diff
@@ -1,26 +1,24 @@
...
+import { lowerCaseEqual } from 'lib/utilities/strings';
+import { useRouter } from 'next/router';
+import type { ReactNode } from 'react';
+import { useEffect, useState } from 'react';
 import type { UrlObject } from 'url';
-import log from 'lib/log';
 
 // Pages shared to the public that don't require user login
-const publicPages = ['/', 'invite', 'share', 'api-docs', 'u'];
+const publicPages = ['/', 'invite', 'share', 'api-docs', 'u', 'authenticate'];
```
mattcasey, line 15: is there going to be an '/authenticate' path?

hooks/useWeb3AuthSig.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const generated: AuthSig = {
+      sig: newSignature,
+      derivedVia: 'charmverse.sign',
+      signedMessage: body,
+      address: signatureAddress
+    };
+
+    setSignature(generated, true);
+
+    if (!litCommKey) {
```
mattcasey, line 118: please confirm that we need to set the commkey, i was under the assumption we only need to set the auth sig and that would greatly simplify the logic

lib/utilities/strings.ts
```diff
@@ -138,3 +138,7 @@ export function conditionalPlural ({ word = '', count = 1 }: { word: string, cou
   }
   return word;
 }
+
+export function lowerCaseEqual (firstString: string, secondString: string): boolean {
+  return firstString?.toLowerCase() === secondString?.toLowerCase();
```
mattcasey, line 143: firstString and secondString are always required, or you should fix the function arguments

package.json
```diff
@@ -72,8 +72,10 @@
     "@emotion/react": "^11.7.1",
     "@emotion/server": "^11.4.0",
     "@emotion/styled": "^11.6.0",
+    "@ethersproject/address": "^5.7.0",
```
mattcasey, line 75: i didnt see this lib actually used, do we need it?

mattcasey (approved):
crazy amount of work, great job tackling this and so quickly

components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/BountyProperties/components/BountySignupButton.tsx]: Thanks for spotting!

components/common/RouteGuard.tsx
motechFR: Old leftover before I moved to using the login page :)

hooks/useWeb3AuthSig.tsx
motechFR: Just tested without the comms key pair, seems to work. Deleted as less is more!

motechFR:
Tests

- [X] Login in as first time user
_I'm asked to connect wallet
Sign auto pops up after
Redirected to the signup_

- [X] Create a token gate on ETH Mainnet requiring address

- [X] Create a token gate on Polygon requiring address

- [X] Access join token gate as first time user who never used charmverse
Navigated to link
Redirected to login screen
Clicked connect wallet
After connect, sign pops up

**If clicking "deny" the error doesn't show (probably a binding issue inside automagic WalletSign button**

Redirected back to the join page
Asked to sign again

- [] Create a space after login
Stuck on the grey screen.
Clicking away and coming back fixes this

- [] Access public bounty board as user who never logged in or hand an account
Whole page re renders on connect wallet
Connection fails

lib/utilities/strings.ts
motechFR: Fixed

mattcasey:
please fix butto nlayout on mobile. they are supposed to both be 100%
image

motechFR:
> please fix button layout on mobile. they are supposed to both be 100% image

Fixed

.env.test.local
```diff
@@ -1,3 +1,4 @@
 DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/charmversetest
 DOMAIN=http://127.0.0.1:3335
-PORT=3335
\ No newline at end of file
+PORT=3335
+ENABLE_TEST_LOGIN=true
```
mattcasey, line 4: remember this probably needs to be added for all tests. thats why i thought NODE_ENV=test is good enough

motechFR: Went with NODE_ENV=test initially, but had too many challenges getting next.js to set it. That's why I went with this flag for now to get job done.

That said, we should definitely standardise in future

PR #1023, opened 2022-09-29
Fix Firefox drag and drop for kanban card

valentinludu:
For some reason, the a wrapped link with a prevent default & react-dnd library & Firefox don't go well with each other.
The onDrag works but onDrop does not.
I am making the assumption from what I've read in the issues column in react-dnd library, that onDrop is not working because the of the preventDefault.
This seems to be an issue just for Firefox.


mattcasey (approved):
works! it's fine not to use the common link here since we don't want to use it for *navigation* anyway

PR #1027, opened 2022-09-29
replace addresses with new wallet table

mattcasey:
This addresses the issue of users ending up with multiple accounts for the same wallet.

My migration script found about 20 users that have more than one account tied to the same address that also have a workspace.. but i'm just prioritizing by the user that joined first

I created this PR by removing 'addresses' from the prisma model, then generating types using `prisma generate`. Then I added it back just so things compile, because i'm not ready to drop the field. I'll run the migration script while this is deploying.

Man, the combination of prisma and Typescript made this much quicker than I thought :)


motechFR (approved):
@mattcasey Reviewed this across the board, and looks really good!

I searched for any inconsistencies but we're joining the wallet should be all fine. UserDetails also seems to cover the case where no wallets exist, so looks fine for me

scripts/migrateWallets.ts
```diff
@@ -0,0 +1,49 @@
...
+        const uniqueAddresses = user.addresses.filter(a => {
+          const used = !!addressMap[a];
+          if (!used) {
+            addressMap[a] = user.id;
+            return true;
+          }
+          return false;
+        });
+        if (uniqueAddresses.length) {
+          // await tx.userWallet.createMany({
```
motechFR, line 30: Small note, we'll probably need this uncommented for the script to run properly

mattcasey: yes you're correct :)

PR #1030, opened 2022-09-30
Pages list jumps to top when selecting a page

Devorein:


msadura:
In my honest opinion UX does not feel any better with this :( Sidebar jumps and dances, and it ends up in a different way it was before navigating anyway 😅 

What I would propose:
- remember scroll position (Element.scrollTop) of sidebar, right before navigating
- when navigation is done and sidebar is rendered, restore scroll immediately without animation
- hopefully it should look like it didnt move

If you would need a hand with this solution let me know :)

Devorein:
That's a far better solution. I will try to implement that right away.

msadura:
Solution looks fine from code perspective and does the job on the UI side, nice job!

It just made me thinking... why do we need it? I mean according to nextjs docs: https://nextjs.org/docs/basic-features/layouts#per-page-layouts layouts are preserved between page renders and scroll should just persist. Maybe we have an issue somewhere else in layout handling? This might deserve another card for the future. Let me know what you think @mattcasey @motechFR @Devorein

mattcasey:
@msadura when you request changes please mark 'request changes'

mattcasey:
@msadura we do use that method from Next.js. I still think they only support scrolling at the document level, not components.

Devorein:
No unfortunately `scroll='false'` doesn't work :(

PR #1031, opened 2022-09-30
Remove 0 from the start of max submissions

valentinludu:
TextField from MUI has an internal state to show the user his input when writing numbers in the input.
That means we don't need to handle a state that shows the value. Just need to show the first value the user sees when he opens the modal.


mattcasey (approved):
Nice! I really appreciate that in these past couple PR's you've found a fix that involved relying on the libraries/environments rather than patching around them

valentinludu:
> Nice! I really appreciate that in these past couple PR's you've found a fix that involved relying on the libraries/environments rather than patching around them

I appreciate it Matt.

I always try to refactor as less code as possible. 

Also, this bug I resolved it before in another project os it was familiar :P

PR #1035, opened 2022-10-01
collapse columns on mobile layout

mattcasey:


msadura (approved):
I wish one day we will not see .bangle-* classes :D

PR #1036, opened 2022-10-01
Remove sourcePermission from pages payload

msadura:
PR removes `sourcePermission` from usePages payload. It was not used directly from that entity and could be safely omitted.
Thanks to this small change pages query is faster and payload is smaller.


lib/pages/server/getAccessiblePages.ts
```diff
@@ -24,6 +26,23 @@ export function includePagePermissions (): Prisma.PageInclude & {
   };
 }
 
+export function includePagePermissionsMeta (): { permissions: { select: PermissionsSelect } } {
+  return {
+    permissions: {
+      select: {
+        pageId: true,
```
msadura, line 33: omitted 'inheritedFromPermission' field for smaller payload

mattcasey (approved):

PR #1038, opened 2022-10-03
Add edit button for share pages

valentinludu:
For the shared pages, I added a button for the user that is logged in and has permissions to edit that page. The button redirects the user to that specific page where he can edit it.

Here I am assuming 2 things:
- that the edit page has the same url without '/share'
- I get the space for that user and check if the user has admin rights


mattcasey (commented):
The link works! I have a few changes:
- admin access is too low in scope. Try using `getPagePermissions` from the usePages hook to determine if user has "edit_content" = true
- use 'href' instead of an onclick handler
- this button doesn't logically belong in "Account" component. I would either move the logic to PublicPage or put it in its own file.
- I'd like to make the button more obvious. Please try the standard Button component (maybe with props: `color='secondary' size='small' variant='outlined'`) and place it to the left of the dark/light mode toggle.
- can you add fontSize='small' to the dark/light mode toggle to match the regular view?

mattcasey (changes requested):
see my comment

valentinludu:
> The link works! I have a few changes:
> 
> * admin access is too low in scope. Try using `getPagePermissions` from the usePages hook to determine if user has "edit_content" = true
> * use 'href' instead of an onclick handler
> * this button doesn't logically belong in "Account" component. I would either move the logic to PublicPage or put it in its own file.
> * I'd like to make the button more obvious. Please try the standard Button component (maybe with props: `color='secondary' size='small' variant='outlined'`) and place it to the left of the dark/light mode toggle. (This button supports an href attribute too)
> * can you add fontSize='small' to the dark/light mode toggle to match the regular view?

Hey Matt,

I wasn't sure where to put it exactly at first. So yes, PublicPage is a better place.
I just redid it like you suggested.
 I still think it looked good without the outline.

Tell me what you think.

edit-btn

mattcasey (approved):
thanks for fixing

mattcasey:
btw it's fine if you take out the outline, we can get more feedback once it's in prod

PR #1041, opened 2022-10-03
remove addresses column on users

mattcasey:
The output of the script looks right to me. We could simply mark everyone deleted, but my OCD says just delete them since we won't ever have this issue again in the future:
```
Addresses with multiple users 143
Affected users 314
Unresolved situations 0
Users to "delete" 171
Users to mark deleted 26
Users to delete because they have no workspace 145
```


scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
-          //   data: uniqueAddresses.map(address => ({ address, userId: user.id }))
-          // });
-        }
-        else {
-          console.log(`Workspace User ${user.id} has no unique addresses: ${user.addresses.join(', ')}`);
-        }
-      }
+  const addressMap = users.reduce<{ [address: string]: typeof users }>((acc, user) => {
+    user.addresses.forEach(a => {
+      acc[a] = acc[a] || [];
```
motechFR, line 19: We wouldn't expect to have duplicate addresses, but if we do, wouldn't this negate the following push statement?

scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
+      acc[a].push(user);
+    });
+    return acc;
+  }, {});
+
+  const addresses = Object.entries(addressMap).filter(([address, _users]) => _users.length > 1);
```
motechFR, line 25: Goal:
Return addresses where there is more than 1 user

scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
+  const duplicateAccounts: typeof users = [];
+
+  // find cases where one user already has a wallet associated - we can mark all the others 'deleted'
+  const unresolved = addresses.filter(([address, _users]) => {
+    const usersWithWallets = _users.filter(u => u.wallets.length > 0);
+    if (usersWithWallets.length === 1) {
+      duplicateAccounts.push(..._users.filter(u => u.id !== usersWithWallets[0].id));
+      return false;
```
motechFR, line 33: Goal:

We found a user with this wallet, we can mark other accounts as duplicate

scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
     }
+    return true;
```
motechFR, line 35: Goal:

Return addresses where no corresponding user wallets were found

scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
+    });
+    return acc;
+  }, {});
+
+  const addresses = Object.entries(addressMap).filter(([address, _users]) => _users.length > 1);
+  const duplicateAccounts: typeof users = [];
+
+  // find cases where one user already has a wallet associated - we can mark all the others 'deleted'
+  const unresolved = addresses.filter(([address, _users]) => {
+    const usersWithWallets = _users.filter(u => u.wallets.length > 0);
```
motechFR, line 30: For safety (even though I don't think we have multiple wallets)

u.wallets.some(w => w.address === a)

scripts/migrateWallets.ts
```diff
@@ -1,47 +1,105 @@
...
+    // only one user has a workspace
+    if (withSpace.length === 1) {
+      duplicateAccounts.push(..._users.filter(u => u.id !== withSpace[0].id));
+      return false;
+    }
+    // none of the users have a workspace, just keep the first one
+    else if (withSpace.length === 0) {
+      duplicateAccounts.push(..._users.filter(u => u.id !== _users[0].id));
+      return false;
+    }
```
motechFR, line 50: Since the source data came from unresolved where the users did not have a wallet, should we also create the wallet here?

motechFR (approved):
@mattcasey looks reasonable. As I understand, we're deleting users with no workspace, and

scripts/migrateWallets.ts
mattcasey: i only care if they have any wallets, which i think is more cautious than checking individual addresses

scripts/migrateWallets.ts
mattcasey: found a user with *any* wallet

scripts/migrateWallets.ts
mattcasey: no, we can just delete them.

mattcasey: actually i think you may be right, i should create a wallet (or just delete these users...)

PR #1042, opened 2022-10-03
add eslint rule and sort imports

mattcasey:
- i added "import/internal-regex" to capture lib, pages, etc. other internal folders which would otherwise be categorized as 'external'
- i chose to alphabetize the imports too, this way there is exactly one 'correct' position for any import relative to the others
- i chose not to put type imports in their own group, altho that is an option
- imports that don't get an assignment are not auto-sorted (ex: `import '@testing-library/jest-dom';`) because technically we might need these to import before/after some other dependency. This required some manual fixes


msadura (approved):
💪

PR #1043, opened 2022-10-04
Remove exported test files

msadura:
This will clean files that are created during tests related to exporting pages to files


mattcasey:
Thanks, but I think we should also add these to .gitignore somehow. The test runner could blow up due to an exception or memory error, and the cleanup might not run.
I think it's better if the test files also got written to something like a .test folder, or we didn't test writing to disk at all.

msadura:
> It would also be good to add these to .gitignore somehow, or just don't even write to disk when tests are run (stub out fs?). It's possible the test runner fails due to memory leaks or something like that and cleanup doesn't get run

@mattcasey you are genius! Instead of cleaning up files I just mocked `fs` and updated tests :)

mattcasey (approved):

PR #1045, opened 2022-10-04
Bugfix/add new space

msadura:
Fix for blank page when creating workspace from sidebar.

Additional stuff:
- `useSpaces` will now return object instead of an array
- `createNewSpace` - moved to `useSpaces` to have it in a single place to reuse


hooks/useSpaces.tsx
```diff
@@ -33,7 +40,19 @@ export function SpacesProvider ({ children }: { children: ReactNode }) {
...
 
-  const value = useMemo(() => [spaces, setSpaces, isLoaded] as IContext, [spaces, isLoaded]);
+  const createNewSpace = useCallback(async (newSpace: Prisma.SpaceCreateInput) => {
+    const space = await charmClient.createSpace(newSpace);
+    setSpaces((s) => [...s, space]);
+    // refresh user permissions
+    const _user = await charmClient.getUser();
+    setUser(_user);
+    // give some time for spaces state to update or user will be redirected to /join in RouteGuard
+    setTimeout(() => {
```
msadura, line 50: lack of this timeout in one of the components caused the issue 😓

package.json
```diff
@@ -16,7 +16,7 @@
     "migrate:prod": "prisma migrate deploy",
     "migrate:staging": "npx dotenv -e .env.staging -- prisma migrate deploy",
     "server:dev": "next dev",
-    "debug:e2e": "dotenv -e .env.test.local -- npx playwright test --headed --timeout=0",
+    "debug:e2e": "dotenv -e .env.test.local -- npx playwright test **YOUR_TEST_PATH.spec.ts** --headed --timeout=0",
```
mattcasey, line 19: i think this makes the command useless? YOu can just add the file at the end. What i've been doing is:
```
npm run debug:e2e __e2e__/login.spec.ts
```
Should we add this to the README?

hooks/useSpaces.tsx
```diff
@@ -1,19 +1,26 @@
...
 import type { ReactNode } from 'react';
-import { createContext, useContext, useEffect, useMemo, useState } from 'react';
+import { useCallback, createContext, useContext, useEffect, useMemo, useState } from 'react';
 
 import charmClient from 'charmClient';
 
 import { useUser } from './useUser';
 
-type IContext = [spaces: Space[], setSpaces: (spaces: Space[]) => void, isLoaded: boolean];
+type IContext = {
```
mattcasey, line 10: thanks for the cleanup of array => map

__e2e__/sidebarAddNewSpace.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    // 1. Make sure there is a button to add a new workspace
+    const addNewSpaceBtn = page.locator('data-test=sidebar-add-new-space');
+    await expect(addNewSpaceBtn).toBeVisible();
+
+    // 2. Open create space dialod
+    await addNewSpaceBtn.click();
+
+    await expect(page.locator('data-test=create-space-form')).toBeVisible();
+
+    const nameInput = page.locator('data-test=workspace-name-input >> input');
```
mattcasey, line 59: For e2e, I want to be EXTREMELY minimal on what we are testing. Anything that can be done via Cypress or Jest should be done there instead, for example testing that the domain is updated when the name changes should be done in Cypress. I think we should remove all the logic around inputs in this test. Just open the modal and click 'create workspace' and make sure that we get to an actual page on the other side

mattcasey (changes requested):
one request: reduce the scope of the e2e test

package.json
msadura: ah I though you need manually put you filename in there 😅 I will fix that

msadura:
@mattcasey  updates pushed - removed input tests, added readme info

mattcasey (approved):

PR #1046, opened 2022-10-04
fix e2e tests

mattcasey:
First I tried to mock the wallet via web3, but finally came to the conclusion that this will never work unless we can mock the exact message that is sent to be signed, including the nonce, timestamp, etc. Otherwise, `verifyMessage(...)` returns a different address than the one that is used to mock the signature. I was able to generate a signature that is valid according to the server, at least!


__e2e__/importWorkspacePages.spec.ts
```diff
@@ -1,95 +0,0 @@
...
-import { baseUrl, createUserAndSpace } from './utils';
-
-let browser: Browser;
-
-test.beforeAll(async () => {
-  // Set headless to false in chromium.launch to visually debug the test
-  browser = await chromium.launch();
-});
-
-test('view a card created via workspace import', async () => {
```
mattcasey, line 17: i removed this test because it seems to be testing too many things. For example, we should be able to validate the data that importWorkspacePages returns from the backend only. Whether or not that data can be rendered by the UI is a separate concern.

.env.test.local
```diff
@@ -1,4 +1,4 @@
 DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/charmversetest
 DOMAIN=http://127.0.0.1:3335
 PORT=3335
-ENABLE_TEST_LOGIN=true
\ No newline at end of file
+APP_ENV=test
```
mattcasey, line 4: we cant set NODE_ENV because of next.js, but this is their recommendation.

__e2e__/importWorkspacePages.spec.ts
mattcasey: Happy to discuss f2f though

__e2e__/utils/web3.ts
```diff
@@ -0,0 +1,127 @@
...
+  context?: Partial<T>;
+  init (params: InitParams<T>): void;
+}
+
+type MockContext = { address: string, chainId?: number, privateKey: string };
+
+// load web3 mock library https:// massimilianomirra.com/notes/mocking-window-ethereum-in-playwright-for-end-to-end-dapp-testing
+export async function mockWeb3<T extends MockContext> ({ page, context, init }: MockWeb3Options<T>) {
+
+  const wallet = Wallet.createRandom();
```
mattcasey, line 68: to simplify things, always create a mock wallet and verify it. this might get trickier if we want to have a user w/multiple wallets or something.. but in the normal case, this will just take care of wallet verification w/o any extra work in the individual tests.

.env.test.local
motechFR: Yep sounds good, works great for the test environment-specific cases instead of individual bypasses

PR #1048, opened 2022-10-05
Fix sign up for discord

Devorein:


pages/index.tsx
```diff
@@ -59,8 +59,11 @@ export default function LoginPage () {
     // redirect user once wallet is connected
     if (isDataLoaded) {
       // redirect once account exists (user has connected wallet)
-      if (user && (isLogInWithDiscord
-        || (account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account)))) {
+      if ((user && (isLogInWithDiscord
+        || (account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account))))
+        // If the user has discord connected
+        || (user?.discordUser)
```
mattcasey, line 65: this could make it impossible to verify wallets in certain cases. we need to fix isLogInWithDiscord

components/common/RouteGuard.tsx
```diff
@@ -102,7 +102,7 @@ export default function RouteGuard ({ children }: { children: ReactNode }) {
       return { authorized: true };
     }
     // condition: no user session and no wallet address
-    else if ((!user && !account)) {
+    if (!user) {
```
mattcasey, line 105: i think we should keep 'else if' it is about readability: if/else if/ else if/ else if. it links all the conditions.

pages/index.tsx
```diff
@@ -56,21 +56,21 @@ export default function LoginPage () {
...
-    // redirect user once wallet is connected
+    // console.log('Data loaded', user, isDataLoaded)
+    // redirect user once they can be redirected
     if (isDataLoaded) {
       // redirect once account exists (user has connected wallet)
-      if (user && (isLogInWithDiscord
-        || (account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account)))) {
+      if ((user && ((account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account))))
+        // If the user has discord connected and no wallet
+        || (user?.discordUser && !user.wallets.length)
```
mattcasey, line 65: As we discussed, please check whether we are logging in via discord. checking 'static' data about the user is making too many assumptions.

pages/index.tsx
```diff
@@ -56,21 +56,21 @@ export default function LoginPage () {
...
       }
-      else if (account && walletAuthSignature && lowerCaseEqual(walletAuthSignature?.address as string, account)) {
-        loginUser();
-      }
       else {
         setShowLogin(true);
       }
     }
-  }, [account, walletAuthSignature, isDataLoaded]);
+  }, [account, walletAuthSignature, isDataLoaded, user]);
```
mattcasey, line 73: you shouldn't need to watch user, since isDataLoaded is set once the user is loaded already

pages/index.tsx
```diff
@@ -56,21 +56,21 @@ export default function LoginPage () {
...
   useEffect(() => {
 
-    // redirect user once wallet is connected
+    // console.log('Data loaded', user, isDataLoaded)
+    // redirect user once they can be redirected
     if (isDataLoaded) {
       // redirect once account exists (user has connected wallet)
-      if (user && (isLogInWithDiscord
-        || (account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account)))) {
+      if ((user && ((account && user.wallets.some(w => w.address === account) && lowerCaseEqual(walletAuthSignature?.address as string, account))))
```
mattcasey, line 63: I found the logic from before the wallet sig update: https://github.com/charmverse/app.charmverse.io/blob/83272d0fc27e23b9a146216fa87d11725593f302/pages/index.tsx#L51.
It was much simpler and I'd like to return it to that. For one, we were not even relying on discord status.
How about something like this:
```
const isLoggedIn = !!user;
const isVerificationStep = account && (!walletAuthSignature || !lowerCaseEqual(walletAuthSignature.address, account)

useEffect(() => {
  // redirect user once wallet is connected
    if (isDataLoaded) {
      // redirect once account exists (user has connected wallet)
      if (isLoggedIn && !isVerificationStep) {
        redirectUserAfterLogin();
      }
      else {
        setShowLogin(true);
      }
    }
}, []);
...
```

pages/index.tsx
```diff
@@ -56,21 +56,21 @@ export default function LoginPage () {
...
+        // If the user has discord connected and no wallet
+        || (user?.discordUser && !user.wallets.length)
+      ) {
         redirectUserAfterLogin();
       }
-      else if (account && walletAuthSignature && lowerCaseEqual(walletAuthSignature?.address as string, account)) {
-        loginUser();
```
mattcasey, line 67: i guess you already talked with Mo, i thought it was odd we removed this tho :)

mattcasey:
This became an urgent fix since users were not able to log in to the app, so I went ahead and deployed a fix already.

I addressed the failing test and applied some suggestions, kept them in a separate PR: https://github.com/charmverse/app.charmverse.io/pull/1051

PR #1051, opened 2022-10-05
Fix discord login

mattcasey:


pages/index.tsx
```diff
@@ -79,11 +80,7 @@ export default function LoginPage () {
   return (
     isLogInWithDiscord ? null : getLayout(
       <>
-        <LoginPageContent walletSigned={() => {
-          showMessage('Wallet verified. Logging you in', 'success');
```
mattcasey, line 83: i removed this message because it disappears as soon as we redirect. there's barely enough time to read it

PR #1053, opened 2022-10-06
Add e2e tests for log in and sign up thru Discord

mattcasey:
This was a sort-of deep rabbit hole. I wanted to reproduce as much of the flow as I could and avoid writing too much extra logic in our app to enable testing.

In the e2e test, I spin up a dummy server to mock the Discord API using Playwright fixtures. There's an example on Playwright doing exactly this: https://playwright.dev/docs/test-advanced, only I used KOA because it's already installed and makes it easy to add routes.

I also introduced the concept of page objects: https://playwright.dev/docs/pom. We used this at my last co and it made the e2e tests very legible. Playwright recommends adding these as fixtures as well.


msadura (approved):
really great job on this! Both server mocking and POMs looks really neat

PR #1055, opened 2022-10-06
Updated user sign/signup flow

Devorein:


lib/spaces/getSpacesPublicInfo.ts
```diff
@@ -0,0 +1,23 @@
+import type { Space } from '@prisma/client';
+
+import { prisma } from 'db';
+
+/**
+ * For now, it is acceptable to return the entire space document to unauthenticated users
+ * Supports lookup by space ID or space name or domain
+ */
+export async function getSpacesPublicInfo (spaceIdOrDomainOrPageId: string): Promise<Space[]> {
```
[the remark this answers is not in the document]
mattcasey: Also, it is never 'spaceId' so spaceIdOrDomainOrPageId is a bad variable name here

charmClient/charmClient.ts
```diff
@@ -273,8 +273,8 @@ class CharmClient {
     }));
   }
 
-  async getPublicSpaceInfo (spaceId: string): Promise<Space> {
-    return http.GET<Space>(`/api/spaces/${spaceId}/public`);
+  async getPublicSpacesInfo (search: string): Promise<Space[]> {
+    return http.GET<Space[]>('/api/spaces/public', { search });
```
mattcasey, line 277: 'public' doesnt make sense on its own, imo... it was like 'give me the public info of this space'. I think a better endpoint would be something like 'spaces/searchByTitle' and 'spaces/searchByDomain' or 'spaces/search?title=?domain=`?

lib/spaces/getSpacesPublicInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+  return prisma.space.findMany({
+    where: {
+      OR: [{
+        name: {
+          contains: spaceIdOrDomainOrPageId
+        }
+      }, {
+        domain: {
+          contains: spaceIdOrDomainOrPageId
```
mattcasey, line 18: domain must be an exact match

lib/spaces/getSpacesPublicInfo.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * For now, it is acceptable to return the entire space document to unauthenticated users
+ * Supports lookup by space ID or space name or domain
+ */
+export async function getSpacesPublicInfo (spaceIdOrDomainOrPageId: string): Promise<Space[]> {
+  return prisma.space.findMany({
+    where: {
+      OR: [{
+        name: {
+          contains: spaceIdOrDomainOrPageId
```
mattcasey, line 14: Please use Postgres text search, its much more powerful. You can just use `search: value`. see getAccessiblePages for an example.

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }, [spaces]);
+
+  useEffect(() => {
+    if (router.query.domain) {
+      setSpaceDomain(stripUrlParts(router.query.domain as string));
+    }
+  }, [router.query]);
+
+  const debouncedGetPublicSpaces = useMemo(() => {
+    return debounce((_spaceDomain: string) => {
```
mattcasey, line 46: theres no need to debounce search by domain. it should just be a one-time query

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+      setSpacesInfo([]);
+    }
+
+  }, [spaceDomain]);
+
+  function onChangeDomainName (event: ChangeEvent<HTMLInputElement>) {
+    setSpaceDomain(stripUrlParts(event.target.value));
+  }
+
+  const popperFactoryMemoized = useMemo(() => {
```
mattcasey, line 71: replace w/a styled component

mattcasey (changes requested):
i think it's worth changing os we have special requests on the frontend to search by title or domain

lib/spaces/getSpacesPublicInfo.ts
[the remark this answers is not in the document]
Devorein: Yep created two separate functions for two different purposes. One returning either the space or null when searching for exact match with the domain, the other returning an array when searching by the space name

charmClient/charmClient.ts
Devorein: Right I've updated the types and endpoints

lib/spaces/getSpacesPublicInfo.ts
Devorein: Updated

lib/spaces/getSpacesPublicInfo.ts
Devorein: Updated. Thank you.

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
Devorein: This was for searching by the space name, the argument name is a bit confusing since we supported both domain and title initially but fixed it now.

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+    router.push(`/${joinedSpace.domain}`);
+  }
+
+  useEffect(() => {
+    if (spaces.some(space => space.domain === router.query.domain)) {
+      router.push(`/${router.query.domain}`);
+    }
+  }, [spaces]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
Devorein: Updated.

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
Devorein: Updated.

Devorein:
I've tested it out on our staging environment, and everything works great, except for discord login (which I think we still don't support automatically)

PR #1061, opened 2022-10-07
Small fix - invites modal

valentinludu:
Fix for a console warning from React:
Warning: React does not recognize the `anchorEl` prop on a DOM element.

Looks like it is not a good idea to spread all the bindPopover props on a mui Modal. It also includes anchorEl and the Modal does not accept that prop.


mattcasey (approved):

mattcasey:
approved.. i wonder if we could be using bindPopper instead, that lib is built specifically for MUI and also has a 'bindMenu' option. however, i am find with the explicit approach!

valentinludu:
> approved.. i wonder if we could be using bindPopper instead, that lib is built specifically for MUI and also has a 'bindMenu' option. however, i am find with the explicit approach!

Looks like bindMenu and bindPopper still have the anchorEl tag exported.

I merged it like this. Will think of something in the future.

PR #1063, opened 2022-10-07
Update walletsign message

motechFR:
This is a more human friendly message.

Also added provisions so that old signatures are not returned anymore


motechFR:
Closing PR as it was accidentally not branched off main

PR #1064, opened 2022-10-07
Update wallet message

motechFR:
More human-friendly sign in message

Also added provisions so the getter method does not expose a wallet address if it is not valid

charmverse signature

### DIFF
--- a/patches/lit-siwe+1.1.8.patch
+++ b/patches/lit-siwe+1.1.8.patch
@@ -1,0 +1,24 @@
+diff --git a/node_modules/lit-siwe/dist/client.js b/node_modules/lit-siwe/dist/client.js
+index 38d7780..a4e3210 100644
+--- a/node_modules/lit-siwe/dist/client.js
++++ b/node_modules/lit-siwe/dist/client.js
+@@ -88,9 +88,9 @@ class SiweMessage {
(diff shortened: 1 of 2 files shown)


patches/lit-siwe+1.1.8.patch
```diff
@@ -0,0 +1,24 @@
...
+-        let prefix = [header, this.address].join("\n");
++        let prefix = [header, this.address].join("\n\n");
+         const versionField = `Version: ${this.version}`;
+         if (!this.nonce) {
+             this.nonce = (0, exports.generateNonce)();
+@@ -121,6 +121,7 @@ class SiweMessage {
+         if (this.statement) {
+             prefix += "\n";
+         }
++        console.log('Generating message', prefix)
```
mattcasey, line 21: console log

mattcasey (changes requested):
The cost of this update seems more than the value of a slightly different message. This looks like it would:
A) break validation of existing signatures
B) introduce a change in the message from the original RFC: https://eips.ethereum.org/EIPS/eip-4361#informal-message-template

We could use better message in in our UI than just "verify wallet," but I don't think patching a library and overriding the default message used in the RFC is a safe idea.  This is a minor improvement with unknown consequences.

PR #1066, opened 2022-10-08
check that users wallet is connected before testing token gate

mattcasey:
@Devorein i think you mentioned this 'bug' in discord. users that don't have a connected walelt should not be able to 'test' a token gate

### DIFF
--- a/components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
+++ b/components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
@@ -13,12 +13,13 @@ import { useWeb3React } from '@web3-react/core';
 import { checkAndSignAuthMessage, humanizeAccessControlConditions } from 'lit-js-sdk';
 import { useRouter } from 'next/router';
 import type { MouseEvent } from 'react';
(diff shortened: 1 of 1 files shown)


components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
```diff
@@ -178,9 +180,9 @@ export default function TokenGatesTable ({ isAdmin, onDelete, tokenGates }: Prop
                   />
                 </TableCell>
                 <TableCell align='center'>
-                  <Tooltip arrow placement='top' title={litClient ? 'Test this gate using your own wallet' : 'Lit Protocol client has not initialized'}>
+                  <Tooltip arrow placement='top' title={litClient ? (account ? 'Connect your wallet to test' : 'Test this gate using your own wallet') : 'Lit Protocol client has not initialized'}>
```
Devorein, line 183: It should be `!account`, since we want to show the `Connect wallet` message if you don't have any account connected.

Devorein (changes requested):
Just one small logic error. Everything else looks correct.

Devorein (approved):

PR #1068, opened 2022-10-09
add checks for signup redirect

mattcasey:


pages/index.tsx
```diff
@@ -31,9 +30,11 @@ export default function LoginPage () {
   const isLoggedIn = !!user;
   const walletIsVerified = !!account && !!walletAuthSignature && lowerCaseEqual(walletAuthSignature.address, account);
   const isVerificationStep = !!account && !walletIsVerified;
+  const hasCompletedSignUp = spaces.length && isSpacesLoaded;
 
   function redirectToDefaultPage () {
-    if (typeof router.query.returnUrl === 'string') {
```
msadura, line 36: you count use helper const like `const redirectTo = router.query.returnUrl;` (used 3 times)

pages/index.tsx
```diff
@@ -31,9 +30,11 @@ export default function LoginPage () {
...
+    // allow redirect url if user is going to a workspace (token gate)
+    if (typeof router.query.returnUrl === 'string' && (!hasCompletedSignUp || isSpaceDomain(router.query.returnUrl))) {
```
msadura, line 37: hmm if I understand this logic correctly it should be `&& (hasCompletedSignUp || isSpaceDomain(router.query.returnUrl)` ?

pages/index.tsx
```diff
@@ -31,9 +30,11 @@ export default function LoginPage () {
...
       router.push(router.query.returnUrl);
     }
     else if (defaultWorkspace === '/nexus') {
```
msadura, line 40: not sure if it would be part of this tasks, but this `/nexus` redirect is also error prone. when you log out of space A and then create a new account (with another wallet on the same browser) you will be redirected directly to `/nexus` without any workspace

msadura (changes requested):
I would consider rethinking signup flow  check a bit (but probably a case for another card). Currently we will redirect you to token gate page after creating a new account, but if token gate url was incorrect user will may end up stuck and clueless?

mattcasey:
> Currently we will redirect you to token gate page after creating a new account, but if token gate url was incorrect user will may end up stuck and clueless?

We do not, actually send you directly to the token gate from signup. What happens is users are being redirected to a content page inside a workspace, which does a 2nd redirect to the token gate. I am fine with that, it is a good separation of concerns IMO as you could already be logged in and we still need that logic. The only thing I'm going to be doing is skipping the redirect if it's to a non-workspace page

pages/index.tsx
mattcasey: ya i hate how this is written... if the user is being redirected, and they have not completed signup OR if this redirect is to an existing workspace, then redirect them. I think maybe i can change the order of conditions to make it more readable

pages/index.tsx
mattcasey: i'm going to address this in a separate PR, but @motechFR and I thought 'defaultWorkspace' should be namespaced to the user

0xepicode (approved):
Thx for clarifications!

msadura (approved):

PR #1069, opened 2022-10-10
Fix comment icon not visible on document pages on small screens

valentinludu:
Because of the min-width, on screens that have 1200-1300px width, comment icon is not visible on document pages.

I tested on multiple pages, but will appreciate if you someone else can test.

Also, I know on small screens a document page doesn't look too good, and I understand the reason for that min-width, but you can look at notion to see it has the same behaviour as the one suggested with this fix.

https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=10afbcef-beff-4143-bc98-b6e6652d16fe


mattcasey:
I recently added that min-width because, as you noticed, the document is unreadable at small screens when the sidebars are out. But Notion does the same thing (extremely squished content). However, I noticed that on mobile, we introduce a weird whitespace that causes the horizontal scrollbar to appear. That may be a different bug but can you take a look? See below:
image

mattcasey (approved):
removing min-width is fine, i think there's a different issue though that the comments suddenly fall off the page around 7-800px (see my comment). Can you try and fix it? I know our layout can be pretty nuts, so I approved the PR as it's not a dealbreaker

valentinludu:
> removing min-width is fine, i think there's a different issue though that the comments suddenly fall off the page around 7-800px (see my comment). Can you try and fix it? I know our layout can be pretty nuts, so I approved the PR as it's not a dealbreaker

Did some small change with big impact on small screens. Please tell me if still happens on Firefox or Chrome.

mattcasey:
@valentinludu now the margins are not even, and we lose some space on mobile. I'm going to ask Chris what he thinks we should do about mobile, I know that he is planning another round of mobile improvements.

mattcasey:
ok, just spoke w/Chris... let's remove the min-width, and he says he'd like to hide the comment icons on mobile. He'd also like that when you click the comments on mobile it triggers the sidebar instead, but ideally we only show the comment that you clicked.

mattcasey (approved):

PR #1075, opened 2022-10-11
Stop clearing input in join space autocomplete

Devorein:
image
Previously when the autocomplete textinput was out of focus, it was cleared. Thus the user had to write the workspace name again, which lead to suboptimal ux

### DIFF
--- a/components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
+++ b/components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
@@ -72,6 +72,9 @@ export function JoinDynamicSpaceForm () {
         }}
         getOptionLabel={(space) => space.name}
         fullWidth
+        filterOptions={(x) => x}
+        clearOnEscape={false}
+        clearOnBlur={false}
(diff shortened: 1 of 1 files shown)


components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
```diff
@@ -72,6 +72,9 @@ export function JoinDynamicSpaceForm () {
         }}
         getOptionLabel={(space) => space.name}
         fullWidth
+        filterOptions={(x) => x}
```
msadura, line 75: is `filterOptions` needed as well?

msadura (approved):

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
Devorein: These are two sources where I found the fix
https://github.com/mui/material-ui/issues/25096
https://mui.com/material-ui/react-autocomplete/#search-as-you-type

PR #1076, opened 2022-10-11
Feat/mixpanel events

msadura:
- a few more priority events
- fixed creating token gate with nested conditions  https://app.charmverse.io/charmverse/page-2737112508981192
- updated Space `updatedAt` field to update on save

### DIFF
--- a/lib/metrics/mixpanel/interfaces/PageEvents.ts
+++ b/lib/metrics/mixpanel/interfaces/PageEvents.ts
@@ -5,7 +5,7 @@ import type { BaseEvent, BaseEventWithoutGroup } from 'lib/metrics/mixpanel/inte
 
 import type { ResourceEvent } from './ResourceEvent';
 
-type PageEvent = ResourceEvent;
+export type PageEvent = ResourceEvent;
 type CustomPageType = 'settings' | 'proposals_list' | 'bounties_list' | 'nexus' | 'profile' | 'integrations'
 
 type CustomPageViewEvent = BaseEventWithoutGroup & {
@@ -14,10 +14,14 @@ type CustomPageViewEvent = BaseEventWithoutGroup & {
   type: CustomPageType;
 }
 
-type PageViewEvent = BaseEvent & {
+export type PageViewEvent = BaseEvent & {
   resourceId?: string;
   type: PageType;
 };
+
+export type PageWithNameEvent = PageEvent & {
+  pageName: string;
+}
 
 export interface PageEventMap {
   page_view: PageViewEvent | CustomPageViewEvent;
@@ -26,6 +30,8 @@ export interface PageEventMap {
   restore_page: PageEvent;
   export_to_markdown: PageEvent;
   edit_page: PageEvent;
+  create_page: PageWithNameEvent;
+  create_a_database: PageWithNameEvent;
 }
 
 export type PageEventNames = keyof PageEventMap;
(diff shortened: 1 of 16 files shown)


prisma/schema.prisma
```diff
@@ -147,7 +147,7 @@ model Space {
   deletedAt                   DateTime?
   createdAt                   DateTime                          @default(now())
   createdBy                   String                            @db.Uuid
-  updatedAt                   DateTime
+  updatedAt                   DateTime                          @default(now())
```
msadura, line 150: this was missing to update space `updatedAt` field on save.

pages/api/pages/index.ts
```diff
@@ -105,6 +107,9 @@ async function createPageHandler (req: NextApiRequest, res: NextApiResponse<IPag
         spaceId
       });
     }
+    else if (page.type === 'page' || page.type === 'board') {
+      trackPageAction(page.type === 'page' ? 'create_page' : 'create_a_database', { userId, pageId: page.id, withName: true, delay: 60000 });
```
mattcasey, line 111: what about creating other page types like bounties and proposals, do we track those someplace?

mattcasey: why is there a minute delay?

prisma/schema.prisma
mattcasey: does this update it any time we update the record?

mattcasey (approved):

pages/api/pages/index.ts
msadura: 1) I was asked for those 2 types for now
2) Delay is kinda workaround I saw somewhere in proposals - this event should send `pageName` but when you create new page, title will be empty... so we wait a minute assuming it is enough for user to add a title, then when event is being send, latest version of a page is queried from db

lib/metrics/mixpanel/interfaces/PageEvents.ts
```diff
@@ -14,18 +14,24 @@ type CustomPageViewEvent = BaseEventWithoutGroup & {
   type: CustomPageType;
 }
 
-type PageViewEvent = BaseEvent & {
+export type PageViewEvent = BaseEvent & {
   resourceId?: string;
```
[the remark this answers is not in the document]
msadura: pages like `settings` / `nexus` / `profile` etc will nt have any other resource id

lib/utilities/wait.ts
```diff
@@ -0,0 +1,4 @@
+export function wait (ms: number) {
+  // eslint-disable-next-line no-promise-executor-return
+  return new Promise(resolve => setTimeout(resolve, ms));
```
motechFR, line 3: Smooth and simple

pages/api/pages/index.ts
motechFR, line 111: I think it would be better to have a mapping from page type to track event type

prisma/schema.prisma
motechFR: What you just added will configure the record initial state.

For future updates,  @updatedAt is the prisma tool for auto generated date stamps on save
https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat

image

motechFR (changes requested):
@msadura looks good.

See my note re: Prisma for a fix to make for proper autosave

prisma/schema.prisma
msadura: interesting, I saw we have `@default(now())` in all `updatedAt` fields, while there is no `@updatedAt`, but feels like `@updatedAt` is a proper way

motechFR (approved):

PR #1077, opened 2022-10-11
Fix token gates

motechFR:
Ensures the test button for token gates works

### DIFF
--- a/components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
+++ b/components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
@@ -9,8 +9,7 @@ import TableCell from '@mui/material/TableCell';
 import Tooltip from '@mui/material/Tooltip';
 import Typography from '@mui/material/Typography';
 import type { TokenGate } from '@prisma/client';
-import { useWeb3React } from '@web3-react/core';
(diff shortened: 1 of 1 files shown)


Devorein (approved):

components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
```diff
@@ -100,8 +99,7 @@ export default function TokenGatesTable ({ isAdmin, onDelete, tokenGates }: Prop
       if (!litClient) {
         throw new Error('Lit Protocol client not initialized');
       }
-      const chain = getLitChainFromChainId(chainId);
-      const authSig = await checkAndSignAuthMessage({ chain });
+      const authSig = walletAuthSignature ?? await sign();
```
mattcasey, line 102: when would a user be able to have a wallet that isn't signed?

mattcasey (approved):

components/settings/invites/components/TokenGates/components/TokenGatesTable.tsx
mattcasey: i'm good leaving this in for now in case its needed in the future of course :)

motechFR: Good point! Doesn't hurt to have this in there in case authsig got cleared for wahetever reason.

PR #1081, opened 2022-10-11
disable click listener on sidebar when its invisible

mattcasey:
For some reason, ClickAwayListener is causing the CenterPanel component to re-render all the time. We can fix this by disabling the mouseEvent that gets triggered, and I wonder if there's something going on w/CenterPanel that we could also fix..


components/common/BoardEditor/focalboard/src/components/viewSidebar/viewSidebar.tsx
```diff
@@ -64,7 +64,7 @@ export default function ViewOptionsSidebar (props: Props) {
   const currentProperties = visiblePropertyIds.filter(id => props.board.fields.cardProperties.some(c => c.id === id)).length;
 
   return (
-    <ClickAwayListener onClickAway={props.closeSidebar}>
+    <ClickAwayListener mouseEvent={props.isOpen ? 'onClick' : false} onClickAway={props.closeSidebar}>
```
mattcasey, line 67: this is the one-line fix

motechFR (approved):
Approved!

PR #1082, opened 2022-10-12
Member directory v1

Devorein:


hooks/useMemberProperties.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const MemberPropertiesContext = createContext<Readonly<Context>>({
+  properties: undefined,
+  addProperty: () => Promise.resolve({} as any),
+  updateProperty: () => Promise.resolve({} as any),
+  deleteProperty: () => Promise.resolve()
+});
+
+export function MemberPropertiesProvider ({ children }: { children: ReactNode }) {
```
mattcasey, line 25: double check we need this globally

msadura: Moved to member directory page level, good call

Devorein:
1. I noticed that we have the `Status` column in the bounties nexus table, which seems more appropriate to be the action. It always shows `Review` as the action. Bounty have their own `status`, which is shown on the bounty's page
image
image
image

components/members/components/MemberDirectoryTableView.tsx
```diff
@@ -0,0 +1,154 @@
...
+                      return (
+                        <TableCell>
+                          <Typography variant='body2'>{member.properties.find(_property => _property.memberPropertyId === timezoneProperty?.id)?.value ?? 'N/A'}</Typography>
+                        </TableCell>
+                      );
+                    }
+                    case 'name': {
+                      return (
+                        <TableCell>
+                          <Link href={`/u/${member.path}`}>
```
mattcasey, line 97: member.path is optional, should be `member.path || member.id`

components/members/MemberDirectoryPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export default function MemberDirectoryPage () {
+  const { members } = useMembers();
+  const [searchedMembers, setSearchedMembers] = useState<Member[]>(members);
+  const { properties } = useMemberProperties();
+  const [currentView, setCurrentView] = useState<typeof views[number]>('gallery');
+  const [isPropertiesDrawerVisible, setIsPropertiesDrawerVisible] = useState(false);
+
+  return properties && searchedMembers ? (
+    <CenteredPageContent style={{
```
mattcasey, line 37: this component has a min-height of 80vh and it makes the page look huge. I'm not sure this page needs a min-height, or why there is one on CenteredPageContent :P it also seems unecessary on the proposals page.

image

components/members/MemberDirectoryPage.tsx
```diff
@@ -0,0 +1,93 @@
...
+const StyledButton = styled(Button)`
+  padding: ${({ theme }) => theme.spacing(0.5, 1)};
+
+  .Icon {
+    width: 20px;
+    height: 20px;
+  }
+`;
+
+const views = ['gallery', 'table'] as const;
```
mattcasey, line 27: For "post v1" we should use the URL To track which view you're on. view=gallery|table, and gallery could just be the default.

components/members/components/AddMemberPropertyButton.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const { mutateMembers } = useMembers();
+  const propertyNamePopupState = usePopupState({ variant: 'popover', popupId: 'property-name-modal' });
+  const [selectedPropertyType, setSelectedPropertyType] = useState<null | MemberPropertyType>(null);
+  const [propertyName, setPropertyName] = useState('');
+  const { properties, addProperty } = useMemberProperties();
+  const [propertyOptions, setPropertyOptions] = useState<PropertyOption[]>([]);
+
+  return (
+    <>
+      <Button
```
mattcasey, line 29: Add Property should be aligned left (notion example below)

image

components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -0,0 +1,122 @@
...
+}: {
+  member: Member;
+}) {
+  const { properties = [] } = useMemberProperties();
+  const timezoneProperty = properties.find(property => property.type === 'timezone');
+  const nameProperty = properties.find(property => property.type === 'name');
+
+  return (
+    <Link
+      href={`/u/${member.path}`}
```
mattcasey, line 24: use member.id as a fallback for memberpath

components/members/components/MemberDirectoryTableView.tsx
mattcasey: can we also add 'color=inherit'? the blue link color is unecessary

mattcasey (approved):
Approved but please fix the paths for members so that we use member.id as a fallback

components/members/components/MemberDirectoryTableView.tsx
Devorein: Alright updating.

components/members/MemberDirectoryPage.tsx
Devorein: Updated. Just required a few lines

components/members/components/AddMemberPropertyButton.tsx
Devorein: Updated. Thanks

components/members/MemberDirectoryPage.tsx
Devorein: This is a bit out of scope as the layout needs to change. A new ticket was created just for this

msadura: https://app.charmverse.io/charmverse/page-7098501094691398 - ticket with description for the issue

PR #1088, opened 2022-10-12
Fix discord-only user not able to visit profile

Devorein:


components/profile/components/UserDetails/UserDetails.tsx
```diff
@@ -94,7 +94,7 @@ function UserDetails ({ readOnly, user, updateUser }: UserDetailsProps) {
     if (user?.wallets) {
       types.push({
         type: IDENTITY_TYPES[0],
-        username: ENSName || user.wallets[0].address,
+        username: ENSName || user.wallets[0]?.address,
```
mattcasey, line 97: actually shouldnt we check for user.wallets.length on line 94?

scripts/updateUserIdentity.ts
```diff
@@ -27,7 +27,7 @@ import { prisma } from '../db';
     if (!user.wallets.length) {
       return;
     }
-    const address = user.wallets[0].address;
+    const address = user.wallets[0]?.address;
```
mattcasey, line 30: we already check for wallets length above

scripts/updateUserIdentity.ts
```diff
@@ -58,7 +58,7 @@ import { prisma } from '../db';
           data: {
             identityType: IDENTITY_TYPES[0],
             // @ts-ignore
-            username: user.ensName || shortenHex(user.wallets[0].address)
+            username: user.ensName || shortenHex(user.wallets[0]?.address)
```
mattcasey, line 61: uneccessary

PR #1091, opened 2022-10-13
Make entire area of new proposal button clickable

mattcasey:
Using MUI styles for the button UX: https://mui.com/material-ui/api/button-group/
image


motechFR (approved):
@mattcasey I also took the liberty of copying this over to the database pages

PR #1094, opened 2022-10-13
generate bounty csv data only when user clicks download

mattcasey:
Slight optimization, but should help the page load faster especially if you have many bounties


mattcasey:
Ugh, the library itself is broken... need sthis PR to be merged: https://github.com/react-csv/react-csv/pull/325

PR #1098, opened 2022-10-14
Gnosis multisign table

valentinludu:
- Added responsive table for Gnosis Table.

- Improved UX -> when you see My Nexus but you are not connected to Metamask, you see an Alert to connect to Metamask  instead of 'Import your Gnosis safes...'

### DIFF
--- a/components/nexus/GnosisTasksList.tsx
+++ b/components/nexus/GnosisTasksList.tsx
@@ -3,10 +3,21 @@ import ExpandLess from '@mui/icons-material/ExpandLess';
 import ExpandMore from '@mui/icons-material/ExpandMore';
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import PeopleIcon from '@mui/icons-material/People';
-import { Alert, Box, Card, Chip, Collapse, Divider, Grid, Typography } from '@mui/material';
+import Alert from '@mui/material/Alert';
+import Box from '@mui/material/Box';
+import Chip from '@mui/material/Chip';
+import Collapse from '@mui/material/Collapse';
+import Divider from '@mui/material/Divider';
+import Grid from '@mui/material/Grid';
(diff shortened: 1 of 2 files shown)


Devorein:
1. On a mobile screen, the cells are a bit squished. I believe attaching a `min-width: {sx: 150, md: 'inherit' }` to the `TableCell` will help.
image
2. I think we should add a bit more gap between each gnosis-safe table. Right now, the Gnosis safe label with the URL looks kind of like a footer of the above table.
image
3. There was a background color for each gnosis-safe table before, and a little bit less gap between each row.
image

valentinludu:
> 1. On a mobile screen, the cells are a bit squished. I believe attaching a `min-width: {sx: 150, md: 'inherit' }` to the `TableCell` will help.
> 2. I think we should add a bit more gap between each gnosis-safe table. Right now, the Gnosis safe label with the URL looks kind of like a footer of the above table.
> 3. There was a background color for each gnosis-safe table before, and a little bit less gap between each row.
.....

Hey Safwan,

Totally agree. 
1. I did the change for the bigger columns where there was these need of more spacing on mobile.
2. I added some margins between them so now they have more room to breath.
3. I think the background was not intended because it wasn't defined in sx prop. I think this was a default bg for the Card that was before. To be 100% sure, I will ask Chris today if he wants it or not. 
Also the spacing inside the cells that define the space between rows and cells are giving the table more space to breath. I discussed with Chris this last week and he totally agreed it is better then before. This spacing is visible on all tables in the Nexus.

mattcasey:
@valentinludu the background was intended... do we have a mockup somewhere of what our tables should look like?

valentinludu:
Discussed with Chris and Matt.

I ended up adding a grey hover state for all the new tables for consistency.

Devorein (approved):

PR #1106, opened 2022-10-19
adding tags to every resource in staging

yiwork:
So these resources wouldn't count as a datadog host in billing.


mattcasey (approved):

.cdk/staging-stack.ts
```diff
@@ -89,12 +89,20 @@ export class CdkDeployStack extends Stack {
       },
     ];
 
+    const resourceTags: cdk.CfnTag[] = [
+      { 
+        key: 'env',
```
mattcasey, line 94: is this supposed to be "Env"?

yiwork: So it sets the key name to be "env". key name can be anything right?

PR #1108, opened 2022-10-19
Nexus -> Keep votes visible after cast

valentinludu:
Show votes in the votes tasks list even if the user has voted.

https://app.charmverse.io/charmverse/page-025450911219631722


mattcasey (approved):
approved. what we really want to do is persist all previous votes, i'm going to check to see if we have a story for that.

mattcasey:
@valentinludu for the proper fix, check out where we retrieve tasks in pages/api/tasks/list.ts. I think getMentionedTasks and getVoteTasks could be updated to return a map of 'marked' and 'unmarked' (seen and unseen) tasks, just like getProposalTasks.

msadura (approved):

Devorein:
I found a few places where we might have some issues

1. Notification marker doesn't go away even after viewing/casting by vote
2. Once a user has cast a vote, maybe the CTA text should be `View` as well? The user would be able to see that the deadline hasn't passed.
3. Casting a vote in nexus doesn't update the poll percentage. It does close the modal, but opening it shows the same percentage as before.
4. I think we should return the votes based on the ascending order of deadline since you would want to see votes whose deadline is within 5 hrs on top of those votes whose deadline is five days later

valentinludu:
> I found a few places where we might have some issues
> 
> 1. Notification marker doesn't go away even after viewing/casting by vote
> 2. Once a user has cast a vote, maybe the CTA text should be `View` as well? The user would be able to see that the deadline hasn't passed.
> 3. Casting a vote in nexus doesn't update the poll percentage. It does close the modal, but opening it shows the same percentage as before.
> 4. I think we should return the votes based on the ascending order of deadline since you would want to see votes whose deadline is within 5 hrs on top of those votes whose deadline is five days later

Good point Safwan.

I just updated the PR with all of the fixes.

PS: Looks like the bugs were there before my PR, but I fixed them so we can test the Poll tab clean.

Devorein (approved):

PR #1112, opened 2022-10-20
Real time board updates

motechFR:
This sets up basic websocket connectivity and allows us to publish updates from one board to another.

Current product MVP: Card properties

Remaining technical issue:
- [ ] There seems to be more than 1 connection at a time on client side. This can be monitored by going to pages/api/socket.ts and uncommenting the line that auto-prints connections.
- [ ] We are sending the full block payload from the server, rather than the incremental update

For consuming websockets, I have added RxJS as Observables play really nicely with this sort of update streams.

The latest commit (bc68705) shows an example of how we extend this system across the stack for new event types. The case here is page metadata.


components/common/BoardEditor/focalboard/src/store/cards.ts
```diff
@@ -43,15 +43,17 @@ const cardsSlice = createSlice({
...
         }
         else if (card.fields.isTemplate) {
-          state.templates[card.id] = card;
+          const existingCard = state.templates[card.id] ?? {};
+          state.templates[card.id] = { ...existingCard, ...card };
         }
         else {
-          state.cards[card.id] = card;
+          const existingCard = state.cards[card.id] ?? {};
+          state.cards[card.id] = { ...existingCard, ...card };
```
mattcasey, line 56: Simpler version: `Object.assign(state.cards[card.id] || {}, card);`

lib/websockets/broadcaster.ts
```diff
@@ -0,0 +1,123 @@
...
+
+// Export the singleton instance
+declare global {
+  // eslint-disable-next-line no-var, vars-on-top
+  var relay: Broadcaster | undefined;
+}
+
+export const relay = global.relay ?? new WebsocketBroadcaster();
+
+global.relay = relay;
```
mattcasey, line 123: I don't think we can use a global with next.js, it casues memory leaks

hooks/useSocketClient.tsx
```diff
@@ -0,0 +1,141 @@
...
+});
+
+let socket: Socket;
+
+export function WebSocketClientProvider ({ children }: { children: ReactNode }) {
+
+  const [isConnected, setIsConnected] = useState(socket?.connected ?? false);
+  const [lastPong, setLastPong] = useState<string | null>(null);
+  const [messageLog, setMessageLog] = useState<LoggedMessage[]>([]);
+  const [eventFeed] = useState<EventFeed>({
```
mattcasey, line 48: this useState looks like it's not doing anything since it doesn't have a dependency, and will be re-run on every render. Are you meaning to cache it? Maybe you meant `useMemo(..., [])` (remember an empty array at the end)?

hooks/useSocketClient.tsx
```diff
@@ -0,0 +1,141 @@
...
+    block_updated: new BehaviorSubject<WebsocketMessage<'block_updated'> | null>(null),
+    page_meta_updated: new BehaviorSubject<WebsocketMessage<'page_meta_updated'> | null>(null)
```
mattcasey, line 50: I like this pattern w/RXjs. I've never used it before, but I think having TS-protection on the socket interfaces will make it more maintainable.

hooks/useSocketClient.tsx
```diff
@@ -0,0 +1,141 @@
...
+    });
+
+  }
+
+  function sendPing () {
+    socket.emit('ping');
+  }
+
+  function sendMessage (message: any) {
+    setMessageLog((prev) => [{ type: 'ping', message }, ...prev]);
```
mattcasey, line 115: This log is never purged and looks like it could really kill memory. Maybe only enable it in development?

hooks/usePages.tsx
```diff
@@ -209,6 +211,18 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     return freshPageVersion;
   }
 
+  useEffect(() => {
```
mattcasey, line 214: looks so simple :)

hooks/useSocketClient.tsx
```diff
@@ -0,0 +1,141 @@
...
+    block_updated: new BehaviorSubject<WebsocketMessage<'block_updated'> | null>(null),
+    page_meta_updated: new BehaviorSubject<WebsocketMessage<'page_meta_updated'> | null>(null)
+  });
+
+  const { user } = useUser();
+
+  useEffect(() => {
+
+    if (user) {
+      connect();
```
mattcasey, line 58: i believe this can just be called outside the hook (so it is only run once). It doesn't look like we actually need to wait for the user?
Also, this will re-connect any time anything about the user is changed, I think you'd at least want to just watch when user exists: `useEffect(..., [Boolean(user)])`

components/common/BoardEditor/FocalBoardProvider.tsx
```diff
@@ -12,13 +14,26 @@ import { initialLoad } from './focalboard/src/store/initialLoad';
...
+
   useEffect(() => {
     log.debug('Load focalboard data');
     if (space) {
       dispatch(initialLoad({ spaceId: space.id }));
     }
   }, [space?.id]);
 
+  useEffect(() => {
+    eventFeed.block_updated.subscribe({
```
mattcasey, line 28: even though this is probably never 'unmounted' i think it would be a good practice to unsubscribe in the useEffect callback...

it looks like you can do this:
```
useEffect(() => {
  const sub = eventFeed...subscribe();
  return () => {
     sub.unsubscribe();
  }
 , []);
 

```

hooks/useSocketClient.tsx
msadura: I am not a fan of rxjs and I think simple pub/sub would be enough for our needs. But since it looks simple I will not object :P 
One thing that looks weird in my opinion, is that generic socket client holds the list of all the events that you can subscribe to. I am not sure if that will be easy to maintain as it might grow fast.
Why don't we just have a list of possible events somewhere and then create a `BehaviorSubject` when some piece of code actually subscribes to it for the first time?

hooks/useSocketClient.tsx
msadura: since this feed is jsut for internal use and is not supposed to cause any renders it can be stored event  in `React.ref`

hooks/usePages.tsx
```diff
@@ -209,6 +211,18 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
...
+    eventFeed.page_meta_updated.subscribe({
```
msadura, line 215: hmm I think this is not a best solution - you tight subscribing closely with library picked for the solution. Lets say we would like to replace rxjs implementation with something else - all the code would need to be updated.

What I would propose: 
- expose `subscribe` (or `subscribeToFeed` / `subscribeToEvent`) method from `useWebSocketClient `. 
- subscribe would take `eventName` and `callback` function and under the hood it would just do something like:

```
eventFeed[eventName].subscribe({
   next: (value => {
       callback(value)
      })
})
```

additionally it would be nice to have specific value type for each eventName - I am not sure why such check is needed `typeof value?.payload.id === 'string'` ?

msadura (commented):
This looks really good! I would really like to have improved the way of subscribing to event feed, other than that it looks awesome!

hooks/usePages.tsx
mattcasey: I would +1 a simpler version that doesn't require rxjs. It's another 45kb minified - it's already a subdependency  of web3-react, but we will be replacing it with Rainbowkit eventually.

hooks/useSocketClient.tsx
mattcasey: Had the same feeling about rxjs and +1 your idea

hooks/usePages.tsx
msadura: In `useWebSocketClient`:

```
function subscribe(eventName, callback) {
eventFeed[eventName].subscribe({
   next: (value => {
       callback(value)
      })
})
}
```

in component where you want to subscribe:

`subscribe('page_meta_updated', someTypedCallbackFuntionThatWillReceiveValue)`

components/common/BoardEditor/focalboard/src/store/cards.ts
motechFR: Thanks, resolved

hooks/useSocketClient.tsx
motechFR: Migrated event feed to a useRef since we never change it after initialisation

hooks/useSocketClient.tsx
motechFR: Done team! We have gone with simple pub / sub and removed need for RxJS whilst still insulating consumers from Websocket logic. Thanks @msadura

hooks/useSocketClient.tsx
motechFR: Now limited to 50, and only pushes when env === 'development' :)

hooks/useSocketClient.tsx
motechFR: Updated this section so we reconnect & resubscribe on each space change.

components/common/BoardEditor/FocalBoardProvider.tsx
motechFR: Both useEffects now call unsubscribe in their unmounting phase

hooks/usePages.tsx
motechFR: This is now implemented, thanks both for the input

hooks/useSocketClient.tsx
```diff
@@ -0,0 +1,141 @@
...
+  sendMessage: () => null,
+  // Development only
+  messageLog: [],
+  clearLog: () => null,
+  eventFeed: {} as any
+});
+
+let socket: Socket;
+
+export function WebSocketClientProvider ({ children }: { children: ReactNode }) {
```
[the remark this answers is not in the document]
motechFR: Can do later, for now anyways this file does the job of insulating the rest of the client code from Websocket logic, which is good enough

lib/websockets/pubSub.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class PubSub<T, V> {
+  subscribers = new Map<T, CallbackType<V>[]>();
+
+  subscribe<E extends T = T, S extends V = V> (type: E, callback: CallbackType<S>) {
+    if (!this.subscribers.has(type)) {
+      this.subscribers.set(type, []);
+    }
+
+    this.subscribers.get(type)?.push(callback as any);
```
msadura, line 12: I realised one easy potential micro-imprvement here. subscribe could return:

`return () => this.unsubscribe(type, callback)` 

this way in place we subscribe for a feed we would have nice way of cleanup, for example:

```
  useEffect(() => {
    const unsubscribe = eventFeed.subscribe('block_updated', handleBlockUpdate);

    return () => unsubscribe();
  }, []);
  ```

File: hooks/useSocketClient.tsx [hunk 12]
```diff
@@ -0,0 +1,132 @@
...
+
+  function clearLog () {
+    setMessageLog([]);
+  }
+
+  const value: IContext = useMemo(() => ({
+    sendMessage,
+    messageLog,
+    clearLog,
+    eventFeed
```
msadura, line 121: This is nice, I am just wondering if we should expose whole `eventFeed` or just `subscribe / unsibscribe` methods. This way is probably more flexible, while having just `subscribe / unsubscribe` gives kinda shorter, no-brainer way of subscribing. I do not have strong opinion here so just dropping a comment.

hooks/usePages.tsx
```diff
@@ -209,6 +212,16 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
     return freshPageVersion;
   }
 
+  const handlePageUpdate = useCallback((value: WebsocketPayload<'page_meta_updated'>) => {
+    mutatePage(value, false);
+  }, []);
+
+  useEffect(() => {
+    eventFeed.subscribe<'page_meta_updated', WebsocketPayload<'page_meta_updated'>>('page_meta_updated', handlePageUpdate);
```
msadura, line 220: I am wondering if there would be a way for `subscribe` method to determine what will be payload for specific event instead of doing this explicitly. Something like in `lib/metrics/mixpanel/trackUserAction.ts` - you just pick an event and payload is typed automatically. This would be "nice to have" but not a priority.

msadura (approved):
I left 3 comments, but not a high priority, so please check them but I am approving anyway :)

lib/websockets/pubSub.ts
motechFR: Good one, added this in

hooks/useSocketClient.tsx
motechFR: Went a step further and combined this with the previous unsubscribe idea. We now expose only subscribe, which returns a bound unsubscribe method.

hooks/usePages.tsx
motechFR: This was on my todo list, all done now. Definitely a must have in my view when exposing events and payloads so that it's automagic for the developer

PR #1113, opened 2022-10-21
remove leftover websocket code

mattcasey:
Removes some websocket code I had added that we don't need, and also web socket code from focalboard


components/common/BoardEditor/focalboard/src/octoClient.ts
```diff
@@ -1,13 +1,21 @@
...
+export interface Subscription {
+  blockId: string;
+  workspaceId: string;
+  subscriberId: string;
+  blockType: string;
+  subscriberType: string;
+  notifiedAt?: number;
+  createdAt?: number;
+  deletedAt?: number;
+}
```
motechFR, line 18: I think we might not even need this either

motechFR (approved):
Looks good! Always love a nice negative diff PR

components/common/BoardEditor/focalboard/src/octoClient.ts
mattcasey: removed

PR #1116, opened 2022-10-22
include /members and /proposals pages as default redirect page

mattcasey:
The bug is that if you're on /<domain>/members, then go back to just /<domain>, we won't send you back to /members and we just send you to the first page in the workspace.

And out of scope, but it seems odd we have to maintain a whitelist at all. This was broken for /proposals too without anyone knowing.


msadura (approved):

msadura:
Agree it is weird such list is needed, probably because everything else is treated as "/[domain]" route?

PR #1117, opened 2022-10-23
Show member directory onboarding form on joining workspace

Devorein:


msadura (commented):
1) Onboarding modal appears only when I join workspace with existing user. When I create a new user and join it does not work
2) Product wise, I do not think that attacking user with some sort of modal form with any description what is this can be called "onboarding". I will bring this up on discord.

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useEffect } from 'react';
+
+import Modal from 'components/common/Modal';
+import { MemberPropertiesPopupForm } from 'components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import { useMemberPropertyValues } from 'hooks/useMemberPropertyValues';
+import { useUser } from 'hooks/useUser';
+import { setUrlWithoutRerender } from 'lib/utilities/browser';
+
+function MemberPropertiesOnBoardingForm ({ memberId, spaceId, onClose }: { onClose: () => void, spaceId: string, memberId: string }) {
```
mattcasey, line 12: I think it would be good to memoize this, since it appears on all pages. Or don't render it at all until the modal is open (you would check if isOpen is true before embedding the component)

components/common/TokenGateForm/JoinDynamicSpaceForm.tsx
```diff
@@ -29,7 +29,7 @@ export function JoinDynamicSpaceForm () {
   const [selectedSpace, setSelectedSpace] = useState<null | Space>(null);
 
   async function onJoinSpace (joinedSpace: Space) {
-    router.push(`/${joinedSpace.domain}`);
+    router.push(`/${joinedSpace.domain}?onboarding=true`);
```
mattcasey, line 32: Can you avoid using URL to pass 'state'. ?onbarding=true is not really part of navigation, it's just saying 'show the user a modal'. It seems like this var could be handled inside of a hook or using localStorage even.

pages/_app.tsx
```diff
@@ -235,7 +236,7 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
                                   handleClose={() => setIsOldBuild(false)}
                                 />
                                 {getLayout(<Component {...pageProps} />)}
-
+                                <MemberPropertiesOnBoardingModal />
```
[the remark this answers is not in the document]
Devorein: Updated thnx.

components/_app/GlobalComponents.tsx
```diff
@@ -1,12 +1,14 @@
 import HexagonalAvatarMask from 'components/common/HexagonalAvatarMask';
 import Snackbar from 'components/common/Snackbar';
-import SetAvatarPopup from 'components/profile/components/SetAvatarPopup/SetAvatarPopup';
+import { MemberPropertiesOnBoardingModal } from 'components/members/MemberPropertiesOnBoardingModal';
+import { useOnboarding } from 'hooks/useOnboarding';
 
 export default function GlobalComponents () {
+  const { onboarding } = useOnboarding();
```
msadura, line 7: please use proper naming for booleans. should be `isOnboardingOpen` or basically name that you can clearly understand it is a boolean value. `onboarding` is too broad.

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import Modal from 'components/common/Modal';
+import { MemberPropertiesPopupForm } from 'components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import { useMemberPropertyValues } from 'hooks/useMemberPropertyValues';
+import { useOnboarding } from 'hooks/useOnboarding';
+import { useUser } from 'hooks/useUser';
+import type { LoggedInUser } from 'models';
+
+function MemberPropertiesOnBoardingForm ({ user, spaceId, onClose }: { onClose: () => void, spaceId: string, user: LoggedInUser }) {
```
msadura, line 11: I am not sure what is the reason for splitting those 2 components `MemberPropertiesOnBoardingForm` and `MemberPropertiesOnBoardingModal` but if you need them separately please have them in 2 separate files. Having multiple components in 1 file is just getting too confusing

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { user } = useUser();
+  const { setOnboarding } = useOnboarding();
+
+  function onClose () {
+    setOnboarding(false);
+    memberDirectoryFormModal.close();
+  }
+
+  return (
+    <Modal size='large' open={true} onClose={onClose}>
```
msadura, line 37: This feels like a hack. MemberProperty form is based on modal already, but you wrap it in another modal instance. 
Maybe `MemberPropertiesPopupForm` should be refactored and `MemberPropertiesForm` (without popup) extracted to separate component and reused in both places

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
```diff
@@ -65,8 +71,16 @@ export function MemberPropertiesPopupForm ({ memberId, spaceId, updateMemberProp
 
   return (
     <Dialog open={!!spaceId} onClose={onClose} fullWidth>
-      <DialogTitle>Edit workspace profile</DialogTitle>
+      <DialogTitle>{title}</DialogTitle>
       <DialogContent dividers>
+        {showUserDetailsForm && user && (
```
msadura, line 76: This is just conceptually wrong. Since this is member properties form, why would you like more things to it and add more props and conditions. As I mentioned in previous comment - lets make properties form just a component to reuse, then you can reuse it in "onboarding modal" and user details will also belong to onboarding modal.

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,47 @@
+import { usePopupState } from 'material-ui-popup-state/hooks';
```
msadura, line 1: I believe onboarding is one word so file names should be `...OnboardingModal.tsx`, especially that useOnboardind uses this convention

components/_app/GlobalComponents.tsx
```diff
@@ -1,12 +1,14 @@
...
   return (
     <>
       <HexagonalAvatarMask id='hexagon-avatar' />
-      <SetAvatarPopup />
+      {onboarding && <MemberPropertiesOnBoardingModal />}
```
msadura, line 11: This is jut container component for other global components, so it should not have logic and checking if onboarding is opened should go to `MemberPropertiesOnBoardingModal` (just render null there if it is not opened)

`{onboarding && <MemberPropertiesOnBoardingModal />}`

msadura (changes requested):

components/members/MemberPropertiesOnBoardingModal.tsx
msadura: ah and this is actually `MemberOnboardingModal` since it is not only related to properties

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: `MemberPropertiesOnBoardingForm` was separated from its parent component as inside we are using `useMemberPropertyValues` hook which requires `userId` to be present. I kept them in the same module since they are quite small at the moment, but I will move them to separate modules :D

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: Yep. This was updated. Thank you. It caused a ghost modal to appear behind the onboarding form, now its been removed.

components/_app/GlobalComponents.tsx
Devorein: Agreed. I will make the changes.

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: Yes I agree `MemberOnboardingModal` is better

components/_app/GlobalComponents.tsx
Devorein: I had to change things a bit and updated it to be an object not just a `bool`. Now its containing the structure 

`onboarding-<userId>`: {
	[spaceId]: boolean
}

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
Devorein: Its was slightly tricky to move them in separate components as everything had to be placed inside a Dialog. I've used `children` prop instead

components/profile/components/UserDetails/SocialIcons.tsx
```diff
@@ -27,38 +28,47 @@ export function SocialIcons ({
...
-  const hasAnySocialInformation = (model: Social) => model.twitterURL || model.githubURL || model.discordUsername || model.linkedinURL;
+  const theme = useTheme();
 
   return (
     <Stack direction='row' alignItems='center' gap={2} {...props}>
-      {social && social.twitterURL && (
-        <Link href={social.twitterURL} target='_blank' display='flex'>
-          <TwitterIcon style={{ color: '#00ACEE', height: '22px' }} />
-        </Link>
+      {social && (
```
msadura, line 35: I believe you do not need to check `social &&` on each item, but you can check it once and do not render any of them

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import { useOnboarding } from 'hooks/useOnboarding';
+import { useUser } from 'hooks/useUser';
+
+import { MemberOnboardingForm } from './components/MemberOnboardingForm';
+
+export function MemberPropertiesOnBoardingModal () {
```
msadura, line 9: MemberPropertiesOnBoardingModal -> MemberPropertiesOnboardingModal + change filename

msadura: ps. why not simply member onboarding modal?

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const memberDirectoryFormModal = usePopupState({ variant: 'popover', popupId: 'member-directory-onboarding' });
```
msadura, line 10: this is not used and can be removed

components/members/MemberPropertiesOnBoardingModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+      hideOnboarding(space.id);
+      memberDirectoryFormModal.close();
+    }
+  }
+
+  if (!space || !user) {
+    return null;
+  }
+
+  if (!onboarding[space.id]) {
```
msadura, line 27: you can squeeze both conditions this into 1:

`!user || !onboarding[space?.id]`

msadura (changes requested):

components/profile/components/UserDetails/SocialIcons.tsx
Devorein: Updated thank you. The reason I had to have them for each media was because Alex said to show those icons even when they are not connected. I replaced `social && social.twitterURL` with `social?.twitterURL`

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: Lets go with that

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: Nice catch. Thank you :D

components/members/MemberPropertiesOnBoardingModal.tsx
Devorein: Updated

PR #1122, opened 2022-10-24
Disable notion import for non-admins

mattcasey:
Security feature requested by a user in Discord.

This PR also adds a new 'disabledTooltip' prop to the Button component, which makes it easy to add a message. This adds the Tooltip component to all our buttons, if there's a performance issue we could optimize it later.


motechFR (approved):
Looks good! For the security of API, 2 integration tests, one with a 200 and one with a 401 are always welcome (although requireSpaceMembership / admin : true) does the job for sure

mattcasey:
Added some tests, but making it return 200 would require overriding the calls to Notion api

PR #1124, opened 2022-10-24
update and patch library for prosemirror tables

mattcasey:
This lib got updated in another PR.. I decided to go with it and update our patch file rather than try to figure out why the library was getting updated.


msadura:
looks good but CI fails

msadura (approved):

PR #1128, opened 2022-10-25
Feat/member property permissions

msadura:
https://app.charmverse.io/charmverse/page-717681698804959

Property permissions:
- returning only accessigble props from api
- api for creating / removing proeprty permission by role
- UI updates
- tests

https://user-images.githubusercontent.com/5198394/197764800-25e000a6-55c5-4e7f-8e31-6f147f0a3819.mp4


prisma/migrations/20221024201445_member_property_permission/migration.sql
```diff
@@ -0,0 +1,27 @@
+-- CreateEnum
```
motechFR, line 1: General note re:migrations, if you iterated the data model multiple times in PR, its best to delete and recreate the migrations so we have a single migration file before merge

prisma/schema.prisma
```diff
@@ -811,3 +819,19 @@ model MemberPropertyValue {
 
   @@unique([memberPropertyId, spaceId, userId])
 }
+
+model MemberPropertyPermission {
+  id                            String                        @id @default(uuid()) @db.Uuid
+  userId                        String?                       @db.Uuid
+  spaceId                       String?                       @db.Uuid
```
motechFR, line 826: Here, you're using the spaceId key to group permissions. This is a bit confusing for developers, as we might think it's an assignable group who can receive permissions.

It's better to drop this field, and lookup the space of the linked Member property

lib/members/createMemberPropertyPermission.ts
```diff
@@ -0,0 +1,20 @@
+
+import { MemberPropertyPermissionLevel } from '@prisma/client';
+
+import { prisma } from 'db';
+import type { CreateMemberPropertyPermissionInput } from 'lib/members/interfaces';
+
+export function createMemberPropertyPermission (data: CreateMemberPropertyPermissionInput) {
+  return prisma.memberPropertyPermission.create({
```
motechFR, line 8: Security best practice: Add a check and error so we can only assign permissions to users and roles who are in the same space as this property.

lib/members/getAccessibleMemberPropertiesBySpace.tsx
```diff
@@ -0,0 +1,103 @@
...
+                userId
+              }
+            }
+          }
+        },
+        {
+          OR: [
+            // User has permission to view property
+            {
+              spaceId: { in: spaceIdQuery },
```
motechFR, line 35: After removing spaceId from MemberPropertyPermissions, you can update this to MemberProperty {spaceId}

lib/members/getAccessibleMemberPropertiesBySpace.tsx
```diff
@@ -0,0 +1,103 @@
...
+            }
+          }
+        }
+      }
+    }
+  });
+}
+
+export function accessiblePropertiesByPermissionsQuery ({ spaceIds, userId }:
+  { spaceIds: string[], userId: string }): Prisma.MemberPropertyPermissionListRelationFilter {
```
motechFR, line 81: Philosophically, there's only ever 1 possible spaceId for a specific set of MemberProperties, best to update this to single string

components/members/components/MemberDirectoryProperties/MemberPropertiesSidebar.tsx
```diff
@@ -160,25 +157,29 @@ export function MemberPropertySidebarItem ({
...
+                  color='secondary'
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    deleteConfirmation.open();
+                  }}
+                />
+              </Tooltip>
             )}
             <ConfirmDeleteModal
               title='Delete property'
```
motechFR, line 185: I'm not sure we need to have a Modal when deleting the permission. The UI immediately changes to "Everyone at X space"

motechFR:
@msadura This looks great, all our core paradigms have been ported over to Member properties.

For security, can you please add integration tests for the create and delete endpoints?

We need a 201 (admin user) and 401 (non admin user) for each endpoint.

Finally, for the list endpoint, can you copy / paste the tests from the getAccessibleProperties and adapt them to the server setup? You covered all the key cases, so no edits required

lib/members/getAccessibleMemberPropertiesBySpace.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const spaceIdQuery = typeof spaceId === 'string' ? [spaceId] : spaceId;
+
+  return prisma.memberProperty.findMany({
+    where: {
+      AND: [
+        // User must be a member of space
+        {
+          space: {
+            id: { in: spaceIdQuery },
```
motechFR, line 23: Is there a case where spaceId is actually an array of more than 1? If so, we should rename it to spaceIds, otherwise drop the in: query and ensure consumers pass a single spaceId

motechFR (approved):
Approved, but plz remember integration tests ;)

msadura:
@motechFR PR is ready for re-review.

-removed unised stuff from permissions model
- added business logic check before creating permission
- business logic tests and integrations tests for listing props with permissions, creating / deleting permissions

__integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    (await request(baseUrl)
+      .delete(`/api/spaces/${space.id}/members/properties/permissions`)
+      .set('Cookie', nonAdminCookie)
+      .send({
+        permissionId: permission.id
+      })
+      .expect(401)).body as MemberPropertyPermissionWithRole;
+  });
+
+  it('should return error for user from other space when trying to delete space', async () => {
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts:97]: text correction for better clarity

should return error for user from other space when trying to delete **permissions inside** the space

__integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+      .post(`/api/spaces/${space.id}/members/properties/permissions`)
+      .set('Cookie', nonAdminCookie)
+      .send({
+        memberPropertyId: property1.id,
+        roleId: role1.id
+      })
+      .expect(401)).body as MemberProperty[];
+  });
+
+  it('should return error for user from other space', async () => {
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts:59]: should return error for user from other space when they try to create a permission

__integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  const { user, space: generatedSpace } = await generateUserAndSpaceWithApiToken(undefined, true);
+  space = generatedSpace;
+  adminUser = user;
+  nonAdminUser = await generateSpaceUser({ spaceId: generatedSpace.id, isAdmin: false });
+  nonAdminCookie = await loginUser(nonAdminUser.id);
+  adminCookie = await loginUser(adminUser.id);
+  role1 = await generateRole({ spaceId: space.id, roleName: 'test role 1', createdBy: adminUser.id });
+});
+
+describe('GET /api/space/[id]/members/properties/permissions - Create and delete property permissions', () => {
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/spaces/[id]/members/properties/permissions.spec.ts:28]: I think you forgot to rename this when copying over from other file.

Also, the delete endpoints should be under a separate describe group, but same file is fine (following API structure)

PR #1129, opened 2022-10-25
Feat - better discord account association

Rieranthony:
Worked on this ticket: https://app.charmverse.io/charmverse/page-09977184042575171

The issue was that if a user previously created an account using discord oauth and would try to sync their discord with another account, they were prompted with an error.

This features allows to merge the two accounts in a non-blocking flow.

I had to create a migration to add the `deletedAt` field on the `User` table, I'm using it to soft delete the old user account.

### DIFF
--- a/lib/users/mergeUserDiscordAccounts.ts
+++ b/lib/users/mergeUserDiscordAccounts.ts
@@ -8,7 +8,7 @@ export async function mergeUserDiscordAccounts ({ currentUserId, toDeleteUserId,
   discordId: string;
 }): Promise<DiscordUser> {
 
-  const [discordUser] = await Promise.all([
+  const [discordUser] = await prisma.$transaction([
     prisma.discordUser.update({
       where: {
         discordId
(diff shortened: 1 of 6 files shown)


lib/users/mergeUserDiscordAccounts.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { prisma } from 'db';
+
+export async function mergeUserDiscordAccounts ({ currentUserId, toDeleteUserId, discordId }: {
+  currentUserId: string;
+  toDeleteUserId: string;
+  discordId: string;
+}): Promise<DiscordUser> {
+
+  const [discordUser] = await Promise.all([
```
mattcasey, line 11: Prisma has a nice utility for handling multiple updates at once. It runs them in parallel but also commits it as one transaction. I think this would work:
```
const [discordUser] = await prisma.$transaction([
  prisma.discordUser.update(...),
  prisma.user.update(...)
]);
```

mattcasey (approved):
Nice, had one suggestion to turn the merging into a single DB transaction

Devorein (approved):

PR #1130, opened 2022-10-25
open wallet if user tries to create a tg without a wallet connected

Devorein:


mattcasey (changes requested):
I tested and it connects, but the Lit token gate modal should pop up after I have connected my wallet, since the action I clicked on was "Add a token gate". I think you could add a new state variable:
```
Component () {
  ...
  const [clickedAddTokenGate, setClickedAddTokenGate] = useState(false);
  
  useEffect(() => {
    if (account && clickedAddTokenGate) {
      handleTokenGate();
      setClickedAddTokenGate(false);
    }
  , [account, clickedAddTokenGate]);

  return (...
        <MenuItem
          onClick={(e) => {
            if (account) {
                handleTokenGate(e);  
            }
            else {
              openWalletSelectorModal();
              setClickedAddTokenGate(true);
            }
          }}
        />
   ...);
}

```

msadura:
> I tested and it connects, but the Lit token gate modal should pop up after I have connected my wallet, since the action I clicked on was "Add a token gate". I think you could do something like this:
> 
> ```
> Component () {
>   ...
>   const [clickEvent, setClickEvent] = useState(null);
>   
>   useEffect(() => {
>     if (account && clickEvent) {
>       handleTokenGate(clickEvent);
>       setClickEvent(null);
>     }
>   , [account, clickEvent]);
> 
>   return (...
>         <MenuItem
>           onClick={(e) => {
>             if (account) {
>                 handleTokenGate(e);  
>             }
>             else {
>               openWalletSelectorModal();
>               setClickEvent(e);
>             }
>           }}
>         />
>    ...);
> }
> ```

this will work, but I am wondering - why handleTokenGate needs event itself? It is kinda weird to put DOM event in state and reuse it later on

components/settings/invites/components/InviteLinks/components/InviteActions.tsx
```diff
@@ -119,7 +121,19 @@ function InviteActions ({ isAdmin, openInvites, openTokenGate }: InviteActionsPr
             />
           </Box>
         </MenuItem>
-        <MenuItem disabled={!account} onClick={handleTokenGate} disableRipple dense {...restTokenGateProps}>
+        <MenuItem
+          onClick={(e) => {
```
msadura, line 125: we should try to avoid creating functions with logic directly in JSX / handlers. Please move declaration to the top and just use declared function here

Devorein:
> I tested and it connects, but the Lit token gate modal should pop up after I have connected my wallet, since the action I clicked on was "Add a token gate". I think you could do something like this:
> 
> ```
> Component () {
>   ...
>   const [clickEvent, setClickEvent] = useState(null);
>   
>   useEffect(() => {
>     if (account && clickEvent) {
>       handleTokenGate(clickEvent);
>       setClickEvent(null);
>     }
>   , [account, clickEvent]);
> 
>   return (...
>         <MenuItem
>           onClick={(e) => {
>             if (account) {
>                 handleTokenGate(e);  
>             }
>             else {
>               openWalletSelectorModal();
>               setClickEvent(e);
>             }
>           }}
>         />
>    ...);
> }
> ```

Unfortunately that doesn't work as the whole react component is unmounted after switching the wallet and the state is thus lost. Maybe we use a query parameter or a value in the localstorage?

msadura:
Unfortunatelly connecting wallet causes remounting the whole app. Because of that I needed a way to tell if there is a pending action (modal to be opened) and save it in local storage. I ended up having a hook for it.

mattcasey (approved):

hooks/usePendingLocalAction.ts
```diff
@@ -0,0 +1,21 @@
+import { useLocalStorage } from 'hooks/useLocalStorage';
+
+export function usePendingLocalAction (actionKey: string, validityTime = 5000) {
```
mattcasey, line 3: this might come in handy in other places

mattcasey (approved):

PR #1134, opened 2022-10-25
fix warning of prop being passed to html element

mattcasey:
Fixes this warning:
image

### DIFF
--- a/components/common/PageLayout/components/Sidebar/Sidebar.tsx
+++ b/components/common/PageLayout/components/Sidebar/Sidebar.tsx
@@ -95,7 +95,7 @@ const sidebarItemStyles = ({ theme }: { theme: Theme }) => css`
   font-weight: 500;
   padding-top: 4px;
   padding-bottom: 4px;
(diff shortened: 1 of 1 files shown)


components/common/PageLayout/components/Sidebar/Sidebar.tsx
```diff
@@ -95,7 +95,7 @@ const sidebarItemStyles = ({ theme }: { theme: Theme }) => css`
   font-weight: 500;
   padding-top: 4px;
   padding-bottom: 4px;
-  &:hover {
+  &:hover {f
```
msadura, line 98:
```suggestion
  &:hover {
```

msadura (approved):

components/common/PageLayout/components/Sidebar/Sidebar.tsx
mattcasey: i just put that there to see if you were reading ;)

PR #1139, opened 2022-10-26
re-enable e2e tests

mattcasey:
There's some additional code related to fixing Discord login which I'm just choosing to leave in here because it is going to be used in a PR to fix discord login.. https://github.com/charmverse/app.charmverse.io/pull/1140


pages/[domain]/index.tsx
```diff
@@ -24,7 +24,7 @@ export default function RedirectToMainPage () {
     if (isCommonDefaultPage || isDynamicDefaultPage) {
       router.push(defaultPage);
     }
-    else {
+    else if (!loadingPages) {
```
Comment (mattcasey) [inline pages/[domain]/index.tsx:27]: This should fix other tests: we were sending users to /members all the time on first page load because pages were still being loaded

PR #1144, opened 2022-10-27
feat: make roles collapsable

HagiaSofiya:
User story:
https://app.charmverse.io/charmverse/page-49366552253645923?viewId=0b064948-b9d6-418b-983d-8ac1844fd2af&cardId=f814bd8e-e26a-4677-b213-d5dba289e4e8

The arrow indicator is right aligned instead of aligned to role count.


mattcasey (changes requested):
member rows are repeated 4 times

mattcasey (approved):

PR #1145, opened 2022-10-27
Disable lib tests in server tests

msadura:
Restoring commented line of code which I changed while testing things locally
 and additionally  I think someone did not pushed updated package-lock file

### DIFF
--- a/jest.config-server.ts
+++ b/jest.config-server.ts
@@ -5,8 +5,8 @@ import { jestConfig } from './jest.config';
 const integrationConfig = {
   ...jestConfig,
   testMatch: [
-    '**/__integration-tests__/**/*.spec.ts',
-    '**/lib/**/*.spec.ts'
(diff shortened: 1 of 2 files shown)


motechFR (approved):

jest.config-server.ts
```diff
@@ -5,8 +5,8 @@ import { jestConfig } from './jest.config';
 const integrationConfig = {
   ...jestConfig,
   testMatch: [
-    '**/__integration-tests__/**/*.spec.ts',
-    '**/lib/**/*.spec.ts'
+    '**/__integration-tests__/**/*.spec.ts'
+    // '**/lib/**/*.spec.ts'
```
mattcasey, line 9: we can probably just remove this line, right @motechFR ?

msadura: it is useful when you want to run lib tests via vscode extension (unless I am doing something wrong)

PR #1147, opened 2022-10-27
Add timezone for member directory

Devorein:


components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -50,10 +50,10 @@ function MemberDirectoryGalleryCard ({
               {member.roles.length === 0 ? 'N/A' : member.roles.map(role => <Chip label={role.name} key={role.id} size='small' variant='outlined' />)}
             </Stack>
           </Stack>
-          {/* <Stack flexDirection='row' gap={1}>
+          <Stack flexDirection='row' gap={1}>
             <AccessTimeIcon fontSize='small' />
-            <Typography variant='body2'>{member.properties.find(property => property.memberPropertyId === timezoneProperty?.id)?.value ?? 'N/A'}</Typography>
-          </Stack> */}
+            <Typography variant='body2'>{member.profile?.timezone ? convertTZ(member.profile.timezone) : 'N/A'}</Typography>
```
msadura, line 55: this logic of displaying timezone is repeated in 3 places. I think It would be better to have `<TimezoneDisplay />` component that would pick timezone and default value and it would run `convertTZ` logic on its own

components/profile/components/TimezoneModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import Button from 'components/common/Button';
+import Modal, { DialogTitle } from 'components/common/Modal';
+
+interface Timezone {
+  tz: string;
+  offset: string;
+}
+
+const getTimeZoneOptions = () => {
```
msadura, line 13: move it to separate util function

package.json
```diff
@@ -142,6 +142,7 @@
     "mixpanel": "^0.17.0",
     "mjml": "^4.13.0",
     "mjml-react": "^2.0.8",
+    "moment-timezone": "^0.5.38",
```
msadura, line 145: This is a no-go for me. Moment is legacy and deprecated solution. From what I see we use something called luxon - TBH I would rather go with something widely used like date fns, but anyway luxon seem to have support for timezones as well https://github.com/moment/luxon/blob/master/docs/zones.md

components/profile/components/TimezoneModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const timezoneOptions = useMemo(() => getTimeZoneOptions(), []);
+  const [timezone, setTimezone] = useState<null | Timezone | undefined>(initialTimezone ? {
+    tz: initialTimezone,
+    offset: moment.tz(initialTimezone).format('Z')
+  } : null);
+
+  useEffect(() => {
+    if (initialTimezone) {
+      setTimezone({
```
msadura, line 50: this logic is called 3 times. You can create helper function like setInitialTimezone and reuse it

components/profile/components/TimezoneModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    >
+      <DialogTitle onClose={onClose}>Setup your timezone</DialogTitle>
+      <form onSubmit={(e) => {
+        e.preventDefault();
+        onSave(timezone?.tz ?? null);
+        close();
+      }}
+      >
+        <Stack gap={1}>
+          <Autocomplete<Timezone>
```
msadura, line 79: I think this autocomplete deserves its own method

msadura (changes requested):

mattcasey:
image

- Set up is two words
image
- this is the timezone display, i dont know what this means. we should just use America/Denver (the value of timezone) which the user selects from the dropdown

lib/utilities/browser.ts
```diff
@@ -208,3 +208,10 @@ export function highlightDomElement (domElement: HTMLElement, postHighlight?: ()
     postHighlight?.();
   }, 1000);
 }
+
+export function convertTZ (tzString: string) {
```
mattcasey, line 212: just show the timezone, I'm not sure what this is doing.. getting the current time?

components/profile/components/TimezoneModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+    offsetTmz.push({
+      offset: tzOffset,
+      tz: timeZone
+    });
+  });
+
+  return offsetTmz;
+};
+
+export default function TimezoneModal ({
```
mattcasey, line 28: we're supposed to stop using default exports

package.json
mattcasey: I agree, even if the timezone component uses it internally, we may not be using the same version always and we might want to just replace it in the future. Looks like we're just using it to calculate the offset

mattcasey:
It would be nice if, during onboarding, we selected your current timezone as default

package.json
Devorein: Agreed. I will use `Intl.supportedValuesOf('timeZone')` as Matt suggested.

Devorein: I will go with `luxon` as its only available in the latest versions of chrome and ff.
image

Devorein: Luxon doesn't provide the db of all the tz, I had to rely on another package to do so. https://github.com/moment/luxon/issues/353#issuecomment-514203601

components/profile/components/TimezoneModal.tsx
Devorein: Updated thank you.

components/profile/components/TimezoneModal.tsx
Devorein: Agree. I've created a new component `TimezoneAutocomplete` for re-use purpose

components/profile/components/TimezoneModal.tsx
Devorein: It used default since we had to lazy load the component. I can remove it, now that we don't use `moment-timezone` anymore

lib/utilities/browser.ts
Devorein: Removed this function and opted for an inline version with correct date formatting

Devorein:
1. Will tackle that one in the onboarding PR
2. It was too much hassle trying to get the styling correct, and it was using no specific class so the styling had to rely on position of children. And the code was extremely small, its just MUI autocomplete with moment-tz tz's.
3. That is only in the latest browsers. Opted for another [package](https://bundlephobia.com/package/tzdata@1.0.34)

components/profile/components/TimezoneModal.tsx
Devorein: Updated

components/members/components/TimezoneDisplay.tsx
```diff
@@ -0,0 +1,23 @@
...
+  defaultValue = 'N/A',
+  timezone,
+  showTimezone = false
+}: {
+  timezone?: string | null;
+  defaultValue?: string;
+  showTimezone?: boolean;
+}) {
+  return (
+    <>
```
msadura, line 15: it would be safer to wrap this into Stack rather than fragment - currently depending on parent it might endy up as a row or as a col

components/profile/components/TimezoneAutocomplete.tsx
```diff
@@ -0,0 +1,52 @@
...
+  setTimezone,
+  timezone
+}: {
+  timezone?: Timezone | null;
+  setTimezone: (timezone: Timezone | null) => void;
+}) {
+  const currentTz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+  const currentTzOffset = toHoursAndMinutes(DateTime.local().offset);
+
+  const timezoneOptions = useMemo(() => getTimezonesWithOffset(), []);
```
msadura, line 22: getTimezonesWithOffset are not react-dependent so you can move timezoneOptions above component declaration and you will not need memoization

lib/utilities/dates.ts
```diff
@@ -105,3 +105,41 @@ export function coerceToMilliseconds (timestamp: DateInput): number {
...
+  }
+  else {
+    return num.toString().padStart(2, '0');
+  }
+}
+
+export function toHoursAndMinutes (totalMinutes: number) {
+  const hours = Math.floor(totalMinutes / 60);
+  const minutes = totalMinutes % 60;
+  return `${padToTwoDigits(hours)}:${padToTwoDigits(minutes, false)}`;
```
msadura, line 129: actually luxon support `Duration` object which you can use for formatting without introducing weird custom pad method:

```
const hours = Duration.fromObject({ hours: totalMinutes / 60 }).toFormat("hh");
const minutes = Duration.fromObject({ minutes: totalMinutes % 60 }).toFormat("mm");
return `${hours}:${minutes}`);
```

msadura (changes requested):
just a few cosmetic changes

components/members/components/TimezoneDisplay.tsx
Devorein: Good point. Updated.

PR #1148, opened 2022-10-27
My Nexus discussion comments

valentinludu:
Update Nexus discussions
https://app.charmverse.io/charmverse/page-32276379464710536

Also tested email notifications.


motechFR (approved):
@valentinludu Tested and approved!

Quick question, could you add some test cases specific to these 2 changes?

'it should notify the page author when someone comments on the page'

'it should notify the person who created a comment thread when someone replys in this thread'

motechFR:
@valentinludu Tests look good, thanks for adding them in.

2 stylistic comments:
- Please try to generate data as close as possible to where it will be used inside tests. It brings following benefits
-- This also allows us to name variables in more developer-friendly way ie "pageAuthor" instead of "user1"
_I'm aware we have a number of tests with shared data, this is often for reusing a space or admin user, but overall data should be as segregated as possible_

-- It insulates the tests from shared data and makes them more maintainable down the line

- I know that we use "marked" for seen notifications, but we should aim to be as human friendly as possible in our writing. Maybe rename marked / unmarked to "seenNotifications" & "newNotifications" ?

valentinludu:
> @valentinludu Tests look good, thanks for adding them in.
> 
> 2 stylistic comments:
> 
> * Please try to generate data as close as possible to where it will be used inside tests. It brings following benefits
> 
> -- This also allows us to name variables in more developer-friendly way ie "pageAuthor" instead of "user1" and understand when / why that variable is used. _I'm aware we have a number of tests with shared data, this is often for reusing a space or admin user, but overall data should be as segregated as possible_
> 
> -- It insulates the tests from shared data and makes them more maintainable down the line
> 
> * I know that we use "marked" for seen notifications, but we should aim to be as human friendly as possible in our writing. Maybe rename marked / unmarked to "seenNotifications" & "newNotifications" ?

Thanks for the advice Mo.

Just updated right now.

PR #1149, opened 2022-10-27
My Nexus Bounty email notification

valentinludu:
https://app.charmverse.io/charmverse/page-6689592977229408

Tested with Mailgun & also visually tested at api/email-templates


mattcasey:
@valentinludu FYI the background stuff doesnt run in staging

lib/emails/templates/PendingTasks.tsx
```diff
@@ -249,6 +286,30 @@ function ProposalTaskMjml ({ task }: { task: ProposalTask }) {
...
+  const pageWorkspaceTitle = `${task.pageTitle || 'Untitled'} | ${task.spaceName}`;
+  return (
+    <MjmlText>
+      <a
+        href={`${charmverseUrl}/${task.spaceDomain}/${task.pagePath}`}
+        style={{
+          display: 'block'
+        }}
+      >
+        <div style={{ ...h2Style, fontSize: '18px', fontWeight: 'bold', marginBottom: 10, color: '#000' }}>
```
mattcasey, line 299: we support dark mode, so please don't hard-code colors. i noticed we do this in other places but those can be removed as well. if the point was to hide a color link, we should use 'color: inherit' instead

valentinludu: I just removed that color and left the primary one everywhere.

mattcasey (approved):

PR #1150, opened 2022-10-27
Bug fix: adding member property breaks the UI

mattcasey:
We had the incorrect types on the client side for the api endpoints, which don't always return a list of permissions.

### DIFF
--- a/hooks/useMemberProperties.tsx
+++ b/hooks/useMemberProperties.tsx
@@ -36,24 +36,28 @@ export function MemberPropertiesProvider ({ children }: { children: ReactNode })
   const addProperty = useCallback(async (propertyData: CreateMemberPropertyPayload) => {
     if (space) {
       const createdProperty = await charmClient.members.createMemberProperty(space.id, propertyData);
+      const propertyWithPermissions = { ...createdProperty, permissions: [] };
       mutateProperties(state => {
-        return state ? [...state, createdProperty] : [createdProperty];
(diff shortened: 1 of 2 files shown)


hooks/useMemberProperties.tsx
```diff
@@ -36,24 +36,28 @@ export function MemberPropertiesProvider ({ children }: { children: ReactNode })
...
 
   const updateProperty = useCallback(async (propertyData: Partial<MemberProperty> & { id: string }) => {
     if (space) {
       const updatedProperty = await charmClient.members.updateMemberProperty(space.id, propertyData);
       mutateProperties(state => {
-        return state ? state.map(p => p.id === updatedProperty.id ? { ...updatedProperty } : p) : [updatedProperty];
+        if (state) {
+          return state.map(p => p.id === updatedProperty.id ? { ...p, ...updatedProperty } : p);
+        }
+        return state;
```
mattcasey, line 55: I think this was safe... when handling a response to update, we just don't update the list of props unless three's actually something to update. I didn't want to just create an empty list of permissions

hooks/useMemberProperties.tsx
```diff
@@ -36,24 +36,28 @@ export function MemberPropertiesProvider ({ children }: { children: ReactNode })
...
       });
 
       return updatedProperty;
     }
-  }, [space]);
+  }, [space?.id]);
```
mattcasey, line 60: made deps more explicit  so we don't change the method just because you update the workspace name, etc

PR #1152, opened 2022-10-27
fix errors from Share Button effect

mattcasey:
In dev mode, I usually get an error that the ShareButton is trying to update state on an unmounted component. I replaced useState/useEffect pattern with useSWR to get rid of this message.

Please double-check I"m refreshing the page and permissions as needed!


motechFR (approved):
@mattcasey Works well.

One small (pre-existing) issue, I need to close and reopen share button to see updates made in a separate tab as the component is memoized.

mattcasey:
@motechFR the memoization only prevents re-rendering unless the props change. since we're only passing in pageId and headerHeight, i don't tihnk it should change whether it needs to be updated or not... i'll take another look into what you're talking about

motechFR:
> 

I was trying to have the useSWR fire on renavigating to the page.

Maybe a real time event to add?

PR #1153, opened 2022-10-28
Add sort to Member Directory views

Devorein:


components/members/components/MemberDirectorySort.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <InputLabel>Sort</InputLabel>
+      <Select
+        variant='outlined'
+        value={sortedProperty}
+        onChange={(e) => {
+          setSortedProperty(e.target.value);
+        }}
+      >
+        {
+          properties.filter(property => !property.type.match(/role|profile_pic/)).sort((propA, propB) => propA.name > propB.name ? -1 : 1)
```
msadura, line 25: Please do not use regexp matchers for types because you are basically loosing type checking. You should instead check `property.type !== 'role'  && property.type !== 'profile_pic' ` or `!['profile_pic', 'role'].includes(property.type)`

additionally please assign this computation above jsx to const for better readability.

components/members/MemberDirectoryPage.tsx
```diff
@@ -29,15 +32,71 @@ const StyledButton = styled(Button)`
 const views = ['gallery', 'table'] as const;
 type View = typeof views[number];
 
+function sortMembers (members: Member[], property: MemberProperty) {
```
msadura, line 35: please move it outside of this file

msadura (changes requested):

msadura:
Having empty sort select looks weird, don't we have some kind of default sort? Or there should be some placeholder.

Screenshot 2022-10-28 at 22 20 36

Additionally "sort" font is different than on proposal, it is tiny thing but we should keep ui consistent:
Screenshot 2022-10-28 at 22 20 15

msadura:
Extra additional point - it would be nice to have property type icons in this dropdown:
Screenshot 2022-10-28 at 22 22 49

components/members/components/MemberDirectorySort.tsx
Devorein: Agree. Updated

components/common/ViewOptions.tsx
```diff
@@ -0,0 +1,10 @@
+import styled from '@emotion/styled';
+
+export const StyledViewOptions = styled.div`
```
msadura, line 3: Styling MUI with classes like this is just a hack. I like the idea of exporting this component to be reusable but correct way would be to make it like this (pseudo-code like):

```
export function ViewOption({ children, label })
    <Stack display="row" gap={1} //any needed styling>
      <InputLabel>{label}</InputLabel> // with proper font styling
      <>
         {children}
      </>
   </Stack>
 ```

this way we will be able to reuse it in all the places we need

Devorein: Nice one. Updating

PR #1155, opened 2022-10-30
Fixed resyncing gnosis safes

Devorein:
Fix for [this](https://discord.com/channels/894960387743698944/950861057398833212/1036220214964138004) bug reported by Alex

### DIFF
--- a/lib/gnosis/gnosis.ts
+++ b/lib/gnosis/gnosis.ts
@@ -67,7 +67,7 @@ export async function getSafesForAddress ({ signer, chainId, address }: GetSafes
 }
 
 export async function getSafesForAddresses (signer: ethers.Signer, addresses: string[]) {
-  const safes = await Promise.all(Object.values(RPC).filter(network => network.chainId !== 4).map(network => {
+  const safes = await Promise.all(Object.values(RPC).map(network => {
     return Promise.all(addresses.map(address => getSafesForAddress({ signer, chainId: network.chainId, address })));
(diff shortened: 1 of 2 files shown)


lib/gnosis/gnosis.ts
```diff
@@ -67,7 +67,7 @@ export async function getSafesForAddress ({ signer, chainId, address }: GetSafes
 }
 
 export async function getSafesForAddresses (signer: ethers.Signer, addresses: string[]) {
-  const safes = await Promise.all(Object.values(RPC).map(network => {
+  const safes = await Promise.all(Object.values(RPC).filter(network => network.chainId !== 4).map(network => {
```
msadura, line 70: Please avoid using this sort of magic numbers in code. I can understand it only because I was and issue on discord, otherwise it would be just confusing.

1) create `getGnosisConnectors` method
2) It should have an array of keys from `RPC` const supported by gnosis. 
3) Based on that return proper array of connectors for gnosis. If you want to just filter RPC view chainId at east explain that in comment.
4) Ps. There is `RPCList` in connectors which provides array of connectors

msadura (changes requested):

lib/gnosis/gnosis.ts
mattcasey: Actually, let's just remove rinkeby

Devorein: Updating

mattcasey (approved):

PR #1156, opened 2022-10-31
Feat/select field

msadura:
reusable Select and multiselect fields for member properties. https://app.charmverse.io/charmverse/page-4910186832290666

Note: this does not include ad-hoc creating new values in select mode - this and connecting to focalboard is incoming in separate pr

https://user-images.githubusercontent.com/5198394/198970787-5e337c49-397d-4a66-ae5d-b0730de99872.mp4


Devorein:
1. The select properties value isn't responsive on public profile and on gallery card. I think we can let them move onto the next row if its overflowing
image
image

2. In the table view the select chips are showing the id, the color and label is missing
image

msadura:
@Devorein mentioned issues are fixed. As for searching by values I would extract it to separate ticket. This will not be trivial and query might bet slow. We are storing id of option rather than value, this is why this task will be more complex

Devorein (approved):

PR #1157, opened 2022-10-31
Fix react key property

msadura:
For some reason our eslint setup did not warn about missing `key` prop for children arrays. This is one of basic good react practises so I think it is worth to add it.
This pr adds eslint rule + fixes all the places where keys were missing.


mattcasey:
This is amazing, thanks. I've been fixing them one at a time when I notice in the dev console.

mattcasey (approved):

PR #1158, opened 2022-10-31
Member directory properties reorder

Devorein:


components/members/components/MemberDirectoryProperties/MemberPropertiesSidebar.tsx
```diff
@@ -111,15 +114,89 @@ export function MemberPropertySidebarItem ({
...
+      }
+    },
+
+    // listen to hover events to determine if the mouse is over the top portion of the node
+    hover (_item: MemberPropertyWithPermissions, monitor) {
+      if (!ref.current) {
+        return;
+      }
+      const _isOverCurrent = monitor.isOver({ shallow: true });
+      let _isAdjacent = true;
```
mattcasey, line 160: Looks like you copied a bunch of code from TreeNode but you're not using it. You can instead just add `isOver: monitor.isOver()` to the collection in useDrag().

The background of 'isAdjacent': in the sidebar, we give the appearance that you are dragging between items, so we want to ignore the hover unless you are within 5px of the top of the row (and we show a little blue div over the top edge). One thing we don't support yet is dragging to the bottom of the last item. I am fine if we just do what you have now but I think you should take out all the logic related to 'isAdjacent'.

mattcasey: There's a lot of good examples here https://react-dnd.github.io/react-dnd/examples btw

mattcasey (changes requested):
see comments

msadura:
1) When I try to drag something it looks like dragging 3items:
Screenshot 2022-11-03 at 10 57 39

2) Currently discovering drag&drop is rather a matter of coincidence by the user. I think we should follow notion and display draggable icon on hover:
Screenshot 2022-11-03 at 11 08 56

components/common/PageLayout/components/PageNavigation/components/TreeNode.tsx
```diff
@@ -191,7 +191,7 @@ function DraggableTreeNode ({ item, onDropAdjacent, onDropChild, pathPrefix, add
 }
 
 // pulled from react-merge-refs
-function mergeRefs (refs: any) {
+export function mergeRefs (refs: any) {
```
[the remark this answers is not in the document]
Devorein: Updated

components/members/components/MemberDirectoryProperties/MemberPropertiesSidebar.tsx
```diff
@@ -111,15 +114,89 @@ export function MemberPropertySidebarItem ({
...
+
+  const [{ canDrop, isOverCurrent }, drop] = useDrop<MemberPropertyWithPermissions, any, { canDrop: boolean, isOverCurrent: boolean }>(() => ({
+    accept: 'item',
+    drop: async (droppedProperty, monitor) => {
+      const didDrop = monitor.didDrop();
+      if (didDrop) {
+        return;
+      }
+      if (isAdjacentRef.current) {
+        await updateProperty({
```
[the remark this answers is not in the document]
Devorein: Fixed logic to update index of all affected properties

mattcasey (changes requested):
Please make sure properties are being sorted on the gallery cards. Doesn't seem to change at the moment, but they do get sorted on the table view!

lib/members/generateDefaultPropertiesInput.ts
```diff
@@ -27,5 +27,5 @@ export function generateDefaultPropertiesInput ({ userId, spaceId }: { userId: s
     });
   });
 
-  return defaultPropertiesInput;
+  return defaultPropertiesInput.sort((propA, propB) => propA.name > propB.name ? 1 : -1).map((property, index) => ({ ...property, index }));
```
mattcasey, line 30: we shouldn't sort by name if we are sorting by index

lib/members/updateMemberProperty.ts
```diff
@@ -19,10 +22,45 @@ export function updateMemberProperty ({ data, userId, id }: UpdatePropertyInput)
...
-  return prisma.memberProperty.update({
+  if (newIndex !== null && newIndex !== undefined) {
+    const memberProperty = await prisma.memberProperty.findUnique({
+      where: {
+        id
+      },
+      select: {
+        index: true
+      }
+    }) as MemberProperty;
```
mattcasey, line 33: It's dangerous to use a type assertion to hide a possible null value. You can use findUniqueOrThrow, then we would get a useful error and no type assertion necessary

lib/members/updateMemberProperty.ts
```diff
@@ -19,10 +22,45 @@ export function updateMemberProperty ({ data, userId, id }: UpdatePropertyInput)
...
+
+    const currentIndex = memberProperty.index;
+
+    transactions.push(prisma.memberProperty.updateMany({
```
mattcasey, line 37: the new indexing logic should be handled on the frontend. There's no guarantee that the current state of indexes matches what's in the frontend, especially if multiple users happen to be trying to change order the same time, or if someone has a bad network connection.
I think it deserves its own API route.. just send back a list of indexes for all member properties, like we do for the page tree?

lib/members/updateMemberProperty.ts
```diff
@@ -19,10 +22,45 @@ export function updateMemberProperty ({ data, userId, id }: UpdatePropertyInput)
...
+      where: {
+        spaceId,
+        index: currentIndex < newIndex ? {
+          gt: currentIndex,
+          lte: newIndex as number
```
mattcasey, line 42: please try to avoid using type assertions

lib/members/updateMemberProperty.ts
```diff
@@ -19,10 +22,45 @@ export function updateMemberProperty ({ data, userId, id }: UpdatePropertyInput)
...
+        }
+      }
+    }));
+  }
+
+  transactions.push(prisma.memberProperty.update({
     where: {
       id
     },
     data: { ...data, updatedBy: userId, id }
```
mattcasey, line 62: we should be setting updatedAt as well

lib/members/generateDefaultPropertiesInput.ts
Devorein: The default properties are sorted by their name first and that alphabetical sort order also becomes their index. This line was necessary as there were two properties with the same index, but I've updated it

lib/members/updateMemberProperty.ts
Devorein: Updated

lib/members/updateMemberProperty.ts
Devorein: `@updatedAt` attribute in `schema.prisma` should take care of that, its been added for the `MemberProperty` model

lib/members/updateMemberProperty.ts
Devorein: Updated

lib/members/updateMemberProperty.ts
Devorein: Added a few unit tests to make sure its working as expected.

jest.config.ts
```diff
@@ -32,8 +34,8 @@ export const jestConfig = {
 
   // The glob patterns Jest uses to detect test files
   testMatch: [
-    '**/lib/**/*.spec.ts',
-    '**/testing/**/*.spec.ts'
+    path.join(__dirname, '**/lib/**/*.spec.ts')
```
mattcasey, line 37: why is this needed

PR #1159, opened 2022-10-31
Expand member properties when navigating to a profile from Member Directory

Devorein:


components/profile/components/SpacesMemberDetails/SpacesMemberDetails.tsx
```diff
@@ -13,6 +15,19 @@ type Props = {
...
+  const { query, pathname } = useRouter();
+
+  useEffect(() => {
+    if (query.workspace && !isLoading) {
+      const expandedWorkspaceAccordion = document.getElementById(`workspace-properties-accordion-${query.workspace}`);
+      if (expandedWorkspaceAccordion) {
+        expandedWorkspaceAccordion.scrollIntoView({
+          behavior: 'smooth'
+        });
+        setUrlWithoutRerender(pathname, { workspace: null });
```
mattcasey, line 27: it's actually fine to leave the URL imo. We shouldn't use the URL just to pass params around - if that's the case, we can use something else like localStorage but I think that's overkill.

PR #1161, opened 2022-10-31
Marketing analytics

motechFR:
Final todo: Send track page from public page in a way that it doesn't create duplicates

Need to test bounty and proposals, to see if they send this.

Main issue is we don't use the same CharmEditor in public and internal mode


pages/api/profile/index.ts
```diff
@@ -32,29 +36,64 @@ async function createUser (req: NextApiRequest, res: NextApiResponse<LoggedInUse
...
+
   if (!profile) {
-    return res.status(404).json({ error: 'No user found' });
+    return handleNoProfile();
   }
+
+  // Clean up the anonymous id if the user has a profile
+  req.session.anonymousUserId = undefined;
+
+  await req.session.save();
```
mattcasey, line 95: can you avoid this call if anonymouseUserId was not set? this call is called all the time, we should avoid extra async ops that are unecesssary

marketing-site/analytics.html
```diff
@@ -0,0 +1,39 @@
+<!--
```
mattcasey, line 1: please add a README to this folder explaining what the folder is for

hooks/useUserAcquisition.ts
```diff
@@ -0,0 +1,73 @@
...
+      });
+    }
+
+  }
+
+  function setCampaign () {
+    const currentCampaign = router.query.utm_campaign;
+
+    if (currentCampaign) {
+      setCookie('appCampaign' as SignupCookieType, currentCampaign, {
```
mattcasey, line 46: please use setCookie/getCOokie from our utils in lib/utilities/browser - no need to add a new npm dependency

lib/metrics/mixpanel/utils-signup.ts
```diff
@@ -0,0 +1,67 @@
+
```
mattcasey, line 1: all files should use camelCase

charmClient/apis/trackApi.ts
```diff
@@ -1,10 +1,38 @@
...
+
+      return Promise.resolve({ success: 'ignored' });
+    }
+
+    this.lastEvent = {
+      payload: payloadAsString,
+      timestamp: now
+    };
+
+    return http.POST<{ success: 'ok' | 'ignored' }>(`/api/events?event=${event}`, payload);
```
mattcasey, line 35: the backend never returns 'ignored'?

pages/api/profile/index.ts
```diff
@@ -32,29 +36,64 @@ async function createUser (req: NextApiRequest, res: NextApiResponse<LoggedInUse
...
   }
 
+  // Null out the anonmyous user id after successful login
+  req.session.anonymousUserId = undefined;
   req.session.user = { id: user.id };
   await updateGuildRolesForUser(user.wallets.map(w => w.address), user.spaceRoles);
   await req.session.save();
 
+  signupCookieNames.forEach(cookie => {
+    deleteCookie(cookie, {
```
mattcasey, line 57: we have a pattern for handling cookies already on the backend
```
import Cookies from 'cookies';
....
const cookies = new Cookies(req, res);
...
cookies.set(cookie, null);
````

File: hooks/useUserAcquisition.ts [hunk 7]
```diff
@@ -0,0 +1,73 @@
...
+        maxAge: maxCookieAge,
+        path: '/'
+      });
+    }
+  }
+
+  function setLandingPage () {
+    const currentPage = window.location.href.split('?')[0];
+
+    if (currentPage) {
```
mattcasey, line 32: when would currentPage be undefined? seems impossible

pages/index.tsx
```diff
@@ -26,6 +27,13 @@ export default function LoginPage () {
   const { spaces, isLoaded: isSpacesLoaded } = useSpaces();
   const discordCookie = getCookie(AUTH_CODE_COOKIE);
 
+  const { refreshSignupData } = useUserAcquisition();
+
+  useEffect(() => {
```
mattcasey, line 32: this function needs to run on ever page, not just if users hit the domain w/no path (pages/index.tsx). I believe you just have to put this in app.tsx instead

mattcasey (changes requested):
see comments

marketing-site/analytics.html
motechFR: yep, done

hooks/useUserAcquisition.ts
motechFR: Gotcha, will know for future!

charmClient/apis/trackApi.ts
motechFR: Yep! Fair point

pages/api/profile/index.ts
motechFR: Gotcha, will know in future

hooks/useUserAcquisition.ts
motechFR: I was going along with patter for other setters here, but you're right. Removed this :)

pages/index.tsx
motechFR: Done!

mattcasey (approved):

PR #1162, opened 2022-10-31
Internal links fix

motechFR:
This fixes an issue where internal links between pages of a space would fail in the public pages mode.

Also added an e2e test to click on nested page element, make sure second page renders, and click back to first page


__e2e__/utils/mocks.ts
```diff
@@ -291,3 +291,48 @@ export async function generateUserAndSpace ({ isAdmin, spaceName = 'Example Spac
     privateKey: wallet.privateKey
   };
 }
+
+export function generatePage (options: Partial<Page> & Pick<Page, 'spaceId' | 'createdBy'> & { pagePermissions?: Prisma.PagePermissionCreateManyPageInput[] }): Promise<IPageWithPermissions> {
```
msadura, line 295: I've got the feeling this e2e/utils/mocks.ts file is mix of a e2e and "general" testing utils. For example `generatePage` is not related to any browser behavior so I would put it under `testing/utils`. 
`testing/seedData.ts` got too big in my opinion so I started adding new utils under specified domain (i.e. spaces, members).
TLDR: I would put this function in `testing/utis/pages.ts`

msadura (approved):

__e2e__/utils/mocks.ts
motechFR: Good spot! Thanks for the clean code healthcheck :)

PR #1166, opened 2022-11-01
Re-added rinkeby to rpc list and filtered out in gnosis safe call

Devorein:


lib/gnosis/gnosis.ts
```diff
@@ -66,8 +66,14 @@ export async function getSafesForAddress ({ signer, chainId, address }: GetSafes
   return [];
 }
 
+export function gnosisSupportedNetworks () {
+  // ChainId 4 Rinkeby is not supported by gnosis anymore
+  const unsupportedChainIds: number[] = [4];
```
mattcasey, line 71: consts like this should be defined outside of functions at the top of the file

Devorein: oops by bad
updating

mattcasey (approved):

PR #1169, opened 2022-11-02
Create an unique taskid for discussion notifications

valentinludu:
The problem was that a commentId can't be a unique id in the UserNotifications table.
FIY: mentionId is unique so I left it as it is.

### DIFF
--- a/lib/discussion/__tests__/getDiscussionTasks.spec.ts
+++ b/lib/discussion/__tests__/getDiscussionTasks.spec.ts
@@ -277,4 +277,52 @@ describe('getDiscussionTasks', () => {
       && item.pageId === page.id
     ));
   });
+
+  it('Should return unique taskIds for two users that have the same comment to be notified about', async () => {
+    const pageAuthor = await generateSpaceUser({ spaceId: space1.id, isAdmin: false });
+    const pageCommenter = await generateSpaceUser({ spaceId: space1.id, isAdmin: false });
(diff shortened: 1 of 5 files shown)


lib/discussion/__tests__/getDiscussionTasks.spec.ts
```diff
@@ -277,4 +277,52 @@ describe('getDiscussionTasks', () => {
       && item.pageId === page.id
     ));
   });
+
+  it('Should return unique taskIds for two users that have the same comment to be notified about', async () => {
```
mattcasey, line 281: the unique constraint is taskId + userId, I don't think this was the issue

lib/discussion/getDiscussionTasks.ts
```diff
@@ -140,10 +141,11 @@ export async function getDiscussionTasks (userId: string): Promise<DiscussionTas
 
     const commentTask = {
       ...commentTaskWithoutUser,
-      createdBy: usersRecord[commentTaskWithoutUser.userId]
+      createdBy: usersRecord[commentTaskWithoutUser.userId],
+      taskId: `${commentTaskWithoutUser.commentId}.${userId}` // Creating an unique identifier for Prisma UserNotifications
```
mattcasey, line 145: this cannot be the right fix since the unique constraint already includes userId

mattcasey (changes requested):
userId should not be baked into taskId because it's handled in the table constraint already

PR #1170, opened 2022-11-02
Member directory property visibility

Devorein:


prisma/schema.prisma
```diff
@@ -820,6 +821,20 @@ model MemberPropertyValue {
...
+  table
+}
+
+model MemberPropertyVisibility {
+  id               String                       @id @default(uuid()) @db.Uuid
+  view             MemberPropertyVisibilityView
+  memberPropertyId String                       @db.Uuid
+  memberProperty   MemberProperty               @relation(fields: [memberPropertyId], references: [id], onDelete: Cascade)
+
+  @@unique([view, memberPropertyId])
```
mattcasey, line 835: (in case we keep this table) Looks like you need to add a migration for this constraint. When i test locally and run `prisma migrate dev`, prisma is asking me to generate a migration file.

components/common/BoardEditor/focalboard/src/widgets/icons/table.tsx
```diff
@@ -1,22 +1,27 @@
 
+import { useTheme } from '@emotion/react';
+import type { SVGProps } from 'react';
 import React from 'react';
 
-export default function TableIcon (): JSX.Element {
+export default function TableIcon (props: SVGProps<any> & { disabled?: boolean }): JSX.Element {
+  const theme = useTheme();
```
mattcasey, line 7: Suggestion: 'currentColor' means the color can be defined by the parent element. There's no need to set it manually here, and you'll be able to use this pattern with any icon. For example, using the Typography component "should work" this way:
```
  <Typography component='span' color={disabled ? 'secondary' : 'inherit'}>
    <GalleryIcon />
  </Typography>
```
Theoretically you shouldn't have to change anything inside this icon component, and the solution would work for any svg icon we use.

components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -17,6 +17,12 @@ function MemberDirectoryGalleryCard ({
 }) {
   const { properties = [] } = useMemberProperties();
   const nameProperty = properties.find(property => property.type === 'name');
+  const timezoneProperty = properties.find(property => property.type === 'timezone');
+  const rolesProperty = properties.find(property => property.type === 'role');
+
+  const isNameHidden = nameProperty?.memberPropertyVisibilities.find(pv => pv.view === 'gallery');
```
mattcasey, line 23: .some() is a bit better than .find() when you just want a boolean return

components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -39,24 +45,34 @@ function MemberDirectoryGalleryCard ({
...
-          <Stack flexDirection='row' gap={1}>
-            <TimezoneDisplay
-              showTimezone
-              timezone={member.profile?.timezone}
-            />
-          </Stack>
+          )}
           {properties.map(property => {
             const memberPropertyValue = member.properties.find(memberProperty => memberProperty.memberPropertyId === property.id);
+            const hiddenInGallery = property.memberPropertyVisibilities.find(propertyVisibility => propertyVisibility.view === 'gallery');
```
mattcasey, line 72: if this was just a list, the logic would be simplified:
```

File: prisma/schema.prisma [hunk 5]
```diff
@@ -820,6 +821,20 @@ model MemberPropertyValue {
   @@unique([memberPropertyId, spaceId, userId])
 }
 
+enum MemberPropertyVisibilityView {
+  gallery
+  table
```
mattcasey, line 826: we should also include and support 'profile'

components/members/components/MemberDirectoryProperties/MemberPropertiesSidebar.tsx
```diff
@@ -154,9 +160,54 @@ export function MemberPropertySidebarItem ({
             gap={0.5}
             className='icons'
             sx={{
-              opacity: 0
+              opacity: 0,
+              alignItems: 'center'
             }}
           >
+            <Tooltip title={`Property is ${disabledInGalleryView ? 'invisible' : 'visible'} in gallery view`}>
```
mattcasey, line 167: The menu has gotten really crowded, I think this should go on its own row, at the top of the collapsible section. On the left it could say "View visibility?" We also need a 3rd icon for 'profile' view.

image

mattcasey (changes requested):
Functionally, it's working well. I left some comments and think we should just have a column of strings rather than a whole new table

prisma/schema.prisma
```diff
@@ -790,19 +790,20 @@ enum MemberPropertyPermissionLevel {
...
+  updatedBy                  String                     @db.Uuid
+  spaceId                    String                     @db.Uuid
+  space                      Space                      @relation(fields: [spaceId], references: [id], onDelete: Cascade)
+  name                       String
+  type                       MemberPropertyType
+  options                    Json?
+  index                      Int                        @default(-1)
+  memberPropertyValues       MemberPropertyValue[]
+  permissions                MemberPropertyPermission[]
+  memberPropertyVisibilities MemberPropertyVisibility[]
```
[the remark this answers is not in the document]
Devorein: I've replaced the table with `enabledViews` property in `MemberProperty`

prisma/schema.prisma
Devorein: Not required, since the table was removed

components/common/BoardEditor/focalboard/src/widgets/icons/table.tsx
Devorein: Thanks, tried and it works. I will keep that in mind from now on

components/members/components/MemberDirectoryGalleryView.tsx
Devorein: Its a lot easier now since the table has been removed its been replaced with the following
`const isNameHidden = !nameProperty?.enabledViews.includes('gallery')`

components/members/components/MemberDirectoryGalleryView.tsx
Devorein: Updated thnx for the suggestion :D

prisma/schema.prisma
Devorein: Yes added to the enum and also made sure its functionally working

components/members/components/MemberDirectoryProperties/MemberPropertiesSidebar.tsx
Devorein: Updated

PR #1174, opened 2022-11-03
Improve discord login

motechFR:
Enables user to login via discord even if they have an active wallet

### DIFF
--- a/components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
+++ b/components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
@@ -5,7 +5,6 @@ import Box from '@mui/material/Box';
 import Grid from '@mui/material/Grid';
 import TextField from '@mui/material/TextField';
 import { DateTimePicker } from '@mui/x-date-pickers';
-import { useWeb3React } from '@web3-react/core';
 import { getChainById } from 'connectors';
 import { DateTime } from 'luxon';
 import { useEffect, useState } from 'react';
@@ -19,6 +18,7 @@ import PrimaryButton from 'components/common/PrimaryButton';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
 import useIsAdmin from 'hooks/useIsAdmin';
 import { usePages } from 'hooks/usePages';
+import { useWeb3AuthSig } from 'hooks/useWeb3AuthSig';
 import type { PageMeta } from 'lib/pages';
 import { generateMarkdown } from 'lib/pages/generateMarkdown';
 import type { SnapshotReceipt, SnapshotSpace, SnapshotVotingModeType, SnapshotVotingStrategy } from 'lib/snapshot';
@@ -46,7 +46,7 @@ const MIN_VOTING_OPTIONS = 2;
 
 export default function PublishingForm ({ onSubmit, page }: Props) {
 
-  const { account, library } = useWeb3React();
+  const { account, library } = useWeb3AuthSig();
 
   const [space] = useCurrentSpace();
 
--- a/components/common/TokenGateForm/TokenGateOption.tsx
+++ b/components/common/TokenGateForm/TokenGateOption.tsx
@@ -6,11 +6,11 @@ import Chip from '@mui/material/Chip';
 import Grid from '@mui/material/Grid';
 import Typography from '@mui/material/Typography';
 import type { SxProps, Theme } from '@mui/system';
-import { useWeb3React } from '@web3-react/core';
 import { humanizeAccessControlConditions } from 'lit-js-sdk';
 import { useEffect, useState } from 'react';
 
 import LoadingComponent from 'components/common/LoadingComponent';
+import { useWeb3AuthSig } from 'hooks/useWeb3AuthSig';
 import type { TokenGateWithRoles } from 'lib/token-gates/interfaces';
 
 interface Props {
@@ -21,7 +21,7 @@ interface Props {
 
 export default function TokenGateOption ({ tokenGate, validGate, isVerifying }: Props) {
 
-  const { account } = useWeb3React();
+  const { account } = useWeb3AuthSig();
   const [description, setDescription] = useState<string>('');
 
   useEffect(() => {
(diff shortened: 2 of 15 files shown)


pages/api/profile/index.ts
```diff
@@ -101,18 +104,34 @@ async function getUser (req: NextApiRequest, res: NextApiResponse<LoggedInUser |
...
+      include: sessionUserRelations
+    }) as LoggedInUser;
+
+    if (!updatedProfile) {
+      return handleNoProfile(req, res);
+    }
+    else {
+      user = updatedProfile;
     }
-    user = await getUserProfile('id', req.session.user.id);
```
mattcasey, line 115: Why are we not calling getUserProfile here still?

pages/index.tsx
```diff
@@ -35,12 +37,17 @@ export default function LoginPage () {
   const returnUrl = router.query.returnUrl as string | undefined;
 
   function redirectToDefaultPage () {
-    // send to signup for users without a workspace unless they are being redirected to an existing workspace
-    if (spaces.length === 0 && !isSpaceDomain(returnUrl?.replaceAll('/', ''))) {
+
+    // Send the user in priority to the invites page if they logged in looking to join a space
+    if (returnUrl?.match('join') || returnUrl?.match('invite')) {
```
mattcasey, line 42: this logic looks like it should already be handled by `!isSpaceDomain(returnUrl)`.. which is basically saying, there is a redirectUrl and the first part of the path is not a workspace domain.

__e2e__/utils/mocks.ts
```diff
@@ -91,7 +91,11 @@ export async function createDiscordUser ({ userId, discordUserId }: { userId: st
     data: {
       account: {},
       discordId: discordUserId,
-      userId
+      user: {
+        connect: {
```
mattcasey, line 95: Please avoid adding includes to test mocks. it slows down all the tests that use these, and most of the time they already have the user data

pages/api/profile/is-connectable-wallet.ts
```diff
@@ -0,0 +1,33 @@
...
+import nc from 'next-connect';
+
+import { prisma } from 'db';
+import { onError, onNoMatch } from 'lib/middleware';
+import { withSessionRoute } from 'lib/session/withSession';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .post(isConnectableWallet);
```
mattcasey, line 12: change to GET

pages/index.tsx
```diff
@@ -53,41 +60,48 @@ export default function LoginPage () {
...
+  useEffect(() => {
+    if (verifiableWalletDetected && signature && !user) {
+      setLoggingIn(true);
+      loginFromWeb3Account(signature)
+        .finally(() => setLoggingIn(false));
+    }
+  }, [verifiableWalletDetected]);
 
-  if (!showLogin) {
+  if (!showLogin || loggingIn) {
```
mattcasey, line 93: please call setShowLogin(true) instead of adding an override

pages/index.tsx
```diff
@@ -53,41 +60,48 @@ export default function LoginPage () {
...
 
-  async function loginFromWeb3 () {
-    await loginFromWeb3Account();
-    redirectToDefaultPage();
-  }
-
   useEffect(() => {
     setTitleState('Welcome');
     if (discordCookie) {
+      // console.log('Logged in with Discord');
```
mattcasey, line 66: please remove comments

pages/index.tsx
mattcasey, line 42: I don't think this logic is necessary, it would mean our join/invite logic for Discord is broken in prod right now. From what I can tell
- invite page should never be a returnUrl, because the invite page redirects directly back to itself (see InviteLinkPage component)
- likewise, /join is just a link you get to from /signup

pages/api/profile/index.ts
```diff
@@ -101,18 +104,34 @@ async function getUser (req: NextApiRequest, res: NextApiResponse<LoggedInUser |
...
 
-  if (req.body.addresses) {
-    for (const address of req.body.addresses) {
-      await prisma.userWallet.createMany({
-        data: [{
-          userId: req.session.user.id,
-          address
-        }]
-      });
+  const addressesToAdd = req.body.addressesToAdd as AuthSigWithRawAddress[];
```
mattcasey, line 109: this seems to warrant its own endpoint...

mattcasey (changes requested):
see comments

pages/api/profile/index.ts
motechFR: Re-added this in thanks for the spot.

Not sure why we are throwing an error, LoggedInUser | null would be better, so that calling code can explicitly handle cases where a user doesn't exist

__e2e__/utils/mocks.ts
motechFR: This is used in some of our tests. I updated this create assertion to meet our normal prisma pattern of using the connect in a create statement.

Surprised that userId works, but have reverted this as per your request.

pages/api/profile/is-connectable-wallet.ts
motechFR: I've deleted this endpoint entirely as it was used in an earlier iteration

pages/index.tsx
motechFR: Removed this

__e2e__/sessions/signup.spec.ts
```diff
@@ -70,6 +71,20 @@ test('signup - ignores the logic to redirect user after connect', async ({ sandb
 
 });
 
+test('signup - not ignore the logic to redirect user after connect if the user has 0 spaces, and the redirect is to a join page', async ({ sandboxPage }) => {
```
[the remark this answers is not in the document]
motechFR: Agreed, extracted to its own file

pages/api/profile/index.ts
motechFR: Done, moved to add-wallets endpoint

PR #1175, opened 2022-11-03
Hide Organization Edit Pencils by default

Devorein:


components/profile/components/SpacesMemberDetails/components/SpaceDetailsAccordion.tsx
```diff
@@ -16,9 +18,28 @@ type Props = {
   expanded?: boolean;
 };
 
+const StyledAccordionSummary = styled(AccordionSummary)`
+  ${({ theme }) => !isTouchScreen() && `
+    &:hover .icons {
+      opacity: 1;
+      transition: ${theme.transitions.create('opacity', {
+    easing: theme.transitions.easing.easeInOut,
+    duration: theme.transitions.duration.enteringScreen
```
msadura, line 27: formatting looks really weird, but this is just a tiny thing

components/profile/components/UserDetails/UserDetails.tsx
```diff
@@ -49,6 +51,41 @@ export interface UserDetailsProps {
   sx?: SxProps;
 }
 
+const StyledStack = styled(Stack)`
+  &:hover .icons {
```
msadura, line 55: looks like it does pretty much the same as styles Accordion. Moreover we will most likely  use this behavior in other places. I would place those in separate file like `hoverIconsStyle` export to reuse with emotion `css` . Then you can import such css and reuse in every place you need it

msadura (changes requested):

components/profile/components/SpacesMemberDetails/components/SpaceDetailsAccordion.tsx
Devorein: yes tried a lot to fix this but as soon as the file is saved, vscode formats it wildly

components/profile/components/UserDetails/UserDetails.tsx
Devorein: Agree. Create a separate file, exported the `hoverIconsStyle` constant and reused in the places its used

msadura (approved):

PR #1178, opened 2022-11-04
Feat/remove tg members

msadura:
https://app.charmverse.io/charmverse/page-42654196334361116

potential concerns:
- how often should we run this ? (for now I set every 10mins) @mattcasey 
- I am not sure what will be performance of this verification when we run it against xxx users on prod


msadura:
Updated logic to not only remove users but also roles assigned via invalid token gates + tests

prisma/schema.prisma
```diff
@@ -438,18 +440,31 @@ model FavoritePage {
...
-  user                   User              @relation(fields: [userId], references: [id], onDelete: Cascade)
-  spaceRoleToRole        SpaceRoleToRole[]
+  jwt                    String?
+  tokenGateId            String?    @db.Uuid
+  tokenGate              TokenGate? @relation(fields: [tokenGateId], references: [id], onDelete: SetNull)
+  space                  Space      @relation(fields: [spaceId], references: [id], onDelete: Cascade)
+  user                   User       @relation(fields: [userId], references: [id], onDelete: Cascade)
+  grantedRoles           String[]   @default([])
+
+  @@unique([tokenGateId, userId, spaceId], name: "tokenGateUserSpace")
```
mattcasey, line 455: don't forget to add indexes :)

mattcasey (approved):

background/tasks/verifyTokenGateMemberships/task.ts
```diff
@@ -0,0 +1,16 @@
...
+import log from 'lib/log';
+import { verifyTokenGateMemberships } from 'lib/token-gates/verifyTokenGateMemberships';
+
+export async function task () {
+
+  log.debug('Running Verify Token Gate memberships cron job');
+
+  try {
+    const results = await verifyTokenGateMemberships();
+    log.debug('Number of members removed due to invalid token gate conditions', results.deletedUsers);
```
mattcasey, line 11: remember to add some metrics.. something like:
```
count('cron.token-gate-verification.deleted-users', results.deletedUsers);
```

lib/token-gates/verifyTokenGateMemberships.ts
```diff
@@ -0,0 +1,75 @@
...
+      },
+      spaceRoleToRole: {
+        include: {
+          role: true
+        }
+      }
+    }
+  });
+
+  const promises = usersWithTokenGates.map(async spaceRole => verifyTokenGateMembership({
```
mattcasey, line 59: i think it would be better to use a loop and not do these in parallel, there's no need to complete asap. and it might add a little buffer to not hit the Lit Protocol all at once

mattcasey (approved):
Looks good, remember to add metrics and I suggested using a loop over members instead of calling lit all at once for everybody

PR #1180, opened 2022-11-04
Don't fetch member values in member directory

Devorein:


components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
```diff
@@ -59,7 +60,7 @@ export function MemberPropertiesPopupForm ({ cancelButtonText = 'Cancel', childr
     reset(defaultValues);
   }, [defaultValues]);
 
-  if (!data && spaceId) {
+  if (!data && spaceId && showLoading) {
     return <LoadingComponent isLoading />;
```
mattcasey, line 64: you could just do isLoading={showLoading}

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
```diff
@@ -59,7 +60,7 @@ export function MemberPropertiesPopupForm ({ cancelButtonText = 'Cancel', childr
     reset(defaultValues);
   }, [defaultValues]);
 
-  if (!data && spaceId) {
+  if (!data && spaceId && showLoading) {
```
mattcasey, line 63: i can't tell the difference when this is used or not (its disabled in the popup for properties right?) I think it is a bad pattern to not show loading if we do show it in other places.

mattcasey (approved):
Approved but I think you should remove the loading indicator override. Although locally I don't even see it even when it's enabled

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
msadura: got the same question - why this is introduced?

components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -167,11 +166,14 @@ function MemberDirectoryGalleryCard ({
         <MemberPropertiesPopupForm
           onClose={() => {
             setIsModalOpen(false);
-            mutateMembers();
           }}
           memberId={member.id}
+          showLoading={false}
           spaceId={currentSpace.id}
-          updateMemberPropertyValues={updateSpaceValues}
+          updateMemberPropertyValues={async (spaceId, values) => {
```
msadura, line 173: please move declaration of this function to the top. We should stop declaring function directly in props because it is kinda bad practise

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
Devorein: This was needed to disable the loading indicator when showing the properties form in member directory table and gallery views. Otherwise, the loading indicator can be seen on the right side causing the whole layout to shift.

components/members/components/MemberDirectoryGalleryView.tsx
Devorein: Updated. Thank you.

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopupForm.tsx
mattcasey: We agreed the loading icon should just appear inside the dialog instead

PR #1184, opened 2022-11-07
Rearrange commands

valentinludu:
Chris requested 2 small changes after the latest updates on the UI:
1. Reorder commands list
2. Added red dot instead of eye icon to unread discussions


components/common/CharmEditor/components/inlinePalette/editorItems/text.tsx
```diff
@@ -69,27 +105,42 @@ export function items (): PaletteItemTypeNoGroup[] {
         };
       }
     },
+    ...(pageType !== 'card_template' && !disableNestedPage ? [{
```
msadura, line 108: this notation with ... at the beginning looks a bit unclear. Could you please move deckaration of this array to a cons, and just spread it here?

msadura (approved):

components/common/CharmEditor/components/inlinePalette/editorItems/text.tsx
valentinludu: Done

mattcasey (approved):
Approved, but it would be better to put the nexus-related story in a separate PR next time

PR #1186, opened 2022-11-07
move setSpace into useSpaces method

mattcasey:
just some cleanup and reducing the reliance on useCurrentSpace()


msadura (approved):
Nice improvement!

PR #1188, opened 2022-11-08
add Twitter embed

mattcasey:
Supports dark mode too:

image

### DIFF
--- a/theme/cssVariables.ts
+++ b/theme/cssVariables.ts
@@ -4,6 +4,7 @@ import * as colors from './colors';
 
 const globalCSS = css`
   :root {
+    color-scheme: light;
     ${Object.entries(colors.lightModeColors).map(([key, value]) => `--bg-${key}: ${value};`).join('\n')}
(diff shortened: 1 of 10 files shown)


theme/cssVariables.ts
```diff
@@ -59,6 +60,8 @@ const globalCSS = css`
 
   /* dark theme */
   [data-theme='dark'] {
+    // use auto to avoid opaque background on Twitter iframe ('dark' would be better)
+    color-scheme: auto;
```
mattcasey, line 64: I set color-scheme here instead of using CssBaselnie. The reason is that Twitter's iframe uses 'color-scheme: auto' and if that does not match the parent (at least in Firefox), then it adds an opaque white background, and we see it on the rounded corners.

If this causes other issues, we could just always have 'light' mode for tweets, which is what Notion does anyway

msadura (approved):
interesting feature, nice stuff

Devorein (approved):

PR #1191, opened 2022-11-08
Remove Rinkeby from our config completely

Devorein:


components/[pageId]/DocumentPage/components/BountyProperties/components/BountyApplicantsTable/BountyPaymentButton.tsx
```diff
@@ -234,23 +235,30 @@ export default function BountyPaymentButton ({
     }
   };
 
+  const isChainSupported = !unsupportedChainIds.includes(bounty.chainId);
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/BountyProperties/components/BountyApplicantsTable/BountyPaymentButton.tsx:238]: After this PR, by definition we should not longer need this config 'unsuportedChainIds'. There's lots of other chains we don't support that are already handled (for example i just added Optimism last week)

scripts/transferChain.ts
```diff
@@ -0,0 +1,24 @@
...
+        in: unsupportedChainIds
+      }
+    },
+    // Transfer to goerli
+    data: {
+      chainId: 5
+    }
+  });
+
+  await prisma.paymentMethod.deleteMany({
```
mattcasey, line 17: just want to confirm: we don't link to payment methods right? it's just we let you use them to select a chainId for bounties, but the bounty update above will address that

components/[pageId]/DocumentPage/components/BountyProperties/components/BountyApplicantsTable/BountyPaymentButton.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/BountyProperties/components/BountyApplicantsTable/BountyPaymentButton.tsx:238]: Got it. I will remove those.

scripts/transferChain.ts
Devorein: Exactly, we only use payment method for selecting the chainId. Nothing is linked according to the schema

PR #1199, opened 2022-11-09
Feat/select editable options

msadura:
- creating new slect options inside of input (for admins)
- editing existing options (for admins)

https://user-images.githubusercontent.com/5198394/200868633-1c31997c-578e-42c8-b23a-96d903b7067d.mp4


components/profile/components/SpacesMemberDetails/components/useMutateMemberPropertyValues.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { SelectOptionType } from 'components/common/form/fields/Select/interfaces';
+import useIsAdmin from 'hooks/useIsAdmin';
+import type { PropertyValueWithDetails } from 'lib/members/interfaces';
+
+export function useMutateMemberPropertyValues (
+  mutatePropertyValue: KeyedMutator<PropertyValueWithDetails[]>,
+  spaceId: string | null
+) {
+
+  const isAdmin = useIsAdmin(spaceId || undefined);
```
mattcasey, line 13: I would just make 'null' another option in useIsAdmin, it feels weird to have to pass in `null || undefined` just to make TS happy :)

Kind of tangential, maybe `useUser()` should return a function `isAdmin()` since useIsAdmin is just calling useUser and then a utility method

theme/utils/getThemeColorFromString.ts
```diff
@@ -0,0 +1,15 @@
+import type { BrandColor, SupportedColor } from 'theme/colors';
+import { propertyColorsMap, brandColorNames } from 'theme/colors';
+
+export function getThemeColorFromString (color: string): SupportedColor {
```
mattcasey, line 4: I am not 100% sure I know what this method is needed for. Based on your other PR, I think it is to map focalboard colors (with names like propColorBlue) to the color names from MUI? In that case, I think it belongs inside the BoardEditor folder, and could be a lot more explicit, like `convertFocalboardToMUIColor()` since we should only be using this for focalboard code. It will help when we get around to cleaning up and reducing code for the focalboard theme.

theme/colors.ts
```diff
@@ -65,3 +66,16 @@ export const lightModeColors: BrandColorScheme = {
 };
 
 export const brandColorNames = Object.keys(darkModeColors) as BrandColor[];
+
+export const propertyColorsMap: { [key: string]: SupportedColor } = {
```
mattcasey, line 70: Want to suggest renaming this to "focalboardColorsMap" to be explicit

mattcasey (approved):
Looks good overall - I think I noticed a bug but mentioned it in your other PR applying this to focalboard

PR #1201, opened 2022-11-09
Datadog FE logs

valentinludu:
Made a context for datadog. Maybe in the future we will have more logging logic there and maybe we need to access it someplace else. The simple solution was to add the useEffect inside the UserContext but I prefer the context approach.

Mentions:
- works only for production
- sends only the user.id
- does not record user input. This is recommended by datadog so we don't send by accident some sensitive data written by the user in the editor/input field

### DIFF
--- a/components/_app/DatadogLogger.tsx
+++ b/components/_app/DatadogLogger.tsx
@@ -1,29 +1,21 @@
 import { datadogLogs } from '@datadog/browser-logs';
 import { datadogRum } from '@datadog/browser-rum';
-import { createContext, useContext, useEffect, useMemo } from 'react';
+import { useEffect } from 'react';
 import type { ReactNode } from 'react';
 
-import { useUser } from './useUser';
+import { isProdEnv } from 'config/constants';
 
-type IContext = {
-  datadogLogs: typeof datadogLogs | null;
-  datadogRum: typeof datadogRum | null;
-};
-
-export const LoggerContext = createContext<Readonly<IContext>>({
-  datadogLogs,
-  datadogRum
-});
+import { useUser } from '../../hooks/useUser';
 
 const DD_SITE = 'datadoghq.com';
 const DD_SERVICE = 'charmverseapp';
 
-export function LoggerProvider ({ children }: { children: ReactNode }) {
+export default function DatadogLogger ({ children }: { children: ReactNode }) {
   const { user } = useUser();
 
   // Load DD_LOGS
   useEffect(() => {
-    if (process.env.NEXT_PUBLIC_DD_CLIENT_TOKEN && process.env.NODE_ENV === 'production') {
+    if (process.env.NEXT_PUBLIC_DD_CLIENT_TOKEN && isProdEnv) {
       datadogLogs.init({
         clientToken: process.env.NEXT_PUBLIC_DD_CLIENT_TOKEN,
         site: DD_SITE,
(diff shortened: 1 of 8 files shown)


hooks/useLogger.tsx
```diff
@@ -0,0 +1,93 @@
...
+      datadogRum.onReady(() => {
+        datadogRum.setUser({ id: user.id });
+      });
+    }
+
+    return () => {
+      datadogLogs.clearUser();
+      datadogRum.clearUser();
+    };
+  }, [user]);
```
mattcasey, line 79: `user?.id` would be a little more specific dependency, so things like changing my username won't trigger an update

hooks/useLogger.tsx
```diff
@@ -0,0 +1,93 @@
...
+import type { ReactNode } from 'react';
+
+import { useUser } from './useUser';
+
+type IContext = {
+  datadogLogs: typeof datadogLogs | null;
+  datadogRum: typeof datadogRum | null;
+};
+
+export const LoggerContext = createContext<Readonly<IContext>>({
```
mattcasey, line 13: I don't think you need to create a context - nothing is consuming it, and what you're exporting is already accessible by just importing the datadog log/rum libraries elsewhere.

But I see that you need to call this below the 'UserProvider' component somewhere, so I guess a component still makes sense. But I would rename it to something like "DatadogLogger" and remove the use of React Context

pages/_app.tsx
```diff
@@ -269,28 +270,27 @@ export default function App ({ Component, pageProps }: AppPropsWithLayout) {
...
-                    </PageTitleProvider>
-                  </PrimaryCharmEditorProvider>
-                </PagesProvider>
-              </PaymentMethodsProvider>
-            </BountiesProvider>
-          </MembersProvider>
-        </WebSocketClientProvider>
-      </SpacesProvider>
-
+      <LoggerProvider>
```
mattcasey, line 274: this isn't really providing data, it doesn't really belong inside of this group of components.. but we might have logs there. I just had another idea, will post in my review

mattcasey (changes requested):
First, kind of minor but I think you should change the name of this to "useDatadogLogger". It's better to be explicit since we already have a log utility and i think it could be confusing for someone looking at this. It would also be better to put this in the `components/_app` folder, since it should only be used in _app.tsx. Then I have two different proposals:
1) remove the use of React context, it's not actually used.
2) Keep the Context, but move this component even higher up, as high as it will go in the tree, and instead of exporting the datadog libraries as context, export a `setUserId` method which useUser can call. That way we capture more logs, and userId is optional. user does not exist if you're not logged in, but we would still want those logs

hooks/useLogger.tsx
valentinludu: Done

hooks/useLogger.tsx
valentinludu: Done

pages/_app.tsx
valentinludu: Done

valentinludu:
> First, kind of minor but I think you should change the name of this to "useDatadogLogger". It's better to be explicit since we already have a log utility and i think it could be confusing for someone looking at this. It would also be better to put this in the `components/_app` folder, since it should only be used in _app.tsx. Then I have two different proposals:
> 
> 1. remove the use of React context, it's not actually used.
> 2. Keep the Context, but make the component a parent of UserProvider and instead of exporting the datadog libraries, export a `setUserId` method which useUser can call

I removed context, made it a component and added it as a child of DataProviders.

Devorein (approved):

mattcasey (approved):

components/_app/DatadogLogger.tsx
```diff
@@ -0,0 +1,74 @@
...
+      });
+    }
+
+    return () => {
+      datadogLogs.clearUser();
+      datadogRum.clearUser();
+    };
+  }, [user?.id]);
+
+  return <div>{children}</div>;
```
msadura, line 73: This component does not provide any additional context, so there is no point in wrapping everything else with it because it only brings unnecessary hierarchy complexity.

I see 2 solutions here:
1) Move this component to  `GlobalComponents` - it can basically run its own hooks and return null, because you only care abotu the logic
2) Instead of having DatadogLogger component, you can simply create just a hook `useDatadogLogger` and move logic from this component (useEffects) to hook and just use it in `GlobalComponents` component.

msadura (changes requested):

components/_app/DatadogLogger.tsx
msadura: Ah I also noted comments from Matt - they are pretty much the same I mentioned. I think Matt proposed 2 potential solutions, while you made a mix of them. I would stick to just one (since you removed context already I think there is no point in wrapping everything with DatadogLogger component)

msadura (approved):

PR #1202, opened 2022-11-09
fix: make date editable from table view

HagiaSofiya:
https://app.charmverse.io/charmverse/page-49366552253645923?viewId=44c79006-57dd-4e78-808f-b21ae0323b8c&cardId=4c086e40-ceab-4ea5-be8e-9184b2edf64f

Screenshot 2022-11-09 at 5 08 02 PM

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.scss
+++ b/components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.scss
@@ -43,8 +43,8 @@
     }
 
     .Button {
+        display: flex;
         height: 100%;
-        width: calc(100% - 16px);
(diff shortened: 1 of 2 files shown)


components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.scss
```diff
@@ -43,8 +43,8 @@
     }
 
     .Button {
-        width: 100%;
         height: 100%;
+        width: calc(100% - 16px);
```
msadura, line 47: Just wondering, can't this be based on flex rather than this sort of calculation?

msadura (approved):

PR #1203, opened 2022-11-09
Disable gnosis safes when the token is not native

mattcasey:
image


mattcasey:
closing this because i figured out how to support erc20

PR #1206, opened 2022-11-10
Feat/handle page share

msadura:
https://app.charmverse.io/charmverse/page-2915519139705518

- handling `/share/page-id` url
- showing 404 when spacde-domain is not found
- e2e tests

### DIFF
--- a/components/share/PublicPage.tsx
+++ b/components/share/PublicPage.tsx
@@ -35,6 +35,7 @@ import { usePageTitle } from 'hooks/usePageTitle';
 import { useSpaces } from 'hooks/useSpaces';
 import { useUser } from 'hooks/useUser';
 import { useWeb3AuthSig } from 'hooks/useWeb3AuthSig';
+import type { PublicPageResponse } from 'lib/pages';
 import { findParentOfType } from 'lib/pages/findParentOfType';
 
 import { lowerCaseEqual } from '../../lib/utilities/strings';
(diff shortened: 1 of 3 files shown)


components/share/PublicPage.tsx
```diff
@@ -73,15 +74,25 @@ export default function PublicPage () {
...
+    let foundPage: PublicPageResponse | null = null;
 
     try {
-      foundSpace = await charmClient.getSpaceByDomain(spaceDomain);
-      if (foundSpace) {
-        setSpaces([foundSpace]);
+      if (validate(router.query.pageId?.[0] || '')) {
+        foundPage = await charmClient.getPublicPage(pageIdOrPath);
+        foundSpace = foundPage.space;
+        router.replace(`/share/${foundSpace?.domain}/${foundPage.page.path}`);
```
mattcasey, line 86: Suggestion, maybe for next time: I think this logic would be better to do on the backend (in getServerSideProps). then you can do the redirect before the page even loads. We don't use a lot of server-side next.js, but this would be a good use case.

mattcasey (approved):
Nice little fix... had one suggestion, maybe if you want to try out a new Next.js feature you could move the logic to the backend in getServerSideProps

components/share/PublicPage.tsx
msadura: oh damn, good call! Unfortunately we do not use much of nextjs serverside magic and I keep forgetting about it

PR #1207, opened 2022-11-11
fix discord tests to use unique user id

mattcasey:


.github/workflows/test_and_deploy.yml
```diff
@@ -7,7 +7,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    branches: ['**']
```
mattcasey, line 10: updated so all PRs run tests

PR #1209, opened 2022-11-11
Remove padding for embedded boards

valentinludu:
Padding was added for all boards and we don't need it for boards embedded in pages or other places.

https://app.charmverse.io/charmverse/page-16806054051346364

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
@@ -54,8 +54,11 @@
     // }
     > div {
       position: relative;
-      padding-bottom: var(--page-layout-pb);
       height: 100%;
+
+      .full-page & {
+        padding-bottom: var(--page-layout-pb);
+      }
     }
 
     &.sidebar-visible {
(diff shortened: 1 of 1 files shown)


Devorein (approved):

Devorein:
Probably a bit more padding would be helpful
image

components/common/BoardEditor/focalboard/src/components/centerPanel.scss
```diff
@@ -76,12 +74,19 @@
...
+      .view-actions {
+        opacity: 1;
+      }
+    }
+    .view-actions {
+      opacity: 0;
+    }
+    > .container-container {
+      > div {
+        padding-bottom: 5px;
```
mattcasey, line 88: instead of adding a new rule, please just fix the original CSS to be more specific. Less is always more :)

For example, you can use "full-page" class to specify padding-bottom only on full page databases in centerPanel.scss.

mattcasey:
> Probably a bit more padding would be helpful
No component should add extra whitespace around itself. This breaks the vertical rhythm of components, and is really dependent on the content you are using.

mattcasey (changes requested):
please fix the specificity of the original CSS rule, rather than adding an override

valentinludu:
> Probably a bit more padding would be helpful image

Following Matt's comment I will remove padding and leave the user to add empty lines as spaces if he/she wishes to.

components/common/BoardEditor/focalboard/src/components/centerPanel.scss
valentinludu: Done.

msadura (approved):

mattcasey (approved):

Devorein (approved):

PR #1210, opened 2022-11-11
Standardize the page actions menu

Devorein:


components/common/BoardEditor/focalboard/src/components/kanban/kanbanCard.tsx
```diff
@@ -64,6 +61,10 @@ const CurrencyIcon = styled.span`
   justify-content: center;
 `;
 
+const StyledBox = styled(Box)`
+  ${({ theme }) => hoverIconsStyle({ theme, isTouchScreen: isTouchScreen() })}
```
msadura, line 65: 1) instead of passing `isTouchScreen` here you should just invoke it directly in `hoverIconsStyle` file
2) you should not need to pass theme to `hoverIconsStyle` - it uses `css` which should have access to theme anyway

components/common/PageActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <ListItemText>Copy link</ListItemText>
+      </MenuItem>
+      <MenuItem dense onClick={onClickOpenInNewTab}>
+        <LaunchIcon fontSize='small' sx={{ mr: 1 }} />
+        <ListItemText>Open in new tab</ListItemText>
+      </MenuItem>
+      <Divider />
+      {
+      pageCreator && (
+        <Stack sx={{
```
msadura, line 101:
```suggestion
        <Stack px={1}
```

components/bounties/components/BountyCard.tsx
```diff
@@ -14,11 +21,31 @@ interface Props {
   onClick?: () => void;
 }
 
+const StyledBox = styled(Box)`
+  ${({ theme }) => hoverIconsStyle({ theme, isTouchScreen: isTouchScreen() })}
```
mattcasey, line 25: Suggestion for hoverIconsStyle: I hadn't seen this util before, but it could just be a styled component itself. That reduces a lot of boilerplate:

```
const HoverIconWrapper = styled(Box)<{ isTouchScreen: boolean }>`
  ...
`;
const StyledBox = styled(IconWrapper);

```

components/bounties/components/BountyCard.tsx
```diff
@@ -44,7 +71,31 @@ function BountyCard ({ bounty, page, onClick }: Props) {
           <BountyStatusBadge bounty={bounty} hideStatus={true} truncate />
         </Box>
       </Box>
-    </Box>
+      <IconButton
+        size='small'
+        className='icons'
+        sx={{
```
mattcasey, line 77: you're already using classname to style this, i would keep using and it not use sx. this keeps the styling rules in one place, and potentially more reusable.

components/common/PageActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+export function PageActions ({
+  open,
+  pageId,
+  onClickDelete,
+  pageUpdatedAt,
+  pageCreatedBy,
+  anchorEl,
+  onClick,
+  children,
+  pageType
```
mattcasey, line 28: This is a lot of props to map everywhere, I would suggest either: just pass pageId in and call usePages() to get the rest of the data, OR, if performance is a concern, replace them all with a 'page' prop which just conveniently mirrors the page object model

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -126,38 +122,111 @@ export default function Header ({ open, openSidebar }: HeaderProps) {
...
+        <ListItemText primary='Copy link' />
       </ListItemButton>
+      <Divider />
+      <Tooltip title={!pagePermissions.delete ? 'You don\'t have permission to delete this page' : ''}>
+        <div>
+          <ListItemButton
+            disabled={!pagePermissions.delete}
+            onClick={onDeletePage}
+          >
+            <DeleteIcon
```
mattcasey, line 219: use DeleteOutlined instead of the solid trash can (and outlined versions wherever we can)

components/common/PageActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+        <ListItemText>Open in new tab</ListItemText>
+      </MenuItem>
+      <Divider />
+      {
+      pageCreator && (
+        <Stack sx={{
+          px: 1
+        }}
+        >
+          <Typography variant='subtitle2'>
```
mattcasey, line 105: use variant 'caption' and color 'secondary', and please fix the padding so that it lines up vertically with the lines above (attached example of Notion, the padding on left, bottom, and right is consistent around the whole dropdown):
image

components/bounties/components/BountyCard.tsx
```diff
@@ -44,7 +71,31 @@ function BountyCard ({ bounty, page, onClick }: Props) {
...
+          top: 0,
+          right: 0,
+          p: 1,
+          m: 1
+        }}
+        onClick={handleClick}
+      >
+        <MoreHorizIcon color='secondary' fontSize='small' />
+      </IconButton>
+      <PageActions
```
mattcasey, line 88: PageActions should open out to the right side when it's on a card:
image

components/proposals/components/ProposalActionsMenu.tsx
```diff
@@ -60,7 +62,7 @@ export default function ProposalActionsMenu ({ deleteProposal, editProposal, pro
             <ListItemText>Delete proposal</ListItemText>
           </MenuItem>
         )}
-      </Menu>
+      </PageActions>
```
mattcasey, line 65: it looks like you can get rid of 'deleteProposal' code and just pass in `onClickDelete`. The only difference will be "Delete proposal" vs "Delete", but it's better to be consistent

components/common/PageActions.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export function PageActions ({
+  open,
+  pageId,
+  onClickDelete,
+  pageUpdatedAt,
+  pageCreatedBy,
+  anchorEl,
+  onClick,
+  children,
```
mattcasey, line 27: I would remove 'children' as a prop, and instead just add optional props for 'duplicatePage', 'editPage', etc. This reduces code, makes the API simpler, and ensures consistency whenever we show delete, duplicate, etc.

components/bounties/components/BountyCard.tsx
```diff
@@ -44,7 +71,31 @@ function BountyCard ({ bounty, page, onClick }: Props) {
...
       </Box>
-    </Box>
+      <IconButton
+        size='small'
+        className='icons'
+        sx={{
+          position: 'absolute',
+          top: 0,
+          right: 0,
+          p: 1,
```
mattcasey, line 81: looks better with less padding, please remove p: 1

mattcasey (changes requested):
Please fix my comments about UI appearance. The code cleanup can be done later (plus some things like hoverIconStyle would change a lot of other files), I think this is a great start in the right direction.

I also noticed you didn't update the menu in the PageDialog component. Do you want to save that for later? I see the menu in Header isn't using PageActions, but maybe that could be extracted and used in PageDialog since it is pretty different.

components/common/BoardEditor/focalboard/src/components/kanban/kanbanCard.tsx
mattcasey: I noted someplace below, maybe this could just be its own styled component

components/bounties/components/BountyCard.tsx
Devorein: The problem is that not all the components where `hoverIconsStyle` is used is a `Box`. So we would end up with two extra tags, if we go this route (when for example we need a `TableRow`)

components/common/BoardEditor/focalboard/src/components/kanban/kanbanCard.tsx
Devorein: Going with Marek's suggestion as that reduces the boilerplate code
```tsx
const StyledBox = styled(Box)`
  ${hoverIconsStyle}
`;
```

components/bounties/components/BountyCard.tsx
msadura: hmm if that is the only issue, then propably `withComponent` will solve it: https://emotion.sh/docs/styled#change-the-rendered-tag-using-withcomponent

components/bounties/components/BountyCard.tsx
Devorein: Updated. Thank you for the suggestion.

components/common/PageActions.tsx
Devorein: Thank you. I've replaced all those props with `page: PageMeta`

components/common/PageLayout/components/Header/Header.tsx
Devorein: Thank you. I've used it everywhere in page header.

components/common/PageActions.tsx
Devorein: Updated thnx.

components/bounties/components/BountyCard.tsx
Devorein: Thnx updated.

components/proposals/components/ProposalActionsMenu.tsx
Devorein: Updated thnx.

components/common/PageActions.tsx
Devorein: Updated thnx.

components/bounties/components/BountyCard.tsx
Devorein: Updated thnx.

mattcasey (changes requested):
Some bugs I noticed:
- [ ] make sure we're using DeleteOutline for all Delete icons (we're not in the Header component)
- [ ] rename "Add to favorite" -> "Add to favorites"
- [ ] remove "Create a poll" completely, Alex/Chris agreed we are giving too many options
- [ ] make all pages exportable via Markdown except 'board' page types
- [ ] deleting a board page doesn't update the UI so you see the red banner until I refresh the page
- [ ] replace View Feedback and keep the 3 links: View comments/polls/suggestions for now. It also doesn't seem to do anything currently (sidebar doesn't open?)

components/bounties/components/BountyCard.tsx
mattcasey: @Devorein I actually don't think you need a box either, just use a div. if someone's using styled components they really shouldn't be using box properties too

PR #1211, opened 2022-11-11
Disable submit snapshot btn after submitting

valentinludu:
If the user changes the snapshot eth address, the button should be disabled after submitting it.
Also if the user deletes the input value and adds exactly the same value, the button will be disabled.

https://app.charmverse.io/charmverse/page-8924516118275596

### DIFF
--- a/components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
+++ b/components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
@@ -61,6 +61,8 @@ export default function ConnectSnapshot () {
 
   const values = watch();
 
+  const snapshotIsDefault = space?.snapshotDomain === values.snapshotDomain;
+
   async function onSubmit (formValues: FormValues) {
 
     setFormError(null);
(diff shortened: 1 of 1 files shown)


components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
```diff
@@ -61,6 +61,8 @@ export default function ConnectSnapshot () {
 
   const values = watch();
 
+  const snapshotIsDefault = space?.snapshotDomain === values.snapshotDomain;
```
motechFR, line 64: I think this variable could be clearer.

How about

const snapshotDomainUnchanged = space?.snapshotDomain === values.snapshotDomain;

I think this will facilitate future maintenance :)

motechFR (approved):

components/common/PageLayout/components/Header/components/Snapshot/ConnectSnapshot.tsx
valentinludu: Agree. Just updated

PR #1215, opened 2022-11-13
Member directory immutable join date property

Devorein:


components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -123,6 +124,18 @@ function MemberDirectoryGalleryCard ({
                   );
                 }
 
+                case 'join_date': {
```
msadura, line 127: why type of data is called "join_date" ? I believe data type should be just `date` and in case of "join_date" field it should be just filled with a join date on the backed side. This way we can have universal data type we can reuse in the future.

lib/members/getSpacesPropertyValues.ts
```diff
@@ -25,6 +27,10 @@ export async function getSpacesPropertyValues ({ memberId, requestingUserId, spa
     const spaceRolesMap = await getSpaceMemberRoles({ spaceIds, memberId });
     propertyValues = propertyValues.map(pv => pv.type === 'role' ? { ...pv, value: spaceRolesMap[pv.spaceId]?.map(r => r.name) || [] } : pv);
   }
+  if (visibleMemberProperties.find(mp => mp.type === 'join_date')) {
+    const joinDatesMap = await getSpaceMemberJoinDates({ spaceIds, memberId });
```
msadura, line 31: This logic adds new db query, which opersates on the same data as `getSpaceMemberRoles` above. I think it would be betters to make a common way on adding roles and join dates when needed, but based on 1 query which would fetch data once and jusrt reuse that data to add roles and (or) join dates

msadura (changes requested):

lib/members/getSpacesPropertyValues.ts
Devorein: Agree. I've updated the code so that its all done in a single query. Also made sure all the tests pass.

components/members/components/MemberDirectoryGalleryView.tsx
Devorein: Most of the read only properties are actually joined on the frontend, for example, `name, discord, twitter, timezone` etc🤔

lib/members/constants.ts
```diff
@@ -79,5 +84,6 @@ export const DEFAULT_MEMBER_PROPERTIES_ORDER = [
   'bio',
   'discord',
   'twitter',
-  'timezone'
+  'timezone',
```
mattcasey: @Devorein this "DEFAULT_MEMBER_PROPERTIES_ORDER" is unecessary, why can't we just use the order that's in the map?

Devorein: The order in the map is not guaranteed to be in the same order they are defined right? That was the only reason.

lib/members/getSpaceMemberJoinDates.ts
```diff
@@ -0,0 +1,28 @@
...
+    }
+  });
+
+  const joinDatesMap = spaceRoles.reduce(
+    (acc, spaceRole) => {
+      const joinDate = spaceRole.createdAt;
+      acc[spaceRole.spaceId] = joinDate;
+      return acc;
+    },
+   {} as Record<string, Date>
```
mattcasey: Using generics is the correct way to type reduce:
```
const result = array.reduce<Record<string, Date>>(reduceFunction, {});

```

scripts/addJoinDateMemberDirectoryProperty.ts
```diff
@@ -0,0 +1,21 @@
...
+async function addJoinDateMemberDirectoryProperty () {
+  const spaces = await prisma.space.findMany({
+    orderBy: {
+      createdAt: "asc"
+    }
+  });
+
+  await prisma.memberProperty.createMany({
+    data: spaces.map(space => ({
+      createdBy: space.createdBy,
```
mattcasey, line 12: what will the sort order of this be? won't it suddenly appear at the top since it will be -1?

msadura (approved):

scripts/addJoinDateMemberDirectoryProperty.ts
Devorein: Thnx for catching that. I will add the index when creating the join_date property

mattcasey (approved):

PR #1216, opened 2022-11-14
Sshrayber figma

HagiaSofiya:
https://app.charmverse.io/charmverse/page-49366552253645923?viewId=44c79006-57dd-4e78-808f-b21ae0323b8c&cardId=bd01584a-09f2-45e3-a887-0b00538940f2

Screenshot 2022-11-14 at 12 22 54 AM

Material UI does not have a Figma Icon. I used Feather icons via [React Icons](https://react-icons.github.io/react-icons).

### DIFF
--- a/components/common/CharmEditor/components/iframe/IframeComponent.tsx
+++ b/components/common/CharmEditor/components/iframe/IframeComponent.tsx
@@ -254,7 +254,7 @@ function ResizableIframe ({ readOnly, node, updateAttrs, onResizeStop }:
   }
   else if (node.attrs.type === 'figma') {
 
-    const src = `https://www.figma.com/embed?embed_host=astra&url=${node.attrs.src}`;
+    const src = `https://www.figma.com/embed?embed_host=charmverse&url=${node.attrs.src}`;
     return (
       <BlockAligner onDelete={onDelete}>
         <VerticalResizer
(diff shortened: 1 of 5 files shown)


msadura (approved):
Looks good! I would just see @mattcasey input on adding `react-icons`. Personally I like that package and I think we can use some of the icons in the future.

components/common/CharmEditor/components/iframe/IframeComponent.tsx
```diff
@@ -226,6 +252,35 @@ function ResizableIframe ({ readOnly, node, updateAttrs, onResizeStop }:
       </BlockAligner>
     );
   }
+  else if (node.attrs.type === 'figma') {
```
mattcasey, line 255: you need to also handle the 'readOnly' case. Try switching the editor mode to 'Viewing' you will see that it ends up as a broken link

mattcasey (changes requested):
- Works well, and even supports login!
- Blocking: needs to handle 'readonly' mode
- I am fine adding a new icon framework as long as we aren't uploading the entire st of icons. we should let the whole dev team know this is now available
- I'd like to discuss in a call whether 'figma' should just be a new type on 'embed' or its own Prosemirror plugin type (similar to 'tweet'). I find it's cleaner if we dont have subtypes, and was planning on extracting video out of the current embed type.

components/common/CharmEditor/components/iframe/IframeComponent.tsx
HagiaSofiya: @mattcasey Updated!

PR #1218, opened 2022-11-14
Cards should not re-appear when switching board views

Devorein:


valentinludu:
I found a case that is not covered:
Open a link with a popup -> close that popup.
Click on the + sign to create another type of board -> click Table or Gallery.
The popup will appear.
Screenshot 2022-11-14 at 19 08 13

PR #1219, opened 2022-11-14
Make page updates via web sockets more efficient

mattcasey:
- dont re-query pages on updates
- dont send contents update to web sockets

### DIFF
--- a/hooks/usePages.tsx
+++ b/hooks/usePages.tsx
@@ -235,6 +235,8 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
         ...(existingPages ?? {}),
         ...pagesToUpdate
       };
+    }, {
+      revalidate: false
     });
   }, []);
 
@@ -254,6 +256,8 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
         ...(existingPages ?? {}),
         ...newPages
       };
+    }, {
+      revalidate: false
     });
   }, []);
 
@@ -268,6 +272,8 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
       });
 
       return newValue;
+    }, {
+      revalidate: false
     });
   }, []);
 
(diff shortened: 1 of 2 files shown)


hooks/usePages.tsx
```diff
@@ -215,26 +215,30 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
   }
 
   const handlePagesUpdate = useCallback((value: WebsocketPayload<'pages_meta_updated'>) => {
-    const pagesToUpdate = value.reduce((pageMap, updatedPageMeta) => {
 
-      const existingPage = pages[updatedPageMeta.id];
```
mattcasey, line 220: Pretty sure that pages could be out of sync, so i moved the logic to check existingPages below

pages/api/pages/[id]/index.ts
```diff
@@ -117,22 +117,21 @@ async function updatePageHandler (req: NextApiRequest, res: NextApiResponse<IPag
...
 
-  if (hasNewParentPage) {
-    const updatedPage = await setupPermissionsAfterPageRepositioned(pageId);
-    return res.status(200).json(updatedPage);
-  }
+  const { content, contentText, ...updatedPageMeta } = req.body as Page;
 
-  // Update page track profile, unless it was content update
+  // Update page track profile and meta data state, unless it was content update
   if (!('content' in updateContent)) {
```
Comment (motechFR) [inline pages/api/pages/[id]/index.ts:123]: Are we sure that there is never page meta and content in same update?

i think a more resilient way might be to say (“if title, index or parentId”)

motechFR (approved):
Looks good

msadura (approved):

pages/api/pages/[id]/index.ts
Comment (mattcasey) [reply pages/api/pages/[id]/index.ts:123]: yes, we are (also we were already making this assumption when updating mixpanel pages). it will be even moreso once web sockets are deployed

PR #1223, opened 2022-11-15
Update bounty notifications script

valentinludu:


mattcasey (approved):
I noticed we were not checking that 'action' was not null in the tasks, so I made an update for that as well

PR #1224, opened 2022-11-15
Fix for Gnosis safes + bounties

mattcasey:
An error is occurring inside the MultiPaymentModal component after a payment is made:
https://www.loom.com/share/83dad81337584e38a0ed02d6bd53080e

This fix addresses two issues:
1) we were creating a map of transactions based on 'applicationId'. I think it was somehow out of sync the the state of bounties, and so when we looked up the transaction, it was undefined
2) I noticed we were assuming the first gnosis safe in the payment button's onPaymentSuccess handler, which also is run/used when you pay from a single bounty. This is now fixed

### DIFF
--- a/components/bounties/components/MultiPaymentModal.tsx
+++ b/components/bounties/components/MultiPaymentModal.tsx
@@ -10,7 +10,6 @@ import Button from 'components/common/Button';
 import { DialogTitle, Modal } from 'components/common/Modal';
 import UserDisplay from 'components/common/UserDisplay';
 import { useMembers } from 'hooks/useMembers';
-import type { TransactionWithMetadata } from 'hooks/useMultiBountyPayment';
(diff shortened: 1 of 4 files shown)


components/bounties/components/MultiPaymentModal.tsx
```diff
@@ -182,9 +185,7 @@ export default function MultiPaymentModal ({ bounties }: { bounties: BountyWithD
               <MultiPaymentButton
                 chainId={gnosisSafeChainId}
                 safeAddress={gnosisSafeAddress}
-                transactions={selectedApplicationIds.map(selectedApplicationId => (
-                  applicationTransactionRecord[selectedApplicationId](gnosisSafeAddress)
```
mattcasey, line 186: this is the line that was triggering an exception. Minified, it was `F[e] is not a function` which is also the error you get if it was undefined

PR #1228, opened 2022-11-16
Make all cards same height in gallery view

Devorein:


components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -60,6 +60,8 @@ function MemberDirectoryGalleryCard ({
         href={`/u/${member.path || member.id}`}
         color='primary'
         sx={{
+          height: '100%',
```
msadura, line 63: Since `StyledLink` is already a styled component please move all styles there

msadura (approved):

PR #1230, opened 2022-11-16
Fix double click issue on iOS

motechFR:
We were forced to double click links to navigate elsewhere to the app inside iOS.

The specific issue came from a component inside Kanban.tsx that was preventing click events being propagated


mattcasey (approved):
it looks like that library is used to scroll the window when dragging a compontent from react-dnd... i think this is from focalboard's original code and we have changed the container component

PR #1231, opened 2022-11-16
Feat/select options editable by members

msadura:
https://app.charmverse.io/charmverse/page-3557311048353411

### DIFF
--- a/jest.config-server.ts
+++ b/jest.config-server.ts
@@ -5,8 +5,8 @@ import { jestConfig } from './jest.config';
 const integrationConfig = {
   ...jestConfig,
   testMatch: [
-    '**/__integration-tests__/**/*.spec.ts',
-    '**/lib/**/*.spec.ts'
+    '**/__integration-tests__/**/*.spec.ts'
+    // '**/lib/**/*.spec.ts'
   ]
 };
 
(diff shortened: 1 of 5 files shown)


jest.config-server.ts
```diff
@@ -5,8 +5,8 @@ import { jestConfig } from './jest.config';
 const integrationConfig = {
   ...jestConfig,
   testMatch: [
-    '**/__integration-tests__/**/*.spec.ts',
-    '**/lib/**/*.spec.ts'
+    '**/__integration-tests__/**/*.spec.ts'
+    // '**/lib/**/*.spec.ts'
```
Devorein, line 9: Yikes we need to uncomment this 👀

lib/members/updateMemberProperty.ts
```diff
@@ -13,6 +13,33 @@ type UpdatePropertyInput = {
...
+  });
+
+  const memberProperty = await prisma.memberProperty.findUnique({
+    where: {
+      id
+    }
+  });
+
+  if (!spaceRole) {
+    throw new InvalidInputError('Invalid workspace data.');
```
Devorein, line 30: If the user is not a member of the workspace its better to throw `UserIsNotSpaceMemberError`
image

Devorein (changes requested):

jest.config-server.ts
msadura: actually it was commented all the time, I just uncommented this in my previous pr by mistake

Devorein (approved):

jest.config-server.ts
mattcasey: I think I asked this once before, can we just remove this line?

msadura: We can probably remove it, but I do not see other way to run `libs/...` tests via vscode jest runner extension. Maybe I am just missing some piece of the puzzle.

PR #1232, opened 2022-11-16
Handle empty value for gallery and table view in member directory

Devorein:


components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -115,30 +115,30 @@ function MemberDirectoryGalleryCard ({
               }
               switch (property.type) {
                 case 'bio': {
-                  return (
+                  return member.profile?.description && (
                     <Stack key={property.id}>
```
msadura, line 119: While fiddling with this part of a code, can you also add keys to Each proeprty type? Console warns:

Screenshot 2022-11-16 at 14 34 47

msadura (approved):

components/members/components/MemberDirectoryGalleryView.tsx
Devorein: Yes absolutely. Thank you for reporting. Will do asap

PR #1234, opened 2022-11-16
Lowercase accounts

motechFR:
Force Web3 accounts to be lowercase in the client and server to make things simpler


mattcasey (approved):

motechFR:
This may cause an issue for users logging in. I suggest we use a direct SQL operation for this one as manually calling prisma.update for each address will take too long.

After research, I tested this command locally and it worked locally (using psql client)

```SQL
UPDATE "UserWallet" SET address=LOWER(address);
```

PR #1237, opened 2022-11-16
Test send notifications background task

valentinludu:


jest.config-server.ts
```diff
@@ -6,8 +6,10 @@ const integrationConfig = {
   ...jestConfig,
   testMatch: [
     '**/__integration-tests__/**/*.spec.ts',
-    '**/lib/**/*.spec.ts'
+    '**/lib/**/*.spec.ts',
+    '**/background/**/*.spec.ts'
```
mattcasey, line 10: these tests should only be run with jest.config

mattcasey (changes requested):
looks good but one comment. feel free to merge once you remvoe it from config-server (the lib tests were accidentally included.. this config should only be running integration tests on the api endpoints)

jest.config-server.ts
valentinludu: Removed
'**/lib/**/*.spec.ts',
'**/background/**/*.spec.ts'
from 
jest.config-server.ts

mattcasey (approved):

PR #1238, opened 2022-11-16
Show more for multi-text property

Devorein:
[Add "Show More" for long text](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=b29383fb-f66a-412a-a5b7-0200c6914fd4)


mattcasey (approved):
look sgood.. i think you could reduce the max chars even further, maybe 100 or 150?

PR #1241, opened 2022-11-16
add secrets for S3

mattcasey:
I also want to replace 'stg' with 'prd' as default. The last two devs that had to add variables were confused that we use /prd/ in ebextensions_cron and asked me where to set prd for the webapp.


.ebextensions/00_env_vars.config
```diff
@@ -4,23 +4,27 @@
...
-        COMPOSE_PROFILES: "stg"
-        IMGTAG: ""
\ No newline at end of file
+        NODE_ENV: "prod"
+        NPM_SCRIPT: "start:prod"
+        COMPOSE_PROJECT_NAME: "prd"
+        COMPOSE_PROFILES: "prd"
+        IMGTAG: ""
+        S3_UPLOAD_BUCKET: "{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:bucket}}"
+        S3_UPLOAD_KEY: "{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:access_key}}"
```
yiwork, line 28: you don't need this if you have instances have IAM profile that permits access to the bucket

.ebextensions/00_env_vars.config
```diff
@@ -4,23 +4,27 @@
...
+        S3_UPLOAD_REGION: "{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:region}}"
+        S3_UPLOAD_SECRET: "{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:access_key_secret}}"
```
yiwork, line 30: Ditto point above

yiwork (changes requested):
remove the s3 access keys unless this is a bucket we don't own

mattcasey:
We decided offline that using client key/secret is fine for now, Connie will see if she can get generating the federated token to work using instance roles instead

PR #1246, opened 2022-11-17
Edited/created by

valentinludu:
https://app.charmverse.io/charmverse/page-5997063735675179

Confirmed with Chris the design is ok.

Interactions are done similar to Notion's site:
- when you hover, relativeDates refresh
- if you don't hover, you see the relativeDate from when the component was mounted last time (loading of the page or rerender after a user edited)


components/common/PageLayout/components/DocumentHistory/DocumentHistory.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+function DocumentHistory ({ page }: { page: PageMeta }) {
+  const updatedAtMemo = useMemo(() => getRelativeTime(new Date(page.updatedAt)), [page.updatedAt]);
+  const createdAtMemo = useMemo(() => getRelativeTime(new Date(page.createdAt)), [page.createdAt]);
+
+  const [{ updatedAt, createdAt }, setTime] = useState({ updatedAt: updatedAtMemo, createdAt: createdAtMemo });
+
+  const { members } = useMembers();
+
+  const createdBy = members.find(member => member.id === page.createdBy)?.username ?? 'unknown';
```
motechFR, line 22: For user-friendliness, let's update this label to "Unknown user"

components/common/PageLayout/components/DocumentHistory/DocumentHistory.tsx
```diff
@@ -0,0 +1,49 @@
...
+    setTime((prevState) => ({ ...prevState, updatedAt: updatedAtMemo }));
+  }, [updatedAtMemo]);
+
+  return (
+    <Tooltip
+      enterTouchDelay={0}
+      enterDelay={0}
+      arrow={false}
+      placement='bottom-start'
+      onOpen={() => setTime({ updatedAt: getRelativeTime(new Date(page.updatedAt)), createdAt: getRelativeTime(new Date(page.createdAt)) })}
```
motechFR, line 35: This seems a bit heavy along with the useEffect and useMemo

Why not simplify to just

```typescript
const updatedAt = useMemo(() => getRelativeTime(new Date(page.updatedAt)), [page.updatedAt]);
 
const createdAt = useMemo(() => getRelativeTime(new Date(page.createdAt)), [page.createdAt]);

```

lib/utilities/getRelativeTime.ts
```diff
@@ -0,0 +1,21 @@
+export const getRelativeTime = (date: Date): string => {
```
motechFR, line 1: We already have a getTimeDifference method in lib/utilities/dates.ts

Would that fulfill your purposes?

If not can you please move this to lib/utilities/dates file>

motechFR (approved):
Works fine for the purpose! I have some requests around simplification, could you please address those before merging?

components/common/PageLayout/components/DocumentHistory/DocumentHistory.tsx
valentinludu: Done

lib/utilities/getRelativeTime.ts
valentinludu: Unfortunately those functions are not suited for what I need.

I added it in the dates file. I also added a description so it is more clear what is the difference between them.

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -77,6 +80,7 @@ export default function Header ({ open, openSidebar }: HeaderProps) {
   const isExportablePage = pageType === 'card' || pageType === 'page' || pageType === 'proposal';
 
   const isBountyBoard = router.route === '/[domain]/bounties';
+  const isSmall = useMediaQuery((_theme: Theme) => _theme.breakpoints.down('sm'));
```
mattcasey, line 83: This.. look really expensive, it's somehow querying the browser width and converting it to a boolean? Instead it woudl be better to use a styled component or even the Box component:
```
<Box display={{ xs: 'none', sm: 'block' }} component='span'>...</Box>
```

mattcasey (changes requested):
remove use of useMediaQuery

components/common/PageLayout/components/Header/Header.tsx
valentinludu: Good point.

I removed useMediaQuery and added the display rule inside the Tooltip -> Typography item.

mattcasey (approved):

PR #1248, opened 2022-11-17
Auto close sidebar on mobile

motechFR:
When a sidebar link is clicked on mobile, it should auto close


mattcasey (approved):
nice UX touch, one thing that might be cleaner is if you watched the router object in Sidebar and just closed the popup when router.path changes

PR #1249, opened 2022-11-17
Feat/daylight integration

msadura:
Integration with daylight abilities api

https://app.charmverse.io/charmverse/page-880216490409613


lib/token-gates/daylight.ts
```diff
@@ -0,0 +1,188 @@
...
+      title: `Join CharmVerse's ${space.name} workspace`,
+      type: 'access',
+      isActive: true,
+      // Daylight's sourceId will be the same as tokenGate id
+      sourceId: tokenGate.id
+    })
+  };
+
+  try {
+    const res = await fetch('https://api.daylight.xyz/v1/abilities', params);
```
mattcasey, line 45: FYI for next time maybe: we have a utility to make workign with fetch easier
```
import fetch from 'adapters/http/fetch.server';
fetch('https://api.daylight.xyz/v1/abilities', params);
```
I'd like it to do more (like the browser-equivalent http/fetch has GET, POST, etc. with json headers added automatically) but haven't had time

lib/token-gates/daylight.ts
```diff
@@ -0,0 +1,188 @@
...
+      };
+    }
+
+    default: {
+      return null;
+    }
+  }
+}
+
+function getDaylightRequirements (conditionsData: TokenGateAccessConditions) {
```
mattcasey, line 155: seems like a good candidate for a test

lib/token-gates/daylight.ts
```diff
@@ -0,0 +1,188 @@
...
+  return {
+    requirements: hasInvalidRequirements ? [] : requirements,
+    operator: conditionsOperator
+  };
+}
+
+function getActionUrl (spaceDomain: string) {
+  // Daylight will not allow to create ability with action url pointing to localhost
+  // for testing we can sue our main domain
+  const base = baseUrl.includes('localhost') ? 'https://charmverse.io' : baseUrl;
```
mattcasey, line 185: i think you meant to use app.charmverse.io?

mattcasey (changes requested):
left some comments, but i think the most important one is just making sure the domain is correct. second would be a couple unit tests, the fetch comment is easy to change later

msadura:
@mattcasey  this is ready to re-review.

jest.config-all.ts
```diff
@@ -0,0 +1,14 @@
+import createJestConfig from 'testing/createJestConfig';
```
msadura, line 1: separate config for running tests via vscode jest runner - includes lib and integration tests

lib/token-gates/daylight.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  const params = {
+    method: 'PUT',
+    headers: HEADERS,
+    body: JSON.stringify({
+      isActive: false
+    })
+  };
+
+  try {
```
msadura, line 66: try/catch here because we want to ignore case when something goes wrong - ability simply will not be created

mattcasey (approved):

PR #1250, opened 2022-11-17
Handle UX arriving in workspace with 0 pages

Devorein:
[Handle UX arriving in workspace with 0 pages](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=8ce4de7b-b2b1-4087-830f-0acf30ed4f91)


pages/api/spaces/[id]/pages.ts
```diff
@@ -27,7 +31,56 @@ async function getPages (req: NextApiRequest, res: NextApiResponse<IPageWithPerm
...
+        },
+        include: {
+          ...includePagePermissionsMeta()
+        }
+      }) as IPageWithPermissions;
+
+      await setupPermissionsAfterPageCreated(createdPage.id);
+      createdPages.push(createdPage);
+    }
+    // else {
```
Comment (mattcasey) [inline pages/api/spaces/[id]/pages.ts:63]: please remove commented section

pages/api/spaces/[id]/pages.ts
```diff
@@ -27,7 +31,56 @@ async function getPages (req: NextApiRequest, res: NextApiResponse<IPageWithPerm
...
+  if (accessiblePages.length === 0 && !search) {
+    const totalPages = await prisma.page.count({
+      where: {
+        spaceId
+      }
+    });
+
+    if (totalPages === 0) {
+      const createdPage = await createPage({
+        data: {
```
Comment (mattcasey) [inline pages/api/spaces/[id]/pages.ts:46]: please use the method untitledPage() from 'seedData'... and btw, it looks like that is the only function we use from there anymore, even though it's 1000 lines. looks like we could delete seedData.ts and just add the function to its own file inside of lib/pages

pages/api/spaces/[id]/pages.ts
Comment (Devorein) [reply pages/api/spaces/[id]/pages.ts:63]: Removed

pages/api/spaces/[id]/pages.ts
Comment (Devorein) [reply pages/api/spaces/[id]/pages.ts:46]: Updated.

hooks/usePages.tsx
```diff
@@ -77,11 +80,23 @@ export function PagesProvider ({ children }: { children: ReactNode }) {
...
+    }
+
     return pagesDict;
   }, { refreshInterval });
 
   const pages = data || {};
 
+  useEffect(() => {
+    if (ref.current?.empty && currentSpace) {
+      router.push(`/${currentSpace.domain}/members`);
```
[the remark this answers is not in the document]
Devorein: Nope as that code is run only when the user lands in the workspace domain, what if the path is `/pages/[domain]/[pageId]`.

PR #1254, opened 2022-11-18
Update comments and votes if in viewing mode

valentinludu:
https://app.charmverse.io/charmverse/page-9709252337021876

### DIFF
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -137,7 +137,12 @@ export function charmEditorPlugins (
         }
         return {
           update: (view, prevState) => {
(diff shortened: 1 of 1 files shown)


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -137,7 +137,12 @@ export function charmEditorPlugins (
         }
         return {
           update: (view, prevState) => {
-            if (!readOnly && onContentChange && !view.state.doc.eq(prevState.doc)) {
+            if (
+              // Update only if page is in editing mode or in viewing mode
+              (!readOnly || (enableComments && enableVoting))
```
mattcasey, line 142: it should probably be enableComments || enableVoting, but I think they both rely on the same permission so this should be fine

mattcasey (approved):
just see my comment, not sure if i am correct or not

PR #1255, opened 2022-11-18
auto-detect figma embeds when using url embed

mattcasey:


components/common/CharmEditor/components/iframe/IframeComponent.tsx
```diff
@@ -19,44 +18,11 @@ import BlockAligner from '../BlockAligner';
...
 
 const name = 'iframe';
 
-interface DispatchFn {
-  (tr: Transaction): void;
-}
-
-// inject a real iframe node when pasting embed codes
-
-export const iframePlugin = new Plugin({
```
mattcasey, line 32: Getting rid of this plugin that we are not using

PR #1256, opened 2022-11-19
Convert to proposal page menu option

Devorein:


components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -143,6 +148,23 @@ export default function Header ({ open, openSidebar }: HeaderProps) {
     setPageMenuOpen(false);
   }
 
+  async function convertToProposal () {
+    if (currentSpace && user && basePage) {
+      await charmClient.pages.updatePage({
```
mattcasey, line 153: Offline we discussed not changing the type. We should just be creating a proposal properties, that way this page doesnt disappear from the sidebar

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -143,6 +148,23 @@ export default function Header ({ open, openSidebar }: HeaderProps) {
...
+        type: 'proposal',
+        id: basePage.id
+      });
+
+      mutatePage({
+        id: basePage.id,
+        type: 'proposal'
+      });
+
+      refreshPage(basePage.id);
```
mattcasey, line 163: why are we calling mutatePage AND refreshPage? mutatePage seems to be the only thing we need?

components/common/PageLayout/components/Header/Header.tsx
mattcasey: And I think this deserves its own endpoint, since we need logic to create proposal attributions. you're not just changing a property on the page object

mattcasey: OK, Chris says it SHOULD remove from the workspace, so type = 'proposal' it is

pages/api/pages/[id]/index.ts
```diff
@@ -99,6 +104,36 @@ async function updatePageHandler (req: NextApiRequest, res: NextApiResponse<IPag
     }
   }
 
+  if (updateContent.type === 'proposal') {
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:107]: please put this logic in a new endpiont. I think it is doing a lot of extra work... for example, what if someone is updating an existing proposal and just happens to pass in 'type'?

pages/api/pages/[id]/index.ts
```diff
@@ -99,6 +104,36 @@ async function updatePageHandler (req: NextApiRequest, res: NextApiResponse<IPag
...
+      allowAdminBypass: true,
+      resourceId: page.spaceId,
+      userId
+    });
+
+    if (!spacePermissions.createVote) {
+      throw new UnauthorisedActionError('You do not have permission to create a page in this space');
+    }
+    else {
+      await createProposal({
```
Comment (mattcasey) [inline pages/api/pages/[id]/index.ts:118]: We don't want to create a new page, just convert the existing one

mattcasey (changes requested):
Convert to proposal:
- please put the logic in a new endpoint. this keeps our api easier to unerstand and prevents weird issues that someone working on the frontend might not expect
- According to Chris, ideally the type of the existing page becomes 'proposal' so that it disappears from the sidebar completely and goes into proposals.

mattcasey (approved):
I implemented the changes I proposed

PR #1263, opened 2022-11-22
Feat/super api token

msadura:
https://app.charmverse.io/charmverse/page-4089706117765626

TODO:
- more tests for logic (verify domain, get unique domain)

Works:
- securing endpoint by `requireSuperApiKey`
- POST `api/v1/spaces` endpoint with validation


lib/spaces/validateDomainName.ts
```diff
@@ -0,0 +1,24 @@
+import { string } from 'yup';
+
+import { DOMAIN_BLACKLIST } from 'lib/spaces/utils';
+
+export const domainSchema = string().ensure().trim().lowercase()
```
motechFR, line 5: Using Yup this was was unexpected, but I like how it makes for a very clear string validation

lib/middleware/requireSuperApiKey.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+  else {
+    throw new ApiError({
+      message: 'Invalid API key',
+      errorType: 'Access denied'
+    });
+  }
+
+  next();
+}
```
motechFR, line 59: I think a simpler way to write this would be:

Const superapiKey = apiKey ? Null : await..

if (superapiKey) {
  Req.superApiKey = superApiKey
  next ()
 
}

Else { throw err }

lib/utilities/errors/errors.ts
```diff
@@ -13,7 +13,8 @@ const ErrorCodes = {
   'Access denied': 401,
   'Insecure operation': 401,
   'External service': 500,
-  'Unexpected result': 500
+  'Unexpected result': 500,
+  Conflict: 409
```
motechFR, line 17: Cleanup idea: We could order this objects keys by http response code?

lib/utilities/errors/errors.ts
```diff
@@ -149,3 +150,13 @@ export class MissingWeb3SignatureError extends SystemError {
     });
   }
 }
+
+export class DataConflictError extends SystemError {
```
motechFR, line 154: Nice one extending our errors system :)

pages/api/sudo/super-api-tokens/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import { prisma } from 'db';
+import { onError, onNoMatch, provisionSuperApiKey, requireKeys } from 'lib/middleware';
+import { requireSudoApiKey } from 'lib/middleware/requireSudoApiKey';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireSudoApiKey)
+  .use(requireKeys(['name'], 'body'))
+  .post(provisionSuperToken)
```
motechFR, line 14: We now have 3 levels of API

API key for monospace
Super api key for multi space
Sudo for provisioning keys

I think it's worth a chat on naming conventions for these 3 levels

prisma/schema.prisma
```diff
@@ -247,6 +249,17 @@ model SpaceApiToken {
   @@index([spaceId])
 }
 
+model SuperApiToken {
+  id        String   @id @default(uuid()) @db.Uuid
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+  token     String   @unique
+  name      String?
+  spaces    Space[]
```
motechFR, line 258: Pretty interesting way of modeling the relationship, makes sense 👍

lib/spaces/validateDomainName.ts
msadura: actually I  just moved this logic from web form component. After initial "what the hell is this" I realized it is really convenient way of data validation

pages/api/sudo/super-api-tokens/index.ts
msadura: I agree we can discuss that. I also think we could have single db table for api keys, and they would have "type" column

lib/middleware/requireSuperApiKey.ts
msadura: yeah much simpler :D thx

pages/api/v1/spaces/index.ts
```diff
@@ -0,0 +1,101 @@
...
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireSuperApiKey)
+  .post(createSpace);
+
+async function createSpace (req: NextApiRequest, res: NextApiResponse<Space>) {
+  const { name, discordServerId, avatar, adminDiscordUserId } = req.body as CreateSpaceInputData;
+
+  if (!name) {
```
motechFR, line 33: FYI, we have a requireKeys middleware that can check all the keys are present in the body or query

pages/api/v1/spaces/index.ts
```diff
@@ -0,0 +1,101 @@
...
+  avatar?: string;
+}
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireSuperApiKey)
+  .post(createSpace);
+
+async function createSpace (req: NextApiRequest, res: NextApiResponse<Space>) {
```
motechFR, line 30: It would be good to move the controller logic inside a lib function for easier testing and resilience

pages/api/v1/spaces/index.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  if (!adminDiscordUserId) {
+    throw new InvalidInputError('Missing discord admin id');
+  }
+
+  // generate a domain name based on space
+  const spaceDomain = await getAvailableDomainName(name);
+
+  // create new bot user as space creator
+  const botUser = await prisma.user.create({
```
motechFR, line 53: Question: Have separate bot user for each space? Or one per key?

pages/api/v1/spaces/index.ts
```diff
@@ -0,0 +1,101 @@
...
+      connect: {
+        id: req.superApiToken?.id
+      }
+    }
+  };
+
+  const space = await createWorkspace({ spaceData, userId: botUser.id });
+
+  // Add bot user to space
+  await addUserToSpace({
```
motechFR, line 83: I think it would make more sense to have this inside the createWorkspace method

motechFR:
Security: If Discord account is switched to a new user, the space roles should be moved over

pages/api/v1/spaces/index.ts
0xepicode: In my opinion it is more flexible from space perspective to have bot user per space, but we can discuss that. @mattcasey @motechFR

pages/api/v1/spaces/index.ts
msadura: I tried it and actually it is kinda weird solution. `requireKeys` checks if key is in payload, but it does not check if it has any value. We can refactor and fix it but it is widely used so I'd not do it in this pr.

testing/utils/members.ts
```diff
@@ -49,3 +49,17 @@ export function generateMemberPropertyValue ({
     }
   });
 }
+
+export function addDiscordUser ({ userId, discordId }: { userId: string, discordId: string }) {
```
mattcasey, line 53: this doesn't look like it's being used

pages/api/v1/spaces/index.ts
mattcasey: I like that `requireKeys()` justs 'requires the keys'... we could add a new middleware `requireTruthyValues()`?

Actually, in the future I think we should try a different approach the Express middleware pattern, which has its limits. For example, [next-yup](https://www.npmjs.com/package/next-yup) lets you have all the flexibility of yup and automatically types the query.

lib/spaces/createWorkspaceApi.ts
```diff
@@ -0,0 +1,75 @@
...
+  // Create workspace
+  const spaceData = {
+    name,
+    updatedBy: botUser.id,
+    domain: spaceDomain,
+    spaceImage: avatar && isValidUrl(avatar) ? avatar : undefined,
+    discordServerId,
+    author: {
+      connect: {
+        id: botUser.id
```
mattcasey, line 42: i think this should actually be the user id, now that we have one

mattcasey (approved):

PR #1267, opened 2022-11-24
Forum main page

valentinludu:


components/forum/components/FilterList/FilterList.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Card variant='outlined'>
+      <CardContent>
+        <ButtonGroup orientation='vertical' sx={{ alignItems: 'flex-start' }} color='inherit' disabled={disabled}>
+          {sortList.map(btn => (
+            <Button key={btn} variant='text' onClick={() => handleClick(btn)}>{btn}</Button>
+          ))}
+        </ButtonGroup>
+        <Divider sx={{ pt: '10px', mb: '10px' }} />
+        <ButtonGroup orientation='vertical' sx={{ alignItems: 'flex-start' }} color='inherit' disabled={disabled}>
+          {categories.map(category => (
```
motechFR, line 28: Could we highlight the currently selected category (s) ? It would be better visually

Screenshot 2022-11-28 at 18 57 15

components/forum/components/ForumFilters/ForumFilters.tsx
```diff
@@ -0,0 +1,55 @@
...
+import FilterList from '../FilterList';
+import FilterSelect from '../FilterSelect';
+
+const sortList = ['Most Popular', 'Newest post', 'Latest Activity'];
+
+interface ForumFiltersProps {
+  type: 'list' | 'select';
+}
+
+export default function ForumFilters ({ type }: ForumFiltersProps) {
```
motechFR, line 17: Can you please remove the export default notation?

We only use default exports for page level components that Next.Js needs for rendering the page.

Additional benefit: Guarantees we always import components with the correct name

hooks/useOnScreen.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    observer.observe(ref.current);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, []);
+
+  return isIntersecting;
+}
```
motechFR, line 20: Question: What is this used for?

pages/api/forum/posts.ts
```diff
@@ -0,0 +1,34 @@
...
+  .use(requireUser);
+
+async function getPosts (req: NextApiRequest, res: NextApiResponse<ForumPost[]>) {
+  const { spaceId, publicOnly, count, page, sort } = req.query as any as AvailableResourcesWithPaginationRequest;
+
+  const publicResourcesOnly = ((publicOnly as any) === 'true' || publicOnly === true);
+
+  const userId = req.session?.user?.id;
+
+  const posts = await getForumPosts({
```
motechFR, line 23: Great work stubbing the responses!

prisma/schema.prisma
```diff
@@ -156,6 +156,7 @@ enum PageType {
   proposal_template
   inline_board
   inline_linked_board
+  post
```
motechFR, line 159: @mattcasey This covers the post type missing in pages

motechFR (approved):
Great start, let's get the back end rolling!

hooks/useOnScreen.ts
valentinludu: To check if a div is visible by the user. If yes, load more posts.

components/forum/ForumPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import ForumFilters from './components/ForumFilters';
+import ForumPosts from './components/ForumPosts';
+
+export default function ForumPage () {
+  const [search, setSearch] = useState('');
+
+  /*
+   // track page_view for the forum main page
+  useEffect(() => {
```
mattcasey, line 19: is there a reason this is commented out?

components/forum/ForumPage.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <TextField
+        variant='outlined'
+        placeholder='Search Posts, Comments and Members'
+        onChange={(e) => setSearch(e.target.value)}
+        fullWidth
+        sx={{ padding: '20px 0' }}
+        InputProps={{ endAdornment: <InputAdornment position='end'><SearchIcon color='secondary' fontSize='small' /></InputAdornment> }}
+      />
+      <Grid container spacing={2}>
+        <Grid item xs={12} md={9}>
```
mattcasey, line 36: I think the minimum breakpoint to show a sidebar should be "lg", "md" seems too narrow":
image

components/forum/components/ForumFilters/ForumFilters.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { push, query } = useRouter();
+  const currentSpace = useCurrentSpace();
+
+  const { data: categories, error, isValidating } = useSWR(currentSpace ? '/forum/categories' : null, () => charmClient.forum.listPostCategories(currentSpace!.id));
+
+  const handleClick = (value: string) => {
+    const isValidSort = value && sortList.some(btn => btn === value);
+    const isValidCategory = value && categories?.some(btn => btn === value);
+
+    push({
```
mattcasey, line 27: links should always be defined by using attributes on the links instead of click handlers.. right now you can't right-click to open a link in a new tab.
I'm not sure the custom Types on our Link component will allow it, but it uses next/link and you can define the href using an object to make  life easier:
https://nextjs.org/docs/api-reference/next/link
image

components/forum/components/ForumFilters/ForumFilters.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (error) {
+    return <Alert severity='error'>An error occured while loading the categories</Alert>;
+  }
+
+  if (!categories) {
+    return null;
+  }
+
+  if (type === 'list') {
```
mattcasey, line 46: This feels like it should really be two components. 'type' is never being passed in dynamically, the only other thing that's being resued is logic to generate links, which could be a shared functino

components/forum/components/ForumFilters/ForumFilters.tsx
```diff
@@ -0,0 +1,55 @@
...
+        query: { category: value }
+      }))
+    }, undefined, { shallow: true });
+  };
+
+  if (error) {
+    return <Alert severity='error'>An error occured while loading the categories</Alert>;
+  }
+
+  if (!categories) {
```
mattcasey, line 42: there should always be at least one category (General) but also, same request as before: always return a component so that layout doesn't jump

components/forum/components/ForumPosts/ForumPosts.tsx
```diff
@@ -0,0 +1,95 @@
...
+export default function ForumPosts ({ search }: ForumPostsProps) {
+  const ref = useRef();
+  const currentSpace = useCurrentSpace();
+  const { query } = useRouter();
+  const querySort = query.sort;
+  const queryCategory = query.category;
+
+  const isVisible = useOnScreen(ref);
+
+  const sortValue = useMemo(() => {
```
mattcasey, line 30: why is this a useMemo?

components/forum/components/ForumPosts/ForumPosts.tsx
```diff
@@ -0,0 +1,95 @@
...
+        return queryCategory[0];
+      }
+      else {
+        return queryCategory;
+      }
+    }
+    return undefined;
+  }, [queryCategory]);
+
+  const { data, error, size, setSize, isValidating } = useSWRInfinite(
```
mattcasey, line 55: never heard of useSWRInfinite, nice find :)

components/forum/components/ForumPosts/ForumPosts.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const isReachingEnd = isEmpty || (data && data[data.length - 1]?.length < count);
+  const isRefreshing = isValidating && data && data.length === size;
+
+  useEffect(() => {
+    if (isVisible && !isReachingEnd && data && !isRefreshing && !isLoadingMore) {
+      setSize(size + 1);
+    }
+  }, [isVisible, isRefreshing, isReachingEnd, data, isLoadingMore]);
+
+  if (error) {
```
mattcasey, line 73: i don't think that we should clear out the entire UI if there's an error loading data.. you might still be able to create a new post for example. i think if anything we could show an ephemeral message via useSnackBar(), or display the error in alert but jsut at the top of the posts area

lib/forum/getForumCategories.ts
```diff
@@ -0,0 +1,8 @@
+import { prisma } from 'db';
+import type { AvailableResourcesRequest } from 'lib/permissions/interfaces';
+
+export async function getForumCategories ({ spaceId, userId }: AvailableResourcesRequest): Promise<string[]> {
+  // const categories = await prisma.forumCategories.findUnique({});
+  const categories = ['Announcements', 'Governance', 'Randoms', 'NFTS', 'Introductions', 'Question & Support', 'New York', 'Devcon VI'];
```
mattcasey, line 6: this is a constant and should be defined outside the function

lib/forum/interfaces.ts
```diff
@@ -0,0 +1,18 @@
+import type { User } from '@prisma/client';
+
+export interface ForumPostContent {
```
mattcasey, line 3: use `type` instead of `interface`

lib/forum/interfaces.ts
```diff
@@ -0,0 +1,18 @@
...
+  type: 'text' | 'image';
+  content: string;
+}
+
+export interface ForumPost {
+  id: string;
+  title: string;
+  content: ForumPostContent;
+  user: Omit<User, 'addresses'>;
```
mattcasey, line 12: We can get user info from useMembers() in the frontend, so this should just be a userId.

Side note: it's much better to use Pick and be explicit about the fields needed. also 'addresses' has been removed from the User model.

pages/[domain]/forum/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { setTitle } from 'hooks/usePageTitle';
+
+export default function ForumPage () {
+
+  setTitle('Forum');
+
+  const space = useCurrentSpace();
+
+  // Show this page only to charmverse users
+  if (!(space?.name.includes('charmverse')) && process.env.NODE_ENV === 'production') {
```
Comment (mattcasey) [inline pages/[domain]/forum/index.tsx:13]: please use `isProdEnv` instead of checking the value, that way this is still accessible by all envs in testing and staging

pages/api/forum/categories.ts
```diff
@@ -0,0 +1,31 @@
...
+  const { spaceId, publicOnly } = req.query as any as AvailableResourcesRequest;
+
+  const publicResourcesOnly = ((publicOnly as any) === 'true' || publicOnly === true);
+
+  // Session may be undefined as non-logged in users can access this endpoint
+  const userId = req.session?.user?.id;
+
+  const categories = await getForumCategories({
+    spaceId: spaceId as string,
+    userId: publicResourcesOnly ? undefined : userId
```
mattcasey, line 25: there isn't a reason to pass a userId for categories that i'm aware of

lib/utilities/dates.ts
```diff
@@ -104,20 +104,22 @@ export function relativeTime (dateInput: DateInput) {
  * @param date Date
  * @returns string
  */
-export const getRelativeTimeInThePast = (date: Date): string => {
+export const getRelativeTimeInThePast = (date: Date, options: ({ ago: boolean } | undefined) = { ago: true }): string => {
```
mattcasey, line 107: 'ago' is a weird name for a prop.. it is probably better to be consistent and have it all the time or not at all. Please pick one

pages/api/forum/posts.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.get(getPosts)
+  .use(requireUser);
+
+async function getPosts (req: NextApiRequest, res: NextApiResponse<ForumPost[]>) {
+  const { spaceId, publicOnly, count, page, sort } = req.query as any as AvailableResourcesWithPaginationRequest;
+
+  const publicResourcesOnly = ((publicOnly as any) === 'true' || publicOnly === true);
```
mattcasey, line 19: again, no need for 'public' logic

pages/api/forum/interfaces.ts
```diff
@@ -0,0 +1,7 @@
+import type { AvailableResourcesRequest } from 'lib/permissions/interfaces';
+
+export interface AvailableResourcesWithPaginationRequest extends AvailableResourcesRequest {
```
mattcasey, line 3: I don't see a lot of benefit to  coupling the types of all our endpoints and we should avoid coupling them.. in this case, userId and publicOnly from AvailableResourcesRequest does not make sense. just call this one "GetForumPostsRequest".  and put it in getForumPosts.ts file next to the function that actually takes the input.

mattcasey: Other note: we never use this pattern of 'interfaces.ts' file inside the api folder. This either belongs in a single 'forum-related' interfaces.ts file inside of lib/forum, but close to the code is the BEST place. interfaces.ts is a hack

components/forum/components/FilterList/FilterList.tsx
valentinludu: Done

components/forum/components/CreateForumPost/CreateForumPost.tsx
```diff
@@ -0,0 +1,37 @@
...
+import TextField from '@mui/material/TextField';
+
+import Button from 'components/common/Button';
+import UserDisplay from 'components/common/UserDisplay';
+import { useUser } from 'hooks/useUser';
+
+export default function CreateForumPost () {
+  const { user } = useUser();
+
+  if (!user) {
```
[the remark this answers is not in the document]
valentinludu: Done

components/forum/ForumPage.tsx
valentinludu: Actually I saw this inside bounty and proposal pages so I figured we will need this at some point.
I will delete it for now and make a note to discuss about page views at the end of all stories.

components/forum/ForumPage.tsx
valentinludu: Agree. Just changed it!

pages/api/forum/posts.ts
valentinludu: Thanks Mo.

components/forum/components/ForumFilters/ForumFilters.tsx
valentinludu: Did a React hook for now.

components/forum/components/ForumFilters/ForumFilters.tsx
valentinludu: Done

components/forum/components/ForumPosts/ForumPosts.tsx
valentinludu: Since they are a dependency of useSWRInfinite, I added useMemo to be sure on component update, the SWR hook does not revalidate if nothing changes in the values.

lib/forum/interfaces.ts
valentinludu: I did this interface only to quickly mock a response. For sure will be improved in the next few days by Mo.
For now I did the change you suggested.

components/forum/components/ForumFilters/ForumFilters.tsx
valentinludu: I deleted entirely this component

components/forum/components/ForumFilters/ForumFilters.tsx
valentinludu: I was thinking that a link is more for redirections to other pages and here I was just updating the query.

valentinludu: I updated the Link here to be aligned for UX

PR #1268, opened 2022-11-25
add Post data models

mattcasey:
I have a modelling question, looking for opinions

1. First question: should Votes and Comments be linked to the Poll database or the Page database?
In our other models, it's a mix. Bounty applications point to the bounty, but proposal votes point to the page.  One reason for this is that applications are very specific to bounties, whereas votes can be applied to non-proposal pages as well. So I'm thinking we should link the Votes and Comments to the pages to be future-proof.

2. Do we call them PageVote or PostVotes? PageComment or PostComment?
If we agree to link the votes/comments to the pages, not the posts table, then it seems to make sense to name the tables accordingly. This might be confusing because we also have 'inline comments' and database cards have comments inside blocks... but it's conceivable we might replace database cards with this model eventually, not the other way around.


lib/pages/server/deleteArchivedPages.ts
```diff
@@ -17,6 +17,12 @@ export async function deleteArchivedPages (maxDay: number) {
     }
   });
 
+  const { count: deletedPostsCount } = await prisma.post.deleteMany({
```
mattcasey, line 20: this PR just also updates something we forget: delete the relational entities when we delete archived pages

prisma/schema.prisma
```diff
@@ -978,3 +984,65 @@ model MemberPropertyPermission {
...
+  upvoted   Boolean
+
+  pageId String @db.Uuid
+  page   Page   @relation(fields: [pageId], references: [id], onDelete: Cascade)
+
+  @@unique([createdBy, pageId])
+  @@index([pageId])
+}
+
+model PostComment {
```
mattcasey, line 1022: I left it "PostComment" for now, but use pageId to link to page instead of the post. I think we should go one way or the other and change one of these...

valentinludu:
> I have a modelling question

Just an opinion:

If PageVote and PageComment will pe in the Page table it will be very confusing because of other votes and comments that are also linked to the page. For this reason I like better PostComment and PostVote.

prisma/schema.prisma
```diff
@@ -978,3 +984,65 @@ model MemberPropertyPermission {
...
+  createdAt DateTime @default(now())
+  createdBy String   @db.Uuid
+  upvoted   Boolean
+
+  commentId String      @db.Uuid
+  comment   PostComment @relation(fields: [commentId], references: [id], onDelete: Cascade)
+
+  @@unique([createdBy, commentId])
+  @@index([commentId])
+}
```
motechFR, line 1048: For simplicity, I think this could be merged with PostVote, and have optional relation to Comment or Page.

We can then have a unit tested lib method (like permissions) that ensures votes are assigned to one or the other.

prisma/schema.prisma
```diff
@@ -978,3 +984,65 @@ model MemberPropertyPermission {
...
+  parentId    String
+
+  pageId          String            @db.Uuid
+  page            Page              @relation(fields: [pageId], references: [id], onDelete: Cascade)
+  PostCommentVote PostCommentVote[]
+
+  @@index([pageId])
+}
+
+model PostCommentVote {
```
motechFR, line 1038: Some alternative ideas (if we're going to merge these 2 tables)

PostRating
PostScore
PostUpDownVote

prisma/schema.prisma
```diff
@@ -978,3 +984,65 @@ model MemberPropertyPermission {
   @@index([memberPropertyId])
   @@index([roleId])
 }
+
+// Begin Forum models
+model PostCategory {
+  id    String @id @default(uuid()) @db.Uuid
+  color String
+  name  String
+  Post  Post[]
```
motechFR, line 993: Should we add a Space relation here? I can see this working same as ProposalCategory with some initial smart defaults

motechFR:
@mattcasey Re: your top level questions

First question: should Votes and Comments be linked to the Poll database or the Page database?
In our other models, it's a mix. Bounty applications point to the bounty, but proposal votes point to the page. One reason for this is that applications are very specific to bounties, whereas votes can be applied to non-proposal pages as well. So I'm thinking we should link the Votes and Comments to the pages to be future-proof.

> Agreed, let's link the up/down/votes and comments to the page. This would also make it easy to enable this functionality in other page types in future (if needed)


Do we call them PageVote or PostVotes? PageComment or PostComment?
If we agree to link the votes/comments to the pages, not the posts table, then it seems to make sense to name the tables accordingly. This might be confusing because we also have 'inline comments' and database cards have comments inside blocks... but it's conceivable we might replace database cards with this model eventually, not the other way around.

>  If we go with having a single up/down vote entity, I'd have the relations on there. Since we are voting on content, I suggest ContentRating, ContentScore or ContentUpDownVote

valentinludu:
> If we go with having a single up/down vote entity, I'd have the relations on there. Since we are voting on content, I suggest ContentRating, ContentScore or ContentUpDownVote

+1 for ContentVote

motechFR:
For comments, +1 re: PageComment :)

prisma/schema.prisma
mattcasey: Woops! I think we absolutely need one

prisma/schema.prisma
mattcasey: If we're going for simplicity, I think having the db ensure assignment is much simpler. It's also more robust

valentinludu (approved):

PR #1269, opened 2022-11-25
Sort members

motechFR:
Alphabetic sorting of members in a workspace.

Native JS allows us to do string comparison using < or > without worrying about casing.

Also added memoization for the search, as well as partial matching for member names (so we don't only look at names starting with a specific string)

### DIFF
--- a/components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
+++ b/components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
@@ -1,8 +1,8 @@
-import { usePluginState, useEditorViewContext } from '@bangle.dev/react';
-import { Typography, Divider, MenuItem } from '@mui/material';
+import { useEditorViewContext, usePluginState } from '@bangle.dev/react';
+import { Divider, MenuItem, Typography } from '@mui/material';
 import { Box } from '@mui/system';
 import type { PluginKey } from 'prosemirror-state';
-import { useCallback, useEffect, memo, useMemo } from 'react';
+import { memo, useCallback, useEffect } from 'react';
 
 import UserDisplay from 'components/common/UserDisplay';
 import { useMembers } from 'hooks/useMembers';
--- a/components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
+++ b/components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
@@ -2,7 +2,7 @@ import { usePluginState, useEditorViewContext } from '@bangle.dev/react';
 import { Typography, Divider, MenuItem } from '@mui/material';
 import { Box } from '@mui/system';
 import type { PluginKey } from 'prosemirror-state';
-import { useCallback, useEffect, memo } from 'react';
+import { useCallback, useEffect, memo, useMemo } from 'react';
 
 import UserDisplay from 'components/common/UserDisplay';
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
```diff
@@ -43,10 +43,12 @@ function MentionSuggestMenu ({ pluginKey }: { pluginKey: PluginKey }) {
     [view, pluginKey]
   );
 
-  const filteredMembers = triggerText.length !== 0 ? members.filter(
-    member => (
-      member.username?.toLowerCase()?.startsWith(triggerText.toLowerCase()))
-  ) : members;
+  const filteredMembers = useMemo(() => {
```
mattcasey, line 46: Please avoid optimization before we know there's an issue.. useMemo is really overkill here, even if you had 1000 members.. it could introduce bugs if someone updates this but forgets to change dependencies

components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
```diff
@@ -43,10 +43,12 @@ function MentionSuggestMenu ({ pluginKey }: { pluginKey: PluginKey }) {
...
+    return triggerText.length !== 0 ? members.filter(
+      member => (
+        member.username?.toLowerCase()?.match(triggerText.toLowerCase()))
```
mattcasey, line 49: username is not optional

hooks/useMembers.tsx
```diff
@@ -22,7 +22,17 @@ export function MembersProvider ({ children }: { children: ReactNode }) {
   const space = useCurrentSpace();
 
   const { data: members, mutate: mutateMembers } = useSWR(() => space ? `members/${space?.id}` : null, () => {
-    return charmClient.members.getMembers(space!.id);
+    return charmClient.members.getMembers(space!.id).then(_members => _members.sort((a, b) => {
```
mattcasey, line 25: I think it would be better to put this logic on the server-side

mattcasey, line 25: i suggest putting this logic on server-side... 1) it's less code to ship to browsers 2) it's closest to the source of truth, so if we call getMembers() anywhere else, it will still be sorted

mattcasey (approved):

mattcasey (changes requested):
Actually, I don't think native sort works:

```

![image](https://user-images.githubusercontent.com/305398/204044439-c00db3a6-dfc5-4643-a320-ba51bf7dfeea.png)

```

components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
mattcasey: Also, this component is probably only re-rendering when triggerText changes anyway?

hooks/useMembers.tsx
motechFR: Agreed and done

components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
motechFR: Fair enough, removed unneeded memoisation and question marks. They were there before this PR, but good cleanup opportunity

mattcasey (approved):

PR #1271, opened 2022-11-25
Real time vote updates

motechFR:
When a vote is created, deleted or cancelled, this is broadcast to all other connected space members


mattcasey (changes requested):
This works, but I don't think it's a good general pattern to couple the logic and extra payload required to update tasks when updating votes, pages, or anything else. It will get really messy if we had to keep track of updating tasks everywhere a task might change (also for proposals, etc.) 

A better pattern could be to have the central websocket listening for certain events and then it could submit a 'new_task' update, that way the task logic stays in a single place.. this'll get trickier because not every event is a task for every person in the workspace

mattcasey (approved):

PR #1272, opened 2022-11-25
fix sorting table rows by title

mattcasey:
We were relying on 'card.title' which is no longer ever set.


motechFR:
I just tested this, it works but there is a quirk.

If you add an item in one view, it will appear sorted in the other and vice verse

motechFR (approved):
tested & approved

PR #1273, opened 2022-11-25
remove default Member Profile database in new workspaces

mattcasey:
Now that we have Member Directory feature, there's no need to show users how to make a member gallery database


motechFR (approved):

motechFR:
Makes sense, tested and works just fine. New space creates pages without that extra board

PR #1275, opened 2022-11-25
rename BoardPage to DatabasePage

mattcasey:
just keeping up with terminology


components/[pageId]/DatabasePage/index.ts
```diff
@@ -0,0 +1,3 @@
+import BoardPage from './DatabasePage';
```
Comment (motechFR) [inline components/[pageId]/DatabasePage/index.ts:1]: Shall we drop the default export and go with a named {export} to eliminate this Typo across the repo ? :)

motechFR (approved):
All good, 1 comment / cleanup opportunity to address

components/[pageId]/DatabasePage/index.ts
Comment (mattcasey) [reply components/[pageId]/DatabasePage/index.ts:1]: yes, thanks

PR #1284, opened 2022-11-28
Add prettier and fix all the errors

msadura:
Quite a lot of changes for code standardisation with prettier :) Not much of a prettier config other than defaults to keep our code consistent with industry standards.
I think it will be impossible to review everything, so I will point main differences.
1) Max line length changed from 150 to 120 chars
2) no spaces in function declaration names `function test()` vs `function test ()`
3) else, else if statements in single line:
```
  } else if (propertyTemplate.type === 'createdBy') {
    return <CreatedBy userID={card.createdBy} />;
  } else if (propertyTemplate.type === 'updatedBy') {
  ```
  
  4) inline type properties needs to be separated with `;` : `type Test = { prop1: string; prop2: string; }`
  
  I made eslint checker compatible with prettier and they will work together. All the errors (32k) has been resolved.

If you are using vscode eslint extention it should format code for you on save so you can type it as you like :)

### CODE REVIEW

**Review by mattcasey — APPROVED** [review]

File: components/proposals/components/NoProposalsMessage.tsx [hunk 1]
```diff
@@ -21,19 +20,21 @@ const Center = styled.div`
...
 
-export default function NoProposalsMessage ({ message }: { message: string }) {
+export default function NoProposalsMessage({ message }: { message: string }) {
   return (
     <EmptyProposalContainerBox>
       <Center>
         <TaskOutlinedIcon
-          fontSize='large'
-          color='secondary'
+          fontSize="large"
```
mattcasey, line 28: this drives me nuts, can we have single quotes in React?

mattcasey (changes requested):
set jsxSingleQuote to true?

msadura:
> set jsxSingleQuote to true?

sure thing. Single quotes for jsx string props are back :)

mattcasey (approved):
thx

PR #1292, opened 2022-11-29
WIP Connect Unstoppable domains

motechFR:
Allows login and registration via UD.

Since we are adding an Identity Type, I took the opportunity to eliminate IdentityTypes[number] from models and use Prisma enum as the source of truth (as well as replacing array refs by actual value)


components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
```diff
@@ -66,6 +71,26 @@ function WalletSelectorModal({
     }
   }, [error, openNetworkModal, closeModal]);
 
+  const uauth = new UAuth({
+    clientID: '0e8c724d-bbb8-4876-9dc7-ddfb466a3a0f',
```
mattcasey, line 75: this should be moved to Secrets, even if it is in the frontend

testing/setupDatabase.ts
```diff
@@ -44,7 +44,7 @@ export async function generateSpaceUser({
 }): Promise<LoggedInUser> {
   return prisma.user.create({
     data: {
-      identityType: IDENTITY_TYPES[1],
+      identityType: 'Discord',
```
mattcasey, line 47: just curious, is there a reason test users need discord? i think this might not work 100% if they don't have a discordUser record as well?

pages/api/profile/with-unstoppable-domains.ts
```diff
@@ -0,0 +1,100 @@
...
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.post(loginViaUnstoppableDomains);
+
+export type UnstoppableDomainsLoginRequest = {
+  authSig: UnstoppableDomainsAuthSig;
+};
+
+async function loginViaUnstoppableDomains(req: NextApiRequest, res: NextApiResponse<LoggedInUser>) {
+  const { authSig } = req.body as UnstoppableDomainsLoginRequest;
```
mattcasey, line 29: this should be put under unstoppable-domains, similar to how we bucket endpoints for discord, telegram, etc.

__e2e__/utils/mocks.ts
```diff
@@ -238,7 +237,7 @@ export async function generateUser({ walletAddress = Wallet.createRandom().addre
 export async function generateDiscordUser() {
   const user = await prisma.user.create({
     data: {
-      identityType: IDENTITY_TYPES[0],
+      identityType: 'Wallet',
```
mattcasey, line 240: +1 removing the redundant wallet type since we have it in prisma.. but i think u could have still done `IdentityType.WalletConnect`, IdentityType is a map when imported as a regular variable.

components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
mattcasey: I added NEXT_PUBLIC_UNSTOPPABLE_DOMAINS_REDIRECT_URI as well

components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
```diff
@@ -66,6 +71,26 @@ function WalletSelectorModal({
...
+  });
+
+  async function handleAuth() {
+    setIsConnectingIdentity(true);
+    try {
+      const authSig = (await uauth.loginWithPopup()) as any as UnstoppableDomainsAuthSig;
+      await charmClient.profile.loginWithUnstoppableDomains({ authSig });
+      window.location.reload();
+    } catch (err) {
+      // console.log('UD Error', err);
```
mattcasey, line 87: log.error?

pages/api/profile/with-unstoppable-domains.ts
```diff
@@ -0,0 +1,100 @@
...
+            domain,
+            user: {
+              connect: {
+                id: userId
+              }
+            }
+          }
+        });
+
+        return getUserProfile('id', userId, tx);
```
mattcasey, line 90: theres really no reason to make this part of the transaction. we should just use them for writing multiple times, or when the query is required before a write

lib/blockchain/verifyUnstoppableDomainsSignature.ts
```diff
@@ -0,0 +1,141 @@
...
+export async function assignUnstoppableDomainAsUserIdentity({
+  userId,
+  tx = prisma,
+  domain
+}: {
+  userId: string;
+  tx?: Prisma.TransactionClient;
+  domain: string;
+}): Promise<LoggedInUser> {
+  const foundDomain = await tx.unstoppableDomain.findUnique({
```
mattcasey, line 121: This is aggressive use of transactions

mattcasey (changes requested):
- fix the secret so it's not hard-coded
- general comment: i might be jumping to conclusions, but I don't see a strong need for prisma transactions here, or to add support to our lib methods. I am having trouble coming up with a scenario where two users are trying to connect the same unstoppable domain at the same time. And it would be much simpler if you just used prisma to write a single query that created the records that are necessary for login. It's not causing too much harm here but i am  worried this pattern gets reused in other places unnecessarily (copy/paste, etc) and it will be very hard to remove later since we won't know what were the original valid use cases

components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
motechFR: Great, done

__e2e__/utils/mocks.ts
motechFR: Our general pattern has been to use the Typescript string types generated by Prisma, but happy to move over to enum.key notation if it makes more sense

pages/api/profile/with-unstoppable-domains.ts
motechFR: Noted, adding to cleanups post this release

motechFR: On second thought removing on this release, if we get a surge of signups, don't want to kill our DB with bunch of long lived transactions.

testing/setupDatabase.ts
motechFR: Sounds like something to cleanup indeed

PR #1293, opened 2022-11-30
Extract video plugin from iframe plugin

mattcasey:
While prepping to add support for Mux, I want to add an "Upload" tab for video, similar to images, and didn't want to further complicate the iframe component. So I broke out the logic for videos to a new Prosemirror plugin and created some standard components for embeds to share. There should be no visible difference for users, but the /video command will now create a 'video' node type, and this remains backwards compatible with 'iframe' nodes that have 'video' type.

My plan is to merge this PR separately from Mux integration.

Tested:
-  existing 'iframe' nodes still display correctly (both 'embed' and 'video' types)
- adding new figma and youtube links from /embed 
- adding a youtube link from /video


components/common/CharmEditor/components/common/EmbeddedUrl.tsx
```diff
@@ -0,0 +1,43 @@
+import { Box, Button, TextField, Typography } from '@mui/material';
```
mattcasey, line 1: my plan is to re-use this component for 'iframe' and 'tweet' components

PR #1296, opened 2022-12-01
Update guild.xyz SDK and address 'roleIds' vs 'roleids'

Devorein:


lib/guild-xyz/getGuildRoleIds.ts
```diff
@@ -8,10 +8,11 @@ export async function getGuildRoleIds(addresses: string[]) {
   const guildMembershipsResponses = await Promise.all(addresses.map((address) => user.getMemberships(address)));
   guildMembershipsResponses.forEach((guildMembershipsResponse) => {
     guildMembershipsResponse?.forEach((guildMemberships) => {
-      if (!guildMemberships.roleids) {
-        log.warn('Guild.xyz response is mossing roleids', { addresses, guildMemberships });
+      const { roleIds } = guildMemberships as unknown as { roleIds: string[] };
```
mattcasey, line 11: @Devorein can we continue support `roleids` as well just in case? but still log a warning if they're using 'roleids'

mattcasey (changes requested):
support both formats for now, since we don't know what's going on really and their demo and TS definitions use the lowercase format. They might realize the mistake and revert it

lib/guild-xyz/getGuildRoleIds.ts
```diff
@@ -8,10 +8,13 @@ export async function getGuildRoleIds(addresses: string[]) {
   const guildMembershipsResponses = await Promise.all(addresses.map((address) => user.getMemberships(address)));
   guildMembershipsResponses.forEach((guildMembershipsResponse) => {
     guildMembershipsResponse?.forEach((guildMemberships) => {
-      if (!guildMemberships.roleids) {
-        log.warn('Guild.xyz response is mossing roleids', { addresses, guildMemberships });
+      const { roleIds, roleids } = guildMemberships as unknown as { roleids: string[]; roleIds: string[] };
```
msadura, line 11:
```
const roleIds = guildMemberships.roleIds || guildMemberships.roleids;
if (roleIds) {
        guildRoleIds.push(...roleIds.map(String));
      }  else {
        log.warn('Guild.xyz response is missing roleIds', { addresses, guildMemberships });
      }
```

PR #1297, opened 2022-12-01
Error updating profile item

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=71a30e2a-fd70-4857-ba3c-0466149467e0)
The error was caused due to not having the `profileItem` id. Unfortunately I couldn't reproduce the error, so resorted to adding guards in fe and be

### DIFF
--- a/components/profile/PublicProfile.tsx
+++ b/components/profile/PublicProfile.tsx
@@ -231,7 +231,9 @@ export default function PublicProfile(props: UserDetailsProps) {
                 <CommunityRow
                   key={community.id}
                   onClick={() => {
-                    toggleCommunityVisibility(community);
+                    if (community.id) {
+                      toggleCommunityVisibility(community);
+                    }
                   }}
                   visible={!community.isHidden}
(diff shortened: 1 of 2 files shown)


mattcasey (changes requested):
1. Looking at the source of communities, it can come from either Collab.land or from DeepDAO. Collab.land auth is not currently enabled, so I would assume that DeepDAO must be returning communities sometimes without an organizationId. Can we add a filter there (inside getAggregatedData) and just exclude these communities? Add a log message that include the whole profile object so we can do further investigation... that should be enough for this ticket, IMO.

2. I don't think it's a good idea to conditionally call toggle, because it is not really addressing the error. It prevents us from knowing if the issue is continuing, and also would be confusing to the user when the button does nothing - it's always better to show them an error if an action fails. But if we do 1, and fix it at the source, I think we can wait to add error handling on the frontend.

PR #1298, opened 2022-12-01
Adding/editing categories for forum

Devorein:


msadura:
I tried to test this, but when I visit forum as an admin I get:
Screenshot 2022-12-02 at 13 15 56

msadura:
Screenshot 2022-12-02 at 13 18 05

Devorein:
Both the above occurred due to color type mismatch. I've disabled colors for categories for now so it should be working as expected.

msadura:
> Both the above occurred due to color type mismatch. I've disabled colors for categories for now so it should be working as expected.

does it mean colors are not in a scope of post category? Or it was not discussed

hooks/useForumFilters.ts
```diff
@@ -45,5 +46,32 @@ export function useForumFilters() {
...
+    const isValidSort = value && sortList.some((btn) => btn === value);
+    const isValidCategory = value && categories?.some((category) => category.name === value);
+
+    push(
+      {
+        pathname: `/${query.domain}/forum`,
+        ...(isValidSort && {
+          query: { sort: value }
+        }),
+        ...(isValidCategory && {
```
msadura, line 59: I don't quite get what this does? Won't 1st query get override by the 2nd spread?

hooks/useForumFilters.ts
```diff
@@ -45,5 +46,32 @@ export function useForumFilters() {
     return `${query.domain}/forum/`;
   };
 
-  return { categories, sortList, error, getLinkUrl, disabled: isValidating };
+  const handleClick = (value: string) => {
```
msadura, line 49: "handleClick" is quite misleading here. It should be "applyFilters" or something like that

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,20 +1,71 @@
...
+
 export default function FilterList({ categoryIdSelected, selectedCategory }: FilterProps) {
-  const { getLinkUrl, categories, sortList, error } = useForumFilters();
-  const { query } = useRouter();
+  const { getLinkUrl, categories, sortList, error, refetchForumCategories } = useForumFilters();
+  const addCategoryPopupState = usePopupState({ variant: 'popover', popupId: 'add-category' });
+  const [forumCategoryName, setForumCategoryName] = useState('');
+  const currentSpace = useCurrentSpace();
+  const admin = isAdmin();
+  async function createForumCategory() {
```
msadura, line 32: those helper methods like "createForumCategory" fits more into `useForumFilters` than component

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,20 +1,71 @@
...
+    if (currentSpace) {
+      await charmClient.forum.createPostCategory(currentSpace.id, {
+        name: forumCategoryName,
+        spaceId: currentSpace.id
+      });
+      refetchForumCategories();
```
msadura, line 38: is there a reason to blindly refetch all the categories from the server rather than mutate cache locally?

msadura (changes requested):
2 small UI improvements:
1) When you edit category it would be nice to close popover on enter
2) There should be some sort of "ellipsis" in category name:
Screenshot 2022-12-05 at 11 11 25

hooks/useForumFilters.ts
Devorein: Right. Don't remember touching that file up until now but will fix it 💯

components/forum/components/Filters/FilterList.tsx
Devorein: I think having a hook link `useForumCategories` would be better. `useForumFilters` shouldn't need to deal with categories imo🤔

components/forum/components/Filters/FilterList.tsx
Devorein: Normally for smaller data we refetch the data, but I agree we should always try to update the cache when possible. I've made the changes required 🥳

components/forum/components/Filters/FilterList.tsx
msadura: yeah that is true, additional `useForumCategories` would be the best. As far as I understand posts are "filtered" by categories and this is why they landed in this hook in the first place, but it will be getting more and more confusing

components/common/Icons/hoverIconsStyle.ts
```diff
@@ -4,7 +4,10 @@ import { css } from '@emotion/react';
 import { isTouchScreen } from 'lib/utilities/browser';
 
 export const hoverIconsStyle =
-  ({ absolutePositioning = false }: { absolutePositioning?: boolean } = {}) =>
+  ({
+    absolutePositioning = false,
+    marginForIcons = true
```
mattcasey, line 9: this should be false by default (see the other prop is false by default).. but it also seems like something that could be handled by the parent component instead. This style mixin should just handle hover effects and not layout

components/common/Link.tsx
```diff
@@ -42,7 +45,7 @@ export default function Link({ external, href, onClick, children, color = 'prima
   }
 
   return external ? (
-    <StyledMuiLink href={href} color={color} rel='noreferrer' underline='none' {...restProps}>
+    <StyledMuiLink href={href as string} color={color} rel='noreferrer' underline='none' {...restProps}>
```
mattcasey, line 48: this type assertion is not necessary. it should throw an error instead

components/forum/ForumPage.tsx
```diff
@@ -46,6 +46,10 @@ export default function ForumPage() {
     }
   }
 
+  useEffect(() => {
+    setCategoryId(categories?.find((c) => c.id === categoryId)?.id);
+  }, [router.query.categoryIds]);
```
mattcasey, line 51: This code is confusing... it's just using categoryId to set categoryId?

components/forum/components/Filters/FilterCategory.tsx
```diff
@@ -0,0 +1,72 @@
+import DeleteOutlinedIcon from '@mui/icons-material/DeleteOutlined';
```
mattcasey, line 1: there is no "Filters" component, please put these all inside components

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,61 +1,177 @@
...
+    setForumCategoryName('');
+    addCategoryPopupState.close();
+  }
+
   if (error) {
-    return <Alert severity='error'>An error occured while loading the categories</Alert>;
+    return <Alert severity='error'>An error occurred while loading the categories</Alert>;
+  }
+
+  if (!currentSpace) {
```
mattcasey, line 45: doesn't look like currentSpace is a requirement to show the UI. But also, I think you could use router.query.domain instead which is immediately available.

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,61 +1,177 @@
...
+    return null;
   }
 
   return (
     <Card variant='outlined'>
-      <CardContent>
+      <CardContent
+        sx={{
```
mattcasey, line 52: `px` is a prop of `CardContent`.. no need to use `sx` which is slightly slower to render

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,61 +1,177 @@
...
+                    />
+                  </Box>
+                )}
+              </MenuItem>
+            </StyledBox>
           ))}
-        </Box>
+        </Stack>
+        <Button
+          sx={{
```
mattcasey, line 133: add 'size=small'?

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,61 +1,177 @@
...
-              onClick={() => categoryIdSelected(category.id)}
-              sx={{ fontWeight: selectedCategory === category.id ? 'bold' : 'initial' }}
-              color='inherit'
-            >
-              {category.name}
-            </Typography>
+        <Stack mb={2}>
+          {[{ id: 'all-categories', name: 'All categories' }, ...(categories ?? [])]?.map((category) => (
+            <StyledBox key={category.id}>
+              <MenuItem
```
mattcasey, line 83: add 'dense'

components/forum/components/Filters/FilterList.tsx
```diff
@@ -1,61 +1,177 @@
...
       </CardContent>
+      <Modal
+        open={addCategoryPopupState.isOpen}
+        onClose={() => {
+          addCategoryPopupState.close();
+        }}
+        title='Add forum category'
+      >
+        <TextField
+          sx={{
```
mattcasey, line 152: add autoFocus={true}

hooks/useForumCategories.ts
```diff
@@ -0,0 +1,64 @@
+import type { PostCategory } from '@prisma/client';
+
+import charmClient from 'charmClient';
+
+import { useCurrentSpace } from './useCurrentSpace';
+import { useForumFilters } from './useForumFilters';
+
+export function useForumCategories() {
+  const currentSpace = useCurrentSpace();
+  const { mutateForumCategories } = useForumFilters();
```
mattcasey, line 10: these methods are only used in one place, and 'useForumCategories' doesnt actually supply the categories. I think we should either:
A) move the methods into useForumFilters()
B) move the methods into the hook component

hooks/useForumFilters.ts
mattcasey: instead of having 'applyFilters' check category and sort list, just make the query the input:
`applyFilters({ sort: value })`

mattcasey: Hm it doesn't look like  we are using applyFilters anywhere?

lib/forums/categories/createPostCategory.ts
```diff
@@ -16,7 +16,7 @@ export function createPostCategory({ name, spaceId, color }: CreatePostCategoryI
           id: spaceId
         }
       },
-      color: categoryColour
+      color: categoryColor
```
mattcasey, line 19: please completely remove color from categories - that's not part of the product design

mattcasey (changes requested):
see comments

components/forum/ForumPage.tsx
Devorein: Thanks. I overlooked that one. We can already get `categoryIds` from the router state, there was no need to store it in a separate state.

hooks/useForumCategories.ts
Devorein: I went with B since `useForumFilters` depends on `useForumCategories` for the `categories`. Refrained from C as we might need to use those api methods later on

components/common/Icons/hoverIconsStyle.ts
Devorein: I've created a [Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=1d9b3860-9345-4ad1-9780-54e6f4a0795a) and assigned myself. Will take care of it as part of quick followup

PR #1301, opened 2022-12-02
Show the forum for CharmVerse team

motechFR:
Quick fix for showing Forums for charmverse team

### DIFF
--- a/pages/[domain]/forum/index.tsx
+++ b/pages/[domain]/forum/index.tsx
@@ -10,7 +10,7 @@ export default function ForumPage() {
   const space = useCurrentSpace();
 
   // Show this page only to charmverse users
-  if (!space?.name.includes('charmverse') && isProdEnv) {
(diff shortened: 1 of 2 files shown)


pages/[domain]/forum/index.tsx
```diff
@@ -10,7 +10,7 @@ export default function ForumPage() {
   const space = useCurrentSpace();
 
   // Show this page only to charmverse users
-  if (!space?.name.includes('charmverse') && isProdEnv) {
+  if (!isProdEnv || (isProdEnv && space?.domain.includes('charmverse'))) {
```
Comment (msadura) [inline pages/[domain]/forum/index.tsx:13]: since this is reused maybe we can have "isCharmVerse" util?

Comment (msadura) [reply pages/[domain]/forum/index.tsx:13]: or some sort of a flag that will check both `isProdEnv` and cv workspace

Comment (motechFR) [reply pages/[domain]/forum/index.tsx:13]: I initially left it outta scope for this one, but agreed, lets plug it in

PR #1307, opened 2022-12-03
add slash command for Dune embeds

mattcasey:
One note: for Figma and Youtube, we support translating normal URLs to embed-friendly ones. But this is impossible with Dune, and when you go to share, it gives you the embeddable URL already:
image

Test URL: https://dune.com/embeds/1394917/2370625/040ebc8a-7ad5-4777-b42f-72307baee48b.

This is the updated slash menu:
image


Devorein (approved):

Devorein:
The horizontal resizing doesn't work to be working (only for Dune embeds), the vertical one does however.

mattcasey:
> The horizontal resizing doesn't work to be working (only for Dune embeds), the vertical one does however.
I think that's unrelated to this PR. But also, not sure we need users to resize iframes horizontally...

PR #1312, opened 2022-12-04
Add voting UI and endpoints for posts

Devorein:


components/forum/components/ForumPost/PostVote.tsx
```diff
@@ -0,0 +1,79 @@
...
+}) {
+  const theme = useTheme();
+
+  async function votePost(_upvoted?: boolean) {
+    const forumPostPageVote = await charmClient.forum.votePost({
+      postId,
+      upvoted: _upvoted
+    });
+
+    setPosts((postPages) => {
```
msadura, line 37: This logis feels to be in wrong place. This is component for single post so it should not do any operations using `setPosts`.  It should rather receive `updatePost` method, or `upvotePostVote` method and it should do all of the logic in a higher component.

lib/forums/posts/voteForumPost.ts
```diff
@@ -0,0 +1,70 @@
...
+  upvoted,
+  userId,
+  pageId
+}: {
+  pageId: string;
+  userId: string;
+  upvoted?: boolean;
+}): Promise<ForumPostPageVote> {
+  if (upvoted === undefined) {
+    await prisma.pageUpDownVote.delete({
```
msadura, line 15: `pageUpDownVote` - upDown vote? 😅  Why not just `PageVote`

lib/forums/posts/interfaces.ts
```diff
@@ -4,4 +4,6 @@ export type ForumPostContent = {
   type: 'text' | 'image';
   content: string;
 };
-export type ForumPostPage = Page & { post: Post };
+export type ForumPostPageVote = { upvotes: number; downvotes: number; upvoted?: boolean };
```
msadura, line 7: I do not understand this logic - you store upvodtes, downvotes and `upvoted` - what about `downvoted` ?

pages/api/forums/posts/[postId]/vote.ts
```diff
@@ -0,0 +1,48 @@
...
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).put(voteForumPostHandler);
+
+async function voteForumPostHandler(req: NextApiRequest, res: NextApiResponse<ForumPostPageVote>) {
+  const { postId } = req.query as any as { postId: string };
+  const userId = req.session.user.id;
+  const { upvoted } = req.body;
+
+  const post = await prisma.post.findUnique({
```
Comment (msadura) [inline pages/api/forums/posts/[postId]/vote.ts:20]: plase move logic from controller to  functoin in `lib/` dir

lib/forums/posts/interfaces.ts
msadura: maybe something like action: 'upvote' | 'downvote' would be more clear

msadura (changes requested):

.vscode/settings.json
```diff
@@ -25,6 +25,7 @@
   "jestrunner.configPath": "jest.config-all.ts",
   "typescript.tsdk": "node_modules/typescript/lib",
   "cSpell.words": [
+    "downvotes",
```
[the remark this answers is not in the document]
Devorein: Its from an extension that checks whether the spelling of a word is correct or not 😛 
I will remove this one as well as the other one I added by accident

lib/forums/posts/voteForumPost.ts
Devorein: Yeah I was a bit confused about that as well (P.S I didn't create the prisma schema 😛)
Probably `PostVote` would've been best, but since its already in prod, I think we have to keep it as is

lib/forums/posts/interfaces.ts
msadura: so you can cast a vote without casing a vote? It does not make sense. Additionally logic above is clear because you described it to me but it is not intuitive at all.

Devorein: Yeah I agree with that, but not sure how else I could've wrote it. Also this is not for casting a vote, its to show whether you've voted on a post or not. We delete the vote on be if `upvoted === undefined`

lib/forums/posts/voteForumPost.ts
```diff
@@ -0,0 +1,93 @@
...
+import { prisma } from 'db';
+
+import type { ForumPostPageVote } from './interfaces';
+
+export async function voteForumPost({
+  upvoted,
+  userId,
+  postId
+}: {
+  postId: string;
```
mattcasey, line 12: ** always use pageId, not postId **

lib/forums/posts/voteForumPost.ts
mattcasey: > since its already in prod, I think we have to keep it as is
Definitely not. we have not released this to customers yet

lib/forums/posts/voteForumPost.ts
```diff
@@ -0,0 +1,93 @@
...
+      where: {
+        createdBy_pageId: {
+          createdBy: userId,
+          pageId
+        }
+      }
+    });
+  }
+
+  const page = await prisma.page.findUnique({
```
mattcasey, line 65: i don't think this method should query the vote state. Can't the UI figure it out on its own?

lib/forums/posts/listForumPosts.ts
```diff
@@ -91,7 +98,20 @@ export async function listForumPosts({
...
   const response: PaginatedPostList = {
-    data: posts,
+    data: pages.map((_page) => {
+      const { upDownVotes, post, ...rest } = _page;
+      const userVote = upDownVotes.find((vote) => vote.createdBy === userId);
+      const forumPostPage: ForumPostPage = {
+        ...rest,
+        post: {
+          ...post!,
+          downvotes: upDownVotes.filter((vote) => !vote.upvoted).length,
```
mattcasey, line 108: instead of returning downvotes and upvotes to the frontend, i think it would be better to just do the math here and return 'voteCount' or something

components/forum/components/ForumPost/ForumPost.tsx
```diff
@@ -34,9 +44,48 @@ function ForumPostContent({
...
+      return postPages
+        ? {
+            cursor: postPages.cursor,
+            data: postPages.data.map((page) => {
+              if (page.postId === postId) {
+                return {
+                  ...page,
+                  post: {
+                    ...page.post,
+                    ...forumPostPageVote,
```
mattcasey, line 77: so here, just add or subtract 1 based on `upvoted`

components/forum/components/ForumPost/ForumPost.tsx
```diff
@@ -34,9 +44,48 @@ function ForumPostContent({
...
+  async function votePost(upvoted?: boolean) {
+    const forumPostPageVote = await charmClient.forum.votePost({
+      postId,
+      upvoted
+    });
+
+    setPosts((postPages) => {
+      return postPages
+        ? {
+            cursor: postPages.cursor,
```
mattcasey, line 70: its probably better to do `...postPages` in case we add something else to the interface in the future, instead of setting cursor and hasNext, etc..

mattcasey (changes requested):
see comments

lib/forums/posts/voteForumPost.ts
Devorein: Thats a very good point. I totally overlook that part but it was getting quite difficult to update it on the frontend. In most of the cases we update the cache based on the mutation response but this is definitely a rare case and needs a different approach

PR #1315, opened 2022-12-05
Notify users in realtime for a new post

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=fffaec65-404b-4848-9395-4a1ecce7cdcb)

image


components/forum/components/ForumPosts/ForumPosts.tsx
```diff
@@ -134,7 +182,9 @@ export default function ForumPosts({ search, categoryId }: ForumPostsProps) {
 
   return (
     <>
-      <CreateForumPost />
+      <Box ref={createPostBoxRef}>
```
mattcasey, line 185: you could just use 'div' here

mattcasey (changes requested):
needs to not use snackbar

hooks/useSnackbar.tsx
```diff
@@ -43,11 +47,14 @@ export function SnackbarProvider({ children }: { children: ReactNode }) {
     if (reason === 'clickaway') {
       return;
     }
-
+    setOrigin({
```
mattcasey, line 50: what happens when snackbar is used again? won't the origin be incorrect?
I think this feature is better off not using snackbar.. the utility is intended for quick messages in the bottom-left, which we use in a lot of places. whereas the 'new posts' option is very specific

mattcasey (changes requested):
I think this feature is better off not using the global snackbar.. the utility is intended for quick messages in the bottom-left, which we use in a lot of places. whereas the 'new posts' option is a very specific UI component. (once you start adding lots of configuration props, overrides, etc. to a component it's a sign that it may be better to just build a custom one)

Also... looks like we were not creating posts originally until this PR? That needs to be fixed, and we need a "Publish" button too

lib/websockets/interfaces.ts
```diff
@@ -60,6 +60,11 @@ type VotesUpdated = {
   payload: ExtendedVote[];
 };
 
+type PostCreated = {
+  type: 'post_created';
+  payload: Post & { page: Page | null };
```
mattcasey, line 65: i don't think this event needs any payload, definitely not all the page content, etc.

pages/api/pages/index.ts
```diff
@@ -121,6 +121,32 @@ async function createPageHandler(req: NextApiRequest, res: NextApiResponse<IPage
...
+            connect: {
+              id: newPageToNotify.id
+            }
+          },
+          status: 'draft'
+        }
+      });
+
+      if (post) {
+        relay.broadcast(
```
mattcasey, line 137: this event should only be triggered when a post is published

hooks/useSnackbar.tsx
Devorein: Not if its the origin is updated beforehand. But I agree this is not the safest approach.

mattcasey (approved):

PR #1322, opened 2022-12-07
fix autofill color on safari

mattcasey:


motechFR (approved):
Tested and approved
EBFBFAD9-5089-45EC-9895-91246BF131B4

PR #1323, opened 2022-12-07
Use central socket server in production

mattcasey:


background/websockets.ts
```diff
@@ -2,7 +2,7 @@ import { createServer } from 'http';
 
 import { Server } from 'socket.io';
 
-import { baseUrl, appEnv } from 'config/constants';
```
mattcasey, line 5: checking baseUrl actually doesn't make sense since the websocket server has a unique domain. Also, we don't set DOMAIN currently so this value is undefined

background/websockets.ts
```diff
@@ -17,12 +17,11 @@ const io = new Server(server, {
     allowedHeaders: ['authorization'],
     credentials: true,
     origin: (requestOrigin, callback) => {
-      if (baseUrl) {
```
mattcasey, line 20: checking baseUrl actually doesn't make sense since the websocket server has a unique domain. Also, we don't set DOMAIN currently so this value is undefined

PR #1324, opened 2022-12-07
move floating menu code internally

mattcasey:


components/common/CharmEditor/components/floatingMenu/Icons.tsx
```diff
@@ -0,0 +1,207 @@
...
+      fill='none'
+      xmlns='http://www.w3.org/2000/svg'
+      {...props}
+    >
+      <path d='M6.34317 7.75732L4.92896 9.17154L12 16.2426L19.0711 9.17157L17.6569 7.75735L12 13.4142L6.34317 7.75732Z' />
+    </svg>
+  );
+}
+
+export function ChevronUp(props: any) {
```
Devorein, line 195: Probably best to use `SVGProps` rather than `any`

Devorein (approved):

components/common/CharmEditor/components/floatingMenu/Icons.tsx
mattcasey: Nice catch. this is actually code from bangle.dev... i'm going to see if we can use mui icons instead

PR #1328, opened 2022-12-08
Integrate forum API with UI

Devorein:


pages/api/forums/posts/[postId]/publish.ts
```diff
@@ -0,0 +1,28 @@
...
+import { onError, onNoMatch, requireUser } from 'lib/middleware';
+import { withSessionRoute } from 'lib/session/withSession';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).put(publishForumPostHandler);
+
+// TODO - Update posts
+async function publishForumPostHandler(req: NextApiRequest, res: NextApiResponse) {
+  const { postId } = req.query as any as { postId: string };
```
Comment (mattcasey) [inline pages/api/forums/posts/[postId]/publish.ts:15]: use pageId not postId

lib/permissions/pages/triggers/page-created.ts
```diff
@@ -9,7 +9,10 @@ import { copyAllPagePermissions } from '../actions/copyPermission';
 import { generateReplaceIllegalPermissions } from '../actions/replaceIllegalPermissions';
 import { upsertPermission } from '../actions/upsert-permission';
 
-export async function setupPermissionsAfterPageCreated(pageId: string): Promise<IPageWithPermissions> {
+export async function setupPermissionsAfterPageCreated(
+  pageId: string,
+  addSpaceLevelPermissions = true
```
mattcasey, line 14: true should never be a default

mattcasey: Also, I thought @motechFR said we are not using page permissions at this time? I can't tell all the changes being made, but this seems a bit out of scope for this PR..

lib/forums/posts/createForumPost.ts
```diff
@@ -42,7 +42,7 @@ export async function createForumPost({
     data: {
       id: postId,
       title,
-      content: content as Prisma.InputJsonObject,
+      content: (content ?? undefined) as Prisma.InputJsonObject,
```
mattcasey, line 45: there's never any content when creating a forum post the first time. do we even need this?

pages/api/forums/posts/[postId]/index.ts
```diff
@@ -55,36 +35,31 @@ async function deleteForumPostController(req: NextApiRequest, res: NextApiRespon
...
+    postId,
+    userId
   });
 
-  if (!post || !post.page) {
-    return res.status(200).json({});
-  }
+  await deleteForumPost(postId);
+
+  res.status(200).json({});
```
Comment (mattcasey) [inline pages/api/forums/posts/[postId]/index.ts:45]: res.status(200).end()

components/forum/components/CreateForumPost/CreateForumPost.tsx
```diff
@@ -5,45 +5,57 @@ import CardContent from '@mui/material/CardContent';
...
+              value={pageTitle}
+              fullWidth
+              onChange={(e) => {
+                setPageTitle(e.target.value);
+              }}
+            />
           </Box>
           <Box display='flex' justifyContent='flex-end'>
-            <Button component='div' float='right'>
+            <Button disabled={!userSpacePermissions?.createPage} component='div' float='right' onClick={addPageCb}>
```
mattcasey, line 58: needs a tooltip when disabled. if we're using our common Button component you can set this with `disabledTooltip`

components/forum/ForumPage.tsx
```diff
@@ -46,6 +49,17 @@ export default function ForumPage() {
     }
   }
 
+  useEffect(() => {
+    if (typeof router.query.postId === 'string') {
+      showPage({
+        pageId: router.query.postId,
```
mattcasey, line 55: change req.query.postId to req.query.pageId

components/forum/ForumPage.tsx
```diff
@@ -46,6 +49,17 @@ export default function ForumPage() {
...
+        onClose() {
+          setUrlWithoutRerender(router.pathname, { postId: null });
```
mattcasey, line 57: if we are relying on router.query.postId to show the popup, then we should update the path normally.

components/forum/ForumPage.tsx
```diff
@@ -19,6 +21,7 @@ export default function ForumPage() {
 
   const currentSpace = useCurrentSpace();
   const { categories } = useForumFilters();
+  const { showPage } = usePageDialog();
```
mattcasey, line 24: let's not use page dialog for now - Chris said we should either link to a full page view, or come up with a new modal that looks like Reddit. This PR has gotten too huge already

mattcasey (changes requested):
This PR is a lot of changes - shouldn't it just be 'wiring up frontend and backend'? Why the addition of page dialog modal as well as page permissions? I'd like to discuss how page permissions will be used. AFAIK we may not even need them

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -244,7 +243,7 @@ function DocumentPage({ page, setPage, insideModal, readOnly = false, parentProp
                       refreshBountyPermissions={refreshBountyPermissions}
                     />
                   )}
-                  {page.postId && <PostProperties pageId={page.id} postId={page.postId} readOnly={readOnly} />}
+                  {page.postId && <PostProperties postId={page.postId} readOnly={!pagePermissions.edit_content} />}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:246]: pageId, not postId

pages/api/forums/posts/[postId]/publish.ts
Comment (mattcasey) [inline pages/api/forums/posts/[postId]/publish.ts:15]: it would probably be better to use pageId here instead of postId, since we have to consider both records anyway.. pageId should be preferred

mattcasey (approved):

PR #1333, opened 2022-12-09
set galleryimage and has content when editing a page thru websockets

mattcasey:
I also ran scripts/updateCardFallbackPreview to fix previous cards and fixed it so that we send updates when galleryImage/hasContent has changed

### DIFF
--- a/lib/pages/server/updatePage.ts
+++ b/lib/pages/server/updatePage.ts
@@ -17,23 +17,14 @@ export async function updatePage(
   updates: Prisma.PageUpdateInput
 ): Promise<IPageWithPermissions> {
   const data: Prisma.PageUpdateInput = {
-    updatedAt: new Date(),
     ...updates,
+    updatedAt: new Date(),
     updatedBy: userId
   };
 
   if (data.id) {
     // avoid overriding page id
     delete data.id;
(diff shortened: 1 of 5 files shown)


lib/pages/server/updatePage.ts
```diff
@@ -27,15 +27,6 @@ export async function updatePage(
     delete data.id;
   }
 
-  // check if content is empty only if it got changed
-  if ('content' in updates) {
```
mattcasey, line 31: I removed this action to set galleryImage and hasContent since it should be no longer used

pages/api/pages/index.ts
```diff
@@ -50,8 +50,6 @@ async function createPageHandler(req: NextApiRequest, res: NextApiResponse<IPage
     throw new UnauthorisedActionError('You do not have permissions to create a page.');
   }
 
-  data.hasContent = !checkIsContentEmpty(data.content as PageContent);
```
mattcasey, line 53: I removed this action to set hasContent - the content is always empty when we first create a page, unless we're maybe duplicating a template but then we should be duplicating the 'hasContent' field as well

msadura (approved):

PR #1334, opened 2022-12-12
Sshrayber group nft

HagiaSofiya:


HagiaSofiya:
Screenshot 2022-12-11 at 11 09 51 PM

components/profile/PublicProfile.tsx
```diff
@@ -211,6 +215,27 @@ export default function PublicProfile(props: UserDetailsProps) {
     }
   }, [collabError]);
 
+  function groupBy<T>(arr: T[], fn: (item: T) => any) {
+    return arr.reduce<Record<string, T[]>>((prev, curr) => {
+      const groupKey = fn(curr);
+      const group = prev[groupKey] || [];
+      group.push(curr);
+      return { ...prev, [groupKey]: group };
```
msadura, line 223: please always avoid using spreading obejcts in reduce https://prateeksurana.me/blog/why-using-object-spread-with-reduce-bad-idea/

```
acc[groupKey] = group;
return acc;
```

should do the job here

msadura (changes requested):

msadura:
UI-wise I would make less verical sapcing between groups

mattcasey (approved):
Approved but agree with Marek's request - can you remove the spread inside reduce? Thanks!

components/profile/PublicProfile.tsx
HagiaSofiya: @msadura Thank you for the article! I am not following the change you are asking for.  Would you mind sending me a sample?

msadura: sure:

```
  function groupBy<T>(arr: T[], fn: (item: T) => any) {
    return arr.reduce<Record<string, T[]>>((acc, curr) => {
      const groupKey = fn(curr);
      const group = prev[groupKey] || [];
      group.push(curr);
      
      acc[groupKey] = group;
      return acc;
    }, {});
 ```

msadura (approved):

mattcasey:
we decided to nix this design for now, and the dev working on it has disappeared

PR #1336, opened 2022-12-12
Fix  - Forum posts request called 2x

valentinludu:
Fixed:
- loadPosts() was called 2 times - now I hide the ref while API call loads so we don't accidentally call it twice.
- Select was throwing a warning in the console because it has an empty array as a value. Now it defaults to 'none'.
- setError(null) -> call it only if there was an error before. So we avoid a render with this.


components/forum/components/ForumPosts/ForumPosts.tsx
```diff
@@ -135,12 +137,12 @@ export default function ForumPosts({ search, categoryId }: ForumPostsProps) {
...
       <CreateForumPost />
       {error && <Alert severity='error'>There was an unexpected error while loading the posts</Alert>}
-
       {posts?.data.map((post) => (
         <ForumPost key={post.id} {...post} />
       ))}
       {isLoadingMore && <ForumPostSkeleton />}
-      <Box ref={ref}>{posts?.hasNext === false && <Alert severity='info'>No more posts to show</Alert>}</Box>
+      {posts?.hasNext === false && <Alert severity='info'>No more posts to show</Alert>}
+      <Box ref={ref} sx={{ display: isLoadingMore ? 'none' : 'block' }} />
```
mattcasey, line 145: Could be written using box properties: `<Box ref={ref} display={isLoadingMore ? 'none' : 'block'} />`

mattcasey (approved):
This issue could also be addressed with useSWRInfinite, right (only refreshing based on the cache key)? Mo removed it when we re-implemented paging on the back-end, but it was only because he could not get it to work and was in a rush. @valentinludu do you think it'd be worth re-writing again at some point?

components/forum/components/ForumPosts/ForumPosts.tsx
valentinludu: Done

valentinludu:
> This issue could also be addressed with useSWRInfinite, right (only refreshing based on the cache key)? Mo removed it when we re-implemented paging on the back-end, but it was only because he could not get it to work and was in a rush. @valentinludu do you think it'd be worth re-writing again at some point?

Totally agree.
I will make a story for this to try to make it work with SWRInfinite.

mattcasey (approved):

PR #1347, opened 2022-12-14
feat: basic serverless setup

Rieranthony:
### Webhook in charmverse using serverless
Architecture drafted in this document: https://app.charmverse.io/charmverse/page-574326526436393


.github/workflows/deploy_serverless_staging.yml
```diff
@@ -0,0 +1,35 @@
...
+    types: [labeled, opened, synchronize]
+    paths:
+      # Trigger the job only when the serverless config or files have been changed
+      - serverless/**
+      - serverless.yml
+
+jobs:
+  deploy:
+    if: |
+      (github.event.action == 'labeled' && github.event.label.name == ':rocket: deploy') ||
```
mattcasey, line 14: I made a new label called ":rocket: serverless" so that we aren't triggering Lambda always. 99% of the time, people will just want to view the webapp

components/settings/api/Api.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import Legend from '../Legend';
+
+interface Props {
+  isAdmin: boolean;
+  spaceId: string;
+  spaceOwner: string;
+}
+
+export const schema = yup.object({
```
mattcasey, line 22: thanks for using yup... we only have it in a few places but i'd like to use it more often

prisma/schema.prisma
```diff
@@ -1093,6 +1096,15 @@ model UnstoppableDomain {
   user   User   @relation(fields: [userId], references: [id], onDelete: Cascade)
 }
 
+model WebhookSubscription {
+	id        String    @id @default(uuid()) @db.Uuid
+  scope     String
+	spaceId   String    @db.Uuid
+  space     Space     @relation(fields: [spaceId], references: [id], onDelete: Cascade)
+	createdAt DateTime  @default(now())
```
mattcasey, line 1104: could you also capture createdBy? might be important for security to know who added it

serverless/webhook/endpoint.ts
```diff
@@ -0,0 +1,32 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+
+export const webhookEndpoint = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  try {
+    const body = event?.body;
+    const headers = event?.headers;
+
+    // eslint-disable-next-line no-console
+    console.log('BODY', JSON.stringify(body));
```
mattcasey, line 9: replace with our logging util:
```
import log from 'lib/log'

log.debug()
```

serverless/webhook/endpoint.ts
```diff
@@ -0,0 +1,32 @@
+import type { APIGatewayProxyEvent, APIGatewayProxyResult } from 'aws-lambda';
+
+export const webhookEndpoint = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
```
mattcasey, line 3: is this webhook supposed to do something?

mattcasey: Ah I see there's a comment in the serverless.yml file after searching around for where we use "webhookEndpoint". I am not sure most people would look there to find the answer to what this folder is. Even a reference from the README would be useful. ("To understand how this code is used, review serverless.yml file" or something like that)

serverless.yml
```diff
@@ -0,0 +1,51 @@
...
+
+provider:
+  name: aws
+  runtime: nodejs14.x
+  deploymentMethod: direct
+
+functions:
+  # This function is purely for testing our implementation, it received a webhook call and logs it
+  # Returns a 200 HTTP response
+  webhookEndpoint:
```
mattcasey, line 48: Suggestion: what about calling this "webhookHealthcheck" or something more clear that this is just for testing purposes?

serverless/webhook/worker.ts
```diff
@@ -0,0 +1,76 @@
+import type { SQSBatchItemFailure, SQSEvent, SQSHandler, SQSRecord } from 'aws-lambda';
+import type { KeyLike } from 'jose';
+import { SignJWT } from 'jose';
+import fetch from 'node-fetch';
+
+import type { WebhookPayload } from './interfaces';
+
+const signJwt = async (subject: string, payload: Record<string, any>, secret: KeyLike | Uint8Array) => {
```
mattcasey, line 8: the `async` seems unecessary

serverless/webhook/worker.ts
```diff
@@ -0,0 +1,76 @@
...
+          throw new Error(`Expect error 200 back. Received: ${response.status}`, {
+            cause: {
+              webhookData,
+              status: response.status
+            }
+          });
+        }
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.log(`Error in processing SQS Worker: ${body}`);
```
mattcasey, line 65: replace with `log.error` from `lib/log`

serverless/webhook/worker.ts
```diff
@@ -0,0 +1,76 @@
...
+      // change it
+      .setExpirationTime('15m')
+      .sign(secret)
+  );
+};
+
+/**
+ * SQS worker, message are executed one by one
+ */
+export const webhookWorker: SQSHandler = async (event: SQSEvent) => {
```
mattcasey, line 25: it would be great to have at least one test that shows how this method is called and returns successfully, maybe just with a mocked SQSEvent

package.json
```diff
@@ -172,6 +173,7 @@
     "next-connect": "^0.13.0",
     "next-swagger-doc": "^0.3.6",
     "node-cron": "^3.0.0",
+    "node-fetch": "^2.6.8",
```
mattcasey, line 176: I think you can remove this. we don't use node-fetch because we're on an updated version of node

mattcasey (approved):

PR #1355, opened 2022-12-15
block key events from propagating out of inline databases

mattcasey:


components/common/CharmEditor/components/inlineDatabase/components/InlineDatabase.tsx
```diff
@@ -104,7 +103,7 @@ export default function DatabaseView({ containerWidth, readOnly: readOnlyOverrid
   const allViews = useAppSelector(getSortedViews);
   const router = useRouter();
 
-  const views = allViews.filter((view) => view.parentId === pageId);
+  const views = useMemo(() => allViews.filter((view) => view.parentId === pageId), [pageId, allViews]);
```
mattcasey, line 106: i memoized the methods here, now we don't re-render the database when re-rendering events in CharmEditor (even hovering over the doc was triggering re-renders)

motechFR (approved):
Works, nice one!

PR #1359, opened 2022-12-16
Feat/user picker delete

msadura:
Screenshot 2022-12-16 at 10 34 13

Initially I thought about replacing user picker component with MUI autocomplete, but turned out it was a matter of showing X button


mattcasey (approved):
Nice.. looks like i had it commented out ever since we ported the code here. Maybe i was hiding another element with' control' class someplace else?

PR #1361, opened 2022-12-16
Import csv database

valentinludu:
Import csv in any page that is a type board.


components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
+import _ from 'lodash';
+import type { ParseResult } from 'papaparse';
+import * as uuid from 'uuid';
+
+import type { IPropertyTemplate, PropertyType } from 'lib/focalboard/board';
+import type { Member } from 'lib/members/interfaces';
+import { focalboardColorsMap } from 'theme/colors';
+
+interface MappedProperties {
```
mattcasey, line 9: use type, not interface

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
...
+    typeof results.data[0] === 'object' &&
+    results.data[0] !== null &&
+    results.meta.fields
+  ) {
+    return true;
+  }
+  return false;
+};
+
+export const mapCardBoardProperties = (cardProperties: IPropertyTemplate[]) =>
```
mattcasey, line 60: we use `function <functionName> () {}` notation vs `const <functionName>` notation

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
+import _ from 'lodash';
```
mattcasey, line 1: Please use individual lodash methods... this will import the entire library unecessarily. i dont think chaining is worth it

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
...
+    return { ...defaultProps, type: 'email' };
+  }
+  if (propValues.some((p) => p === 'true' || p === 'false')) {
+    return { ...defaultProps, type: 'checkbox' };
+  }
+
+  return { ...defaultProps, type: 'text' };
+};
+
+export const createCardFieldProperties = (
```
mattcasey, line 127: this file really needs some tests, i dont really know what it is doing and it looks like there were some edge cases

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
...
+    const allMultiSelectValues = csvData.map((result) => result[prop].split('|')).flat();
+    const uniqueMultiSelectValues = [...new Set(allMultiSelectValues)];
+    const options = createBoardPropertyOptions(uniqueMultiSelectValues);
+
+    return { ...defaultProps, options, type: 'multiSelect' };
+  }
+  if (propValues.every((p) => uuid.validate(p))) {
+    return { ...defaultProps, type: 'person' };
+  }
+  if (propValues.some((p) => monthNames.find((month) => p.includes(month)))) {
```
mattcasey, line 108: i think this is too loose to determine date. we should be bale to convert it to a date object, or don't consider it a date

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export const createNewPropertiesForBoard = (
+  csvData: Record<string, string>[],
+  prop: string,
+  existingBoardProp?: MappedProperties
+): IPropertyTemplate => {
+  const propValues = csvData.map((result) => result[prop]).filter((result) => !!result);
+  const defaultProps = { id: uuid.v4(), name: prop, options: [] };
+  // eslint-disable-next-line
+  const emailReg = /^([A-Za-z0-9_\-\.])+\@([A-Za-z0-9_\-\.])+\.([A-Za-z]{2,4})$/;
```
mattcasey, line 85: eslint is right in this case, it looks like you don't need to escape a period inside a character class

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -0,0 +1,200 @@
...
+  if (propValues.every((p) => !Number.isNaN(+p))) {
+    return { ...defaultProps, type: 'number' };
+  }
+  if (propValues.some((p) => p.startsWith('http') || p.startsWith('www.'))) {
+    return { ...defaultProps, type: 'url' };
+  }
+  if (propValues.some((p) => emailReg.test(p))) {
+    return { ...defaultProps, type: 'email' };
+  }
+  if (propValues.some((p) => p === 'true' || p === 'false')) {
```
mattcasey, line 120: why are we not checking all values? I think using some is unsafe, a single row could have complete garbage in it.

components/common/PageLayout/components/Header/components/DatabasePageOptions.tsx
```diff
@@ -115,6 +133,81 @@ export default function DatabaseOptions({ pagePermissions, closeMenu, pageId }:
...
+    const allAvailableProperties = headers.filter((header) => header !== 'Name');
+
+    const mappedInitialBoardProperties = mapCardBoardProperties(board.fields.cardProperties);
+
+    // Create card properties for the board
+    const newBoardProperties = allAvailableProperties.map((prop) =>
+      createNewPropertiesForBoard(csvData, prop, mappedInitialBoardProperties[prop])
+    );
+    // Merge the fields of both boards
+    const mergedFields = deepMergeArrays(newBoardProperties, board.fields.cardProperties);
```
mattcasey, line 150: I would have not done a deep merge, I'm not sure what properties you are wanting to merge between properties... but I wonder if it could cause issues if we change an existing column, for example with the wrong type. I would have instead just appended new properties, but left the existing ones as they were

components/common/PageLayout/components/Header/components/DatabasePageOptions.tsx
```diff
@@ -115,6 +133,81 @@ export default function DatabaseOptions({ pagePermissions, closeMenu, pageId }:
...
+    const newBlock: Block = {
+      ...board,
+      fields: {
+        ...board.fields,
+        cardProperties: mergedFields
+      }
+    };
+
+    // Update board with new cardProperties
+    await charmClient.updateBlock(newBlock);
```
mattcasey, line 162: is there a reason not to use charmClient.patchBlock?

mattcasey (changes requested):
- make sure all columns are valid for the type we select
- more strict rules on dates and make sure they're formatted as timestamps
- add some tests on databasePageOptions, maybe one for each type? that is expected
- we should always pick a column to be page title, the first one that is text is fine
- fix lodash import to only import methods you need (i think this is important, altho I'm only seeing a 30KB increase on our frontend.. it adds up)

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
valentinludu: I changed the condition so all strings that pass the check return a certain type.
In this function I am just creating the types for the board. I am filtering from the start null, undefined or empty strings.

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
valentinludu: Using luxon I am checking if it's a date from ISO or from a timestamp.

components/common/PageLayout/components/Header/components/DatabasePageOptions.tsx
valentinludu: I tried to use it but did not work. I lost a few hours trying to format my data accordingly but still did not work in my case. I didn't want to lose to much time on it.

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
valentinludu: I will make a separate PR for the tests

components/common/PageLayout/components/Header/components/DatabasePageOptions.tsx
valentinludu: The merge is mostly for the options array when we have the same type.
I will double check to see that properties with the same name but different type are not overwritten.

mattcasey (approved):

PR #1362, opened 2022-12-16
Refactor notion import code & fix known issues

Devorein:


lib/notion/NotionImporter/CharmverseDatabasePage.ts
```diff
@@ -0,0 +1,133 @@
+import type { DatabaseObjectResponse } from '@notionhq/client/build/src/api-endpoints';
```
mattcasey, line 1: use lowercase file names except for component files

lib/notion/NotionImporter/NotionCache.ts
```diff
@@ -0,0 +1,60 @@
...
+export type DatabasePageItem = {
+  charmversePage?: Page;
+  notionPage?: {
+    pageIds: string[];
+    properties: Record<string, IPropertyTemplate>;
+  };
+  type: 'database';
+};
+
+export class NotionCache {
```
mattcasey, line 27: use lowercase file names except for component files.. 'notionImporter' should be lowercased

lib/notion/convertToPlainText.ts
```diff
@@ -0,0 +1,3 @@
+export function convertToPlainText(chunks: { plain_text: string }[]) {
```
mattcasey, line 1: these could go in a 'utils' folder, so its easier to find

lib/notion/NotionImporter/NotionCache.ts
mattcasey: i see what you did.. 'NotionImporter' is the main class. I guess it's a valid exception

mattcasey (approved):
approved.. but please remove the state of a page on notionimporter

PR #1367, opened 2022-12-20
Show newly created post on forum

Devorein:


mattcasey (changes requested):
Instead of hoisting the state up, we are already listening for 'publish events' via the webhook listener. It would be much simpler to have the view just refresh itself.

mattcasey (changes requested):
Since we're already listening to publish events, why not just refresh the view inside of `handlePostPublishEvent()`? This avoids us having to couple the create form with the logic to populate the feed list, which might change in the future.

mattcasey:
@Devorein I added this 'feature' in my PR here: https://github.com/charmverse/app.charmverse.io/pull/1371/files#diff-66b70237d6d836d8ab22f6450ac36090fa113d3128dc939f563548f542c3a641R167. We just refresh the view if we receive a new publish event from the user

PR #1368, opened 2022-12-20
csv import improvement

valentinludu:


mattcasey (approved):

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -168,6 +171,44 @@ export function createCardFieldProperties(
...
+        return {
+          ...acc,
+          [propId]: jsonValue
+        };
+      }
+
+      if (valuesArr.length === 2) {
+        const fromDateString = valuesArr[0];
+        const toDateString = valuesArr[1];
+        const from = new Date(isNumber(fromDateString) ? +fromDateString : fromDateString).getTime();
```
mattcasey, line 192: FWIW I don't think the "+" before fromDateString is doing anything here, Date can take a number

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -168,6 +171,44 @@ export function createCardFieldProperties(
       }
     }
 
+    if (mappedBoardProperties[key]?.type === 'date') {
+      const valuesArr = value.split(' -> ');
+
+      if (valuesArr.length === 1) {
+        const fromDateString = valuesArr[0];
+        const from = new Date(isNumber(fromDateString) ? +fromDateString : fromDateString).getTime();
+        const jsonValue: string = JSON.stringify({
```
mattcasey, line 180: nice catch using stringify, i would have missed this

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
```diff
@@ -168,6 +171,44 @@ export function createCardFieldProperties(
       }
     }
 
+    if (mappedBoardProperties[key]?.type === 'date') {
+      const valuesArr = value.split(' -> ');
+
+      if (valuesArr.length === 1) {
+        const fromDateString = valuesArr[0];
+        const from = new Date(isNumber(fromDateString) ? +fromDateString : fromDateString).getTime();
```
mattcasey, line 179: it would be nice to see this as a unit-tested method as well, then you could loop over all the test cases for 'isValidDate' and make sure we actually produce valid dates... another idea for the future would be to just have a single 'getValidDate' method, that does validation and conversion and just returns null or Date. it could be used by both the property type detection and value mapping

mattcasey (approved):
Thanks for the tests!

One test is failing... I'm guessing the input string value for a date is at the month level or something, but it's including the current hour/minutes for the rest of it.

valentinludu:
> Thanks for the tests!
> 
> One test is failing... I'm guessing the input string value for a date is at the month level or something, but it's including the current hour/minutes for the rest of it.

Because of the server time it was a difference of 2 hours in the timestamp. I made a workaround to return the same date every time. In CV this is not necessary as we need only the date and not the exact time.

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
valentinludu: I added the + because of this behaviour:
- new Date(1639699200000) returns 'Fri Dec 17 2021 02:00:00'
- new Date('1639699200000') returns 'Invalid Date'

So if it is a number that means it is a timestamp and need to convert it into a number.

components/common/PageLayout/components/Header/components/utils/databasePageOptions.ts
valentinludu: A very good suggestion.
I changed the function so it returns Date or null and now I can use it for validation and for getting the Date.

PR #1374, opened 2022-12-21
Feat/collabland user verify

msadura:
User verification via collabland endpoint.

Unfortunately, they did not give us a way of creating our own test guild / member so we need to use hardcoded ones to test it:

GuildId 834476259199156245 (space's discordServerId)
UserId 646733536744833054 (discordUserId)

api token: 21hlXT_f0whhXLg9pyXGN.5aMN4Ya64sA7ylxNh2jtp

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -95,6 +95,7 @@ enum WalletType {
 
 enum RoleSource {
   guild_xyz
+  collabland
 }
 
 enum PagePermissionLevel {
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -722,6 +722,7 @@ model Role {
   id                        String                     @id @default(uuid()) @db.Uuid
   createdAt                 DateTime                   @default(now())
   createdBy                 String                     @db.Uuid
+  externalId                String?
   name                      String
   spaceId                   String                     @db.Uuid
   source                    RoleSource?
--- a/components/common/DiscordGate/hooks/useDiscordGate.ts
+++ b/components/common/DiscordGate/hooks/useDiscordGate.ts
@@ -19,9 +19,8 @@ export function useDiscordGate({ spaceDomain, onSuccess }: Props) {
   const [joiningSpace, setJoiningSpace] = useState(false);
   const { spaces, setSpaces } = useSpaces();
 
-  const { data } = useSWR(
-    spaceDomain && discordUserId ? `discord/gate/${spaceDomain}/${discordUserId || ''}` : null,
-    () => charmClient.discord.checkDiscordGate(spaceDomain)
+  const { data } = useSWR(spaceDomain ? `discord/gate/${spaceDomain}/${discordUserId || ''}` : null, () =>
+    charmClient.discord.checkDiscordGate(spaceDomain)
   );
 
   async function verifyDiscordGate() {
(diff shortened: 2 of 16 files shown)


background/cron.ts
```diff
@@ -26,6 +27,8 @@ cron.schedule('0 */15 * * * *', proposalTask);
 // Verify token gates and remove users who no longer meet the conditions
 cron.schedule('0 */30 * * * *', verifyTokenGateMembershipsTask);
 
+cron.schedule('0 */3 * * * *', processWebhookMessages);
```
msadura, line 30: running webhook messages fetching every 3mins - probably value to discuss

mattcasey: Actually, can you change this. Instead of a cron schedule, we should use long polling and then it can just be a function that calls itself recursively. That way we receive messages almost instantaneously: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/sqs-examples-enable-long-polling.html. It worked well at my last company

background/tasks/processWebhookMessages/task.ts
```diff
@@ -0,0 +1,13 @@
+import { processMessages } from 'lib/aws/webhookSqs';
+import log from 'lib/log';
+import { processWebhookMessage } from 'lib/webhooks/processWebhookMessage';
+
+export async function task() {
+  log.debug('Running process webhook messages cron job');
+
+  try {
+    await processMessages({ processorFn: processWebhookMessage, maxNumOfMessages: 5 });
```
mattcasey, line 9: maxNumOfMessages should be 1, and we should change the queue to FIFO

background/tasks/processWebhookMessages/task.ts
```diff
@@ -0,0 +1,13 @@
...
+  } catch (error: any) {
+    log.error(`Error deleting archived pages: ${error.stack || error.message || error}`, { error });
```
mattcasey, line 11: error message is from another task

lib/webhooks/processWebhookMessage.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const messageHandlers: Record<MessageType, (message: WebhookMessage) => Promise<boolean>> = {
+  remove_role: async (message: WebhookMessage) => {
+    try {
+      const {
+        guild_id: discordServerId,
+        member: { discordId: discordUserId },
+        role
+      } = message?.data as MemberRoleWebhookData;
+      await createAndAssignRolesDiscord({ discordUserId, discordServerId, roles: role });
```
mattcasey, line 14: i think you meant to call unassignRolesDiscord() here?

lib/webhooks/processWebhookMessage.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  },
+  add_role: async (message: WebhookMessage) => {
+    try {
+      const {
+        guild_id: discordServerId,
+        member: { discordId: discordUserId },
+        role
+      } = message?.data as MemberRoleWebhookData;
+      await unassignRolesDiscord({ discordUserId, discordServerId, roles: role });
```
mattcasey, line 27: is this supposed to be createAndAssignRolesDiscord()?

prisma/schema.prisma
```diff
@@ -723,6 +724,7 @@ model Role {
   id                        String                     @id @default(uuid()) @db.Uuid
   createdAt                 DateTime                   @default(now())
   createdBy                 String                     @db.Uuid
+  externalId                String?
```
mattcasey, line 727: lgtm

mattcasey (changes requested):
see comments

msadura:
Fixed mentioned code issues. I also removed webhook cron and I will work on making FIFO queue + long polling. I think we can deal with them in separate PR, to have discord verification merged

PR #1377, opened 2022-12-21
Show summary of post content

mattcasey:
Shows the first 3 rows of text from a post, or the image/tweet/video if it is the first row.

I also cleaned up the prosemirror interface, removed models/Person.ts, and added a new type "ForumPostMeta" for when we don't need the full content, so lots of files were changed.


testing/prosemirror.ts
```diff
@@ -0,0 +1,5 @@
+import { builders } from 'prosemirror-test-builder';
```
mattcasey, line 1: @Devorein this testing library makes it much easier to generate prosemirror doc content :)

PR #1379, opened 2022-12-22
Add SWR infinite for forum main page + improvements

valentinludu:
Changes:
- add useSWRInfinite for loading posts and search
- updated swr version to 2.0 because of some nice to have features like isLoading state for hooks. Before, we needed to write some cases to be sure the component is actually loading. Also fixed some caching differences after the update.
- improve UX: debounce search field so we don't send 10 requests for 10 keys typed. Now we send only 1, after the user finishes.
- improve UX: don't show the fetch button on new posts if the user searched for something. It's not going to see it because he sees the search results instead of the actual list of posts.
- fix: category select was throwing warnings in the console so I fixed it
- fix: down and upvote was throwing an error because of buttons inside of a button.


components/forum/ForumPage.tsx
```diff
@@ -65,6 +67,12 @@ export default function ForumPage() {
     }
   }, [router.query.pageId]);
 
+  const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setSearch(e.target.value);
+  };
+
+  const debounceSearch = debounce(handleChange, 400);
```
msadura, line 74: I think having this sort of debounce might cause memory leak - your debounced function will get updated on each render, but old reference will not be cleared. I am not entirely sure if it will be the case here, but anyway this sort of logic should be enclosed in useEffect. I think this might be a good example: https://www.carlrippon.com/using-lodash-debounce-with-react-and-ts/

valentinludu: Good catch. I totally forgot about the cancel method.

msadura (approved):

PR #1382, opened 2022-12-22
Add category title to the main forum page if category is selected

valentinludu:
https://app.charmverse.io/charmverse/page-8704528490535499


hooks/useForumCategories.ts
```diff
@@ -7,6 +8,8 @@ import { useCurrentSpace } from './useCurrentSpace';
 
 export function useForumCategories() {
   const currentSpace = useCurrentSpace();
+  const router = useRouter();
```
mattcasey, line 11: This has bit us in the past (putting state into a data source hook) I think we should use a different hook if you want to capture the 'current category'.

mattcasey (changes requested):
remove the logic to determine current category from the useForumCategories hook. it won't always be defined by the URL, i think its safer to just let components maintain that logic

hooks/useForumCategories.ts
valentinludu: OK. Then I will keep it as before. Outside, each component will determine how it gets the id.

valentinludu:
> remove the logic to determine current category from the useForumCategories hook. it won't always be defined by the URL, i think its safer to just let components maintain that logic

Done

mattcasey (approved):

PR #1387, opened 2022-12-23
Fix bug creating comments on non-authored posts

Devorein:


components/forum/components/PostPage/PostPage.tsx
```diff
@@ -92,29 +98,32 @@ export function PostPage({ page, spaceId, onSave }: Props) {
...
+    if (postComments) {
+      const postCommentsRecord: Record<string, PostCommentWithVoteAndChildren> = {};
+      postComments.forEach((postComment) => {
+        postCommentsRecord[postComment.id] = {
+          ...postComment,
+          children: []
+        };
+      });
+      postComments.forEach((postComment) => {
+        if (postComment.parentId !== page?.id) {
```
mattcasey, line 112: this would be good logic to put outside of the component, as suggested by @motechFR earlier this week (even in a separate function inside of this file though, we don't need a lib method)

components/forum/components/PostPage/PostPage.tsx
```diff
@@ -146,26 +155,40 @@ export function PostPage({ page, spaceId, onSave }: Props) {
...
-      {topLevelComments.length === 0 && (
-        <Stack gap={1} alignItems='center' my={1}>
-          <CommentIcon color='secondary' fontSize='large' />
-          <Typography color='secondary' variant='h6'>
-            No Comments Yet
-          </Typography>
-          <Typography color='secondary'>Be the first to share what you think!</Typography>
-        </Stack>
+      {isLoading ? (
+        <Box width='100%' display='flex' height={100} justifyContent='center'>
```
mattcasey, line 169: you should use LoadingComponent, it handles a lot of this already

components/forum/components/PostPage/components/PostComment.tsx
```diff
@@ -44,7 +44,7 @@ export function PostComment({
   setPostComments
 }: {
   comment: PostCommentWithVoteAndChildren;
-  setPostComments: KeyedMutator<PostCommentWithVote[]>;
+  setPostComments: KeyedMutator<PostCommentWithVote[] | undefined>;
```
mattcasey, line 47: doesnt look like this needs to offer undefined because we are always setting it?

mattcasey (approved):

PR #1388, opened 2022-12-23
Sort forum posts

valentinludu:
Added:
- newest (default)
- most voted (up & down)
- most commented


components/forum/components/PostList/PostList.tsx
```diff
@@ -30,7 +32,14 @@ const Alert = forwardRef<HTMLDivElement, AlertProps>((props, ref) => {
   return <MuiAlert elevation={6} ref={ref} variant='filled' {...props} />;
 });
 
-export function ForumPostList({ search, categoryId }: ForumPostsProps) {
+function getSortPost(item: string | undefined): SortPosts {
+  if (item === 'newest' || item === 'most_voted' || item === 'most_commented') {
```
mattcasey, line 36: Wouldn't this logic be better on the backend? I think it's better to blow up than try to coerce/fake the value. This code looks like it will hide bugs. This should be simple enough:
```
if (!SORT_TYPES.includes(sortType)) {
  throw new Error('Invalid sort type')
  }

```
And we should use a type union anywhere we set the sort so that there will be less bugs (`type SortType = 'newest' | 'most_voted' | 'most_commented')

lib/forums/posts/listForumPosts.ts
```diff
@@ -13,23 +13,24 @@ export const defaultPostsPerResult = 5;
 
 export type PaginatedPostList = PaginatedResponse<ForumPostMeta & { totalComments: number }> & { cursor: number };
 
-/**
- * @sort ignored for now - the server sorts posts by most recent
- */
+export type SortPosts = 'newest' | 'most_commented' | 'most_voted';
```
mattcasey, line 16: a better way to write this:
```
export const PostOrderOptions = ['newest', 'most_commented', 'most_voted'] as const;
export type PostOrder = typeof PostOrderOptions[number];
```
I'm not sure if you still need `as const`, but this allows us to have the best of both worlds: an array you can check at runtime and a type union for compile time

mattcasey (changes requested):
Nothing blocking really, but I think the type/list of sorts could be written in a more robust way, and I suggest throw an error on invalid input instead of coercing the sort to a proper value

components/forum/ForumPage.tsx
```diff
@@ -25,10 +25,11 @@ export default function ForumPage() {
   const [search, setSearch] = useState('');
   const router = useRouter();
   const currentSpace = useCurrentSpace();
+  const categoryId = router.query.categoryId as string | undefined;
+  const sort = router.query.sort as string | undefined;
```
mattcasey, line 29: could use the type 'SortPosts' here

components/forum/components/PostList/PostList.tsx
valentinludu: I added the default newest just in listForumPosts()

valentinludu: If I throw an error that means the user will see that error message and an empty list.
I think it's a better user experience if we show latest posts(default) without the error.

The categoryId behaves the same. If you enter an invalid categoryId, it's like you didn't entered anything.

valentinludu: Screenshot 2023-01-03 at 13 07 39

mattcasey (approved):

components/forum/components/PostList/PostList.tsx
valentinludu: Decided to show the error!

PR #1393, opened 2022-12-24
Autoverify wallet

motechFR:
- Autosign when using modal in Connect
- Dont' show user an error when signature cancelled, inform them instead

todo on mobile:
- full width connect button
- fix auto sign showing on mobile


mattcasey (approved):
tested!

PR #1394, opened 2022-12-25
Changes to Polls (Embedded Polls)

Devorein:


components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
```diff
@@ -5,12 +5,9 @@ import type { NodeSelection } from 'prosemirror-state';
 
 import { hasComponentInSchema } from 'lib/prosemirror/hasComponentInSchema';
 
-import { queryIsSelectionAroundInlineVote } from '../inlineVote';
-import { markName as inlineVoteMarkName } from '../inlineVote/inlineVote.constants';
-
 import { floatingMenu } from './floating-menu';
 
-const blacklistedComponents = 'image cryptoPrice iframe page pdf mention tabIndent codeBlock inlineDatabase';
+const blacklistedComponents = 'image cryptoPrice iframe page pdf mention tabIndent codeBlock inlineDatabase poll';
```
motechFR, line 10: Can we add a comment about what this blacklist's purpose is?

components/common/CharmEditor/components/poll/PollComponent.tsx
```diff
@@ -0,0 +1,102 @@
...
+        <StyledEmptyPollContainer>
+          <FormatListBulleted fontSize='small' />
+          <Typography>Add a poll</Typography>
+        </StyledEmptyPollContainer>
+      </ListItem>
+    </BlockAligner>
+  );
+}
+
+export default function PollNodeView({ node, readOnly, updateAttrs, selected, deleteNode }: CharmNodeViewProps) {
```
motechFR, line 52: We agreed to have named exports unless a default is needed (such as for Next.JS page-level components)

Can you please update this?

components/common/CharmEditor/components/poll/config.ts
```diff
@@ -0,0 +1 @@
+export const name = 'poll';
```
motechFR, line 1: This only seems to be used in the poll spec and plugin file. I think it would make more sense to delete this, define name in spec and then import the name into plugin.

components/common/CharmEditor/components/poll/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './pollSpec';
+export { default as Component } from './PollComponent';
```
motechFR, line 2: Named function exports will clean this ;)

components/common/CharmEditor/components/suggestions/SuggestionCard.tsx
```diff
@@ -34,6 +34,7 @@ const ACTIONS: Record<string, string> = {
   insertion_codeBlock: 'Added code block',
   insertion_columnBlock: 'Added column layout',
   insertion_cryptoPrice: 'Added crypto price',
+  insertion_poll: 'Added poll',
```
motechFR, line 37: Random question, what is this whole component and mapping for?

hooks/usePageActionDisplay.tsx
```diff
@@ -47,21 +47,8 @@ export function PageActionDisplayProvider({ children }: { children: ReactNode })
       const cachedInlineCommentData: ThreadWithCommentsAndAuthors[] | undefined = cache.get(
         `pages/${currentPageId}/threads`
       );
```
motechFR, line 49: Always like to see a good code deletion!

motechFR (approved):
Tested and approved, I have some comments related to code organisation

components/common/CharmEditor/components/poll/config.ts
Devorein: This is actually the pattern we are following for `nft` and `iframe` blocks, but since its only exporting a single thing I will move it.

components/common/CharmEditor/components/suggestions/SuggestionCard.tsx
Devorein: Its to indicate what type of changes were made in the suggestion card.

PR #1397, opened 2022-12-26
Change view title when changing layout

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=e46da640-4cf1-4113-824d-7e5273fe4e95)
[Convo](https://discord.com/channels/894960387743698944/933026321683066991/1056894698616258601)

Since there are no easy ways to detect whether the view title changed or not (other than adding a column to our Block table to handle that) I'm just checking if the view title is the same as `{currentViewType} view` or not, if so then change it to the changed view type. Adding a new column for such a small ux improvement could be an overkill. Most people don't really change the view layout, they create new views. What notion does is that it doesn't name the view right away (keeps it empty) and only sync the title if its empty, otherwise respect the current title. Please let me know if we want to proceed in any of the other two directions.


msadura:
>  What notion does is that it doesn't name the view right away (keeps it empty) and only sync the title if its empty, otherwise respect the current title.

Actually the easies solution would be to keep title empty in db for "newly created" views, and fallback such title to our generic "Table / board view" in UI. 
At the same time I agree it is not really worth changing whole logic for such small UX update

components/common/BoardEditor/focalboard/src/components/viewSidebar/viewLayoutOptions.tsx
```diff
@@ -40,8 +42,15 @@ function LayoutOptions(props: LayoutOptionsProps) {
     defaultMessage: 'Gallery'
   });
 
+  function updateViewTitle(newViewType: IViewType) {
```
msadura, line 45: `updateViewTitle` - this method does not update anything, so I would rather call it just `getViewTitle`

msadura (approved):

mattcasey (changes requested):
I thought we would go with the other solution we discussed: have no title by default, but have the UI show a default one. Two reasons:
1. If you look at addViewMenu.tsx, the label is very different depending on the user's browser language. And we know we will want to at least support Japanese this coming year which has special characters.
2. It's a lot simpler and less error-prone since we aren't having to reverse-engineer the title.

PR #1399, opened 2022-12-27
fixed workspace onboarding modal appearing multiple times

Devorein:


prisma/schema.prisma
```diff
@@ -552,17 +553,18 @@ model UserTokenGate {
...
+  id               String             @id @default(uuid()) @db.Uuid
+  isAdmin          Boolean            @default(false)
+  createdAt        DateTime           @default(now())
+  space            Space              @relation(fields: [spaceId], references: [id], onDelete: Cascade)
+  user             User               @relation(fields: [userId], references: [id], onDelete: Cascade)
+  spaceRoleToRole  SpaceRoleToRole[]
+  tokenGate        TokenGate?         @relation(fields: [tokenGateId], references: [id])
+  tokenGateId      String?            @db.Uuid
+  joinedViaLink    Boolean?
+  workspaceOnboard WorkspaceOnboard[]
```
msadura, line 567: any reason for creating separate table and relation instead of simply storing `onboarded` boolean column directly in `SpaceRole` ? Feels like having separate table for single property is a bit overkill

charmClient/apis/workspaceOnboardingApi.ts
```diff
@@ -0,0 +1,11 @@
+import * as http from 'adapters/http';
+
+export class WorkspaceOnboardingApi {
+  getWorkspaceOnboarding({ spaceId }: { spaceId: string }) {
+    return http.GET<boolean>(`/api/spaces/${spaceId}/get-onboarding`);
```
msadura, line 5: ``/api/spaces/${spaceId}/onboarding``

hooks/useOnboarding.tsx
```diff
@@ -1,44 +1,53 @@
...
-
-  function showOnboarding(spaceId: string) {
-    setOnboarding((_onboarding) => ({
-      ..._onboarding,
-      [spaceId]: true
-    }));
-  }
+  const currentSpace = useCurrentSpace();
+  const { data: onboarding, mutate: updateOnboarding } = useSWR(
+    user && currentSpace ? `onboarding-${user.id}-${currentSpace.id}` : null,
```
msadura, line 24: I believe we namespace keys using `/` everythere else like `onboarding/${user.id}/${currentSpace.id}`

hooks/useOnboarding.tsx
```diff
@@ -1,44 +1,53 @@
...
-      [spaceId]: false
-    }));
+  async function completeOnboarding() {
+    if (currentSpace) {
+      charmClient.workspaceOnboarding.completeOnboarding({
+        spaceId: currentSpace.id
+      });
+      updateOnboarding(
+        (_onboarding) => {
+          if (_onboarding) {
```
msadura, line 38: I htink simple `(_onboarding) => _onboarding` will do the same

lib/token-gates/applyTokenGates.ts
```diff
@@ -197,6 +197,21 @@ export async function applyTokenGates({
       }
     });
 
+    await prisma.workspaceOnboard.create({
```
msadura, line 200: after seeing this logic copy-pasted all over the place I would suggest having this as spaceRole column instead of overcomplicating thing this much.

msadura (changes requested):

pages/api/spaces/[id]/get-onboarding.ts
```diff
@@ -0,0 +1,21 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
```
Comment (msadura) [inline pages/api/spaces/[id]/get-onboarding.ts:1]: We agreed some time ago to avoid using filenames-like-this. Additionally we never use `get` suffix for `GET` endpoints, so simply `onboarding.ts` will be enough

lib/token-gates/applyTokenGates.ts
motechFR: I agree with this, especially since the only relevant data in this model is onboarded: true

charmClient/apis/workspaceOnboardingApi.ts
motechFR: If we move workspaceOnboard to a spaceRole prop, we can skip the API lookup and look directly at the spaceroles of the LoggedInUser interface

msadura: oh loving that even more!

prisma/schema.prisma
```diff
@@ -1087,3 +1089,13 @@ model GoogleAccount {
   userId    String @db.Uuid
   user      User   @relation(fields: [userId], references: [id], onDelete: Cascade)
 }
+
+model WorkspaceOnboard {
```
mattcasey, line 1093: I think this might be able to have other data in the future, maybe just "UserSpaceInfo"

mattcasey: @Devorein and everyone: we should just use "space" in our names, not "workspace". product has actually determined to stop using 'workspace' as well

prisma/schema.prisma
mattcasey: the thought was that this is unrelated to permissioning, and we might have more fields int he future. permissions need to be fast - and possibly stored in the permissions api eventually - so it seemed better to keep it separate

mattcasey (changes requested):
rename 'workspaceonboarding' to 'userspaceinfo' or something a little more generic

hooks/useOnboarding.tsx
mattcasey: or even `Boolean`

prisma/schema.prisma
Devorein: After discussion we've decided to keep the property inside `SpaceRole` table

hooks/useOnboarding.tsx
Devorein: Since we are moving to use spaceRole we won't need this anymore

prisma/schema.prisma
Devorein: Sheesh I totally forgot. Thanks!!!

mattcasey (approved):

PR #1400, opened 2022-12-27
Full page view for forum post modal

Devorein:
image
Uses `fullWidth` prop for `Dialog` component. Right gap for showing category based information


components/forum/components/PostPage/PostPage.tsx
```diff
@@ -146,7 +146,12 @@ export function PostPage({ page, spaceId, onSave }: Props) {
   }, [postComments, page, commentSort]);
 
   return (
-    <Container top={50}>
+    <Container
+      top={50}
+      style={{
+        marginLeft: 75
```
mattcasey, line 152: we shouldn't need to add an additional margin here, it throws it off center:
image

components/common/BoardEditor/focalboard/src/components/dialog.tsx
```diff
@@ -38,7 +38,19 @@ const Dialog = React.memo((props: Props) => {
...
           }}
         >
-          <div role='dialog' className='dialog'>
+          <div
+            role='dialog'
+            className='dialog'
+            style={
+              fullWidth
+                ? {
+                    maxWidth: '85%',
```
mattcasey, line 47: - Width should still be 100% on mobile
- Max width should be 1280px, following reddit's layout. this prevents the white space from getting too large. 
- Please remove the border-radius
- don't mix CSS class names and inline styles -- since we're using classes, i think this style should be added as a new class

mattcasey (changes requested):
left some comments

mattcasey (approved):

PR #1402, opened 2022-12-27
Forum post standalone page

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=876aa6ac-5a5f-4262-b5f9-730052f1a534)

### DIFF
--- a/components/forum/components/PostDialog/PostDialog.tsx
+++ b/components/forum/components/PostDialog/PostDialog.tsx
@@ -84,7 +84,7 @@ export default function PostDialog(props: Props) {
                 <Button
                   size='small'
                   color='secondary'
-                  href={`/${router.query.domain}/forum/post/${page.id}`}
+                  href={`/${router.query.domain}/forum/post/${page.path}`}
                   variant='text'
                   startIcon={<OpenInFullIcon fontSize='small' />}
                 >
--- a/components/forum/components/PostDialog/PostDialog.tsx
+++ b/components/forum/components/PostDialog/PostDialog.tsx
@@ -1,11 +1,13 @@
 import OpenInFullIcon from '@mui/icons-material/OpenInFull';
+import { Box } from '@mui/material';
 import { usePopupState } from 'material-ui-popup-state/hooks';
+import { useRouter } from 'next/router';
 import { useEffect, useRef } from 'react';
 
 import charmClient from 'charmClient';
 import Dialog from 'components/common/BoardEditor/focalboard/src/components/dialog';
 import RootPortal from 'components/common/BoardEditor/focalboard/src/components/rootPortal';
-import { useUser } from 'hooks/useUser';
+import Button from 'components/common/Button';
 import type { ForumPostPage } from 'lib/forums/posts/interfaces';
 
 import { PostPage } from '../PostPage/PostPage';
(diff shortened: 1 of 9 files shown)


components/forum/components/PostDialog/PostDialog.tsx
```diff
@@ -75,19 +78,21 @@ export default function PostDialog(props: Props) {
...
-          //     </Button>
-          //   </Box>
-          // }
+          toolbar={
+            page && (
+              <Box display='flex' justifyContent='space-between'>
+                <Button
+                  size='small'
+                  color='secondary'
+                  href={`/${router.query.domain}/forum/post/${page.id}`}
```
mattcasey, line 87: should be page.path

components/forum/components/PostPage/PostPage.tsx
```diff
@@ -148,21 +149,23 @@ export function PostPage({ page, spaceId, onSave }: Props) {
...
-          isContentControlled={true}
-          content={form.content as PageContent}
-          onContentChange={updatePostContent}
-        >
-          <PageTitleInput readOnly={readOnly} value={form.title} onChange={updateTitle} />
-          <Box my={2}>
-            <PostCategoryInput spaceId={spaceId} setCategoryId={updateCategoryId} categoryId={categoryId} />
-          </Box>
-        </CharmEditor>
+        {page && user && (
```
mattcasey, line 152: why did you have to add page && user here? we are not using 'user' in the props, and pageId is optional

pages/[domain]/forum/post/[postId].tsx
```diff
@@ -0,0 +1,22 @@
...
+import useSWR from 'swr';
+
+import charmClient from 'charmClient';
+import ErrorPage from 'components/common/errors/ErrorPage';
+import getPageLayout from 'components/common/PageLayout/getLayout';
+import { PostPage } from 'components/forum/components/PostPage/PostPage';
+
+export default function BlocksEditorPage() {
+  const router = useRouter();
+  const pagePath = router.query.postId as string;
```
Comment (mattcasey) [inline pages/[domain]/forum/post/[postId].tsx:11]: rename to router.query.path or router.query.pagePath

pages/[domain]/forum/post/[postId].tsx
```diff
@@ -0,0 +1,22 @@
+import { useRouter } from 'next/router';
```
Comment (mattcasey) [inline pages/[domain]/forum/post/[postId].tsx:1]: Please use page path instead for navigation and rename this file to [pagePath].tsx

mattcasey (changes requested):
- We should set the page title using `hooks/usePageTitle`.
- Breadcrumbs should look like: `Forum > [page title]` (with Forum linking back to main Forum feed). It would be nice to include the category: `Forum > [category title] > [page title]`. I think this is where having category in the path would be nice, because you could just consume useForumCategories() and match on id/path.

components/forum/components/PostPage/PostPage.tsx
Devorein: The readOnly property is false initially as `user` is `null`. Charmeditor doesn't pickup the updated readOnly property and so the editor becomes readOnly as well. I added a key of `user?.id` and it seems to be working now 🥳

PR #1408, opened 2022-12-30
Import markdown

motechFR:
Allows user to simply upload a folder of markdown files


pages/api/pages/import/[spaceId]/zipped.ts
```diff
@@ -0,0 +1,96 @@
...
+            }
+          },
+          select: pageMetaSelect()
+        });
+
+        createdPages.push(..._createdPages);
+      }
+      res.status(201).send(createdPages);
+    } catch (err) {
+      if (!res.headersSent) {
```
Comment (mattcasey) [inline pages/api/pages/import/[spaceId]/zipped.ts:89]: should log the error

pages/api/pages/import/[spaceId]/zipped.ts
```diff
@@ -0,0 +1,96 @@
...
+              path: getPagePath(),
+              type: 'page',
+              updatedBy: userId,
+              createdBy: userId,
+              spaceId: spaceId as string
+            };
+
+            pagesToCreate.push(pageToCreate);
+          } catch (err) {
+            // Do nothing
```
Comment (mattcasey) [inline pages/api/pages/import/[spaceId]/zipped.ts:65]: should log a warning. Also, why wouldn't we want to show an error to the user? They would be expecting a file to appear

mattcasey (changes requested):
I tried uploading the zip from the user: https://app.charmverse.io/charmverse/page-21850517986912954, but just got an error "Unable to import pages".

Also, this doesn't seem to support folders, which are used pretty heavily in the user's sample zip (the folder path is part of the filename).

pages/api/pages/import/[spaceId]/zipped.ts
```diff
@@ -0,0 +1,96 @@
...
+  .post(importZippedController);
+async function importZippedController(req: NextApiRequest, res: NextApiResponse) {
+  const { spaceId } = req.query;
+  const { id: userId } = req.session.user;
+
+  req.on('data', async (chunk) => {
+    try {
+      const content = await unzip.loadAsync(chunk);
+
+      const fileNames = typedKeys(content.files);
```
Comment (mattcasey) [inline pages/api/pages/import/[spaceId]/zipped.ts:39]: I don't see the benefit of typedKeys here over a simple Object.keys. Isn't that for when there is a type union? But `content.files` is just a map of strings. For some reason typedKeys thinks it is a map of `number | string`, but I'm pretty sure that's impossible.

PR #1409, opened 2022-12-30
Accelerate signup, defer ENS

motechFR:
Also contains a script for migrating current wallets and refreshing their ENS Names

### DIFF
--- a/pages/api/profile/index.ts
+++ b/pages/api/profile/index.ts
@@ -3,6 +3,7 @@ import type { NextApiRequest, NextApiResponse } from 'next';
 import nc from 'next-connect';
 import { v4 } from 'uuid';
 
+import { refreshENSName } from 'lib/blockchain/refreshENSName';
 import { updateGuildRolesForUser } from 'lib/guild-xyz/server/updateGuildRolesForUser';
 import { updateTrackUserProfile } from 'lib/metrics/mixpanel/updateTrackUserProfile';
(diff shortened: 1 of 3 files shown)


pages/api/profile/index.ts
```diff
@@ -33,7 +34,9 @@ async function createUser(req: NextApiRequest, res: NextApiResponse<LoggedInUser
 
     const signupAnalytics = extractSignupAnalytics(cookiesToParse);
 
-    user = await createUserFromWallet({ address, id: req.session.anonymousUserId }, signupAnalytics);
+    user = await createUserFromWallet({ address, id: req.session.anonymousUserId, skipENS: true }, signupAnalytics);
+    // Do this separately from signup response to speed things up
+    refreshENSName({ address, userId: user.id });
```
mattcasey, line 39: This still needs to handle errors, but why wouldn't we want to be wait for ENS to be set up before we return a response from the server? I don't think sign-up is really so slow we need this kind of optimization, and the new 'skipENS' option seems a little hacky.

mattcasey (changes requested):
Ran the script successfully!

Before merging this, errors from refreshENSName still need to be handled or they will crash the process.
I don't see why we need this optimization though - ENS seems pretty fast, and it makes the code more complicated. Can you provide some context?

motechFR:
Closing PR after our discussion, since lag is probably just an issue on my end

PR #1411, opened 2022-12-30
Improvements to markdown import

mattcasey:
I made some updates to make it work for the user zip that was shared with us: https://app.charmverse.io/charmverse/page-21850517986912954
- handles upload of larger zip files. `req.on('end')` was not working but I found a solution here: https://stackoverflow.com/questions/70084915/proxy-pass-multipart-in-nextjsalso-nodejs
- exclude the mirror files that MACOSX creates (these are present in the user's sample zip and were crashing the import, since the content included some weird characters. I found this SO post about the error: https://stackoverflow.com/questions/1347646/postgres-error-on-insert-error-invalid-byte-sequence-for-encoding-utf8-0x0). I wonder if there's a more robust approach than excluding files that start with "." :)


motechFR (approved):

motechFR:
Worked great, thanks!

PR #1416, opened 2023-01-03
Mixpanel Events for Forums

Devorein:


components/forum/ForumPage.tsx
```diff
@@ -70,6 +71,14 @@ export default function ForumPage() {
     }
   }, [router.query.pageId]);
 
+  useEffect(() => {
+    if (currentSpace) {
+      charmClient.track.trackAction('main_feed_page_load', {
+        spaceId: currentSpace.id
+      });
+    }
+  }, [currentSpace]);
```
mattcasey, line 80: track only the existence of currentSpace, so it doesnt track changes to the space itself

lib/forums/comments/createPostComment.ts
```diff
@@ -12,7 +15,18 @@ export async function createPostComment({
   postId: string;
   userId: string;
 }) {
-  return prisma.pageComment.create({
+  const page = await getForumPost({ pageId: postId, userId });
```
mattcasey, line 18: this is an expensive query, we should only use it when we need things like comment votes... can you just get the page/post?

lib/forums/comments/deletePostComment.ts
```diff
@@ -1,21 +1,60 @@
...
       createdBy: userId,
       deletedAt: null
+    }
+  });
+
+  if (!pageComment) {
+    throw new UnauthorisedActionError();
+  }
+
+  const page = await getForumPost({ pageId: postId, userId });
```
mattcasey, line 28: please check all uses of 'getForumPost' it looks like here we are only needing to look up a category Id. no need for any of the logic or overhead inside this method

lib/forums/comments/deletePostComment.ts
```diff
@@ -1,21 +1,60 @@
...
+    where: {
+      id: page.post.categoryId
+    },
+    select: {
+      name: true
+    }
+  });
+
+  if (category) {
+    trackUserAction('delete_comment', {
```
mattcasey, line 40: Always track user actions _after_ the event has happened, if possible. Looks like this can be moved down.

lib/forums/posts/voteForumPost.ts
```diff
@@ -29,6 +30,32 @@ export async function voteForumPost({
...
+      where: {
+        id: page.post.categoryId
+      },
+      select: {
+        name: true
+      }
+    });
+
+    if (category) {
+      if (upvoted) {
```
mattcasey, line 43: it would be cleaner to just define the action, so you do'nt need to call trackUserAction twice with all the same inputs:
```
const userAction = upvoted ? 'upvote_post' : 'downvote_post';
trackUserAction(userAction, {... });
```

lib/metrics/mixpanel/interfaces/ForumEvent.ts
```diff
@@ -0,0 +1,57 @@
+import type { BaseEvent } from './BaseEvent';
+import type { ResourceEvent } from './ResourceEvent';
+
+type PostEvent = BaseEvent & ResourceEvent;
+
+export interface PostPageLoadEvent extends PostEvent {}
+
+export interface MainFeedLoadEvent {
```
mattcasey, line 8: I think it'd be better to write this as `MainFeedLoadEvent extends BaseEvent`

lib/metrics/mixpanel/interfaces/ForumEvent.ts
```diff
@@ -0,0 +1,57 @@
+import type { BaseEvent } from './BaseEvent';
+import type { ResourceEvent } from './ResourceEvent';
+
+type PostEvent = BaseEvent & ResourceEvent;
+
+export interface PostPageLoadEvent extends PostEvent {}
```
mattcasey, line 6: use type, not interface

lib/metrics/mixpanel/interfaces/ForumEvent.ts
```diff
@@ -0,0 +1,57 @@
...
+export interface PostVoteEvent extends PostEvent {
+  categoryName: string;
+}
+
+export interface CommentVoteEvent extends PostEvent {
+  categoryName: string;
+  postId: string;
+}
+
+export interface CommentCreateEvent extends PostEvent {
```
mattcasey, line 38: these interfaces should not be exported unless someone is importing them. By exporting it, you make it hard to know if they can be deleted in the future since the linter won't know if they're used or not

mattcasey (changes requested):
- replace interface with type
- remove exports of types that are not imported somewhere
- track actions _after_ the event is complete
- remove use of getForumPost() and replace with a simple query for the category name

PR #1418, opened 2023-01-03
Rename table column header name

Devorein:


mattcasey (changes requested):
- Title should update when I close the popup, just like how Notion does. I did not think to use "Enter" key at first
- Hitting "Enter" in the title input triggers an error:
image

mattcasey (approved):

PR #1420, opened 2023-01-04
Drop page post type

motechFR:
This PR removes dependency of forum posts on the pages collection


prisma/migrations/20230103114420_use_posts_for_forums/migration.sql
```diff
@@ -0,0 +1,127 @@
+/*
```
mattcasey, line 1: can you combine all these migrations into one file? this could get messy if one fails somehow
To do that, just delete the migration files and re-run `prisma migrate dev` locally

prisma/schema.prisma
```diff
@@ -449,7 +443,7 @@ model User {
   memberPropertyValues MemberPropertyValue[]
   userTokenGates       UserTokenGate[]
   unstoppableDomains   UnstoppableDomain[]
-  PageComment          PageComment[]
+  PostComment          PostComment[]
```
mattcasey, line 446: rename to 'postComments'

components/forum/components/PostDialog/PostDialog.tsx
```diff
@@ -34,21 +33,21 @@ export default function PostDialog(props: Props) {
 
   // open modal when page is set
   useEffect(() => {
-    if (page) {
+    if (post) {
       popupState.open();
     }
-  }, [!!page]);
+  }, [post]);
```
mattcasey, line 39: i believe we should only be checking the existing of post.. otherwise this gets triggered any time anything on the post object changes?

mattcasey (changes requested):
I can't close the popup when I create a post and change my mind. I have to fill out the form and submit. Not sure what's going on but check your changes to the PostDialog and PostPage components?

- Also there's still 17 TS errors when I check, fixing them will address some other issues - like deleting archived posts

mattcasey (approved):

PR #1432, opened 2023-01-06
Fixed notion import for embed blocks

Devorein:


motechFR (approved):
Plz delete the old comment as it no longer applies

PR #1433, opened 2023-01-06
Fix CSV export encoding

valentinludu:
If you have a # symbol in the title of a card, the csv file will show everything until that point but not continue to export after that character.

That means:
If I have 3 cards in a database and the 2nd one has the title "My name is #elon", the result of the exported csv will be:

Name,Priority,Category,Assignee,System,Status
"Conduct first community call","Low",Development,"a65810e1-d20e-4c93-aa6c-50c479d9c1cf","","Completed"
"My name is

As you can see, the csv file looks broken and you can't import it anywhere.

Task:
https://app.charmverse.io/charmverse/page-16845167973892972

### DIFF
--- a/components/common/BoardEditor/focalboard/src/csvExporter.ts
+++ b/components/common/BoardEditor/focalboard/src/csvExporter.ts
@@ -22,15 +22,14 @@ class CsvExporter {
     let csvContent = 'data:text/csv;charset=utf-8,';
 
     rows.forEach((row) => {
-      const encodedRow = row.join(',');
+      const encodedRow = row.map((str) => encodeURIComponent(str)).join(',');
       csvContent += `${encodedRow}\r\n`;
     });
 
     const filename = `${Utils.sanitizeFilename(viewToExport.title || 'Untitled')}.csv`;
-    const encodedUri = encodeURI(csvContent);
     const link = document.createElement('a');
     link.style.display = 'none';
-    link.setAttribute('href', encodedUri);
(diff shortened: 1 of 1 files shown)


mattcasey (approved):

mattcasey (approved):
approved but can you add a test?

mattcasey (changes requested):
I tried this update on anorther branch and the values have quotes on either side encoded:
image

valentinludu:
> I tried this update on anorther branch and the values have quotes on either side encoded: image
> 
> Maybe the values could be encoded inside of the `encodeText()` method?

What was the content of the card? (title & properties)

mattcasey (approved):

tsconfig.json
```diff
@@ -5,20 +5,14 @@
...
     "importHelpers": true,
     "baseUrl": ".",
     "strict": true,
     "target": "es2019",
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
+    "lib": ["dom", "dom.iterable", "esnext"],
```
mattcasey, line 15: my VS Code formats these the other way (with values on separate rows).... was this file cleanup intentional?

valentinludu: My VS code is formatting it like that.
Because of `"printWidth": 120` rule.
Those lines were shorter so it was adding them on a single line.

I reverted for now the change to see only my actual changes in this PR.

mattcasey (approved):

PR #1437, opened 2023-01-08
Integrate iframely into embeds and create /bookmark

Devorein:


components/common/CharmEditor/components/iframe/config.ts
```diff
@@ -43,6 +43,22 @@ export const embeds = {
       return url.includes('airtable.com');
     }
   },
+  airtable: {
```
mattcasey, line 46: this is duplicated from above?

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
+import BookmarkIcon from '@mui/icons-material/Bookmark';
```
mattcasey, line 1: capitalize the filename plz

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { Box } from '@mui/material';
+
+import fetch from 'adapters/http/fetch.server';
```
mattcasey, line 4: should not be importing server-side fetch

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+  node,
+  updateAttrs,
+  selected,
+  deleteNode
+}: CharmNodeViewProps & { readOnly?: boolean }) {
+  const { url, html } = node.attrs as BookmarkViewerProps;
+
+  async function updateNode(bookmarkUrl: string) {
+    try {
+      const iframelyResponse = await fetch<{ html: string; error?: string }>(
```
mattcasey, line 27: this should probably just be added to charmClient

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+      //
+    }
+  }
+
+  if (!url || !html) {
+    if (readOnly) {
+      return <div />;
+    }
+    return (
+      <MediaSelectionPopup
```
mattcasey, line 48: thanks for using this component this time :)

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+          process.env.NEXT_PUBLIC_IFRAMELY_API_KEY
+        }&iframe=1&omit_script=1&media=0`
+      );
+      if (iframelyResponse.html) {
+        updateAttrs({
+          url: bookmarkUrl,
+          html: iframelyResponse.html
+        });
+      }
+    } catch (err) {
```
mattcasey, line 38: if we arent doing anything w/catch, then don't use try/catch. that way we can see the errors

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+    return (
+      <MediaSelectionPopup
+        icon={<BookmarkIcon fontSize='small' />}
+        isSelected={selected}
+        buttonText='Add a bookmark'
+        node={node}
+        onDelete={deleteNode}
+      >
+        <Box py={3}>
+          <MediaUrlInput onSubmit={updateNode} placeholder='https://dune.com/skateordao/Gnars' />
```
mattcasey, line 56: please use a placeholder that is fake

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
mattcasey: it also seems like you should just watch the bookmarkUrl attribute and call the API?

components/common/CharmEditor/components/bookmark/bookmarkSpec.ts
```diff
@@ -0,0 +1,31 @@
...
+        },
+        track: {
+          default: []
+        }
+      },
+      group: 'block',
+      draggable: false,
+      parseDOM: [{ tag: 'div.charm-bookmark' }],
+      toDOM: (): DOMOutputSpec => {
+        return ['div.charm-bookmark'];
```
mattcasey, line 24: make sure to include all node attributes when creating the DOM. it's required so that copy/paste works properly

components/common/CharmEditor/components/bookmark/bookmarkSpec.ts
```diff
@@ -0,0 +1,31 @@
...
+      }
+    },
+    markdown: {
+      toMarkdown: () => null
```
mattcasey, line 28: we should at least include the url that was embedded

components/common/CharmEditor/components/bookmark/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './bookmarkPlugins';
```
mattcasey, line 1: let's not use index.ts files for components anymore. That was something our team agreed on

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+export function Bookmark({
+  readOnly = false,
+  node,
+  updateAttrs,
+  selected,
+  deleteNode
+}: CharmNodeViewProps & { readOnly?: boolean }) {
+  const { url, html } = node.attrs as BookmarkViewerProps;
```
mattcasey, line 23: component needs to look for the 'tooltip-marker' mark so that the input opens automatically

mattcasey (changes requested):
left a bunch of comments... please delpoy to staging to test the env var is set correctly. and whose api key are we using? do i have access?

components/common/CharmEditor/components/iframe/config.ts
Devorein: I don't even know how it got here 🤦🏼‍♂️

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
Devorein: Thank you. I had to omit the credentials as otherwise iframely threw CORS error

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
Devorein: Lets do that 🚀

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
Devorein: Since I've moved everything to charmClient its not going to throw an error anymore

components/common/CharmEditor/components/bookmark/bookmarkSpec.ts
Devorein: right I added a todo for it but forgot to complete it. Thanks.

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
Devorein: Actually `MediaSelectionPopup` automatically does that for you

components/common/CharmEditor/components/bookmark/bookmarkNodeView.tsx
mattcasey: I forgot about the CORS thing, but thanks for making that work

mattcasey (approved):

PR #1438, opened 2023-01-08
Fix display of forum votes

mattcasey:
new icons too:
image


motechFR (approved):
Finally fixes that pesky bug with the weird vote counts, thanks!

msadura:
Just small thing - this "red" filled arrow looks a bit too aggressive imo - red is often associated with error, even more with filled bg :D I would make it more yellowish / orage, or try just outlined one

PR #1439, opened 2023-01-08
enable forums for some DAOs

mattcasey:


hooks/useIsCharmverseSpace.tsx
```diff
@@ -2,13 +2,15 @@ import { isProdEnv } from 'config/constants';
...
 
+const defaultDomains = ['charmverse'];
+
 /**
  * Returns true if the current space is charmverse, OR the current environment is not prod
  *
  * Use this hook for reserving some features in prod for charmverse users only
  */
-export function useIsCharmverseSpace() {
+export function useIsCharmverseSpace(allowedDomains = defaultDomains) {
```
motechFR, line 12: Small suggestion; updating this so we don't need to explicitly pass CharmVerse in the array, and allowedDomains just gets added to default domains

motechFR (approved):

hooks/useIsCharmverseSpace.tsx
mattcasey: I think it's better to be explicit. TBH I am not sure we even need this level of abstraction (a hook for 'isCharmVerse')

PR #1442, opened 2023-01-09
Feat/sqs process

msadura:
Final version for webhooks processing:
- long polling for aws sqs
- veryfing permissions (by checking api key) for space related to webhook message


lib/aws/webhookSqs.ts
```diff
@@ -54,26 +54,23 @@ export async function deleteMessage(receipt: string) {
...
-    // delete if processed correctly
-    await deleteMessage(message.ReceiptHandle || '');
-  } catch (e) {
-    log.error('Failed to process webhook message', e);
+  if (message) {
+    let msgBody: Record<string, any> | string = '';
+    try {
+      msgBody = JSON.parse(message.Body || '');
+      // eslint-disable-next-line no-empty
+    } catch (e) {}
```
mattcasey, line 62: looks like we could still log the error, maybe as a warning?

lib/discord/getSpaceFromDiscord.ts
```diff
@@ -0,0 +1,12 @@
+import { prisma } from 'db';
+import { InvalidInputError } from 'lib/utilities/errors';
+
+export async function getSpaceFromDiscord(discordServerId: string) {
+  const space = await prisma.space.findFirst({ where: { discordServerId } });
```
mattcasey, line 5: You can use `prisma.space.findFirstOrThrow()` if you want to assert there is a result. I think all the methods have an `OrThrow()` option

lib/webhooks/processWebhookMessage.ts
```diff
@@ -55,6 +58,27 @@ export async function processWebhookMessage(message: WebhookMessage) {
   }
 
   const handler = messageHandlers[data?.type];
+  const hasPermission = await verifyWebhookMessagePermission(message);
+  if (!hasPermission) {
+    return true;
```
mattcasey, line 63: This case means we got an unverified request right? I think it would be a good idea to add a log here, warning or debug level, with the message body?

mattcasey (approved):
approved... is this actually running yet? I can't see where we call it in the cron.ts file

PR #1445, opened 2023-01-09
Dont use sockets in readonly mode

motechFR:


mattcasey:
We need to use sockets even if it's readonly mode, because comments and suggestions are available when 'readonly' is true

mattcasey (approved):

PR #1446, opened 2023-01-09
Feat/UI indexes

msadura:


mattcasey (approved):
having hard-coded z-indexes bothers me but if it works, it works! if another issue pops up, I'd like to either just remove these or fix the way our HTML is laid out (it seems weird that the zindex on an icon is even able to cause emojis to appear above a menu!)

PR #1447, opened 2023-01-09
Support for notion twitter/nft/video import 

Devorein:
We used the generic embed block before but since these embeds have their own blocks its better to make use of those plus twitter import doesn't work without the tweet block.


lib/twitter/extractTweetAttrs.ts
```diff
@@ -0,0 +1,17 @@
+import type { TweetNodeAttrs } from './interface';
```
mattcasey, line 1: please leave this in the tweet component..  it is specific to Tweet prosemirror components

mattcasey (changes requested):
I would like to keep prosemirror component code inside of CharmEditor components folder unless there's a really good reason, with lib/prosemirror being an exception - even that folder seems like a weird outlier. 
I do think there's room for improvement, but the logic and specs would still belong together.

Devorein:
@mattcasey I'm closing this PR in favor of #1451 since the only code change that was required was in the `NotionBlock` module

PR #1448, opened 2023-01-09
show embeds when iframely cannot generate html

mattcasey:
This example does'nt return HTML and is not supported when you try it on iframely:
https://www.meti.go.jp/english/press/2022/0715_002.html

Iframely link: https://iframely.com/try?url=https://www.meti.go.jp/english/press/2022/0715_002.html

Good news is we still get metadata


mattcasey:
merging for Xandra

mattcasey:
merging for Xandra

PR #1450, opened 2023-01-10
Fix any upload of a file that contains non ascii characters

valentinludu:
This should fix any upload to aws in our app.

https://app.charmverse.io/charmverse/page-706022384512363


pages/api/aws/s3-upload.ts
```diff
@@ -31,7 +32,14 @@ const makeRouteHandler = (options: Options = {}): Handler => {
...
       const bucket = process.env.S3_UPLOAD_BUCKET;
 
-      const filename = decodeURIComponent(req.query.filename as string);
+      let filename = decodeURIComponent(req.query.filename as string);
+      const validCharacters = /^[\000-\177]*$/;
+
+      if (!validCharacters.test(filename)) {
+        const extension = filename.split('.').at(-1);
+        const randomName = uniqueNamesGenerator({ dictionaries: [colors] });
+        filename = `${randomName}.${extension}`;
```
msadura, line 41: Honestly I would not add additional external library for the purpose of generating unique "user friendly" name. I would make it simple and go with `filename = `${uuid()}.${extension}`;`

mattcasey: We are using unique-names-generator already to generate anonymous user names, but I still agree we should just use a random uuid. Or if we're going to add this as some kind of easter egg, we should use the crypto dictionary

valentinludu: I agree.

I changed the name to be a uuid to be simpler.

motechFR (approved):
I just tested this by uploading a japanese png to the CharmEditor and worked just fine

Below a screenshot of network tab showing uuid filename

Screenshot 2023-01-11 at 15 55 19

PR #1452, opened 2023-01-10
Fix add page/database not working in sidebar

Devorein:


components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
```diff
@@ -249,16 +249,19 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
...
+      if (space && user) {
+        const newPage: NewPageInput = {
+          ...page,
+          createdBy: user.id,
+          spaceId: space.id
+        };
+        return addPageAndRedirect(newPage, router);
+      }
+    },
+    [space, user]
```
Devorein, line 263: Not sure whether we need to add `router` as a deps or not

mattcasey: This looks fine, but instead of adding the whole object, it's best to only watch on the least amount of info (changing my username or avatar shouldn't rerender the sidebar)... in this case, you could just add space?.id and user?.id
> Not sure whether we need to add router as a deps or not
- If we were using 'rule of hooks', i guess you'd need to add addPageAndRedirect as well as router? It would also be fine to just remove useCallback. this was more important to optimize when changing page content would trigger rerenders in the sidebar

mattcasey (approved):

mattcasey (approved):

PR #1453, opened 2023-01-10
Show add group name btn if no select property is present in the board

valentinludu:
I am creating an empty select property for the board only if the user clicks "Add a group" and he does not have any select present.

Workflow:
- user clicks Add new group and a new select option is created for the board
- user types a name and clicks ok. The new option is added to the select property

https://app.charmverse.io/charmverse/page-6573898126800706


mattcasey (approved):
works well!

PR #1454, opened 2023-01-10
Alert deleted account user

motechFR:
Multiple improvements:
- Provide a visual modal which tells users with a deleted account to contact the support team
- Implement server-side checks so that login methods are rejected if user tries to login with deleted account
- Cleanup: Ensure all login paths actually extract the signup analytics into their own method


mattcasey (approved):
nice

PR #1455, opened 2023-01-10
Highlighting text with keyboard doesn't show floating menu

Devorein:


components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
```diff
@@ -1,27 +1,21 @@
...
-import type { Node, Plugin, ResolvedPos } from '@bangle.dev/pm';
+import { PluginKey } from '@bangle.dev/core';
+import type { Node, ResolvedPos } from '@bangle.dev/pm';
 import type { NodeSelection } from 'prosemirror-state';
 
 import { hasComponentInSchema } from 'lib/prosemirror/hasComponentInSchema';
 
 import { floatingMenu } from './floating-menu';
 
+export const floatingMenuPluginKey = new PluginKey('floatingMenu');
```
mattcasey, line 10: i thought this causes an issue (when we don't generate a unique plugin key)

mattcasey:
I made another PR which doesn't show the menu after dropping: https://github.com/charmverse/app.charmverse.io/pull/1462/files

PR #1462, opened 2023-01-11
Highlighting text with keyboard doesn't show floating menu v2

mattcasey:
This brings back the call to the selection tooltip update, but checks the state from row-action before calling it...

Some inspiration came from https://github.com/ProseMirror/prosemirror/issues/587

### DIFF
--- a/components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
+++ b/components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
@@ -1,15 +1,6 @@
 import { link } from '@bangle.dev/base-components';
 import type { PluginKey } from '@bangle.dev/core';
-import type {
-  EditorState,
-  EditorView,
-  Transaction,
-  Node,
-  ResolvedPos,
-  DirectEditorProps,
-  Plugin
-} from '@bangle.dev/pm';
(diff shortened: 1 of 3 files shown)


components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
```diff
@@ -73,17 +73,33 @@ export function plugins({
...
     // @ts-ignore
-    const view = selectionTooltipController.spec.view();
-    selectionTooltipController.spec.view = () => {
-      view.update = () => {};
-      return view;
-    };
+    const viewUpdate = controller.spec.view().update;
+
+    Object.assign(controller.spec, {
+      state: {
```
mattcasey, line 84: I wanted to instead add this to the rowAction state, but for some reason the value in `lastState` in `controller.getState(lastState)` would be false when called below. The drop event seems to trigger two updates, the 2nd of which triggers the popup to appear.

Devorein (approved):

Devorein:
It does what is expected, but after dropping the whole block gets selected and clicking on it shows the floating menu for a split second. On the other hand notion does select the whole block, but clicking on it doesn't show the menu. I tried a few solutions but none of them are working, but its better than what we have right now 🚀

PR #1464, opened 2023-01-11
Remove git env file

motechFR:
Remove .env from git

Improve Developer Experience for the .env.example file


.env.example
```diff
@@ -1,38 +1,136 @@
...
+
+# ----------------------------------------------------------------------
+# ----------------------------------------------------------------------
+# ---------- Essential secrets for the application to work -------------
+# ----------------------------------------------------------------------
+# ----------------------------------------------------------------------
+
 
 # Link to your postgres instance
 DATABASE_URL=postgres://user:password@postgreshost.com:5432/databasename
```
mattcasey, line 12: can we actually make this file so that you can just rename it? so for this url, use the one that's in .env and for the others that aren't being used maybe we can comment them out

.env.test.local
```diff
@@ -1,4 +0,0 @@
-DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/charmversetest
```
mattcasey, line 1: we should leave this one, i think

mattcasey: agreed we'll leave this out for now for people that don't use locally running postgres

.env.example
```diff
@@ -1,38 +1,144 @@
...
+# ----------------------------------------------------------------------
 
-# Link to your postgres instance
-DATABASE_URL=postgres://user:password@postgreshost.com:5432/databasename
 
-# Your CoinMarketCap API token for retrieving cryptocurrency prices
-# You can get an API token at pro.coinmarketcap.com
-CMC_API_TOKEN=abcde-122333-wwdcfww
+# Link to your postgres instance
+DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/charmversetest
```
mattcasey, line 12: rename to 'charmverse' ?

.env.example
```diff
@@ -1,38 +1,144 @@
...
 
 # Used for signing the session token
-AUTH_SECRET=wieVRbuyweG23234sjkEDwFuiFEfebY
+AUTH_SECRET=wieVRbuyweG23234sjkEDwFuiFEfebYZ
+
+# CI-specific environment variables -----
+GOOGLE_OAUTH_ENCRYPTION_SECRET=abceedsssd
```
mattcasey, line 18: i think its better to not set this, than put some garbage in

.env.example
```diff
@@ -1,38 +1,144 @@
...
+# Create a separate .env.test.local file with these environment variables included.
+# Also exclude these from .env, so that your test environment data does not affect your local dev data
+DOMAIN=http://127.0.0.1:3335
```
mattcasey, line 21: this file should work "out of the box" for dev. otherwise its too much work. if we need a special test env example, thats fine too

mattcasey (changes requested):
this file should work "out of the box" for dev with a simple `cp .env.example .env` otherwise it's too much work. So no variables should have dummy data or need to be adjusted (they can just be left in commented out). If we need a second example for test env, thats fine too.

PR #1468, opened 2023-01-11
fix deleting proposals that used to be db cards

mattcasey:
When converting cards to proposals, we kept the 'parentId' on the page doc, which broke the page tree resolution because we filter out proposals... 

1. I now unset parentId when converting proposals. This probably was a bug with page permisisons too
2. I tried to simplify the options for page tree resolving because the error was thrown when proposals were not included in the tree. I think it would be much easier to follow if we took out the filter logic in the lowest-level functions. Ultimately, I think the only time we need to filter pages by type is in the sidebar? In all other cases it seems like we'd always want to consider any page in the tree... 

Discord: https://discord.com/channels/894960387743698944/1060677382979194911/1060677382979194911

### DIFF
--- a/lib/pages/__tests__/mapPageTree.spec.ts
+++ b/lib/pages/__tests__/mapPageTree.spec.ts
@@ -296,7 +296,7 @@ describe('reducePagesToPageTree', () => {
 
     const { rootNodes } = reducePagesToPageTree({
       items: boardAndCards,
-      includeCards: true
+      includeAllPageTypes: true
     });
 
     expect(rootNodes.length).toBe(1);
@@ -338,7 +338,7 @@ describe('reducePagesToPageTree', () => {
 
     const { rootNodes } = reducePagesToPageTree({
       items: boardAndCards,
-      includeCards: true
+      includeAllPageTypes: true
     });
 
     expect(rootNodes.length).toBe(1);
@@ -388,7 +388,7 @@ describe('mapPageTree', () => {
     const rootList = mapPageTree({
       items: boardAndCards,
       // Added this prop manually, Typescript will prevent us from calling it
-      includeCards: true
+      includeAllPageTypes: true
     } as any);
 
     expect(rootList.length).toBe(1);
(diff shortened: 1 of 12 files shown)


pages/api/blocks/[id]/index.ts
```diff
@@ -51,13 +51,8 @@ async function deleteBlock(
       throw new ActionNotPermittedError();
     }
 
-    const pageTree = await resolvePageTree({ pageId: rootBlock.id, flattenChildren: true, includeDeletedPages: true });
```
Comment (mattcasey) [inline pages/api/blocks/[id]/index.ts:54]: these options are the same that we use inside modifyChildPages

lib/pages/mapPageTree.ts
```diff
@@ -118,9 +113,14 @@ export function mapPageTree<T extends PageNode = PageNode>({
 export function mapTargetPageTree<T extends PageNode = PageNode>({
   items,
   targetPageId,
-  includeDeletedPages
+  includeDeletedPages,
+  includeAllPageTypes = true
 }: Omit<PageTreeMappingInput<T>, 'rootPageIds'> & { targetPageId: string }): TargetPageTree<T> {
-  const { itemMap, itemsWithChildren } = reducePagesToPageTree({ items, includeCards: true, includeDeletedPages });
```
mattcasey, line 123: This method (mapTargetPageTree) is only called by resolvePageTree. I am not sure why we  include cards but not proposals... seems like a bug?

mattcasey:
Update: after some iterations it appears we don't even need the filter logic. I removed the options except for deleted pages

lib/pages/__tests__/mapPageTree.spec.ts
```diff
@@ -217,22 +217,6 @@ describe('reducePagesToPageTree', () => {
     expect(rootNodes[0].children.length).toBe(0);
   });
 
-  it('should filter out proposal type pages by default', async () => {
```
motechFR, line 220: I did some further testing on the client, this test can indeed be ignored.

Since we are passing an explicit list of rootIds to mapping function, which includes boards, pages and linkedBoards only, we are fine

motechFR:
Another simplification; mapPageTree is only ever used in the Navigation case, so we should probably drop the "includeDeletedPages" option too

mattcasey:
> Another simplification; mapPageTree is only ever used in the Navigation case, so we should probably drop the "includeDeletedPages" option too

That sounds great too! I am feeling like this is already a pretty big change, but will follow up with another  PR

motechFR (approved):

PR #1472, opened 2023-01-12
extend Error class to include stack traces

mattcasey:
now you can log (err.stack) in the `onError` middleware, for example, and see where in the code the error was thrown

### DIFF
--- a/lib/middleware/onError.ts
+++ b/lib/middleware/onError.ts
@@ -18,7 +18,9 @@ export function onError(err: any, req: NextApiRequest, res: NextApiResponse) {
     });
   }
 
-  res.status(errorAsSystemError.code).json({ ...errorAsSystemError, message: errorAsSystemError.message });
+  const { stack, message, ...withoutStack } = errorAsSystemError;
+
+  res.status(errorAsSystemError.code).json({ ...withoutStack, message });
 }
 
 export default onError;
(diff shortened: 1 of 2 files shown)


lib/middleware/onError.ts
```diff
@@ -18,7 +18,9 @@ export function onError(err: any, req: NextApiRequest, res: NextApiResponse) {
     });
   }
 
-  res.status(errorAsSystemError.code).json(errorAsSystemError);
+  const { stack, message, ...withoutStack } = errorAsSystemError;
```
mattcasey, line 21: I am 50/50 on excluding stack trace. It's helpful for debugging, but it looks unprofessional, if that's a thing? I also have heard that you do it to limit an attacker's visibility into the system, but we are open-source.. but so long as the trace is in Datadog, it doesn't really need to be shown here

motechFR, line 21: Further simplification suggestion; only use stack and …withoutStack since we pass message straight to response

lib/utilities/errors/errors.ts
```diff
@@ -48,6 +48,7 @@ export class SystemError<E = any> implements ISystemError<E> {
   error: E;
 
   constructor(errorInfo: ISystemErrorInput<E>) {
+    super(errorInfo.message);
```
motechFR, line 51: Pleasantly surprised the CI worked.

In the past when I extended the native JS error object into system error class, I’d have issues in jest when asserting instance of for errors extending our system error

motechFR (approved):

PR #1473, opened 2023-01-12
simplify logic to show menu item in sidebar

mattcasey:
This is a cleanup of how we scroll to the correct page on first load. Will result in more accurate sidebar :)
- I think we can get rid of the local storage solution to remember the scroll position of sidebar - this was probably due to the page remounting all the time?
- I moved the scroll logic down and removed some DOM queries (these are really expensive compared to regular JS, and were running on every render!)


msadura (approved):
One more awesome improvement! Need for that hack in sidebar really bothered me 😅

Devorein (approved):

PR #1477, opened 2023-01-12
Add category on the PostCard

valentinludu:
https://app.charmverse.io/charmverse/page-18247391138004954


components/forum/components/PostList/PostList.tsx
```diff
@@ -40,6 +41,7 @@ export function ForumPostList({ search, categoryId, sort }: ForumPostsProps) {
   const { members } = useMembers();
   const { user } = useUser();
   const { subscribe } = useWebSocketClient();
+  const { categories } = useForumCategories();
```
mattcasey, line 44: seems like it would be better to use this hook inside PostCard, since we are passing the whole post already

PR #1481, opened 2023-01-12
User friendly forum links

motechFR:
PR for user friendly forum links and some extra cleanup

**1. Forum links navigation format**

All categories:
/forum/

Category view:
/forum/category/<categoryName>

Category view with popup (we could save this as a fast follower):
/forum/category/<categoryName>?postId=<postId>

Single post view*:
/forum/post/<post_path>

**2. Forum links generation**
- Autogenerated by our library; supports japanese characters

**3. Post Summary**
- Now the time icon only shows time post was created at for a better UX

**4. New post from category**
- New post from current category will default to that category
- On "All categories" it defaults to the space default category


Devorein (approved):

components/forum/components/CategoryMenu.tsx
```diff
@@ -195,9 +200,7 @@ export function CategoryMenu() {
           }}
         />
         <Button
-          disabled={
-            forumCategoryName.length === 0 || categories.find((category) => category.name === forumCategoryName)
-          }
+          disabled={categories.find((category) => category.name === newForumCategoryName)}
```
Devorein, line 203: We should disable empty categories from being created

Devorein (approved):
Looks great. Maybe if a user is in a specific category and deletes that category we could update the url to remove the category as the category itself is switched to `All categories`

mattcasey (changes requested):
Looks good, but one thing: we should sanitize category names so that the path is all lowercase, spaces are replaced with hyphens, and special characters stripped out. For example "category / subcategory" as category results in a 404.

Devorein:
It will be quite difficult to serialize and de-serialize them. Maybe this [npm package](https://www.npmjs.com/package/friendly-url) can help.

PR #1482, opened 2023-01-12
Show category related posts

Devorein:
image


.github/workflows/test_and_deploy.yml
```diff
@@ -10,64 +10,64 @@ on:
...
-  #       run: |
-  #         npm run start:test:ci &
-  #         npm run test:server:ci
-
-  #     - name: Run browser tests
-  #       run: npm run test:browser:ci
-
-  #     - name: Run browser e2e tests
-  #       run: npm run test:e2e:ci
+  test:
```
motechFR, line 13: Why has this file changed? I also noticed the test_and_deploy package numbers are bumped for some reason

components/forum/components/PostPage/PostPage.tsx
```diff
@@ -27,6 +28,7 @@ type Props = {
   spaceId: string;
   post: PostWithVotes | null;
   onSave?: () => void;
+  insideDialog?: boolean;
```
motechFR, line 31: I this this would be better named as "showOtherCategoryPosts"

This is because we want to be as explicit as possible on our intention of enabling a specific feature, whereas "insideDialog" would be an arcane case that only the original developer understands

lib/forums/posts/interfaces.ts
```diff
@@ -14,4 +14,5 @@ export type ForumPostMeta = {
   summary: PageContent | null;
   updatedAt: string;
   votes: ForumVotes;
+  path: string;
```
motechFR, line 17: May I suggest a cleanup of ForumPostMeta? It was originally defined when we had the posts + pages

I suggest making it as follows

```typescript
type ForumPostMeta = Pick<Post, 'createdAt' | 'updatedAt' | 'createdBy' | 'id' | 'categoryId' | 'title' | 'path'> & {
summary: PageContent,
votes: ForumVotes
}
```

motechFR (changes requested):
Some change requests :)

.github/workflows/test_and_deploy.yml
Devorein: It could be due to merging with `main`. Its been fixed now. Thank you for catching.

PR #1483, opened 2023-01-12
remove global typescript defs for prisma and log

mattcasey:
This causes serious bugs because Typescript tells you that prisma/log is always available without import


msadura (approved):

msadura:
note: CI still fails

PR #1484, opened 2023-01-12
Show confirmation before user closes post without saving

Devorein:
image

### DIFF
--- a/components/forum/components/PostDialog/PostDialog.tsx
+++ b/components/forum/components/PostDialog/PostDialog.tsx
@@ -46,7 +46,6 @@ export default function PostDialog({ post, spaceId, onClose, open }: Props) {
     if (post) {
       popupState.open();
     }
-    setForm(post ?? { title: '', content: null, contentText: '' });
   }, [!!post]);
 
   // open modal when page is set
--- a/components/forum/components/PostDialog/PostDialog.tsx
+++ b/components/forum/components/PostDialog/PostDialog.tsx
@@ -96,6 +96,8 @@ export default function PostDialog({ post, spaceId, onClose, open }: Props) {
         post={post ?? null}
         spaceId={spaceId}
         onSave={close}
+        contentUpdated={contentUpdated}
+        setContentUpdated={setContentUpdated}
       />
       <ConfirmDeleteModal
         onClose={() => {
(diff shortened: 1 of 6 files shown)


components/forum/components/PostDialog/PostDialog.tsx
```diff
@@ -72,9 +82,38 @@ export default function PostDialog({ post, spaceId, onClose, open }: Props) {
...
+        } else {
+          close();
+        }
+      }}
     >
-      <PostPage post={post ?? null} spaceId={spaceId} onSave={close} />
+      <PostPage
+        form={form}
+        setForm={(_form) => {
+          const updatedFormInputs = typeof _form === 'function' ? _form(form) : _form;
```
msadura, line 96: This brings some questions :D Why `_form` can be a function? This looks like some hacks.
Looks like I missed the part of storing `contentText` - why is this needed?

Devorein: Changed it to be less complex. Since it is a state setter it could be a function `(prevState) => newState` or just `newState`

components/forum/components/PostDialog/PostDialog.tsx
```diff
@@ -72,9 +76,35 @@ export default function PostDialog({ post, spaceId, onClose, open }: Props) {
...
+      <ConfirmDeleteModal
+        onClose={() => {
+          setShowConfirmDialog(false);
+        }}
+        title='Unsaved changes'
+        open={showConfirmDialog}
+        buttonText='Discard'
+        secondaryButtonText='Go back'
+        question='Are you sure you want to close this post? You have unsaved changes'
+        onConfirm={close}
```
motechFR, line 106: An import UX consideration; When I click discard, and leave the post. I'd expect the content to be deleted.

However, when I come back to the post, the content is still there

motechFR (approved):
Does what it says, no further comment

components/forum/components/PostDialog/PostDialog.tsx
Devorein: Thank you for catching that. Will fix 💯

valentinludu (commented):
If -> I write a title, cat, desc -> click close -> click discard -> Click Create new post

I see the Create new post modal with the title and description from before and no category selected. Is that intentional or a bug? I would have expect to see all empty.

valentinludu (commented):
Probably a bug:

IF -> click create new post -> fill the fields in this exact order: category, title, content.

The POST button is disabled and the tooltip says: Title is required. But title is filled. :(

valentinludu (changes requested):
Some small bugs

Devorein:
That was quite difficult. Thank you for spotting the bug. I had to spend quite a lot of time figuring out the issues and fixing them.

PR #1486, opened 2023-01-13
fix placeholder text and avatars in forums

mattcasey:
image

### DIFF
--- a/theme/@bangle.dev/styles.scss
+++ b/theme/@bangle.dev/styles.scss
@@ -100,8 +100,6 @@
 
 .ProseMirror {
   padding: 5px 0;
-  // set background to hide the loading indicator which disappears with a delay
-  background-color: var(--background-default);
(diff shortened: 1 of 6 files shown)


theme/@bangle.dev/styles.scss
```diff
@@ -100,8 +100,6 @@
 
 .ProseMirror {
   padding: 5px 0;
-  // set background to hide the loading indicator which disappears with a delay
-  background-color: var(--background-default);
```
mattcasey, line 104: I think we have a lot of room to clean up styling for charm editor.. for now, i put the background color definitions in CharmEditor and InlineCharmEditor instead of here to avoid UI bugs like what I did here (the white bg is on top of the grey bg):
image

msadura (approved):

valentinludu (approved):

PR #1488, opened 2023-01-13
Standardise the close button UI to the right on dialogs

motechFR:
This PR relocates the close button to the right of the dialog, and actually shows it

Screenshot 2023-01-13 at 13 45 31

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/dialog.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/dialog.tsx
@@ -1,11 +1,10 @@
 import CloseIcon from '@mui/icons-material/Close';
+import IconButton from '@mui/material/IconButton';
 import Modal from '@mui/material/Modal';
 import clsx from 'clsx';
 import React from 'react';
 import { useHotkeys } from 'react-hotkeys-hook';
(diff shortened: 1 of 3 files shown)


components/common/BoardEditor/focalboard/src/components/dialog.tsx
```diff
@@ -41,16 +41,19 @@ function FBDialog(props: Props) {
...
+              {toolsMenu}
               {!props.hideCloseButton && (
                 <IconButton
                   onClick={props.onClose}
                   icon={<CloseIcon />}
+                  style={{
+                    paddingLeft: '20px'
+                  }}
                   title={closeDialogText}
-                  className='IconButton--large'
```
msadura, line 49: I would go with Mui `IconButton` rather than old one. Always an easy way to move away from focalboard nasty "className" variant styling

msadura (changes requested):

components/common/BoardEditor/focalboard/src/components/dialog.tsx
motechFR: Great cleanup! Did as requested

PR #1489, opened 2023-01-13
Disable post polls

motechFR:
This hides post polls from the forum post command palette until we provide proper support


components/forum/components/PostPage/PostPage.tsx
```diff
@@ -173,6 +173,7 @@ export function PostPage({ showOtherCategoryPosts = false, post, spaceId, onSave
               pageActionDisplay={null}
               pageId={post?.id}
               disablePageSpecificFeatures
+              enableVoting={false}
```
msadura, line 176: overall PR looks good, but I am wondering - can we go with either `enableVoting` or `disableVoting` rather than mixing both approaches?

msadura (approved):

components/forum/components/PostPage/PostPage.tsx
motechFR: Thanks for pointing out, and agreed. I made the fix

PR #1497, opened 2023-01-13
Xps engine - update to create a workspace 

mattcasey:
We need to keep track of the unique id from Game7 (XPS Engine) when they create a workspace. The admin will also have a wallet address instead of a discord user id, which still needs to be supported but my thought was to update the api docs so they could see it

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -211,12 +211,13 @@ model Space {
   defaultVotingDuration       Int?
   snapshotDomain              String?
   spaceImage                  String?
-  defaultPostCategoryId       String?                          @db.Uuid @unique
-  defaultPostCategory         PostCategory?                    @relation(fields: [defaultPostCategoryId], references: [id], name: "defaultPostCategory", onDelete: SetNull)
+  defaultPostCategoryId       String?                           @unique @db.Uuid
+  defaultPostCategory         PostCategory?                     @relation(fields: [defaultPostCategoryId], references: [id], name: "defaultPostCategory", onDelete: SetNull)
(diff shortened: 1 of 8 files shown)


msadura:
I guess it looks ok, but I am wondering, if it would be a good idea to have 2 separate endpoints:

1) Create workspace wth discord data - basically this one would meet collabland's requirements
2) "Generic" endpoint with optional params to be used for game7 and other potential integrations.

I might be overthinking, but I've got the feeling that this enpoint used for collabland is kind of "specific".

prisma/schema.prisma
```diff
@@ -211,12 +211,13 @@ model Space {
...
   spaceImage                  String?
-  defaultPostCategoryId       String?                          @db.Uuid @unique
-  defaultPostCategory         PostCategory?                    @relation(fields: [defaultPostCategoryId], references: [id], name: "defaultPostCategory", onDelete: SetNull)
+  defaultPostCategoryId       String?                           @unique @db.Uuid
+  defaultPostCategory         PostCategory?                     @relation(fields: [defaultPostCategoryId], references: [id], name: "defaultPostCategory", onDelete: SetNull)
   defaultPagePermissionGroup  PagePermissionLevel?              @default(full_access)
   defaultPublicPages          Boolean?                          @default(false)
   permissionConfigurationMode SpacePermissionConfigurationMode? @default(custom)
   publicBountyBoard           Boolean?                          @default(false)
+  xpsEngineId                 String?
```
msadura, line 220: just an potential idea - how about calling it `externalId` or something like that, to have a generic identifier that we can use for any other integration in the future?

msadura (approved):

PR #1499, opened 2023-01-15
Send notifications for mentions in post content and comments

Devorein:
@mention in post content and comments send notifications in the discussions tab.

### DIFF
--- a/lib/discussion/getDiscussionTasks.ts
+++ b/lib/discussion/getDiscussionTasks.ts
@@ -23,7 +23,7 @@ interface GetDiscussionsInput {
 }
 
 interface GetDiscussionsResponse {
-  mentions: Record<string, Discussion>;
+  mentions: Discussion[];
   discussionUserIds: string[];
   comments: Discussion[];
 }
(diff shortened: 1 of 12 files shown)


lib/discussion/getDiscussionTasks.ts
```diff
@@ -90,9 +90,11 @@ export async function getDiscussionTasks(userId: string): Promise<DiscussionTask
...
   const { mentions, discussionUserIds, comments } = await Promise.all([
     getComments(context),
-    getMentionsFromComments(context),
-    getMentionsFromPages(context),
-    getMentionsFromCommentBlocks(context)
+    getPageCommentMentions(context),
+    getPageMentions(context),
+    getMentionsFromCommentBlocks(context),
+    getPostMentions(context),
+    getPostCommentMentions(context)
```
msadura, line 97: I know this is not a part a strict part of this PR but anyways something that should be fixes:
1) Do not use spreads in reduce
2) What is the point of making mentions as a map if it is converted to array below anyway?

lib/discussion/getDiscussionTasks.ts
```diff
@@ -440,7 +442,69 @@ async function getMentionsFromComments({
...
+            bountyId: null,
+            bountyTitle: null,
+            type: 'post'
+          };
+        }
+      });
+    }
+  }
+  return {
+    mentions: mentionsMap,
```
msadura, line 501: I do not really undesrand why mentions cannot be jsut a n arary like comments and discussionUserIds to have consistent data model

mattcasey (approved):

lib/discussion/getDiscussionTasks.ts
Devorein: This was to make sure we don't count multiple same mentions as the same. It sounds strange but when prosemirror goes through each blocks to extract mentions out of them, there is a slight chance that it will encounter the same mention twice. The best approach would be have the mention as a map inside `extractMentions.ts` so that everywhere else it can be used as a regular array.

PR #1500, opened 2023-01-16
Update poll deadline

valentinludu:
Main changes
- User can now change deadlines for the polls (inline and in proposals)
- If the user doesn't finish the process, the update will not be triggered. This behaviour is also the same when you create a poll.
- Removed all imports that were not used in some files
- In useVotes I changed the setState behaviour for all functions to update the state only once per action without mutating the state variable.
- If you check package-lock.json you will see more then 1 package update. That is because the last package.json committed update didn't include package-lock

### DIFF
--- a/charmClient/apis/votesApi.ts
+++ b/charmClient/apis/votesApi.ts
@@ -1,7 +1,7 @@
 import type { UserVote } from '@prisma/client';
 
 import * as http from 'adapters/http';
-import type { ExtendedVote, UserVoteExtendedDTO, VoteDTO } from 'lib/votes/interfaces';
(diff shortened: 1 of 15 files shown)


Devorein (approved):
Code looks good but there are some bugs/UI suggestions
1. Its possible to create and change deadline in the past (current day but past hours). Maybe we can do both client and server side validation so that deadline is not a past date.
2. We should hide the ellipsis if the deadline has passed and there is no action that can be done for the poll
image

valentinludu:
> Code looks good but there are some bugs/UI suggestions
> 
> 1. Its possible to create and change deadline in the past (current day but past hours). Maybe we can do both client and server side validation so that deadline is not a past date.
> 2. We should hide the ellipsis if the deadline has passed and there is no action that can be done for the poll


Done

mattcasey (approved):

PR #1503, opened 2023-01-16
Feat/collabland sync

msadura:
Collabland sends webhook events in a different data model than expected, this PR is to adjust handling that model properly.

Event comes with payload which reflects "old" and "updated" member entity like:

```
{
  event: 'guildMemberUpdate',
  payload: [
    {
      guildId: '889381536904904734',
      joinedTimestamp: 1644677954638,
      premiumSinceTimestamp: null,
      nickname: null,
      pending: false,
      communicationDisabledUntilTimestamp: null,
      userId: '829019255625875456',
      avatar: null,
      displayName: 'wishmaster',
      roles: ['942072530896830544', '942072632986177637', '889381536904904734'],
      avatarURL: null,
      displayAvatarURL: 'https://cdn.discordapp.com/avatars/829019255625875456/8e8649f7ffb6008b822872404f7990bc.webp'
    },
    {
      guildId: '889381536904904734',
      joinedTimestamp: 1644677954638,
      premiumSinceTimestamp: null,
      nickname: null,
      pending: false,
      communicationDisabledUntilTimestamp: null,
      userId: '829019255625875456',
      avatar: null,
      displayName: 'wishmaster',
      roles: ['904089148552527942', '942072530896830544', '942072632986177637', '889381536904904734'],
      avatarURL: null,
      displayAvatarURL: 'https://cdn.discordapp.com/avatars/829019255625875456/8e8649f7ffb6008b822872404f7990bc.webp'
    }
  ]
};
```


lib/aws/webhookSqs.ts
```diff
@@ -66,19 +71,23 @@ export async function processMessages({ processorFn }: ProcessMssagesInput) {
...
+      log.error('[SQS] Failed to process webhook message', e);
     } finally {
       await deleteMessage(message.ReceiptHandle || '');
     }
   }
 
   // process next message
-  processMessages({ processorFn });
+  if (queueUrl) {
+    processMessages({ processorFn });
```
mattcasey, line 89: I think it would be nice to see queueUrl passed into processMessages [and subsequently, getNextMessage), since it's being called conditionally.. Also it looks like getNextMessage() is retrieving the queueURL again and already handles it not being defined?

lib/aws/webhookSqs.ts
```diff
@@ -27,10 +27,15 @@ export async function getQueueUrl() {
     return queueUrl;
   }
 
-  const command = new GetQueueUrlCommand({ QueueName: SQS_NAME });
-  const res = await client.send(command);
-  queueUrl = res.QueueUrl || '';
-  return queueUrl;
+  try {
+    const command = new GetQueueUrlCommand({ QueueName: SQS_NAME });
```
mattcasey, line 31: FYI it's normal to just hard-code the entire URL as the env var, you don't need to query AWS to find out what it is.

lib/collabland/assignRolesCollabland.ts
```diff
@@ -0,0 +1,28 @@
...
+    const discordRoles = await getGuildRoles(discordServerId);
+    const rolesToAdd = roleIdsToAdd
+      .map((roleId) => discordRoles.find((role) => role.id === roleId))
+      .filter(Boolean) as ExternalRole[];
+
+    const { space, user } = await getSpaceAndUserFromDiscord({ discordUserId, discordServerId });
+
+    return createAndAssignRoles({ userId: user.id, spaceId: space.id, roles: rolesToAdd });
+  } catch (e) {
+    return null;
```
mattcasey, line 26: please always log the error, maybe as a warning if its not something we need to look at

lib/discord/unassignRolesDiscord.tsx
```diff
@@ -10,12 +9,12 @@ export async function unassignRolesDiscord({
...
   discordUserId: string;
   discordServerId: string;
-  roles: ExternalRole[] | ExternalRole;
+  roles: string[] | string;
 }) {
   const rolesToRemove = Array.isArray(roles) ? roles : [roles];
   const { user } = await getSpaceAndUserFromDiscord({ discordUserId, discordServerId });
   const removeRoles = await prisma?.role.findMany({
-    where: { externalId: { in: rolesToRemove.map((role) => String(role.id)) } }
+    where: { externalId: { in: rolesToRemove.map((role) => String(role)) } }
```
mattcasey, line 17: why do we need to cast the role as a string? seems odd and possibly error-prone

mattcasey (approved):
Looks like it should work, approved! But I added some suggestions

lib/aws/webhookSqs.ts
```diff
@@ -27,10 +27,15 @@ export async function getQueueUrl() {
...
+    const res = await client.send(command);
+    queueUrl = res.QueueUrl || '';
+    return queueUrl;
+  } catch (e) {
+    log.error('[SQS] Error while getting queue url', e);
```
mattcasey, line 36: I added a logger utility to prefix logs in a file. It was kind of an experiment, not sure if it's something we wanted to use:
```
import { getLogger } from 'lib/log/prefix';
const log = getLogger('ws-docs');
log.info('Client connected') // outputs: [ws-docs] Client connected
```

PR #1504, opened 2023-01-16
Fix snapshot publish

motechFR:
- Fix broken text-color mark export
- Add support for exporting the username of a mentioned person
- Add UI error handling and notification


lib/pages/generateMarkdown.ts
```diff
@@ -21,6 +30,8 @@ export async function generateMarkdown(page: Page, withTitle: boolean = false):
       }
     });
 
+    (serializer.options as any).charmOptions = generatorOptions;
```
mattcasey, line 33: This is a little ugly, but if it works it works. I would have used "Context" in the name instead of "Options" since it's not really got anything to do with options.

PR #1506, opened 2023-01-17
fix /bookmark theme in light mode

mattcasey:


components/common/CharmEditor/components/common/MediaUrlInput.tsx
```diff
@@ -41,3 +41,11 @@ export function MediaUrlInput(props: InputProps) {
     </Box>
   );
 }
+
+// automatically add https:// to user input
+function getUrl(userInput: string) {
```
mattcasey, line 46: just making it easier to add sites, so you can type "ethdenver.com" for example and we'll add the https:// for you

components/common/CharmEditor/components/bookmark/BookmarkNodeView.tsx
```diff
@@ -21,8 +22,11 @@ export function BookmarkNodeView({
   deleteNode
 }: CharmNodeViewProps & { readOnly?: boolean }) {
   const { url } = node.attrs as BookmarkNodeAttrs;
+
+  const theme = useTheme();
+
   const { data, error, isLoading } = useSWRImmutable(url ? `iframely/${encodeURIComponent(url)}` : null, () =>
```
Devorein, line 28: I think we should also add `theme` to the key so that if the theme changes, the embed will be re-fetched & re-rendered.
Maybe something like `iframely/${encodeURIComponent(url)}?theme=${theme.palette.mode}`

Devorein (changes requested):

components/common/CharmEditor/components/bookmark/BookmarkNodeView.tsx
mattcasey: good call, thanks

Devorein (approved):

PR #1507, opened 2023-01-17
Fix publishing to snapshot when space has a voting duration

motechFR:


components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
```diff
@@ -339,30 +351,41 @@ export default function PublishingForm({ onSubmit, page }: Props) {
...
+                  <FieldLabel>End date</FieldLabel>
+                  <DateTimePicker
+                    minDateTime={startDate}
+                    value={endDate}
+                    onChange={(value) => {
+                      if (value instanceof DateTime) {
+                        setEndDate(value as DateTime);
+                      }
+                    }}
+                    disabled={!!snapshotSpace.voting.period}
```
mattcasey, line 376: I like that you thought to disable the period as well.

mattcasey (approved):

lib/snapshot/interfaces.ts
```diff
@@ -22,8 +22,11 @@ export interface SnapshotSpace {
   network: string;
   private: boolean;
   strategies: SnapshotVotingStrategy[];
-  // TBC - Check out validation
-  // validation:
+  voting: {
```
mattcasey, line 25: I think this field is optional, based on the first example in their docs. it doesn't always exist?

motechFR: The graphql query is hardcoded, so we would always expect to get these fields back

i tried with voting duration in existent, or defined in snapshot and they came through here

PR #1509, opened 2023-01-17
Fix pages endpoint permissions

msadura:


lib/pages/server/getAccessiblePages.ts
```diff
@@ -294,7 +298,7 @@ export async function getAccessiblePages(input: PagesRequest): Promise<IPageWith
           (permission.roleId &&
             spaceRole &&
             availableRoles.some((r) =>
-              r.spaceRolesToRole.some((s) => s.spaceRoleId === spaceRole.id && r.id === permission.roleId)
+              r.spaceRolesToRole.some((s) => s.spaceRoleId === spaceRole?.id && r.id === permission.roleId)
             )) ||
           permission.spaceId === input.spaceId
```
motechFR, line 303: // This should already be covered by nulling out the userId previously, but for extra future proofness can we rewrite as follows?

if (permission.public) {
        return true;
      } else if (input.userId **&& spaceRole**) {
 // Code
} else {
...
}
spaceRole && permission.spaceId === input.spaceId

motechFR (approved):

PR #1514, opened 2023-01-18
Fix hydration error when running Next in dev mode

mattcasey:
image

Based on trial-and-error, I think the errors are related to our usage of `router.isReady` to determine how to render components, which Next.js says not to do. I added this logic [12 months](https://github.com/charmverse/app.charmverse.io/commit/85bac433b4121e3fb87c87281e545acf444652f8#diff-a9827ce1be32c607c948881a89b99c592f752898f9c49e0dea0b7199eb151029R67) ago when I was first building the app, but it basically means we don't do any rendering on the server! My comment was about hiding components until the space is known, but all our components seem to handle it gracefully.

https://nextjs.org/docs/api-reference/next/router
> isReady: boolean - Whether the router fields are updated client-side and ready for use. Should only be used inside of useEffect methods and not for conditionally rendering on the server. See related docs for use case with [automatically statically optimized pages](https://nextjs.org/docs/advanced-features/automatic-static-optimization)

### DIFF
--- a/components/common/RouteGuard.tsx
+++ b/components/common/RouteGuard.tsx
@@ -20,8 +20,7 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
   const [authorized, setAuthorized] = useState(true);
   const { user, setUser, isLoaded } = useUser();
   const { spaces, isLoaded: isSpacesLoaded } = useSpaces();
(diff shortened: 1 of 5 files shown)


msadura (approved):
awesome catch, this error was annoying :D

PR #1520, opened 2023-01-18
Sync discord roles when space is created

msadura:
Create all roles from discord when space is created with discord id


mattcasey (approved):
nice touch

PR #1525, opened 2023-01-18
Make forum breadcrumb clickable

valentinludu:


components/common/PageLayout/components/Header/components/PageTitleWithBreadcrumbs.tsx
```diff
@@ -134,12 +134,14 @@ function ProposalPageTitle({ basePath }: { basePath: string }) {
 
 function ForumPageTitle({ basePath }: { basePath: string }) {
   const [pageTitle] = usePageTitle();
+  const title = pageTitle === 'Forum' ? 'All Categories' : pageTitle;
```
msadura, line 137: is there any other way to recognize this than comparing page title? Feels a bit fragile.

mattcasey: i think you can just leave it called "Forum" as well. It's a breadcrumb, so the title shouldn't change

mattcasey (changes requested):
Not sure about "All categories" where the page title should go. But the more I think about it, the more it makes sense! I kind of like it on top of the feed but not where breadcrumbs appear. Was this a request by Chris/Alex or can you pass it by them?

valentinludu:
> Not sure about "All categories" where the page title should go. But the more I think about it, the more it makes sense! I kind of like it on top of the feed but not where breadcrumbs appear. Was this a request by Chris/Alex or can you pass it by them?

Chris said he is ok with: Forum/ All Categories 

Screenshot 2023-01-20 at 10 30 13

components/common/PageLayout/components/Header/components/PageTitleWithBreadcrumbs.tsx
valentinludu: > is there any other way to recognise this than comparing page title? Feels a bit fragile.

I improved a bit so now I am checking for the path like this: pathName === '/[domain]/forum'

valentinludu: > This should still just say "Forum" anyway. We don't need to give the page two different titles

If I let pageTitle it will say Forum / Forum. That is why I added the All Categories string.

mattcasey (approved):

PR #1528, opened 2023-01-19
set sidebar state before page renders

mattcasey:
Turns out, we can't rely on local storage to define UI preferences because it's not available for SSR. I removed a check on 'window', but now the sidebar is always open by default at first, even if it's supposed to be closed.
So this change doesn't render the sidbar/nav at all until the first render, which is unfortunate but we could fix using cookies and getServerSideProps


config/constants.ts
```diff
@@ -1,5 +1,5 @@
 // Note: NODE_ENV can only be 'development' or 'production' according to Next.js, but we don't want to mix them with test env
-export const isTestEnv = process.env.NEXT_PUBLIC_APP_ENV === 'test';
+export const isTestEnv = process.env.NEXT_PUBLIC_APP_ENV === 'test' || process.env.NODE_ENV === 'test';
```
mattcasey, line 2: this is unrelated but something we should fix

valentinludu (approved):

PR #1530, opened 2023-01-19
check if value is a primitive for metadata in logs

mattcasey:
fixes a bug in datadog:
image

### DIFF
--- a/lib/log/logLevel.ts
+++ b/lib/log/logLevel.ts
@@ -62,11 +62,7 @@ export function apply(log: Logger, logPrefix: string = '') {
             if (isPrimitiveValue(opt) || opt instanceof Array) {
               _opt = { data: opt };
             } else {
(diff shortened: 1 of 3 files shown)


lib/log/logLevel.ts
```diff
@@ -34,51 +29,13 @@ export function apply(log: Logger, logPrefix: string = '') {
...
-            let error: LogMeta['error'];
-            const maybeError = opt.error || opt;
-            if (maybeError instanceof Error) {
-              error = { ...maybeError, message: maybeError.message, stack: maybeError.stack };
-            }
-            if (opt instanceof Array) {
-              _opt = { data: opt };
-            } else {
-              try {
-                _opt = { ...opt };
```
mattcasey, line 66: this is where the bug was... spreading a string value results in the screenshot

PR #1533, opened 2023-01-19
Feat/unified share urls

msadura:
Unisied url for public pages. Shared pages will be reachable with the same url pattern as for space members:
- regular page / board: `[domain]/[pagePath`
- public bounties board `[domain]/bounties`

How it works:
- `usePublicPage` hooks take all the responsibility for checking if page that user is viewing (based on router params) should be viewed in regular mode (user is logged in and has access to space) or if we should load additional public page data to check if page is available in public mode. 
- `RouteGuard` uses `usePublicPage` and will not redirect to login if user can see page in public mode
- `PageLayout` component also uses `usePublicPage` to verify if it should user regular layout, or public one
- If we are in public mode, `useCurrentSpace` uses public space as current one and all the data gets propagated

If you are wondering about reusing `usePublicPage` hook - its logic is lightweight, it makes request for public page only when needed + swr will cache result so server gets called only once.

NOTE: I did not touched current `/share/[pagePath]` route and component as some users might still be using that pattern.


components/common/RouteGuard.tsx
```diff
@@ -54,10 +55,6 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
 
       setAuthorized(result.authorized);
 
-      if (result.user) {
```
msadura, line 57: this was not used anywhere

hooks/usePages.tsx
```diff
@@ -67,13 +67,13 @@ export function PagesProvider({ children }: { children: ReactNode }) {
...
   const { data, mutate: mutatePagesList } = useSWR(
-    () => (currentSpace ? getPagesListCacheKey(currentSpace.id) : null),
+    () => getPagesListCacheKey(currentSpace?.id),
     async () => {
       if (!currentSpace) {
         return {};
       }
 
-      const pagesRes = await charmClient.pages.getPages(currentSpace.id);
+      const pagesRes = await charmClient.pages.getPages(currentSpace?.id);
```
mattcasey, line 76: because we check for currentSpace on 72, you don't have to add the ? here

components/publicPages/utils.ts
```diff
@@ -0,0 +1,10 @@
+const BOUNTIES_PATH = '/[domain]/bounties';
+const PUBLIC_PAGE_PATHS = [BOUNTIES_PATH, '/[domain]/[pageId]'];
+
+export function isPublicPagePath(path: string): boolean {
```
mattcasey, line 4: what is the difference of these "public pages" with the ones in RouteGuard? Could you think of a slightly different name that might hint? I'm actually not sure how these are public vs the pages like /invite and "/".

components/publicPages/PublicPage.tsx
```diff
@@ -0,0 +1,82 @@
+import { useEffect } from 'react';
```
mattcasey, line 1: the first folder in 'components' should be a path, or else the component should go in 'components/common'

mattcasey: Another place to put it would be next to EditorPage inside components/[pageId]

components/publicPages/hooks/usePublicPage.ts
```diff
@@ -0,0 +1,76 @@
+import { useRouter } from 'next/router';
```
mattcasey, line 1: I'd consider just putting this in the top-level hooks folder since it's needed by different page components (bounties and [pageId])

mattcasey: Would a better name for this be "useSharedPage"? I keep thinking this is for 'public pages' but we also have public pages like the login page, etc. Not a big deal, I guess

components/publicPages/hooks/usePublicPage.ts
```diff
@@ -0,0 +1,76 @@
...
+import { useMemo } from 'react';
+import useSWR from 'swr';
+
+import charmClient from 'charmClient';
+import { isBountiesPagePath, isPublicPagePath } from 'components/publicPages/utils';
+import { useSpaces } from 'hooks/useSpaces';
+
+export const usePublicPage = () => {
```
mattcasey, line 9: this is probably the first time I saw a hook that I wanted some tests on it! What do you think?

components/publicPages/hooks/usePublicPage.ts
```diff
@@ -0,0 +1,76 @@
...
+  const space = publicSpace || publicPage?.space;
+  const hasError = !!publicPageError || !!spaceError;
+  const hasPublicBounties = space?.publicBountyBoard;
+  const hasPublicPageAccess = !!publicPage || !!hasPublicBounties;
+  const accessCheked =
+    (spacesLoaded && (hasPublicPageAccess || (!shouldLoadSpace && !shouldLoadPublicPage) || !!space)) || hasError;
+
+  return {
+    isCheckingAccess: isPublicPageLoading || isSpaceLoading,
+    accessCheked,
```
mattcasey, line 70: Typo.. should be accessChecked

components/common/RouteGuard.tsx
mattcasey: cc @motechFR ... it's not used anywhere but it seems still like it might be important. Technically, it's used anywhere in our code that calls useUser()

mattcasey (approved):
Approved with some minor comments. I would like to move the components from components/publicPages to either components/[pageId], or components/common.
Folders should either be a page path or a component and loosely follow some kind of hierarchy.

mattcasey:
I believe you can just delete the /pages/share folder now and add this rule to next.config.js:
```

  async redirects() {
    return [
      {
        source: '/:domain(^(?!.*\bapi\b).*$)/settings',
        destination: '/:domain/settings/workspace',
        permanent: false
      },
      {
        source: '/:domain(^(?!.*\bapi\b).*$)/bounties/:id',
        destination: '/:domain/bounties?bountyId=:id',
        permanent: false
      },
      {
        // strip out old /share prefix
        source: '/share/:path*',
        destination: '/:path*',
        permanent: true
      }
    ];
  },
  ```
  Here's the docs: https://nextjs.org/docs/api-reference/next.config.js/redirects

components/publicPages/utils.ts
msadura: Awesome idea, having 2 different types of pages being called "public" was bothering indeed

components/publicPages/hooks/usePublicPage.ts
msadura: I can definietly do that, but I would prefer in separate pr - looks like our react testing library needs update, so I would do it, check if update will not caouse any troubles and add tests for hook

PR #1535, opened 2023-01-19
Warn user before losing content in the CharmEditor with sockets

Devorein:


components/common/CharmEditor/components/fiduswriter/ws.ts
```diff
@@ -101,13 +101,25 @@ export class WebSocketConnector {
...
     // this.listeners.onOffline = onOffline;
     // window.addEventListener('offline', onOffline);
+    const handleWindowClose = (e: BeforeUnloadEvent) => {
+      if (this.messagesToSend.length === 0) return;
+      e.preventDefault();
+      (e || window.event).returnValue = warningText;
+      return warningText;
+    };
+
+    window.addEventListener('beforeunload', (e) => {
```
valentinludu, line 120: 1. To be simpler: window.addEventListener('beforeunload', handleWindowClose)
2. Can you think of a way to remove the event listener. Maybe in the close() method.

mattcasey: agree, i think we need to remove the listener too. every time you go to a new page, it will add another event listener

mattcasey: it might be even better to just add a method on the instance "isSavingChanges" and handle the listeners inside of ReactEditor

components/common/CharmEditor/components/fiduswriter/ws.ts
```diff
@@ -101,13 +101,25 @@ export class WebSocketConnector {
       // path: '/api/socket'
     });
     this.createWSConnection();
+    this.init();
```
mattcasey, line 104: please don't add more init logic than we have already

PR #1537, opened 2023-01-19
Use editable area for text and links

motechFR:
Use the existing component, and autoexpand values


mattcasey:
Please fix the text so that it's the smaller font size we had prior:
Current:
image
New one (too big):
image

components/common/BoardEditor/focalboard/src/components/properties/link/link.tsx
```diff
@@ -34,11 +34,10 @@ function URLProperty(props: Props): JSX.Element {
   return (
     <div className='URLProperty property-link url'>
       {(hasValue || props.placeholder) && (
-        <Editable
```
mattcasey, line 37: maybe you just need to brign back this class `octo-propertyvalue`

motechFR: octo-propertyvalue definitely cleaned up a bunch of the text

mattcasey:
the labels and input are out of line still:
image

motechFR:
@mattcasey fixed this, so we do a center align in case of empty content. Let me know if that works

PR #1539, opened 2023-01-19
Searchbar for database data source

Devorein:


motechFR (approved):
Small suggestion: When no results are found, say "No databases found" instead of "No pages found" since user is selecting from "CharmVerse databases"

PR #1544, opened 2023-01-19
improve DX for super tokens

mattcasey:
- make name unique
- allow passing in an existing token value when creating tokens (for when staging is reset)


scripts/makeSuperAPIKey.ts
```diff
@@ -5,13 +5,14 @@ import { provisionSuperApiKey } from 'lib/middleware/requireSuperApiKey';
 // use this file and run against production to generate super api keys
 // Super API keys allow you to make requests to the API to do things like generate a new space
 const name = '';
+const token = ''; // set this to create a specific token
```
msadura, line 8: Nice, I was just creating a new one and then updating in db manyally to use "previous" version of a token 😅

msadura (approved):

PR #1547, opened 2023-01-20
CharmEditor Embed improvements

mattcasey:
- smoother resizing of sizes
- allow users to resize width of iframe
- smarter parsing for width/height attributes from pasted code
- add an "Embed" and "Link" tabs when inputting a iframe src, however we wanted to support embed codes no matter which tab you use - this is just a trick to educate users that they can use embed codes

### DIFF
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -13,7 +13,7 @@ import debounce from 'lodash/debounce';
 import throttle from 'lodash/throttle';
 import { useRouter } from 'next/router';
 import type { CSSProperties, ReactNode } from 'react';
-import { memo, useCallback, useEffect, useRef, useState } from 'react';
+import { memo, useEffect, useRef, useState } from 'react';
 import { useSWRConfig } from 'swr';
 
 import charmClient from 'charmClient';
--- a/components/common/CharmEditor/CharmEditor.tsx
+++ b/components/common/CharmEditor/CharmEditor.tsx
@@ -489,15 +489,6 @@ function CharmEditor({
       color: 'var(--charmeditor-active)'
     }
   });
-
-  const onResizeStop = useCallback(
-    (view: EditorView) => {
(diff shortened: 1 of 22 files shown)


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -489,15 +489,6 @@ function CharmEditor({
       color: 'var(--charmeditor-active)'
     }
   });
-
-  const onResizeStop = useCallback(
```
mattcasey, line 493: this turned out not to be necessary

components/common/CharmEditor/specs/embeddedNodeSpec.ts
```diff
@@ -10,6 +10,7 @@ export function embeddedNodeSpec(_spec: EmbeddedSpecProps): RawSpecs {
     name: _spec.name,
     markdown: _spec.markdown,
     schema: {
+      atom: true,
```
mattcasey, line 13: I think this is what really made resizing not go haywire. Atom tells Prosemirror that the content inside is not editable

components/common/CharmEditor/components/Resizable/Resizable.tsx
```diff
@@ -40,6 +35,10 @@ function Resizable(props: ResizableProps) {
     }
   }, []);
 
+  useEffect(() => {
+    setSize(initialSize);
+  }, [initialSize]);
```
mattcasey, line 40: this also seemed to help smooth out resizing

components/common/CharmEditor/components/video/video.ts
```diff
@@ -8,7 +8,7 @@ export function plugins() {
   return [
     NodeView.createPlugin({
       name: 'video',
-      containerDOM: ['video-embed']
+      containerDOM: ['div', { class: 'video-container' }]
```
mattcasey, line 11: using a custom tag "video-embed" was causing some CSS not to appear, like the blue background when a video was selected

msadura (approved):

PR #1548, opened 2023-01-20
Profile and spaces dropdown menu

valentinludu:
Removed the first section of the sidebar where before were all the spaces and my profile.

You can still see it in Nexus. There will be other stories that will transform nexus into a modal and the sidebar will disappear from there also.

https://app.charmverse.io/charmverse/page-8516743958920909


Devorein:
Works great. I just noticed a few minor issues.
1. The menu item should be full-width on hover
2. Dropdown menu closes even when clicking on "My Spaces" section
3. Closing the space sidebar doesn't persist across page reload
image

Some UI suggestions
1. Probably its better to show the current space icon beside the text. Notion does it.
2. Maybe Sign out should be in red?
image

components/common/PageLayout/components/Sidebar/Sidebar.tsx
```diff
@@ -235,21 +217,18 @@ export default function Sidebar({ closeSidebar, favorites }: SidebarProps) {
     }
   }
 
+  async function logoutCurrentUser() {
```
Devorein, line 220: Maybe we can move it inside `useUser` hook? It's being used in `NexusPageTitle` as well.

Devorein (approved):

valentinludu:
> The menu item should be full-width on hover

I made it bigger so it's full width in the sidebar. In the future the sidebar will have the drag option so the user can see the whole title if he drags the sidebar and makes it bigger.

> Dropdown menu closes even when clicking on "My Spaces" section

Yes, the dropdown closes on any click. It was easy to add the close method on the Menu. The alternative would have been to add the close method on each MenuItem.

> Closing the space sidebar doesn't persist across page reload

This was a bug introduced a few days ago. I fixed it.

> Probably it's better to show the current space icon beside the text. Notion does it.

Yes, I missed that from the design. Added!

> Maybe Sign out should be in red?

We have only errors in red. I would avoid adding this color anywhere.

components/common/PageLayout/components/Sidebar/Sidebar.tsx
valentinludu: I wouldn't combine useUser, useWeb3AuthSig and useRouter. I think they are 3 different hooks for a reason.
Also I had a discussion with Matt a couple of weeks ago and I discovered we had some problems with putting useRouter inside hooks.

PR #1554, opened 2023-01-20
Charmeditor for forum posts and comments

Devorein:
Removed row actions from the `CharmEditor` for now.
image


components/common/BoardEditor/focalboard/src/components/viewSidebar/viewSourceOptions.tsx
```diff
@@ -176,6 +176,10 @@ function CharmVerseDatabases(props: DatabaseSourceProps & { activePageId?: strin
           pages={sortedPages}
           activePageId={props.activePageId}
           onSelectPage={onSelect}
+          style={{
+            height: '250px',
```
mattcasey, line 180: why does a height need to be set here? it should just be max height

mattcasey, line 180: i think the height should be 100%, why does it need a limit

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -271,20 +271,33 @@ export function charmEditorPlugins({
...
 
   /** DONT REMOVE THIS STYLING */
   /** ITS TO MAKE SURE THE USER CAN DRAG PAST THE ACTUAL CONTENT FROM RIGHT TO LEFT AND STILL SHOW THE FLOATING MENU */
   left: -50px;
 
+  ${({ colorMode }) =>
+    colorMode === 'dark'
+      ? `
+          background-color: var(--background-light);
+          .ProseMirror[data-placeholder]::before {
```
mattcasey, line 288: prosemirror should not have to define a background color. that was causing bugs

components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -271,20 +271,33 @@ export function charmEditorPlugins({
...
+      : ''};
+
   /** DONT REMOVE THIS STYLING */
   div.ProseMirror.bangle-editor {
     padding-left: 50px;
     margin-right: -50px;
   }
 
-  code {
+  code {s
```
mattcasey, line 300: typo

components/common/CharmEditor/components/PageList.tsx
```diff
@@ -11,14 +11,16 @@ interface Props {
...
 }
 
 export default function PagesList({
   emptyText = 'No pages found',
   activeItemIndex = -1,
   activePageId,
   pages,
-  onSelectPage
+  onSelectPage,
+  style
```
mattcasey, line 23: use className instead of style, and styled components please

mattcasey (approved):
approved, but could be cleaned up a little

components/common/BoardEditor/focalboard/src/components/viewSidebar/viewSourceOptions.tsx
Devorein: Without a fixed height the inline-database source list makes all the databases visible
image
image

components/common/CharmEditor/CharmEditor.tsx
Devorein: Thank you 🙏🏼

components/common/CharmEditor/components/PageList.tsx
Devorein: This is just a one-off style used only in `viewSourceOptions`, not sure how to make use of `className` in this case. Perhaps `fixedHeight` prop would add a `class` with appropriate styles?

components/forum/components/PostPage/components/PostComment.tsx
```diff
@@ -179,17 +179,26 @@ export function PostComment({
             left: 10
           }}
         />
-        <Box ml={3}>
+        <Box
+          ml={3}
+          sx={{
+            'div.ProseMirror.bangle-editor': {
```
Devorein, line 185: I'm not sure how to use a different padding-left for bangle-editor. @mattcasey you mentioned using classes, do you mean a `bangleEditorclassName` prop passed to `CharmEditor` like `no-padding-left`, that will remove the padding-left?

mattcasey (approved):

PR #1557, opened 2023-01-23
Add mutiline support to focalboard URL property

mattcasey:


motechFR:
@mattcasey Tested and looks good. There is one outstanding issue, which is that in the table view, the row expands fully to the text content. I think we need to fix this UX

Screenshot 2023-01-23 at 14 56 00

motechFR (approved):

PR #1558, opened 2023-01-23
reset view settings state when navigating between databases

mattcasey:
This fixes a few bugs related to state not getting updated properly when navigating between boards.

Fixes:
- when navigating from a databsae that had a source to one that doesn't, we wouldn't show the 'select a source' menu
- when navigating from a db with no source to one that has a source, we would briefly show the 'select a source' menu, because activeView was not set until a useEffect got triggered. Instead, I try to grab the first one from the board views if there are any
- it was not possible to create a new view on a linked inline db. (we'd constantly reset the current view to the first available - removing this logic doesnt seem to break anything, tested w/google forms too)
- in the views tabs, fixed the underline / tab indicator that animates into the first position when loading a new board. It now just appears instant.
- fixed flicker of bold font weight when clicking a Database page or one of its views in the sidebar
- added a loadingState to our Redux data. I thought I'd need this, but I don't. I am leaving it in cuz it could be useful, but open to just deleting it. It was a little hard to figure out, not having used Redux before

### DIFF
--- a/components/[pageId]/DatabasePage/DatabasePage.tsx
+++ b/components/[pageId]/DatabasePage/DatabasePage.tsx
@@ -142,6 +142,8 @@ export function DatabasePage({ page, setPage, readOnly = false, pagePermissions 
             viewId: viewId || ''
           }
         });
+        // call setCurrentViewId in case user clicked "add view", because we didnt update the URL so it wouldnt affect the activeView
+        setCurrentViewId(viewId);
       }
     },
     [router.query]
--- a/components/[pageId]/DatabasePage/DatabasePage.tsx
+++ b/components/[pageId]/DatabasePage/DatabasePage.tsx
@@ -12,7 +12,7 @@ import {
   setCurrent as setCurrentBoard
 } from 'components/common/BoardEditor/focalboard/src/store/boards';
 import { useAppDispatch, useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
-import { initialReadOnlyLoad } from 'components/common/BoardEditor/focalboard/src/store/initialLoad';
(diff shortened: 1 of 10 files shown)


components/[pageId]/DatabasePage/DatabasePage.tsx
```diff
@@ -40,9 +40,11 @@ interface Props {
 export function DatabasePage({ page, setPage, readOnly = false, pagePermissions }: Props) {
   const router = useRouter();
   const board = useAppSelector(getCurrentBoard);
-  const [currentViewId, setCurrentViewId] = useState<string>(router.query.viewId as string);
-  const activeView = useAppSelector(getView(currentViewId));
+  const [currentViewId, setCurrentViewId] = useState<string | undefined>(router.query.viewId as string | undefined);
   const boardViews = useAppSelector(getCurrentBoardViews);
+  // grab the first board view if current view is not specified
+  const activeView =
+    typeof currentViewId === 'string' ? boardViews.find((view) => view.id === currentViewId) : boardViews[0];
```
Comment (mattcasey) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:47]: I had to do this logic in a couple places... "find current view by id OR just pick the first one..." because the
states of currentViewId, activeView, and boardViews sometimes get out of sync while transitioning between two database pages that use the same components. I can think of a few long-term fixes but not sure if they'd work:
1) move currentViewId/activeView into Redux state, where it used to live. I moved it out when I added embedded databases so we could have multiple views on the page at a time, but maybe we could keep a map of current/active views inside the state. This seems like the simplest..
2) wrap the data from Redux in a hook state, so that it is updated in the same digest as currentViewId/activeView
3) get rid of Redux. Don't think we have time for that right now

PR #1559, opened 2023-01-23
use same logic to generate paths for posts and post categories

mattcasey:
Bug: We don't parse out "?" in titles at the moment

https://app.charmverse.io/charmverse/page-49366552253645923?viewId=9937b355-8c33-4069-81b5-640e3c561bdb&cardId=fd8ed5c7-69cc-4706-ba61-31a4106b62fd


mattcasey:
i'm merging this as it's a bug in prod

PR #1566, opened 2023-01-24
Add "Undo" to the "..." menu on pages

Devorein:


mattcasey (approved):

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -89,6 +88,13 @@ export default function Header({ open, openSidebar }: HeaderProps) {
 
   const isBountyBoard = router.route === '/[domain]/bounties';
 
+  const undoEvent = useMemo(() => {
+    if (basePage) {
+      return new CustomEvent('undo', { detail: { pageId: basePage.id } });
+    }
+    return null;
+  }, [basePage]);
```
mattcasey, line 96: i think u could just depend on basePage.id, not too big a deal tho

components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -108,6 +111,24 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
...
+      if (editor) {
+        const detail = (event as CustomEvent).detail as { pageId: string } | null;
+        if (detail && detail.pageId === pageId) {
+          undo()(editor.view.state, editor.view.dispatch);
+        }
+      }
+    }
+
+    if (editorRef && editorRef.current && editor) {
+      editorRef.current.addEventListener('undo', listener);
```
mattcasey, line 125: I think a more explicit event name would be better, and instead of a magic string, have a constant so that we avoid any possible typos or mismatch in name

mattcasey (approved):
left some comments

components/common/PageLayout/components/Header/Header.tsx
Devorein: Ah I don't know why I keep making the same mistake :)

components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
Devorein: I agree. I will add the event name inside a const file and call it `editor-undo`

PR #1572, opened 2023-01-24
add a script to review page history

mattcasey:
Sample output: 
image


mattcasey:
> Made a small tweak, which is to add some defaulting for the date ranges, so script can work without having to explicitly specify dates

Thanks! I thought this could be improved as well

PR #1573, opened 2023-01-25
Fix filter/sort on views, and a bunch of minor UI polish for views

mattcasey:
Clicking Filter or Sort ends up with dropdown being hidden:
image

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
@@ -40,7 +40,7 @@
     position: sticky;
     left: 0;
     background: rgb(var(--center-channel-bg-rgb));
-    z-index: 1;
+    z-index: 2;
   }
 
   > .container-container {
@@ -75,13 +75,14 @@
   .view-actions {
     display: flex;
     transition: opacity 0.2s ease-in-out;
+    opacity: 0;
   }
 
   &.embedded-board .view-actions {
     opacity: 0;
   }
 
-  &.embedded-board:hover .view-actions {
+  &.BoardComponent:hover .view-actions {
     opacity: 1;
   }
 }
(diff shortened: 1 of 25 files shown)


components/common/BoardEditor/focalboard/src/components/centerPanel.scss
```diff
@@ -40,7 +40,7 @@
     position: sticky;
     left: 0;
     background: rgb(var(--center-channel-bg-rgb));
-    z-index: 1;
+    z-index: 2;
```
mattcasey, line 43: @msadura this is the fix for getting filter/sort to appear. It seems fine, I can't find any dropdowns that are negatively affected by this change? I had replaced the focalboard menu with MUI's popover so that I could use a React portal, but this has the downside that the view header would disappear because it relies on mouse hover to be visible. (I have copied Notion's UI to hide them by default until hover, makes for a cleaner page )

components/common/BoardEditor/focalboard/src/components/viewHeader/viewHeader.tsx
```diff
@@ -109,7 +110,11 @@ function ViewHeader(props: Props) {
 
       {/* add a view */}
 
-      {!props.readOnly && views.length <= maxTabsShown && props.addViewButton}
+      {!props.readOnly && views.length <= maxTabsShown && (
+        <Box className='view-actions' pt='4px'>
```
Devorein, line 114: Should we hide the add view button unless hovered? Notion always shows it.

Devorein (changes requested):
Looks good. But sometimes I can't create a view, the whole screen gets stuck (Fixes on reload). Try creating a new board and keep on adding views.

components/common/BoardEditor/focalboard/src/components/centerPanel.scss
msadura: yeah I cannot find anything broken, so probably I went too far changing those zindexes

msadura (approved):

mattcasey:
> Looks good. But sometimes I can't create a view, and the whole screen gets stuck (Fixes on reload). Try creating a new board and keep on adding views. For some reason, I can't reproduce it on staging.

I think this is just it waiting to 'register' the new view, doing all the digest, etc. it has to do. It happened to me, but i waited a minute and was able to continue adding stuff.

image

Devorein (approved):

PR #1574, opened 2023-01-25
Fix space btn hover

valentinludu:


msadura:
IMG_0477

unfortunately it is still broken on mobile devices

valentinludu:
> 
> unfortunately it is still broken on mobile devices

Fixed!

msadura (approved):

PR #1576, opened 2023-01-25
Feat/mobile sidebar

msadura:
- mobile will not use "temporary" (mui variant) drawer, which opens on top of the page without messing the content
- overall updates to mobile sidebar styling - whole content is scrollable, text and buttons are a bit bigger
- drag & resize sidebar on desktop

https://user-images.githubusercontent.com/5198394/214570143-9d651cad-5be8-42ff-a0fc-c122b0f2bfe5.mp4


hooks/useMobileSidebar.ts
```diff
@@ -0,0 +1,9 @@
+import { useTheme } from '@emotion/react';
+import { useMediaQuery } from '@mui/material';
+
+export function useMobileSidebar() {
+  const theme = useTheme();
+  const isMobileSidebar = useMediaQuery(theme.breakpoints.down('md'));
```
valentinludu, line 6: Would be simpler to:
`const isMobileSidebar = useMediaQuery((theme) => theme.breakpoints.down('md'));

And since you have a one liner, it would make sense to remove this file and use it where you need it in the component.

Also the path import would be recommended since we are trying to use it in all our app. -> import useMediaQuery from '@mui/material/useMediaQuery';

valentinludu (approved):

hooks/useMobileSidebar.ts
msadura: sine sidebar is realted rto specific breakpoint I wanted to have explicit function to make it easy to reause in other places without thinking about which breakpoint is used for sidebar related stuff

PR #1578, opened 2023-01-25
Fixed open in new tab in page actions

Devorein:


motechFR (approved):
Tested and works well, also love to see a negative diff PR!

PR #1583, opened 2023-01-25
encode url for redirect

mattcasey:
We were unable to redirect a user, because a returnUrl was set with Japanese characters:
```
/henkaku-ochakai/forum/post/visionについて_ld9xxajl
```
I considered encoding the returnUrl when it was set, but I think this solution handles more cases.


mattcasey:
Deploying because it fixes a pretty critical bug for japanese users, but adding a TODO to write an e2e test. (Looks like e2e tests for tomorrow :)
Example post:
https://pr-1583-encode-redirects.charmverse.co/original-sharding-sturgeon/forum/post/visionについて_ld9xxajl_ldcnc6e3

PR #1589, opened 2023-01-26
Edit profile from member directory

Devorein:


components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -33,8 +34,11 @@ const StyledLink = styled(Link)`
     opacity: 0.8;
   }
   position: relative;
+  cursor: pointer;
 `;
 
+const StyledBox = StyledLink.withComponent(Box);
```
mattcasey, line 40: if you're not using Box properties, is this even needed?

mattcasey (approved):

components/common/Icons/hoverIconsStyle.ts
```diff
@@ -32,7 +32,7 @@ export const hoverIconsStyle =
     top: 0;
     right: 0;
     margin: ${marginForIcons ? theme.spacing(1) : 0};
-    z-index: 0;
+    z-index: 1;
```
mattcasey, line 35: this has me on edge lol.. fixing all the z-index issues w/views right now. maybe a comment on why it has to be 1?

mattcasey (approved):

PR #1590, opened 2023-01-26
Feat/mobile updates v1

msadura:
Mobile updates:
https://app.charmverse.io/charmverse/page-02206983794527928
https://app.charmverse.io/charmverse/page-8443765873921243
https://app.charmverse.io/charmverse/page-4322475312816265
https://app.charmverse.io/charmverse/page-45242899509696066

Screenshot 2023-01-26 at 17 26 53
Screenshot 2023-01-26 at 17 25 43
busercontent.com/5198394/214891282-4fe342ce-c802-4bde-b1fd-6f0992ab5966.png">
Screenshot 2023-01-26 at 17 26 19

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -46,7 +46,7 @@ export const Container = styled(({ fullWidth, ...props }: any) => <Box {...props
   margin: 0 auto ${({ top }) => top + 100}px;
   position: relative;
   top: ${({ top }) => top}px;
-  padding: ${({ theme }) => theme.spacing(0, '20px')};
+  padding: 0 40px 0 30px;
 
   ${({ theme }) => theme.breakpoints.up('sm')} {
     padding: 0 80px;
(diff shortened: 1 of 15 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -78,6 +80,7 @@ function DocumentPage({ page, setPage, insideModal, readOnly = false, parentProp
   const { draftBounty } = useBounties();
   const { currentPageActionDisplay } = usePageActionDisplay();
   const { editMode, setPageProps } = usePrimaryCharmEditor();
+  const isSmallScreen = useMediaQuery((theme: Theme) => theme.breakpoints.down('lg'));
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:83]: in general if possible it's better to use breakpoints.up, so we are always thinking the default is for mobile

Comment (mattcasey) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:83]: FWIW i always avoided useMediaQuery out of concerns this triggered awatched on window resize, b ut apparently it's using a native method called window.matchMedia! We have a few resize listeners including useWindowSize that could be updated to use that instead

mattcasey (approved):
Request for future: always use breakpoints.up() when possible. It's easier to keep in your head if default is always mobile

PR #1594, opened 2023-01-27
Fix editing table columns

mattcasey:
We were using useMemo which caused the click handlers to be using stale state


mattcasey:
@Devorein i think this was your code, FYI

PR #1595, opened 2023-01-27
Database table rows handlebars to duplicate, remove and move

Devorein:


motechFR (changes requested):
@Devorein Core functionality works.

One fix needed for page permissions. If I only have board access as a viewer, I'm still able to create a card.

Once the card opens, I can only view it (which is expected as it inherits board permissions)

We shouldn't allow this action. For the "duplicate page" button, can you please check getPagePermissions for the board?

Also, it's good to add a Tooltip explaining why the user can't perform this action which is disabled.

Delete: "You don't have the permissions to delete this page"

Duplicate: "You don't have the permissions to create new entries on this database"


Screenshot 2023-01-27 at 15 21 49

motechFR (approved):
Works better now, hiding handlebars is fine :)

PR #1597, opened 2023-01-27
Update space tests

motechFR:
This pull request makes updating a space in space settings more reliable

- Extracted updateSpace from API to a lib method which only updates name, domain and spaceImage (also stops us sending arbitrary data from req.body straight to Prisma)
- Ensured the API and lib are resilient against inputs with additional unexpected values
- Added an e2e test for changing the space settings form

### DIFF
--- a/lib/spaces/__tests__/updateSpace.spec.ts
+++ b/lib/spaces/__tests__/updateSpace.spec.ts
@@ -1,0 +1,163 @@
+import type { Space, User } from '@prisma/client';
+import { v4 } from 'uuid';
+
+import { prisma } from 'db';
+import { DuplicateDataError, InvalidInputError } from 'lib/utilities/errors';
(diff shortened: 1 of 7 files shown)


lib/spaces/__tests__/updateSpace.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+  secondUser = await prisma.user.create({
+    data: {
+      username: 'secondUser'
+    }
+  });
+});
+
+afterAll(() => {
+  jest.unmock('lib/metrics/mixpanel/updateTrackGroupProfile');
```
mattcasey, line 30: why are you mocking this? the test runner does not have access to what happens inside the service

mattcasey (approved):

PR #1598, opened 2023-01-27
Implement page export and add default page to all new spaces

motechFR:
Adds our new getting started page as the default everywhere


mattcasey (approved):

Devorein (approved):

__e2e__/sessions/login.spec.ts
```diff
@@ -40,6 +40,21 @@ test('login - redirects a logged in user on the site to their workspace', async
 
   await loginPage.goto();
 
+  // Prepare pages as a glob OR pattern since we might land on any of them
```
mattcasey, line 43: what does this mean? why is it not deterministic which page we will land on?

mattcasey: one additional thing is that I believe it's intentional we were checking for title to appear inside waitForWorkspaceLoaded(), since just checking the URL doesn't necessarily mean you got to the content.

lib/spaces/createWorkspace.ts
```diff
@@ -121,6 +123,41 @@ export async function createWorkspace({ spaceData, userId, createSpaceOption, ex
     });
   }
 
+  await prisma.page.create({
+    data: {
+      ...gettingStartedPage,
+      content: gettingStartedPage.content as Prisma.InputJsonValue,
```
mattcasey, line 129: This is duplicating a lot of logic from convertJsonPagesToPrisma. If you extracted the logic to read the filesystem so that it just took an array of parsed content, then you could just prepend this page as input.

__e2e__/sessions/login.spec.ts
motechFR: Thanks for the tip on page title check, makes sense.

I did this more flexible approach as we might have multiple root pages. However, I can totally see the case for explicitly testing that Getting Started is in fact the first page we arrive at.

lib/spaces/createWorkspace.ts
motechFR: I think we might even want to deprecate convertJsonPages in favour of importWorkspacePages, since we now do exports to a single JSON file instead of a recursive directory tree.

I've captured this task to our backlog

https://app.charmverse.io/charmverse/page-4620858382907256

PR #1600, opened 2023-01-27
prevent paste commands from escaping database

mattcasey:
User was trying to paste a URL into the table of an embedded DB, and it would either paste in the page document outside the table or jsut delete the table :(

https://www.loom.com/share/e24f7ea63e1841a1a02e252e4f4046d0


motechFR (approved):
Tested and worked great

Created an inline database in a page

Added url field

Pasted with keyboard

Also pasted with right click ==> paste

PR #1602, opened 2023-01-28
Mobile dialog

msadura:
I will need  dialog that will be reusable on mobile for different kinds of content. I made this reusable component + connected it with our current `Modal` for easy use. I added `mobileDialog` prop to our `Modal` and you can enable mobile behavior and animation with a single prop. Added example use to create sapce modals.
Note that `Modal` will work as it was for bigger screens.

https://user-images.githubusercontent.com/5198394/215264159-b3e13419-bb67-4a2d-a054-ec1f90f519b5.mp4

This is how it would look like if you pass `title` as `Modal` prop:

Screenshot 2023-01-28 at 12 32 01


Devorein (approved):
Feels really slick 🚀

PR #1607, opened 2023-01-30
Bug fix: document selected by default causes unexpected behavior

mattcasey:
I introduced this bug when I added a selection to the doc when we got the first handshake with web sockets... basically, the selection type was "all" by default.

I was trying to remove some weirdness where blocks would be highlighted by default, but a better fix was to disable focus on read-only content, and also simplify the focusOnInit logic for documents. It looks like Notion never applies focus on the document by default, just the page title if anything. I'd like to revisit the decision to put smart logic for autoFocus (aka focusOnInit) inside ReactEditor in the future

### DIFF
--- a/components/[pageId]/DocumentPage/components/PageTitleInput.tsx
+++ b/components/[pageId]/DocumentPage/components/PageTitleInput.tsx
@@ -78,7 +78,7 @@ export function PageTitleInput({ value, updatedAt: updatedAtExternal, onChange, 
       value={title}
       onChange={_onChange}
       placeholder='Untitled'
-      autoFocus={!readOnly && !isTouchScreen()}
+      autoFocus={!value && !readOnly && !isTouchScreen()}
       multiline
       variant='standard'
       onKeyDown={(e) => {
--- a/components/common/CharmEditor/components/fiduswriter/collab/doc.ts
+++ b/components/common/CharmEditor/components/fiduswriter/collab/doc.ts
@@ -126,9 +126,7 @@ export class ModCollabDoc {
     const stateConfig: EditorStateConfig = {
       schema: this.mod.editor.schema,
       doc: stateDoc,
-      plugins: this.mod.editor.view.state.plugins.concat(plugins),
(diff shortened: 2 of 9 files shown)


components/[pageId]/DocumentPage/components/PageTitleInput.tsx
```diff
@@ -78,7 +78,7 @@ export function PageTitleInput({ value, updatedAt: updatedAtExternal, onChange,
       value={title}
       onChange={_onChange}
       placeholder='Untitled'
-      autoFocus={!readOnly && !isTouchScreen()}
+      autoFocus={!value && !readOnly && !isTouchScreen()}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageTitleInput.tsx:81]: notion doesn't focus the title input unless it's empty

components/common/CharmEditor/components/fiduswriter/collab/doc.ts
```diff
@@ -126,9 +126,7 @@ export class ModCollabDoc {
     const stateConfig: EditorStateConfig = {
       schema: this.mod.editor.schema,
       doc: stateDoc,
-      plugins: this.mod.editor.view.state.plugins.concat(plugins),
-      // if we dont include the selection, the first node gets selected for some reason
-      selection: this.mod.editor.view.state.selection
+      plugins: this.mod.editor.view.state.plugins.concat(plugins)
```
mattcasey, line 129: this was the bug

theme/@bangle.dev/styles.scss
```diff
@@ -151,16 +126,14 @@
...
-.ProseMirror-hideselection *::selection {
-  background: transparent;
-}
-
-.ProseMirror-hideselection *::-moz-selection {
-  background: transparent;
+.ProseMirror-hideselection *::selection,
+.ProseMirror-hideselection *::-moz-selection,
+.ProseMirror-hideselection .ProseMirror-selectednode {
+  background-color: transparent !important;
```
mattcasey, line 132: just some css cleanup

msadura (approved):

PR #1609, opened 2023-01-31
Update snapshot lib, better block validation

msadura:
We used outdated snapshot lib and it basically failed to fetch current block id for selected network. That resulted in wrong voting permissions.

- latest lib version fixes the issue
- I needed to add "app: ''" property to snapshot params to be able to create snapshot using lates lib version
- Added better validation - if in the future fetching latest block number will fail, it will set block number to 0 and trigger form error, rather than setting block to 1.

Screenshot 2023-01-31 at 12 30 45


components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
```diff
@@ -207,6 +205,14 @@ export default function PublishingForm({ onSubmit, page }: Props) {
         });
       }
 
+      if (!snapshotBlockNumber) {
+        throw new SystemError({
+          errorType: 'External service',
+          severity: 'warning',
+          message: 'Please provide correct snapshot block number.'
```
motechFR, line 212: Can we update this message to say "Could not retrieve block number from Snapshot. Please enter this manually"

Idea being to reflect that we expect this to be auto populated

motechFR (approved):
Tested on staging and works

components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
msadura: oh cool idea!

PR #1610, opened 2023-01-31
Fix dropping parent page onto child in nav tree

msadura:
fix for https://app.charmverse.io/charmverse/page-39935349937146336

and a bit of refactor

### DIFF
--- a/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
+++ b/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
@@ -238,8 +238,8 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
         <TreeNode
           key={item.id}
           item={item}
-          onDropChild={onDropChild}
-          onDropAdjacent={onDropAdjacent}
+          onDropChild={isFavorites ? null : onDropChild}
+          onDropAdjacent={isFavorites ? null : onDropAdjacent}
           pathPrefix={`/${router.query.domain}`}
           // pass down so parent databases can highlight themselves
           selectedNodeId={selectedNodeId}
--- a/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
+++ b/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
@@ -96,7 +96,7 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
 
         const droppedPage = _pages[droppedItem.id];
         if (!droppedPage) {
-          throw new Error('canot find dropped page');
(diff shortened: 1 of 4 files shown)


lib/pages/mapPageTree.ts
```diff
@@ -184,3 +184,22 @@ export function flattenTree<T extends PageNode = PageNode>(
 
   return flatNodes;
 }
+
+export function isParentNode({ node, child, items }: { node: PageNode; child: PageNode; items: PageNode[] }): boolean {
+  if (!node || !child || node.id === child.id || !child.parentId) {
```
mattcasey, line 189: if node or child is possibly undefined please update the type signature of the function

mattcasey: Also, if a page doesn't have a parentId, then that means it can only be a parent

components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
```diff
@@ -143,47 +72,60 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
...
 
-  const onDropAdjacent = useCallback((droppedItem: ParentMenuNode, containerItem: MenuNode) => {
-    if (droppedItem.id === containerItem?.id) {
-      return;
-    }
+  const isValidDropTarget = useCallback(
+    ({ droppedItem, targetItem }: { droppedItem: MenuNode; targetItem: MenuNode }) => {
+      // do not allow to drop parent onto children
+      return (
+        droppedItem.id !== targetItem?.id && !isParentNode({ node: droppedItem, child: targetItem, items: pagesArray })
```
mattcasey, line 79: instead of passing in pagesArray, you should pass in `pages` which is a map of the pages. this way when you look up parent nodes it will be O^n faster (did i use that right?)

components/common/PageLayout/components/PageNavigation/components/TreeRoot.tsx
```diff
@@ -0,0 +1,73 @@
+import { useTheme } from '@emotion/react';
+import TreeView from '@mui/lab/TreeView';
+import { useRouter } from 'next/router';
+import type { ComponentProps, ReactNode } from 'react';
+import { useEffect, useRef } from 'react';
+import { useDrop } from 'react-dnd';
+
+import type { MenuNode } from 'components/common/PageLayout/components/PageNavigation/components/TreeNode';
```
mattcasey, line 8: i think this would be a little better as a relative import :)

msadura:
cleaned up the code + fixed `canDrop` logic. thx for insights @mattcasey !

mattcasey (approved):
nice

PR #1611, opened 2023-01-31
Forum permissions

motechFR:
Provide permissions for the forum


pages/api/permissions/forum/compute-post-permissions.ts
```diff
@@ -0,0 +1,24 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
```
mattcasey, line 1: Camel case files please. This is a repeat offense :(

lib/permissions/forum/__tests__/listPostCategoryPermissions.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const foundPublicPermissions = await listPostCategoryPermissions({
+      resourceId: postCategory.id,
+      userId: undefined
+    });
+
+    expect(foundPublicPermissions.length).toBe(0);
+  });
+
+  it('should throw an error if the post category does not exist or the ID is invalid', async () => {
```
mattcasey, line 99: this is an example test i would not expect by default. testing garbage input is not very useful, since the db is bound to throw an error anyway, and we don't really care what type of error is thrown - that's helpful for debugging but has no user impact

__integration-tests__/server/pages/api/permissions/forum/compute-post-permissions.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+import request from 'supertest';
+
+import { computePostPermissions } from 'lib/permissions/forum/computePostPermissions';
+import { upsertPostCategoryPermission } from 'lib/permissions/forum/upsertPostCategoryPermission';
+import { baseUrl, loginUser } from 'testing/mockApiCall';
+import { generateRole, generateUserAndSpaceWithApiToken } from 'testing/setupDatabase';
+import { generateForumPost, generatePostCategory } from 'testing/utils/forums';
+
+describe('POST /api/permissions/forum/compute-post-permissions - Compute permissions for a forum post', () => {
+  it('should return computed permissions for a user and non user, and respond 200', async () => {
```
mattcasey, line 11: this should be split into two test blocks

lib/permissions/forum/__tests__/listPostCategoryPermissions.spec.ts
mattcasey: I also think since this method is just wrapped in an endpoint, this is a case where i would just put all the expects on a single test suite: [list-post-category-permissions.spec.ts](https://github.com/charmverse/app.charmverse.io/pull/1611/files#diff-a82a265f945412ce1bea7b52f8fb2cc609bd15934e95d23bf9bbbb87de345e65). You could call the endpoint w/the same garbage endpoints but expect certain status codes. And that way we have one single place where we test permissions, easier to maintain and fix. I dont find any benefit in arbitrarily splitting the expectations

__integration-tests__/server/pages/api/permissions/forum/compute-post-permissions.spec.ts
motechFR: +1 on this

lib/permissions/forum/__tests__/listPostCategoryPermissions.spec.ts
motechFR: My view on tests is that they should be done on the smallest allowable unit and closest to the source, as this composes upwards.

The API is just a wrapper on the lib, which is an architectural benefit. However, to test the lib as a developer, I prefer to just run an inline unit test, than have to spin up a server and run the test against the endpoint.

With these list endpoints, there is indeed a grey area, as a listing operation is inherently tied to authorisation.

As regards the listPostCategoryPermissions, it actually gets mutated after within the API, which manages the access control.

For the error codes, those are standardised by proxy, since our SystemErrors generate a response code mapping anyways.

mattcasey (approved):

lib/permissions/forum/__tests__/listPostCategoryPermissions.spec.ts
mattcasey: > My view on tests is that they should be done on the smallest allowable unit and closest to the source, as this composes upwards.

I think with unlimited time available, and for the developer trying to make changes, this is the best experience. I think I'm just saying that, if it were me, I find it easier to find, write, and maintain all my use cases in one place.  We are building an API, not a library. I'm not sure there has to be a hard rule though? now I am curious how we are going to organize code and tests with the permissions lib.

lib/permissions/requestOperations.ts
```diff
@@ -0,0 +1,56 @@
...
+type OperationRequest<R extends Resources = Resources> = PermissionCompute & {
+  resourceType: R;
+  operations: (R extends 'post_category' ? PostCategoryOperation : R extends 'post' ? PostOperation : never)[];
+};
+
+/**
+ * Request operations on a resource.
+ * Returns true if all requested operations are allowed, otherwise throws an error.
+ */
+export async function requestOperations<R extends Resources = Resources>({
```
mattcasey, line 21: naming suggestion: 'assert' to mirror the native node.js version of 'expect'. I wouldn't expect a getter/request type method to throw an error

prisma/schema.prisma
```diff
@@ -1136,3 +1140,52 @@ model GoogleCredential {
...
+enum PostCategoryPermissionLevel {
+  category_admin
+  moderator
+  member
+  guest
+  custom
+}
+
+model PostCategoryPermission {
+  id              String                       @id @default(uuid()) @db.Uuid
```
mattcasey, line 1173: @motechFR how does this work for a space that has no roles yet, but wants to use forums?

components/forum/components/ForumFilterListLink.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <MenuItem
+      dense
+      sx={{
+        display: 'flex',
+        flexDirection: 'row',
+        alignItems: 'center',
+        justifyContent: 'space-between'
+      }}
+      // This optimisation allows us to only compute category permissions when the user hovers over the category
+      onMouseEnter={() => (category && !categoryIdToQuery ? setCategoryIdToQuery(category.id) : null)}
```
mattcasey, line 51: why do we need this optimization? mouse event handlers are actually slow and expensive

lib/permissions/requestOperations.ts
motechFR: Thanks for this super detailed feedback!

I’d love to learn more

components/forum/components/ForumFilterListLink.tsx
motechFR: I had to add the create post field for the forum categories to make the flow for only displaying categories you can post to work well.

Your feedback makes sense, let’s move the pre computation up the stack, and return it inside the list forum categories endpoint

prisma/schema.prisma
motechFR: We can assign permissions to space or role.

i have updated create space method so that all the default post categories receive a space / member level permission

PR #1612, opened 2023-01-31
User mini public profile modal

Devorein:


components/profile/components/MemberMiniProfile/MemberMiniProfile.tsx
```diff
@@ -0,0 +1,177 @@
...
+          <DialogContent dividers>
+            <UserDetails user={user} readOnly />
+            {currentSpacePropertyValues && (
+              <Box my={3}>
+                <MemberPropertiesRenderer properties={currentSpacePropertyValues.properties} />
+              </Box>
+            )}
+
+            <Stack gap={3}>
+              <NftsList nfts={nfts} memberId={user.id} />
```
motechFR, line 157: Suggestion:
- Add text saying "add up to 5 NFTs here"
- Then only show one plus sign at a time

components/profile/components/MemberMiniProfile/MemberMiniProfile.tsx
```diff
@@ -0,0 +1,177 @@
...
+          />
+        )}
+        <Typography fontWeight={600}>Member details</Typography>
+      </MemberPropertiesPopup>
+    );
+  }
+
+  return (
+    <Dialog open onClose={onClose} fullScreen={fullScreen}>
+      {isLoading || !user ? (
```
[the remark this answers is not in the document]
motechFR: Another comment, when we fail to fetch NFTs, it would make sense to say "Could not get NFTs" instead of "You don't have NFTs"

components/profile/components/MemberMiniProfile/MemberMiniProfile.tsx
```diff
@@ -0,0 +1,177 @@
...
+            <UserDetails user={user} readOnly />
+            {currentSpacePropertyValues && (
+              <Box my={3}>
+                <MemberPropertiesRenderer properties={currentSpacePropertyValues.properties} />
+              </Box>
+            )}
+
+            <Stack gap={3}>
+              <NftsList nfts={nfts} memberId={user.id} />
+              <OrgsList orgs={orgs} memberId={user.id} />
```
motechFR, line 158: Some UI improvements
- Don't show the + sign if there are no organisations to add
- When I open the popup, it says "The organization you are looking for couldn't be found." whereas I just don't have any organisations to add

motechFR:
@Devorein , one idea we could also embed this functionality inside User Display component and pass a flag to activate it.

We already have linkToProfile option in userDisplay, so this could replace that. Worth discussing with team

motechFR (changes requested):
Works well, key change we need is remove dependency of the mini profile showing based on external API calls.

A more localised error state with more explicit error handling is better "we couldn't get X from Y service"  instead of "you dont have any NFTs"

PR #1613, opened 2023-01-31
Feat/mobile header

msadura:
Fix page header for smallers screens:
- added title ellipsis
- on mobile screens share and page mode buttons will switch to icons to take less space

Screenshot 2023-01-31 at 19 57 10

### DIFF
--- a/components/common/PageLayout/components/Header/components/ShareButton/ShareButton.tsx
+++ b/components/common/PageLayout/components/Header/components/ShareButton/ShareButton.tsx
@@ -1,4 +1,6 @@
-import { Box, Divider, Popover, Tooltip } from '@mui/material';
+import { IosShare, Share } from '@mui/icons-material';
+import type { Theme } from '@mui/material';
+import { Box, Divider, IconButton, Popover, Tooltip, useMediaQuery } from '@mui/material';
 import type { PageType } from '@prisma/client';
 import { bindPopover, usePopupState } from 'material-ui-popup-state/hooks';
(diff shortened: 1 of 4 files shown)


components/common/PageLayout/components/Header/components/ShareButton/ShareButton.tsx
```diff
@@ -20,6 +22,7 @@ function ShareButton({ headerHeight, pageId }: { headerHeight: number; pageId: s
     pageId ? `/api/pages/${pageId}/permissions` : null,
     () => charmClient.listPagePermissions(pageId)
   );
+  const isLargeScreen = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'));
```
mattcasey, line 25: Do you think there's a penalty for all these useMediaQuery calls? Some of these implementations could be a simple class to hide things, or the components could also use ``{ sx: {{ display: { xs: 'none', lg: 'flex' } }} }``, which would create CSS styles instead of re-rendering the component whenever the with changes, and adding a listener to the window

mattcasey (approved):

components/common/PageLayout/components/Header/components/ShareButton/ShareButton.tsx
msadura: if you want to change styles / hide or show component css is enough. But if you want to change some properties or basically display different component then this query is the only way. It will only re-render things when breakpoint changes, so this should not be an issue. I will try to find an information about its performance.

PR #1621, opened 2023-02-01
Update mobile comments sidebar to be displayed as modal

msadura:
Improved comments sidebar - it will be displayed as mobile dialog on small screens.

Additional notes:
- I realised it is a bit annoying to add imports and use `useMediaQuery((theme: Theme) => theme.breakpoints.up('md'));` all over the place, so I added small utility hooks `useLgScreen` , `useMdScreen` and `useSmallScreen` that will return boolean and provide an easy way to tell if we are in given breakpoint.
- @mattcasey you mentioned concerns about useMediaQuery performance - seems like it is not the case, is use `window.matchMedia` which is performant 
- https://mui.com/material-ui/react-use-media-query/
- https://webdevetc.com/blog/matchmedia-events-for-window-resizes/

https://user-images.githubusercontent.com/5198394/216006802-e5e760eb-4d0f-4953-8f9c-5cf791ecdbbf.mp4


mattcasey:
> @mattcasey you mentioned concerns about useMediaQuery performance - seems like it is not the case, is use window.matchMedia which is performant

Yea but also isn't it re-running all that Javascript when you resize the page?

msadura:
> 

Most of your points are valid. Basically to show / hide things css should be used whenever possible. There are 2 cases when we should use js:

- when there is a reason to remove component completely from JS (for example component is complex and it would be cheaper to remove it )
- component behavior might b a bit different - in my PR, I want to replace slider with modal - I am not sure what would happen if both would be rendered and one would have jsut `display: none` property, while both use same `open` state. I can verify that but overall it felt safer to have just one component rendered in  DOM

PR #1622, opened 2023-02-01
Do not cache profile and spaces endpoints

msadura:
https://discord.com/channels/894960387743698944/1070230124932370513

Fix weird case when profile was loaded from cache for logged-out user + remove cache from spaces


motechFR (approved):
Just tested, looks good

PR #1624, opened 2023-02-01
Add subscriptions to forum categories + notification on new post

mattcasey:
Some notes...

- Combined the queries for all forum 'tasks' and renamed to forum notifications and added a new 'forum' type. I don't know if it actually matters what the 'type' of UserNotification should be, but this seems like it was the original intent. We were essentially getting all comments and posts in all the spaces a user is in, might as well do it once. One day, we should rewrite to start from the posts/comments created by a user, I think


prisma/schema.prisma
```diff
@@ -1136,3 +1143,17 @@ model GoogleCredential {
   @@unique([userId, name])
   @@index([userId])
 }
+
+model NotificationPreferences {
```
msadura, line 1147: I think id would be good to name it `SpaceNotificationPreferences` - we have `UserNotification` already, so it would be easy to distinguish

prisma/schema.prisma
```diff
@@ -1136,3 +1143,17 @@ model GoogleCredential {
...
+  userId              String                   @db.Uuid
+  user                User                     @relation(fields: [userId], references: [id], onDelete: Cascade)
+  preferences         Json
+  spaceId             String                   @db.Uuid
+  space               Space                    @relation(fields: [spaceId], references: [id], onDelete: Cascade)
+  forumCategoriesMode CategorySubscriptionMode @default(blacklist)
```
msadura, line 1153: I am not sure how those notifications are going to be working, but from my understanding you will have 1 record with "blacklisted" ids, and one with "whitelisted" for each space?

msadura (approved):

prisma/schema.prisma
mattcasey: actually i renamed it to UserSpaceNotificationSettings :)

prisma/schema.prisma
mattcasey: mentioned this in #developers channel: https://discord.com/channels/894960387743698944/930253269459148900/1070437870969552997

The TL;DR is that you will say "sign me up for all new categories" (blacklist) or "only sign me up for the ones i listed" (whitelist). The UI will update the list of categories accordingly when it switches between the two

msadura (commented):
ah didn't realize this is wip

components/forum/components/CategoryNotificationToggle.tsx
```diff
@@ -0,0 +1,23 @@
+import styled from '@emotion/styled';
+import { IconButton, Tooltip } from '@mui/material';
+import { MdOutlineNotificationsNone, MdOutlineNotificationsOff } from 'react-icons/md';
+
+import { useForumCategoryNotification } from 'hooks/useUserSpaceNotifications';
+
+const OutlinedIconButton = styled(IconButton)`
+  border: 1px solid var(--input-border);
```
msadura, line 8: I was wondering what is this style for, and honestly I do not see any difference even if I remove this style 😅

hooks/useUserSpaceNotifications.ts
```diff
@@ -0,0 +1,44 @@
+import { useMemo } from 'react';
+import useSWRImmutable from 'swr/immutable';
+
+import charmClient from 'charmClient';
+import { useCurrentSpaceId } from 'hooks/useCurrentSpaceId';
+import { useSnackbar } from 'hooks/useSnackbar';
+import { useUser } from 'hooks/useUser';
+
+function useSpaceNotificationsData() {
```
msadura, line 9: Awesome, thx for putting loginc into separate hook

lib/userNotifications/spaceNotifications.ts
```diff
@@ -0,0 +1,56 @@
...
+  ]);
+  const subscriptionMap = categoryIds.reduce<Record<string, boolean>>((acc, categoryId) => {
+    const existsInList = settings.forumCategories.includes(categoryId);
+    acc[categoryId] = settings.forumCategoriesMode === 'whitelist' ? existsInList : !existsInList;
+    return acc;
+  }, {});
+
+  return {
+    forums: {
+      subscribeToNewCategories: settings.forumCategoriesMode === 'blacklist',
```
msadura, line 28: 'blacklist' / 'whitelist' logic is one thing I am not sure if I understand. Does user has the way of changing it right now or this is something to be implemented in the future? I understand overall idea, but I do not see where it is set to whitelist

pages/api/profile/space-notifications/set-forum-category.ts
```diff
@@ -0,0 +1,31 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
```
msadura, line 1: I would name this route simply `/notifications/forum-category` (or even just `/forum`), but you might have wider context here

msadura (approved):

msadura:
Looks really nice. Two things to mention:

1) I already stated it in comment - do we set categories mode to `whitelist` anywhere right now or it is only for the future?

2) Not sure if it just me, but I tried to look for  enable / disable notification button in category dropdown, at first I didn't realize I need to select specific category to view "bell" toggle. I would consider putting additional enable / disable notifications button in category dropdown:
Screenshot 2023-03-08 at 10 18 35

mattcasey:
> I already stated it in comment - do we set categories mode to whitelist anywhere right now or it is only for the future?

It is just for future. Actually right now you can only see subscriptions by going to each category

> Not sure if it just me, but I tried to look for enable / disable notification button in category dropdown

You're right.. going to see if i can just include that

PR #1625, opened 2023-02-02
Fix forum reply input

msadura:
https://discord.com/channels/894960387743698944/1070397116775669760

I am not sure if there was a reason for using `CharmEditor` instead of `InlineCharmEditor` but that + hacky styles broke this.

Screenshot 2023-02-02 at 12 15 28

### DIFF
--- a/components/forum/components/PostPage/components/CommentReplyForm.tsx
+++ b/components/forum/components/PostPage/components/CommentReplyForm.tsx
@@ -61,16 +61,18 @@ export function CommentReplyForm({
     <Stack gap={1}>
       <Box display='flex' gap={1} flexDirection='row' alignItems='flex-start'>
         <UserDisplay user={user} hideName={true} />
-        <InlineCharmEditor
-          colorMode='dark'
-          style={{
-            minHeight: 100
-          }}
-          key={editorKey}
-          content={postContent.doc}
-          onContentChange={updateCommentContent}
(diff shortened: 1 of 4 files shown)


msadura:
+ update create post layout:
Screenshot 2023-02-02 at 12 53 15

motechFR:
@msadura i think this would be a good e2e test

// setup
- create space
- generate post category
- Generate forum post

// assertions
- Navigate to space as non admin user
- go to forum
- Click category, await category url
- Click first post
- click open as page
- Await post url
- click reply
- enter text in charmeditor
- Click reply
- await API call with 200
- Expect a post comment component with inner text matching the given input

mattcasey (changes requested):
we need to use CharmEditor which has all the components we support

components/forum/components/PostPage/components/CommentReplyForm.tsx
```diff
@@ -61,16 +61,18 @@ export function CommentReplyForm({
...
-          style={{
-            minHeight: 100
-          }}
-          key={editorKey}
-          content={postContent.doc}
-          onContentChange={updateCommentContent}
-          placeholderText='What are your thoughts?'
-        />
+        <Box width='calc(100% - 48px)'>
+          <InlineCharmEditor
```
mattcasey, line 65: pretty sure this is supposed to be CharmEditor as well, actually

PR #1626, opened 2023-02-02
feat: fix SQS call for webhook worker

Rieranthony:


Rieranthony:
This implementation has been tested end-to-end and works completely (only for one event, which is `comment.create` 🫡)

mattcasey (approved):

PR #1628, opened 2023-02-02
add anonymous flag to events in mixpanel

mattcasey:


motechFR (approved):
Works as is for client-initiated requests

PR #1631, opened 2023-02-02
Fix/comment sidebar

msadura:
I am really happy with this fix + refactor of COmmentSIdebar logic :P This component was broken mostly due to:

- it was re-rendering like hundreds of time when hovering over CharmEditor (added memorization)
- because of the fact we created new reference to threads with `Object.values(threads).filter(isTruthy)` it was triggering `useEffect` responsible for scrolling into a comment on each render - as I mentioned renders were triggering like crazy and that is why scrolling acted like 💩 
- it is based on selecting document element which cannot be trusted in react ;p

I fixed reported issues in:
https://discord.com/channels/894960387743698944/1067757995431297034
https://discord.com/channels/894960387743698944/1070464508440612865

+ I added really interesting gooks when you need to wait for DOM element based on native js query selector. It uses MutationObserver https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver which is nice because:
- it is native api
- does not need setTimeouts
- I baked it into react hook, logic is triggered in useEffect so it will not break in ssr


msadura:
https://user-images.githubusercontent.com/5198394/216439894-b09ec910-66b0-498d-9d63-fbc82bbfd4aa.mp4

hooks/useDOMElement.ts
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+export function useDOMElement(selector?: string | null) {
```
mattcasey, line 3: 1) remove the listener when the hook is unmounted (in callback of useEffect)
2) this hook is freaking scary, if we put it a few places it means we have mutation observors running DOM queries on every DOM change, no guarantee they will ever trigger and disconnect. At minimum it needs a big warning at the top like DO NOT USE except in these cases
3) please cache results from querySelector
3)

PR #1632, opened 2023-02-02
set autoGenerated on template pages

mattcasey:


lib/spaces/__tests__/createWorkspace.spec.ts
```diff
@@ -222,6 +222,18 @@ describe('createWorkspace', () => {
     }
   });
 
+  it('should set autoGenerated to true for all pages', async () => {
```
motechFR, line 225: Thanks for adding this in!

lib/spaces/createWorkspace.ts
```diff
@@ -114,7 +114,7 @@ export async function createWorkspace({ spaceData, userId, createSpaceOption, ex
     });
     await prisma.$transaction([
       ...seedPagesTransactionInput.blocksToCreate.map((input) => prisma.block.create({ data: input })),
-      ...seedPagesTransactionInput.pagesToCreate.map((input) => createPage({ data: input }))
+      ...seedPagesTransactionInput.pagesToCreate.map((input) => createPage({ data: { ...input, autoGenerated: true } }))
     ]);
   } else if (createSpaceOption === 'templateNftCommunity') {
     await importWorkspacePages({
```
motechFR, line 120: I think we should also make sure that importWorkspacePages sets "autogenerated: true" (I would add the test on that method rather than here)

It would also be good to have another unit test on create space that looks same as what you just added, but passing the "templateCommunity" as the template option

motechFR (approved):

PR #1634, opened 2023-02-03
Fix embeds crash

msadura:
https://discord.com/channels/894960387743698944/1071053289166221403

Fix embeds crash due to breaking rule of hooks. I really think we should consider turning on eslint for that at some point of time :D

Screenshot 2023-02-03 at 14 54 54


mattcasey:
> Fix embeds crash due to breaking rule of hooks

In this case, does rule of hooks know that useEffect is called before a return statement? it wasn't the dependencies right

mattcasey (approved):

PR #1636, opened 2023-02-04
Test: uninstall ts-jest

mattcasey:


motechFR (approved):
Cool idea; our CI went down by an extra 1 or 2 minutes on top of existing matrix improvements

mattcasey:
> Cool idea; our CI went down by an extra 1 or 2 minutes on top of existing matrix improvements
@motechFR i thought you might find it interesting... supposedly, SWC is sooo much faster than ts-jest, but it turns out that the default config from next/jest  uses next's swc compiler already, we were just applying _both_ ts-jest and swc to Typescript files in the 'transform' we were passing to jest. I'm not sure it has an impact, TBH our tests seem to take +1/-1 when I do retries. But this at least simplifies our config and it's one less dependency to install 🤷

PR #1637, opened 2023-02-05
Support og meta data for profiles

Devorein:


pages/u/[userPath].tsx
```diff
@@ -1,23 +1,45 @@
...
-  const router = useRouter();
-
-  const { data: user, error } = useSWR(`users/${router.query.userPath}`, () =>
-    charmClient.getUserByPath(router.query.userPath as string)
-  );
+export const getServerSideProps: GetServerSideProps<{ user: PublicUser | null }, { userPath: string }> = async (
+  context
+) => {
+  const params = context.params;
+  const origin = process.env.NODE_ENV === 'development' ? 'http://localhost:3000' : 'https://app.charmverse.io';
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:18]: Please use `baseUrl` from config/constants

pages/u/[userPath].tsx
```diff
@@ -1,23 +1,45 @@
...
+  try {
+    if (params?.userPath) {
+      const user = await fetch<PublicUser>(`${origin}/api/public/profile/${params.userPath}`);
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:21]: the server should not be querying itself

Comment (mattcasey) [reply pages/u/[userPath].tsx:21]: This is also an extremely expensive query - please don't call any 3rd party APIs when we're just rendering on the server. Information in the DB is enoguh for metadata

pages/u/[userPath].tsx
```diff
@@ -1,23 +1,45 @@
...
+      return {
+        props: {
+          user
+        }
+      };
+    }
+    return {
+      props: {
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:29]: this should do a redirect, not pass 'null' to the frontend

pages/u/[userPath].tsx
```diff
@@ -1,23 +1,45 @@
...
+        user: null
+      }
+    };
+  } catch (err) {
+    return {
+      props: {
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:35]: this should send user to an error page and log the error

mattcasey (changes requested):
see comments

pages/u/[userPath].tsx
Comment (Devorein) [reply pages/u/[userPath].tsx:18]: Thanx I totally forgot that env variable exists 🤦🏼‍♂️

pages/u/[userPath].tsx
Comment (Devorein) [reply pages/u/[userPath].tsx:21]: Got it. Thanks.

pages/u/[userPath].tsx
```diff
@@ -1,43 +1,123 @@
...
+        : { path: userPath };
 
-  const { data: user, error } = useSWR(`users/${router.query.userPath}`, () =>
-    charmClient.getUserByPath(router.query.userPath as string)
-  );
+      const user = await prisma.user.findFirst({
+        where: condition,
+        include: {
+          profile: true
+        }
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:37]: We should filter out only the fields we want because this is a public api... right now you're returning the user's notification email which lives in the user table

pages/u/[userPath].tsx
```diff
@@ -1,43 +1,123 @@
...
+  const { data: user, error } = useSWR(`public-profile-${userPath}`, () => charmClient.getUserByPath(userPath));
 
+  useEffect(() => {
+    let pageTitle = '';
+    if (publicUser) {
+      pageTitle = publicUser.username || '';
+    }
     setPageTitle(pageTitle);
-  }, [user]);
+  }, [publicUser]);
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:86]: publicUser never changes, i don't think u need to use useEffect here?

pages/u/[userPath].tsx
```diff
@@ -1,43 +1,123 @@
...
+        {publicUser.avatar && <meta property='og:image' content={publicUser.avatar} />}
+        {publicUser.avatar && <meta property='twitter:image' content={publicUser.avatar} />}
+
+        <meta property='twitter:title' content={`${publicUser.username} | Profile`} />
+        <meta
+          property='twitter:description'
+          content={publicUser.profile?.description ?? `Check out ${publicUser.username}'s public CharmVerse profile`}
+        />
+      </Head>
+      <PublicProfile readOnly={true} user={user ?? publicUser} />
```
Comment (mattcasey) [inline pages/u/[userPath].tsx:118]: it looks like you could actually do the api call for user profile _inside_ public profile now, you're not even using `user` in this component anymore

mattcasey (changes requested):
make sure we dont send private fields from user table

pages/u/[userPath].tsx
Comment (Devorein) [reply pages/u/[userPath].tsx:37]: Correct, thank you. Learned a lot about next.js pre-fetching mechanism through this PR.

pages/u/[userPath].tsx
Comment (Devorein) [reply pages/u/[userPath].tsx:86]: `setPageTitle` is a state setter, if we don't put it inside an `useEffect` then it would cause an infinite re-render. But more importantly, it could be null if the user doesn't exist. `publicUser: PublicUser | null`

pages/u/[userPath].tsx
Comment (Devorein) [reply pages/u/[userPath].tsx:118]: The reason we have a separate user prop is for passing current user when looking at you profile.

We use the current user when looking at your profile
image

We fetch the user based on query `userPath` here
image

image

I can move the user fetching api call inside but that would also make a request when viewing your own profile. Plus it wont have the necessary fields required to distinguish between public and current user profile so you wont even be able to edit it. Maybe I can completely remove using `useSWR` inside `[userPath]` since it will be pre-fetched.

PR #1639, opened 2023-02-05
Request user email when signing up to a new space

Devorein:
1. We want users to provide their email when joining a new space if they haven't already
2. If they already have it, we only show the onboarding modal
3. There is "Skip" and "Next" for the email form, both of which take you to the next step (filling your space properties)
@msadura Thanks for the reminder to always add PR context 🙏🏼


msadura:
can you please provide context?

msadura:
1) When you create your first space this modal appears twice (I left the email empty and pressed next)

2) Its UI is different from other modals. We should standardize how the modal UI looks. THis is a part of onboarding so it should look more like it (for example having a header with separator)

Screenshot 2023-02-13 at 09 41 23
Screenshot 2023-02-13 at 09 41 13

ccarella:
I agree that this should be in the same modal.

__e2e__/sidebarAddNewSpace.spec.ts
```diff
@@ -57,6 +57,12 @@ test.describe.serial('Add a new space from sidebar and load it', async () => {
     const createdSpace = (await response.json()) as Space;
     await page.waitForURL(`**/${createdSpace.domain}`);
 
+    const memberEmailInput = await page.locator('data-test=member-email-input >> input');
```
motechFR, line 60: I think you can skip awaiting the locator, since it's a reference to a dom element we'll find.

We would just await the actions (like **.fill** or expect(x).toBeVisible() ) where in this case, playwright engine awaits for element to be ready until proceeding with action

__e2e__/sidebarAddNewSpace.spec.ts
```diff
@@ -57,6 +57,12 @@ test.describe.serial('Add a new space from sidebar and load it', async () => {
...
+    await memberEmailInput.fill('john.doe@gmail.com');
+
+    const memberEmailNextButton = await page.locator('data-test=member-email-next');
```
motechFR, line 63: Would it make sense to move these to a standard po class (the ones in our utilites)?

pages/api/permissions/index.ts
```diff
@@ -63,6 +64,7 @@ async function addPagePermission(req: NextApiRequest, res: NextApiResponse<IPage
 
   const createdPermission = await prisma.$transaction(
     async (tx) => {
+      console.time('01-tx-----');
```
motechFR, line 67: Please remember to clean up console.time and timeEnd before merging

components/members/MemberEmailForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useForm } from 'react-hook-form';
+import * as yup from 'yup';
+
+import charmClient from 'charmClient';
+import FieldLabel from 'components/common/form/FieldLabel';
+import { useUser } from 'hooks/useUser';
+import debounce from 'lib/utilities/debounce';
+
+export const schema = yup.object({
+  email: yup.string().ensure().trim().email().max(50)
```
motechFR, line 15: I think the max(50) is a bit of an overkill. We should be fine just with the combo of .trim and .email

components/members/MemberEmailForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+  );
+
+  return (
+    <Box mt={2}>
+      <Typography fontWeight='bold'>Email</Typography>
+      <Typography variant='subtitle2' mb={1}>
+        CharmVerse can use your email address to let you know when there is a conversation or activity you should be
+        part of.
+      </Typography>
+      <TextField
```
motechFR, line 60: Can you add a placeholder email? All other items in member onboarding form have one, that way it looks consistent

components/profile/components/MemberMiniProfile/MemberMiniProfile.tsx
```diff
@@ -77,15 +78,18 @@ function MemberProfile({ title, member, onClose }: { title?: string; member: Mem
...
+            <UserDetails
+              sx={{
+                mt: 0
+              }}
+              // currentUser doesn't have profile thus is not considered as publicUser inside UserDetails
+              // giving the ability to update the profile properties
+              user={user.id === currentUser.id ? currentUser : user}
+              updateUser={setUser}
+            />
+            <MemberEmailForm />
```
motechFR, line 91: Check with product team, but I'd be +1 to move email up. Looks like a pretty important field, and I almost missed it on first test

mattcasey (approved):

PR #1640, opened 2023-02-06
Revert domain name of a space

valentinludu:
This migration file will rename a specific user's domain name back to what it was.

### DIFF
--- a/scripts/migrations/2023_02_06_updateRecoverCapitalSpaceDomain.ts
+++ b/scripts/migrations/2023_02_06_updateRecoverCapitalSpaceDomain.ts
@@ -1,0 +1,37 @@
+import { prisma } from 'db';
+
+
+export async function updateRecoverCapitalSpace() {
+  const defaultSpace = await prisma.space.findMany({
+    where: {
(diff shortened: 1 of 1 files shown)


Devorein (approved):

msadura (changes requested):
Honestly I do not feel like this script should belong to migration scripts. 
As far as I understand migrations scripts it is about changing db structure and adapting fields. This one is just a specific operation for specific user.
I do not see a reason for committing id and name of our user to our codebase. If you think this script can be useful for the future I would just name it "changeSpaceDomain", place it directly in scripts + remove user-specidfic data.

motechFR:
+1 on Marek's comment. It's important to not commit any PRs with actual UUIDs or domains.

In terms of the migration, this script is a one-off support task, so it belongs in scripts.

motechFR:
Santisation of the script:

amend data and then recommit witgh


git commit --amend -m "Update script"

git push --force

valentinludu:
Opened a PR to validate with the team.
Will run this locally. 
Closing this.

PR #1641, opened 2023-02-06
Fix/bookmark style

msadura:
Update styling of custom bookmark (twitter) to looks like iframely bookmarks:

Screenshot 2023-02-06 at 12 04 43
Screenshot 2023-02-06 at 12 03 41


motechFR (approved):
Tested and looks good.

Used twitter.com, and an individual tweet

PR #1642, opened 2023-02-06
Check strict rules for space domain name

valentinludu:
We need to restrict the user from typing special characters into domain name.


mattcasey (approved):

mattcasey (approved):
Approved, and also realized that we don't do any validation on the backend. That's important especially since we're allowing services to create spaces now from the API. I'm going to add a story in backlog

PR #1643, opened 2023-02-06
Default space full access permission for post categories

motechFR:


msadura (approved):

pages/api/spaces/[id]/post-categories/index.ts
```diff
@@ -49,9 +51,17 @@ async function createPostCategoryController(req: NextApiRequest, res: NextApiRes
...
-  const postCategory = await createPostCategory({
-    ...req.body,
-    spaceId: spaceId as string
+  const postCategory: PostCategory = await prisma.$transaction(async (tx) => {
+    const createdCategory = await createPostCategory({
+      name: req.body.name,
+      spaceId: spaceId as string,
+      tx
+    });
+    await assignDefaultPostCategoryPermissions({
```
Comment (mattcasey) [inline pages/api/spaces/[id]/post-categories/index.ts:60]: why don't we just put this inside createPostCategory, since it's supposed to be a default?

Comment (mattcasey) [reply pages/api/spaces/[id]/post-categories/index.ts:60]: it looks like we only use createPostCategory in one place

Comment (motechFR) [reply pages/api/spaces/[id]/post-categories/index.ts:60]: I did go with this approach initially, but it adds permissions logic to the lib/forums.

one great win on the latest iteration has been to have absolutely all permissions logic in lib / permissions / forums

Comment (mattcasey) [reply pages/api/spaces/[id]/post-categories/index.ts:60]: I don't see that as a huge benefit

Comment (mattcasey) [reply pages/api/spaces/[id]/post-categories/index.ts:60]: I don't really see that as a great win at the moment, permissions and categories don't exist without one another.. however since this may need to be an API call, I can see your point

mattcasey (approved):

pages/api/spaces/[id]/post-categories/index.ts
Comment (mattcasey) [reply pages/api/spaces/[id]/post-categories/index.ts:60]: > one great win on the latest iteration has been to have absolutely all permissions logic in lib 

There's definitely some benefits to that... I think it's fine in this case and doesn't need to be reworked.

On the other hand, I'd also consider that the category is tightly coupled and kind of useless in our system without permissions. And vice versa - permissions are tightly coupled to the entities, they even have different tables for different entities. We use createBounty and createPage to create permissions sometimes, and I think that's a good reason to have standard lib methods.

I can also imagine how this separation may be more useful when permissions is an API, but I'm not sure if we will be able to actually decouple the database tables, and we will still have to create permissions even if the space is a 'public space' so that it works when you make it private.

Comment (motechFR) [reply pages/api/spaces/[id]/post-categories/index.ts]: I think your proposed best of both worlds solution is wisest given current state and upcoming goals.

I've gone ahead and added a permissions assignment call from the lib/permissions-defined method into the lib/forums createPostCategory method.

I've also removed usage from createPostCategory from all unit + integration tests except its own, and replaced by generatePostCategory from testing/utils/forums, which only generates a category without permissions.

Here's the comment I added on the test for

  // This assertion is important because it ensures that consumer tests can set up the state of permissions in a deterministic way, without risk of side effects from updates to default permissions logic used in the real createPostCategory method.

PR #1645, opened 2023-02-06
FIxed readonly board and page issues

Devorein:
1. disable updating callout icon
2. disable deleting a bookmark
3. Undo for page actions (w/ view and comment permission)
4. Checking todo

### DIFF
--- a/components/common/CharmEditor/components/callout/components/Callout.tsx
+++ b/components/common/CharmEditor/components/callout/components/Callout.tsx
@@ -1,14 +1,14 @@
-import type { NodeViewProps } from '@bangle.dev/core';
 import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
-import { Box, Menu } from '@mui/material';
-import { alpha } from '@mui/system';
+import { IconButton, Menu } from '@mui/material';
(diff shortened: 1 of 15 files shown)


components/common/CharmEditor/components/callout/components/Callout.tsx
```diff
@@ -65,14 +70,10 @@ export default function Callout({ children, node, updateAttrs }: NodeViewProps &
...
-            justifyContent: 'center',
-            '&:hover': {
-              // Copied from IconButton.js in @mui/system
-              background: alpha(theme.palette.action.active, theme.palette.action.hoverOpacity),
-              transition: 'background 100ms ease-in-out'
-            }
+            justifyContent: 'center'
           }}
           onClick={handleClick}
+          disabled={readOnly}
```
motechFR, line 76: We should override the disabled styles for the callout as otherwise the emoji is greyed out

Screenshot 2023-02-06 at 17 30 31

motechFR:
We should also add readonly support for these plugins:

- CryptoPrice (Im able to change the currency pairs in readonly mode)

Screenshot 2023-02-06 at 17 47 35

PR #1647, opened 2023-02-06
Show mini profile when clicking username/avatar

Devorein:


mattcasey (approved):
Looks good... are there any cases where we wouldn't show a profile? I wonder if we even need the attribute

PR #1651, opened 2023-02-06
Fix issue of post not being editable on initial page load

motechFR:
We have an ongoing issue where CharmEditor readonly mode cannot be changed once the editor mounts. It needs to be remounted with a fresh state.

I've updated usePostPermissions hook to undefined initially, so we can know it's loading in consumer components.

In the post page, I then added a loading and error state.

### DIFF
--- a/__e2e__/forum/commentOnForumPost.spec.ts
+++ b/__e2e__/forum/commentOnForumPost.spec.ts
@@ -1,4 +1,5 @@
 import { expect, test as base } from '@playwright/test';
+import type { Post, PostComment, Space, User } from '@prisma/client';
 
 import { prisma } from 'db';
 import { upsertPostCategoryPermission } from 'lib/permissions/forum/upsertPostCategoryPermission';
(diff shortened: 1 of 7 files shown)


motechFR:
@mattcasey Id welcome your input on a way to target the contenteditable=true prosemirror div that hosts the content

msadura:
> 

I've been playing yesterday with e2e editor tests, and I can share my aprial work in https://github.com/charmverse/app.charmverse.io/compare/feat/editor-e2e-tests?expand=1

on a page with single Charm editor you can simply target it by `this.page.locator('div[contenteditable]');` This works for regular "document" page but I am not sure if that is enough for forum. Maybe editor should be wrapped in some sort of test-id and you can locate `contenteditable` child div?
Unless I misunderstood you problem 😅

motechFR:
Thanks @msadura for experimenting with this, I integrated part of your approach into the forum post page utils

msadura (approved):

PR #1653, opened 2023-02-07
Add user preferences and locale

msadura:
This PR introduces  hook for dates formatting. For now it is quite simple - return formated date using either locale from user preferences or default one (from system). 

NOTE: It does not handle switching locale in profile and saving it in `UserDetails` - I do not want to make one large PR, so I want to separate it from this one. Current formatting will work correctly with default system locale.


testing/normalizeWhitespace.ts
```diff
@@ -0,0 +1,4 @@
+export function normalizeWhitespace(str: string) {
```
msadura, line 1: This one is funny - tests on CI were failing, because `toLocaleString` uses non-breaking space in some cases and assertions were not working properly. This simply replaces any sort of space with "regular" space. Only for testing purposes.

mattcasey (approved):
looks good! tested in staging

PR #1654, opened 2023-02-07
fix validation for spaces on domain names

valentinludu:


mattcasey (approved):
just tested, i couldn't believe we don't check for this :/

PR #1658, opened 2023-02-07
Fix reordering of spaces when a new one is added

valentinludu:
The order of spaces jumps randomly when we create a space.


mattcasey (commented):
Is it possible to make the frontend more robust, so that a user can still sort spaces without having the backend updated? This seems very fragile, for example when you join a space, we need to make sure to update spacesOrder there as well.

Whereas, I would just assume if a space is not in 'spacesOrder', it hsould just be sorted at the bottom alphabetically until a user changes the order again.

valentinludu:
> Is it possible to make the frontend more robust, so that a user can still sort spaces without having the backend updated? This seems very fragile, for example when you join a space, we need to make sure to update spacesOrder there as well.
> 
> Whereas, I would just assume if a space is not in 'spacesOrder', it should just be sorted at the bottom alphabetically until a user changes the order again.

Right!

Found a better solution.

The actual problem was when the user did create/delete/join a workspace, the GET api/spaces was getting the spaces without the users order array.

mattcasey (approved):
much better thanks

PR #1666, opened 2023-02-07
Add page permissions index by id

motechFR:
We have an issue deleting many permissions by ID. This should speed things up


mattcasey:
Not sure this will help, but shouldn't hurt either. I think using @id on a field (making it primary) implicitly adds an index

PR #1667, opened 2023-02-07
Support requesting posts when no user is available

mattcasey:


motechFR (approved):
Works, thanks for the fix

PR #1675, opened 2023-02-08
Show mini profile when joining a space

Devorein:


__e2e__/sidebarAddNewSpace.spec.ts
```diff
@@ -57,6 +57,13 @@ test.describe.serial('Add a new space from sidebar and load it', async () => {
     const createdSpace = (await response.json()) as Space;
     await page.waitForURL(`**/${createdSpace.domain}`);
 
+    let memberProfileNftList = await page.locator('data-test=member-profile-nft-list');
```
motechFR, line 60: @Devorein Can you please write a new test in a file I suggest calling memberOnboarding?

We should keep our test assertions small and focused. Here we just want to focus on adding new spaces.

Your new test can focus on the member onboarding experience.

PR #1676, opened 2023-02-08
Convert Forum Post to Proposal

Devorein:


components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -289,6 +280,20 @@ export default function Header({ open, openSidebar }: HeaderProps) {
     }
   }
 
+  async function convertToProposal(pageId: string) {
+    setPageMenuOpen(false);
+    if (isForumPost) {
+      const proposalPage = await charmClient.forum.convertToProposal(pageId);
+      setPages((_pages) => ({
```
mattcasey, line 287: just let web sockets create the page... this is an over-optimization

pages/api/forums/posts/[postId]/convert-to-proposal.ts
```diff
@@ -0,0 +1,73 @@
...
+  const { page: updatedPage } = await createProposal({
+    createdBy: userId,
+    spaceId: post.spaceId,
+    content: post.content ?? undefined,
+    title: post.title
+  });
+
+  updateTrackPageProfile(updatedPage.id);
+
+  relay.broadcast(
```
Comment (mattcasey) [inline pages/api/forums/posts/[postId]/convert-to-proposal.ts:62]: you're broadcasting new page here so no need to call setPages() on the frontend

mattcasey (changes requested):
I think it is better to have a single method for the page to be updated. We really shouldn't need to call setPages() anywhere outside the usePages hook

mattcasey (approved):

pages/api/forums/posts/[postId]/convert-to-proposal.ts
```diff
@@ -0,0 +1,73 @@
...
+import { createProposal } from 'lib/proposal/createProposal';
+import { withSessionRoute } from 'lib/session/withSession';
+import { UnauthorisedActionError } from 'lib/utilities/errors';
+import { relay } from 'lib/websockets/relay';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).post(convertToProposal);
+
+async function convertToProposal(req: NextApiRequest, res: NextApiResponse<PageMeta>) {
```
Comment (mattcasey) [inline pages/api/forums/posts/[postId]/convert-to-proposal.ts:18]: we should add a test on new endpoints

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -420,14 +425,59 @@ export default function Header({ open, openSidebar }: HeaderProps) {
...
+    const isPostCreator = postCreator?.id === user?.id;
     pageOptionsList = (
       <List data-test='forum-post-actions' dense>
         <CopyLinkMenuItem closeMenu={closeMenu} />
         <Divider />
         <DeleteMenuItem onClick={deletePost} disabled={!forumPostInfo.permissions?.delete_post} />
         <UndoMenuItem onClick={undoEditorChanges} disabled={!forumPostInfo?.permissions?.edit_post} />
         <ExportMarkdownMenuItem onClick={exportMarkdownPage} />
+        <Tooltip
+          title={!canCreateProposal || !isPostCreator ? 'You do not have the permission to convert to proposal' : ''}
```
motechFR, line 437: This should only depend on canCreateProposal

2 reasons
- Someone might be a post author but not be able to create proposals
- Someone might be able to create proposals, and find this post worthy of being a proposal, but not be the author of the post

components/common/PageLayout/components/Header/Header.tsx
```diff
@@ -420,14 +425,59 @@ export default function Header({ open, openSidebar }: HeaderProps) {
...
+                sx={{
+                  mr: 1
+                }}
+              />
+              <ListItemText primary='Convert to proposal' />
+            </ListItemButton>
+          </div>
+        </Tooltip>
+        <Divider />
+        <Tooltip title={!isPostCreator ? "You don't have permission to delete this post" : ''}>
```
motechFR, line 455: Make sure to pull in latest main as this now consumes forum permissions.

Overall, you should never manually compute permissions on the UI, and only use the permissions from the hook

lib/permissions/forum/computePostPermissions.ts
```diff
@@ -41,6 +42,15 @@ export async function computePostPermissions({
...
 
+  // If post has been converted to a proposal make it read-only
+  if (post.proposalId) {
+    permissions.addPermissions(['view_post']);
+    if (isAdmin || post.createdBy === userId) {
+      permissions.addPermissions(['delete_post']);
+    }
+    return permissions.operationFlags;
+  }
+
```
motechFR, line 53: This won't work as we're not computing if the user might be a moderator.

This is a case where we can write the same logic in 2 places without abstraction.

// First branch: isAdmin
if (isAdmin) {
  if (post.proposald) {
  const proposalPermissions = new AvailablePostPermissions()
  proposalPermissions.addPermissions(['view_post', 'delete_post'])
  return proposal permissions
 } else {
  // Return current admin value
}
}

motechFR: Then, around line 97;
// Move the permissions search up to here.

// Computing user level based on actual permissions, but we just want a subset
if (proposalId) {
 const hasModerator = hasModeratorSpaceWidePermissions || permissions.some(p => (p.permissionLevel === 'moderator' || p.permissionLevel === 'category_admin))

const convertedToProposalPermissions = new AvailablePostPermissions()
 if (hasModerator) {
 convertedToProposalPermissions.addPermissions(['delete_post]);


} 

convertedToProposalPermissions.addPermissions(['view_post'])

return convertedToProposalPermissions.operationFlags
}


Actually I'll handle this one

lib/permissions/forum/computePostPermissions.ts
```diff
@@ -41,6 +42,15 @@ export async function computePostPermissions({
 
   const permissions = new AvailablePostPermissions();
 
+  // If post has been converted to a proposal make it read-only
+  if (post.proposalId) {
```
motechFR, line 46: Please add these 5 test cases when there is a proposal, I'll get them to pass:

// Add these below existing unit tests

- Returns only the delete_post and view_post permission for admins
- Returns delete_post and view_post for authors
- Returns delete_post and view_post for moderators with space-wide moderate
- Returns delete_post and view_post for moderators with category_level moderate
- Returns only view_post for normal space members, even if they have a full access permission for this categoru

pages/api/forums/posts/[postId]/convert-to-proposal.ts
```diff
@@ -0,0 +1,82 @@
...
+import { ActionNotPermittedError, NotFoundError, onError, onNoMatch, requireUser } from 'lib/middleware';
+import type { PageMeta } from 'lib/pages';
+import { computePostPermissions } from 'lib/permissions/forum/computePostPermissions';
+import { computeSpacePermissions } from 'lib/permissions/spaces';
+import { createProposal } from 'lib/proposal/createProposal';
+import { withSessionRoute } from 'lib/session/withSession';
+import { UnauthorisedActionError } from 'lib/utilities/errors';
+import { relay } from 'lib/websockets/relay';
+
+import { convertPostToProposal } from '../../../../../lib/forums/posts/convertPostToProposal';
```
Comment (motechFR) [inline pages/api/forums/posts/[postId]/convert-to-proposal.ts:15]: You can delete the extra ../../../ I think ;)

lib/proposal/createProposal.ts
```diff
@@ -43,6 +44,9 @@ export async function createProposal(pageProps: ProposalPageInput, proposalProps
           type: 'proposal',
           updatedAt: new Date(),
           updatedBy: createdBy
+        },
+        include: {
+          permissions: true
```
motechFR, line 49: let's not include these permissions based on your feedback since the UI will handle refresh

__integration-tests__/server/pages/api/posts/[postId]/convert-to-proposal.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      count: 5
+    });
+
+    await request(baseUrl)
+      .post(`/api/forums/posts/${posts[0].id}/convert-to-proposal`)
+      .set('Cookie', nonAdminCookie)
+      .expect(401);
+  });
+
+  it('should succeed if the user has permission, and respond 200', async () => {
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/posts/[postId]/convert-to-proposal.spec.ts:68]: Can you also add a 401 test? These are super short to write

lib/forums/posts/convertPostToProposal.ts
```diff
@@ -0,0 +1,12 @@
+import { prisma } from 'db';
+
+export async function convertPostToProposal({ postId, proposalId }: { postId: string; proposalId: string }) {
```
motechFR, line 3: Can you move the logic inside convert post to proposal?

Let this function automagically handle the conversion. You already have the code in the API.

Then, just add a unit test that checks that the return type is a PageWithProposal and that the content is the same as the post content

motechFR (changes requested):
Please make requested changes.

Moving the createProposal logic inside convertProposal optional for now.

We do need to make sure permissions are clean before merging this

__integration-tests__/server/pages/api/forums/posts/[postId]/comments/index.spec.ts
```diff
@@ -3,25 +3,38 @@ import type { PostCategory, Space, User } from '@prisma/client';
...
+let otherUser: User;
 let userCookie: string;
+let otherUserCookie: string;
 
 let accessiblePostCategory: PostCategory;
 let disallowedPostCategory: PostCategory;
 beforeAll(async () => {
   const { space: _space, user: _user } = await generateUserAndSpace({ isAdmin: false });
-  const { space: _space2, user: _user2 } = await generateUserAndSpace({ isAdmin: false });
+  const { user: _user2 } = await generateUserAndSpace({ isAdmin: false });
```
Comment (motechFR) [inline __integration-tests__/server/pages/api/forums/posts/[postId]/comments/index.spec.ts:22]: Can you please replace this by generateSpaceUser method? It creates the user as a space member

PR #1681, opened 2023-02-09
Optimize opening cards in inline databases

mattcasey:
A customer said they are unable to open pages inside an inline database (450 rows) because they have a weak laptop and the browser stalls. They can open pages when viewing the database as a full page, but I can't reproduce to know exactly why.

I definitely made some kind of improvements when opening the cards: https://www.loom.com/share/c720fdbafba74dfe873786b2e045401e

There's a lot of little things I tried:

- The biggest change: the CardDialog component from the inline db currently triggers a re-render on the parent CharmEditor component when it's opened. This is because the editor inside the modal is sharing the same React Context with the parent page. So I created a provider wrapper called `<DocumentPageProvider />` to create isolated contexts in the `<PageLayout />` and `<CardDialog />` components. I wasn't sure if we'd have to rewrite the methods in `useVotes()` and `useThreads()` to read from different contexts, but it just seems to work 👍 if there is a bug with this PR, it will be that we start mixing up context somehow between the two.
- In focalboard tables, clicking away from a page title or text property would trigger an update to the backend. This is due to an onBlur event that was not correctly checking if the value changed.
- Moved getPagePermissions() call into the PageActions component, removing call to usePages() in a bunch of places
- Increased refresh interval for pages from 5 minutes to 30 minutes and removed the 'updateOnFocus' option. This is really just a fallback now since we introduced web sockets, and it is an expensive update. The updateOnFocus feature makes the site feel slow when you are returning from another tab.
- I separated out `currentPageId` from `usePages()`. This way, less things get re-rendered when just changing page Id. I moved it into two hooks: `usePageFromPath()` and `useCurrentPage()`
- I noticed we were not using all providers on the `<SharedPublicLayout />` page - so we'll show there is a comment but you can't view it. This is now fixed

To test:

- Create page with an inline database with some cards. Try interacting with them as well as content in the parent page. Make sure the page actions menu at top-right have correct context.

Here's the React tree structure to help visualize how the `<DocumentPageProviders />` component is used:
```
// Page layout component
PageLayout
	DocumentPageProviders (global context - also used by header, sidebar, page meta, etc)
		Header
		EditorPage
			// Case 1: Is document page w/inline database and documents inside modal:
			DocumentPage
				Charmeditor
					InlineDatabase
						CenterPanel
			// Case 2: is database page w/document pages inside modal:
			DatabasePage
				CenterPanel

// Public page layout component
SharedPageLayout
	DocumentPageProviders
		// ... same as PageLayout


// Database view component
CenterPanel
	CardDialog
		DocumentPageProviders (modal context)
			PageDialog
				DocumentPage

// Context used by CharmEditor and page context menu
DocumentPageProviders
> CurrentPageProvider
> CharmEditorProvider - (for charmeditor settings)
> ThreadsProvider
> VotesProvider
> PageActionDisplayProvider
```


components/[pageId]/DocumentPage/DocumentPageProviders.tsx
```diff
@@ -0,0 +1,20 @@
+import { CharmEditorProvider } from 'hooks/useCharmEditor';
+import { CurrentPageProvider } from 'hooks/useCurrentPage';
+import { PageActionDisplayProvider } from 'hooks/usePageActionDisplay';
+import { ThreadsProvider } from 'hooks/useThreads';
+import { VotesProvider } from 'hooks/useVotes';
+
+// context that is needed for DocumentPage to work
+export function DocumentPageProviders({ children }: { children: React.ReactNode }) {
```
Comment (msadura) [inline components/[pageId]/DocumentPage/DocumentPageProviders.tsx:8]: I really like having contexts for page enclosed here

msadura (approved):

PR #1682, opened 2023-02-09
Fix forum comment input text overflow

msadura:
Fix text overflowing in forum post input. Moved styling to charm editor internals, it should be applied when we turn on `disableRowHandles`

Zrzut ekranu 2023-02-9 o 08 58 32


Devorein (approved):
Approved but can you please remove it from `PostComment` component as well 🙏🏼

mattcasey (approved):

PR #1691, opened 2023-02-09
Space wide forum moderators

motechFR:
Use space permissions to allow users with a role to moderate all forums


lib/permissions/meta/preset-templates.ts
```diff
@@ -23,7 +24,8 @@ const collaborative: SpacePermissionTemplate = {
     createPage: true,
     createBounty: true,
     createVote: true,
-    createForumCategory: false
+    createForumCategory: false,
+    moderate_forums: false
```
mattcasey, line 28: shouldn't this be camel-case?

motechFR: I added a comment in prisma schema. I agree it's not the most beautiful, but this is one case where I think it's better to adopt our new team wide pattern rather than going consistent with the legacy pattern

PR #1699, opened 2023-02-10
Open cards full-height to avoid layout jumping

mattcasey:
We show the card before content is loaded, and it usually jumps full height once it is loaded. I think it's a better experience if we just use a fixed height. I also made the dialog be full height in mobile.

https://www.loom.com/share/354dbcbb67704b65864b7cee330218c0

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/dialog.scss
+++ b/components/common/BoardEditor/focalboard/src/components/dialog.scss
@@ -44,6 +44,7 @@
     left: 0;
     right: 0;
     bottom: 0;
+    margin: 0 auto;
     max-width: 100%;
 
     &.fullWidth {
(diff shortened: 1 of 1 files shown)


components/common/BoardEditor/focalboard/src/components/dialog.scss
```diff
@@ -63,7 +63,7 @@
       top: 50%;
       width: 100%;
       margin: 0;
-      max-height: 75%;
+      height: calc(100% - 144px);
```
mattcasey, line 66: same height as notion :P

msadura: slightly unrelated, but we should use the same value for forum post form dialog - it takes 100% height and looks weird

msadura (approved):

PR #1700, opened 2023-02-10
Show correct members

motechFR:
@mattcasey Last step of fixing space modal!

I noticed Members were still out of sync, but roles were fine. Upon further research, I understood.

The members depends on a global context. So even though we were setting useSpaceId locally, the useMembers hook was bound to global context and reading from global space.

I readded it below CurrentSpace and it works now.

Below are the global providers. MembersProvider is only provider it made sense to add, but please do double check.

Screenshot 2023-02-10 at 06 58 23

For hooks which just return results (like useRoles) no work needed, they'll pick up the closest provided context.


mattcasey (approved):
nice simple fix :)

PR #1703, opened 2023-02-10
After csv import -> show table columns

valentinludu:
Before, we were importing a csv table without showing the columns on a table view.

Now, the properties will be visible for table view, but not for other views.

Notion does the same.


Devorein (approved):

mattcasey (approved):
Thanks! I forgot about this one

PR #1704, opened 2023-02-10
Close quick find modal after clicking the page you want to go to

valentinludu:


msadura:
looks good, btw will it also work when you use up/down arrows to select and press enter?

valentinludu:
> looks good, btw will it also work when you use up/down arrows to select and press enter?

yes

mattcasey:
I was just noticing this one today for some reason

mattcasey (approved):

PR #1711, opened 2023-02-11
adjust order of static items in sidebar

mattcasey:
Just some tweaks from Chris's mockups that didn't make it 

image


msadura (approved):
nice stuff

valentinludu (approved):

PR #1713, opened 2023-02-12
Format focalboard dates based on user preference locale

Devorein:


components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.tsx
```diff
@@ -50,11 +53,12 @@ function DateRange(props: Props): JSX.Element {
   const { className, value, showEmptyPlaceholder, onChange } = props;
   const intl = useIntl();
   const popupState = usePopupState({ variant: 'popover', popupId: 'dateRangePopup' });
+  const { formatDate } = useDateFormatter();
 
   const getDisplayDate = (date: Date | null | undefined) => {
     let displayDate = '';
     if (date) {
-      displayDate = Utils.displayDate(date, intl);
+      displayDate = formatDate(date);
```
msadura, line 61: Check lines 174 and 183 in this file - they are still using some custom date format instead of this function

msadura: If you update that code you should not even need `locale` variable - overall you want to use our formatted to avoid creating custom date formats and keep app consistent

components/common/BoardEditor/focalboard/src/widgets/editableDayPicker.tsx
```diff
@@ -18,28 +17,22 @@ const parseValue = (value: string): Date | undefined => {
...
 
 function EditableDayPicker(props: Props): JSX.Element {
   const { className, onChange } = props;
   const intl = useIntl();
   const [value, setValue] = useState(() => parseValue(props.value));
   const [dayPickerVisible, setDayPickerVisible] = useState(false);
-
-  const locale = intl.locale.toLowerCase();
+  const { userPreferences } = useUserPreferences();
+  const { formatDate: formatLocaleDate } = useDateFormatter();
```
msadura, line 28: I dont understand why you have `formatLocaleDate ` here and below there is custom `formatDate` - this custom formatDate should be just removed, as we do not want to have random formats

pages/_app.tsx
```diff
@@ -242,17 +243,26 @@ export default function App({ Component, emotionCache = clientSideEmotionCache,
...
+              </ReactDndProvider>
+            </SettingsDialogProvider>
+          </SnackbarProvider>
         </ThemeProvider>
       </ColorModeContext.Provider>
     </CacheProvider>
   );
 }
 
+function LocaleProvider({ children }: { children: ReactNode }) {
```
msadura, line 257: I believe we agreed to try not to mess _app.tsx when not needed. Please move this provider to `components/_app` dir

pages/_app.tsx
```diff
@@ -242,17 +243,26 @@ export default function App({ Component, emotionCache = clientSideEmotionCache,
...
+  const { userPreferences } = useUserPreferences();
+  return (
+    <LocalizationProvider dateAdapter={AdapterLuxon as any} adapterLocale={userPreferences.locale ?? undefined}>
```
msadura, line 260: I believe `as any` is not needed

msadura, line 260: Honestly I somehow do not like this solution as it will bind luxon to mui and it will stay with us forever :D I think we might want to create out own adapter which will use our formatting functions. Anyway it is not in a scope of this task, but we might consider adding this to our long-term backlog. cc: @mattcasey

msadura (changes requested):

components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.tsx
Devorein: Got it. Thank you.

components/common/BoardEditor/focalboard/csvExporter/csvExporter.ts
```diff
@@ -76,7 +76,13 @@ class CsvExporter {
       _row.push(`"${this.encodeText(card.title)}"`);
       visibleProperties.forEach((template: IPropertyTemplate) => {
         const propertyValue = card.fields.properties[template.id];
-        const displayValue = (OctoUtils.propertyDisplayValue(card, propertyValue, template, intl) || '') as string;
+        const displayValue = (OctoUtils.propertyDisplayValue(
+          card,
+          propertyValue,
+          template,
+          (date) => Utils.displayDate(new Date(date), intl),
+          (date) => Utils.displayDateTime(new Date(date), intl)
```
mattcasey, line 84: these are the wrong methods, right?

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -33,11 +33,11 @@ type Props = {
 function PropertyValueElement(props: Props): JSX.Element {
   const [value, setValue] = useState(props.card.fields.properties[props.propertyTemplate.id] || '');
   const [serverValue, setServerValue] = useState(props.card.fields.properties[props.propertyTemplate.id] || '');
-
+  const { propertyDisplayValue } = usePropertyDisplayValue();
   const { card, propertyTemplate, readOnly, showEmptyPlaceholder, board, updatedBy, updatedAt, displayType } = props;
   const intl = useIntl();
   const propertyValue = card.fields.properties[propertyTemplate.id];
-  const displayValue = OctoUtils.propertyDisplayValue(card, propertyValue, propertyTemplate, intl);
+  const displayValue = propertyDisplayValue(card, propertyValue, propertyTemplate);
```
mattcasey, line 40: I think usePropertyDisplayValue wraps propertyDisplayValue...  are we calling it twice here?

components/common/BoardEditor/focalboard/src/widgets/editableDayPicker.tsx
Devorein: Thanks. Ended up deleting this whole component as its not used anywhere 🥳

components/common/BoardEditor/focalboard/src/octoUtils.tsx
```diff
@@ -200,4 +202,18 @@ class OctoUtils {
   }
 }
 
+export function usePropertyDisplayValue() {
+  const { formatDateTime, formatDate } = useDateFormatter();
```
mattcasey, line 206: this seems like an unecessary abstraction, couldnt you hace just passed formatDateTime into the original method instead of `intl`

pages/_app.tsx
```diff
@@ -210,11 +211,11 @@ export default function App({ Component, emotionCache = clientSideEmotionCache,
     <CacheProvider value={emotionCache}>
       <ColorModeContext.Provider value={colorModeContext}>
         <ThemeProvider theme={theme}>
-          <LocalizationProvider dateAdapter={AdapterLuxon as any}>
```
mattcasey, line 213: have you tested all the places this would be used?

pages/_app.tsx
Devorein: yep its not throwing any error 🥳

pages/_app.tsx
Devorein: Yes we are using the `DateTimePicker` in three places and I've tested all of them out.

components/common/BoardEditor/focalboard/csvExporter/csvExporter.ts
Devorein: They were wrong, I've fixed it now. exported csv will now format dates based on user preference locale.

components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.tsx
```diff
@@ -176,7 +175,7 @@ function DateRange(props: Props): JSX.Element {
                 }}
                 onChange={setFromInput}
                 onSave={() => {
-                  const newDate = DateTime.fromFormat(fromInput, 'DD/MM/YYYY', { locale: intl.locale }).toJSDate();
+                  const newDate = new Date(formatDate(fromInput, { withYear: true }));
```
msadura, line 178: I don't get why you first format date to string and then create a date object from it?

PR #1718, opened 2023-02-13
Fix triple-click in charmeditor

mattcasey:


components/common/CharmEditor/components/floatingMenu/floatingMenu.plugins.ts
```diff
@@ -72,25 +76,48 @@ export function plugins({
...
-        },
-        apply(tr: Transaction) {
-          return tr.getMeta('row-handle-is-dragging');
+    Object.assign(controller, {
+      // add listener to onBlur() to hide tooltip
+      props: {
+        ...controller.props,
+        handleDOMEvents: {
+          ...controller.spec.props?.handleDOMEvents,
+          blur: (view: EditorView, event: MouseEvent) => {
```
mattcasey, line 85: it's kind of hard to tell in the diff, but this is the event listener that was added

patches/@bangle.dev+tooltip+0.31.6.patch
```diff
@@ -0,0 +1,24 @@
...
+-            let { head, from } = selection;
++            let { head, from, to } = selection;
+             // since head is dependent on the users choice of direction,
+             // it is not always equal to `from`.
+             // For textSelections we want to show the tooltip at head of the
+@@ -282,7 +282,9 @@ function getSelectionReferenceElement(view) {
+             // But for NodeSelection we always want `from` since, if we go with `head`
+             // coordsAtPos(head) might get the position `to` in head, resulting in
+             // incorrectly getting position of the node after the selected Node.
+-            const pos = selection instanceof NodeSelection ? from : head;
```
mattcasey, line 18: I was annoyed that when I'd select a paragraph, the floating menu would appear in the middle rather htan on top. After a lot of debugging, it was actually due to this decision to use the 'head' part of a selection, vs the 'from'. If you're using right-to-left text direction, the floating menu might ALWAYS be at the bottom.

Devorein (approved):

PR #1719, opened 2023-02-14
Fix error when pasting SVG

mattcasey:
The error turned out to be some code in the image component that tries to upload binary image sources. In this case, it was an svg url which doesn't have binary data to convert. I was able to reproduce by copying the same content that Chris had mentioned:

image

### DIFF
--- a/components/common/CharmEditor/components/ResizableImage.tsx
+++ b/components/common/CharmEditor/components/ResizableImage.tsx
@@ -166,14 +166,19 @@ function ResizableImage({
         </ImageSelector>
       );
     }
-  } else if (imageSource.startsWith('data') && !uploadingImage && !readOnly && !uploadFailed) {
+  } else if (
+    imageSource.startsWith('data') &&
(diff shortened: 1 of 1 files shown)


components/common/CharmEditor/components/ResizableImage.tsx
```diff
@@ -147,6 +147,27 @@ function ResizableImage({
     view.dispatch(view.state.tr.deleteRange(start, end));
   }
 
+  useEffect(() => {
```
mattcasey, line 150: I also wrapped this in a useEffect since we call setUploadingImage. It seems dangerous to do that outside of a useEffect or callback?

Devorein: Good catch. It was only due to `!uploadingImage` condition it didn't run into an infinite loop.

Devorein:
Are we planning on handling svg image upload in the near future? Even base64 encoded svg are not being uploaded eg:- `data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHhtbDpzcGF`, due to invalid filename error. The generated filename becomes `uuid.svg+xml` which is invalid. I removed `+xml` and it worked.

Devorein (approved):

PR #1720, opened 2023-02-14
Allow forums to be made public

motechFR:
To test,

Go to a forum home as a space admin, select a category, click "manage permissions" and toggle the "Public" option.

Create a post within this category.

Go to the space forum home from a private browsing window. You should see the public category, and the post


motechFR:
@msadura I reused the sharedPage hook, and simplified alot of stuff. One change, we don't try to be smart, and just always load the space from this hook. It's a pretty trivial operation, but at least guarantees we always get the space

msadura (approved):
I like your changes and AI am happy it was not hard to add another "public" page to that hook :D

PR #1722, opened 2023-02-14
Fix long property names and values

Devorein:
Previous issues
1. Didn't show ellipsis for long property names (only occurred in readonly mode)
2. Showed ellipsis for property values

Now (Note this is in read only mode)
image

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
@@ -182,7 +182,9 @@ function CardDetailProperties(props: Props) {
         return (
           <div key={`${propertyTemplate.id}-${propertyTemplate.type}-${propertyValue}`} className='octo-propertyrow'>
             {props.readOnly && (
-              <div className='octo-propertyname octo-propertyname--readonly'>{propertyTemplate.name}</div>
+              <div className='octo-propertyname octo-propertyname--readonly'>
+                <Button>{propertyTemplate.name}</Button>
+              </div>
             )}
             {!props.readOnly && (
               <MenuWrapper isOpen={propertyTemplate.id === newTemplateId}>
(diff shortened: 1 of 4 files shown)


components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -182,7 +182,9 @@ function CardDetailProperties(props: Props) {
         return (
           <div key={`${propertyTemplate.id}-${propertyTemplate.type}-${propertyValue}`} className='octo-propertyrow'>
             {props.readOnly && (
-              <div className='octo-propertyname octo-propertyname--readonly'>{propertyTemplate.name}</div>
+              <div className='octo-propertyname octo-propertyname--readonly'>
+                <Button>{propertyTemplate.name}</Button>
```
msadura, line 186: why button in readonly mode?

mattcasey (approved):

components/common/BoardEditor/focalboard/src/widgets/TextInput.tsx
```diff
@@ -18,16 +18,17 @@ const StyledInput = styled(InputBase)`
...
   const { className, ...props } = useEditable(_props, ref, elementRef);
   return (
     <StyledInput
-      {...props}
       inputProps={{
         className,
         maxRows
       }}
+      {...props}
+      multiline={multiline}
```
mattcasey, line 31: this doesnt look like it would have any effect, since multiline was already part of props

theme/focalboard/focalboard.main.scss
```diff
@@ -117,7 +117,6 @@ html {
   .octo-propertyvalue {
     font-size: 12px;
     color: rgb(var(--center-channel-color-rgb));
-    margin: 2px 0;
```
mattcasey, line 120: this breaks the alignment between text and labels, where they are no longer horizotnally aligned.. i'm putting it back for now.

PR #1729, opened 2023-02-15
Feat/show more data

msadura:
Paginating displayed data in table rows / kanban boards. 
https://app.charmverse.io/charmverse/page-3431887540171268

Screenshot 2023-02-15 at 17 12 00
Screenshot 2023-02-15 at 17 11 51
Screenshot 2023-02-15 at 17 11 43

### DIFF
--- a/components/PageSizeInputPopup.tsx
+++ b/components/PageSizeInputPopup.tsx
@@ -1,0 +1,45 @@
+import { MoreHoriz } from '@mui/icons-material';
+import { IconButton, TextField } from '@mui/material';
+import { Stack } from '@mui/system';
+import { useState } from 'react';
+
+import FieldLabel from 'components/common/form/FieldLabel';
+import PopperPopup from 'components/common/PopperPopup';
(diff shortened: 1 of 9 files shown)


msadura:
EDIT: updated UI to be more notion-like

mattcasey (changes requested):
There is a bug that the 'load more' action gets triggered when i am just setting the page size - or even if i change my mind, it triggers when i close the page size menu. I wanted to change page size to 10 from 100 and couldn't figure out why it wasn't going to 10 :)

mattcasey:
FYI the initial idea was going to be to save the page size to the view, the way Notion does, and have it in the View settings as well the way Notion does. But I actually think this maybe fine for now. Just want to bring it up to Chris.

mattcasey:
> I actually was not sure how we want to "save" that - in db? Local storage? Is it per page setting or global for board / table?

I think its for everyone, similar to how we store view filters, etc. for now. I think Notion does it that way,  since it's part of view settings. for that it would just be adding another value  to block.fields on the 'view block'

mattcasey (approved):

PR #1730, opened 2023-02-15
Fixed toggle public bounties

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=5d2ff43c-1f10-4894-bf74-49b9e3d76c43)

Stacktrace
```
Transaction API error: Transaction already closed: A query cannot be executed on an expired transaction. The timeout for this transaction was 5000 ms, however 16928 ms passed since the start of the transaction. Consider increasing the interactive transaction timeout or doing less work in the transaction.
```

Added timeout of 20000 ms for the interactive transaction. Went with 15000 at first but that was also not bullet-proof. We are also using timeout of 20000ms inside `pages\api\permissions\index.ts\addPagePermission` function. Curious to know whether there is any other better approach, maybe increasing timeout by 5000ms everytime it fails?

### DIFF
--- a/lib/spaces/togglePublicBounties.ts
+++ b/lib/spaces/togglePublicBounties.ts
@@ -151,37 +151,42 @@ export async function togglePublicBounties({ spaceId, publicBountyBoard }: Publi
   }
 
   try {
-    const spaceAfterUpdate = await prisma.$transaction(async (tx) => {
-      const updatedSpace = await tx.space.update({
-        where: { id: spaceId },
-        data: {
-          publicBountyBoard
-        }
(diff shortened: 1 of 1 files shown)


mattcasey (changes requested):
>Curious to know whether there is any other better approach, maybe increasing timeout by 5000ms everytime it fails?

Before increasing the timeout, we should try to explain what operations are taking a long time. I don't see why just changing a space-wide option should be so expensive..

PR #1732, opened 2023-02-15
Return space before public space

motechFR:


hooks/useCurrentSpace.ts
```diff
@@ -13,6 +13,6 @@ export function useCurrentSpace() {
 
   if (accessChecked) {
     // IF we are viewing a public page, we want to return the public space as current one
-    return publicSpace || space;
+    return space ?? publicSpace;
```
mattcasey, line 16: why would there be two values at the same time and why is one wrong?

mattcasey (approved):

PR #1734, opened 2023-02-15
fix toggle for public bounties

mattcasey:
I removed the call to retrieve pages when updating bounties. Conceptually, it's nice, but it takes too long to resolve the whole page tree. I think it is extremely unlikely (knock on wood) that someone would be moving bounty cards around while someone else is toggling this setting in a way that would break permissions.

### DIFF
--- a/lib/spaces/interfaces.ts
+++ b/lib/spaces/interfaces.ts
@@ -5,7 +5,7 @@ import type { Space } from '@prisma/client';
 // If we use it, we will need to update the space-related hooks so they also use this subset. This way, we can guarantee consistent behaviour between logged-in and public mode.
 export type PublicSpaceInfo = Pick<Space, 'domain' | 'id'>;
 
-export interface PublicBountyToggle<T extends boolean = boolean> {
+export interface PublicBountyToggle {
   spaceId: string;
-  publicBountyBoard: T;
+  publicBountyBoard: boolean;
 }
--- a/lib/spaces/togglePublicBounties.ts
+++ b/lib/spaces/togglePublicBounties.ts
@@ -5,8 +5,7 @@ import { prisma } from 'db';
 import type { PageNodeWithChildren, PageNodeWithPermissions } from 'lib/pages';
 import { multiResolvePageTree } from 'lib/pages/server/resolvePageTree';
 import { hasSameOrMorePermissions } from 'lib/permissions/pages';
-
-import { InvalidInputError } from '../utilities/errors';
+import { DataNotFoundError, InvalidInputError } from 'lib/utilities/errors';
 
 import type { PublicBountyToggle } from './interfaces';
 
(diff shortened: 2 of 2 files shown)


lib/spaces/togglePublicBounties.ts
```diff
@@ -4,22 +4,12 @@ import { v4, validate } from 'uuid';
...
 import { multiResolvePageTree } from 'lib/pages/server/resolvePageTree';
-import { hasSameOrMorePermissions, IPagePermissionWithSource } from 'lib/permissions/pages';
+import { hasSameOrMorePermissions } from 'lib/permissions/pages';
 
-import { DataNotFoundError, InvalidInputError } from '../utilities/errors';
+import { InvalidInputError } from '../utilities/errors';
 
 import type { PublicBountyToggle } from './interfaces';
 
-async function generatePublicBountyPermissionArgs({
```
mattcasey, line 13: removed this so that we can call generatePublicBountyPermissionArgs() in one place, whether publicBountyBoard is true or false

lib/spaces/togglePublicBounties.ts
```diff
@@ -150,41 +140,33 @@ export async function togglePublicBounties({ spaceId, publicBountyBoard }: Publi
...
-          publicBountyBoard
-        }
-      });
-
-      if (publicBountyBoard === true) {
-        const [deleteArgs, createArgs, childCreateArgs] = await generatePublicBountyPermissionArgs({
-          publicBountyBoard,
-          spaceId
-        });
+  const [deleteArgs, createArgs, childCreateArgs] = await generatePublicBountyPermissionArgs({
```
mattcasey, line 143: moved this async code outside of the transaction

lib/spaces/interfaces.ts
```diff
@@ -5,7 +5,7 @@ import type { Space } from '@prisma/client';
 // If we use it, we will need to update the space-related hooks so they also use this subset. This way, we can guarantee consistent behaviour between logged-in and public mode.
 export type PublicSpaceInfo = Pick<Space, 'domain' | 'id'>;
 
-export interface PublicBountyToggle<T extends boolean = boolean> {
+export interface PublicBountyToggle {
```
motechFR, line 8: Not sure why we had a T there in first place, but good cleanup!

lib/spaces/togglePublicBounties.ts
motechFR: Makes sense, good performance optimisation. A next level would be having much faster resolvePageTree in future :)

motechFR (approved):
@mattcasey Approved

Ran the unit tests, they're all green so that's good

We are getting a failing one, related to the error if space id doesn't exist. So we now get a PrismaError instead of a SystemError

Does seem like an edge case, so we could just delete this test. I think our API creates an unknown error from SystemError anyways

mattcasey:
> Ran the unit tests, they're all green so that's good

We are getting a failing one, related to the error if space id doesn't exist. So we now get a PrismaError instead of a SystemError

Does seem like an edge case, so we could just delete this test. I think our API creates an unknown error from SystemError anyways

Thanks @motechFR ... i took a look and we can actually get a special code from prisma, and still return DatanotFound in that case

PR #1736, opened 2023-02-16
New UI for Profile view

valentinludu:
[Task Item](https://app.charmverse.io/charmverse/page-5367853706449786?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565)


pages/api/spaces/[id]/members/index.ts
```diff
@@ -146,6 +148,44 @@ async function getMembers(req: NextApiRequest, res: NextApiResponse<(Member | Pu
     }
   });
 
+  const currentSpaceRole = spaceRoles.find((sr) => sr.spaceId === spaceId);
+
+  // If the user is not onboarded, populate the user profile with one nft as a profile picture and the first 5 nfts he has as pinned
+  if (!currentSpaceRole?.onboarded && userId) {
```
Comment (valentinludu) [inline pages/api/spaces/[id]/members/index.ts:154]: @devs 
I need your opinions on this one.
I think this is the best place to populate user avatar and pinned nfts.

Comment (mattcasey) [reply pages/api/spaces/[id]/members/index.ts:154]: @valentinludu This route is a GET endpoints to return list of members in a space. It makes no sense to me that it would update the user profile... 1) it's a GET, we should never mutate on GET requests and 2) its to retrieve members of the space, and it is already being used in too many places. We really should have one route to retrieve list of users for general display in docs and focalboard properties, one for member directory (with all the member properties) and one for user profile. So I would not anything else to this endpoint

Comment (mattcasey) [reply pages/api/spaces/[id]/members/index.ts:154]: What are you trying to do? Is this something that should be done during onboarding, or in a migration?

components/common/Modal/SettingsDialog.tsx
```diff
@@ -135,7 +137,7 @@ function SpaceSettingsModalComponent() {
             sx={{ backgroundColor: (theme) => theme.palette.sidebar.background }}
           >
             <Typography p='10px' variant='body1'>
-              {user?.username}
+              {shortWalletAddress(user?.username)}
```
mattcasey, line 140: we should never have to do this in the frontend . usernames should already be 'shortened' when we assign a wallet address  to them in the backend.

components/common/Modal/SettingsDialog.tsx
```diff
@@ -227,18 +229,35 @@ function SpaceSettingsModalComponent() {
...
+          <Button
+            variant='text'
+            color='inherit'
+            onClick={onClose}
+            sx={{
+              position: 'absolute',
+              right: 15,
+              top: 10,
+              zIndex: 1,
+              '&:hover': { backgroundColor: 'transparent' }
```
mattcasey, line 242: Not sure why you want to use a hover effect on mobile, but if you do, you _must_ use the media query `@media (pointer: fine)` to disable it on iOS or else it will require double-clicking.

components/common/form/interfaces.ts
```diff
@@ -25,4 +25,5 @@ export type FieldProps = {
   inline?: boolean;
   error?: string;
   disabled?: boolean;
+  customOnChange?: (value: any) => Promise<void>;
```
mattcasey, line 28: feels like `customOnChange` should have the same signature as `onChange` and i would probably just use `onChange`. "custom" makes me think I am overriding a default handler, but it's just being called in addition.

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopup.tsx
```diff
@@ -112,47 +116,31 @@ export function MemberPropertiesPopup({
...
+                    name={property.memberPropertyId}
+                    control={control}
+                    rules={fieldRendererConfig.rules}
+                    render={fieldRendererConfig.renderer}
+                  />
+                ) : null;
+              })}
+            </Box>
             {postComponent}
-            <DialogActions>
```
mattcasey, line 142: we should add some padding at the bottom of the modal now that the buttons are gone

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopup.tsx
```diff
@@ -112,47 +116,31 @@ export function MemberPropertiesPopup({
...
-                      rules={fieldRendererConfig.rules}
-                      render={fieldRendererConfig.renderer}
-                    />
-                  ) : null;
-                })}
-              </Box>
-            </form>
+            <Box display='flex' flexDirection='column'>
+              {data.map((property) => {
+                const fieldRendererConfig = getFieldRendererConfig({
```
mattcasey, line 121: this is a weird pattern, but i know we were using it before so doesn't need to be fixed here. I think it should just be a single React component, i don't kno why we need to have a method that returns some properties that passed to another React component

components/common/form/fields/FieldWrapper.tsx
```diff
@@ -18,18 +20,7 @@ export function FieldWrapper({ children, label, inline, iconLabel }: Props) {
       {(label || !!iconLabel) && (
         <Box maxWidth={150} width={150} alignItems='center' display='flex' gap={1}>
```
mattcasey, line 21: i don't think we need this max width anymore. a long label looks very squished in the member properties modal?

components/common/form/fields/NumberInputField.tsx
```diff
@@ -7,10 +7,15 @@ import type { ControlFieldProps, FieldProps } from 'components/common/form/inter
 type Props = ControlFieldProps & FieldProps;
 
 export const NumberInputField = forwardRef<HTMLDivElement, Props>(
-  ({ label, iconLabel, inline, error, ...inputProps }: Props, ref) => {
+  ({ label, iconLabel, inline, error, onChange, customOnChange, ...inputProps }: Props, ref) => {
+    const onChangeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+      onChange(event);
+      customOnChange?.(event.target.value);
```
mattcasey, line 13: I'm not really sure why we needed customOnChange, but i find it extremely confusing to have a component that takes onChange as well as customOnChange.. i tihnk a better pattern could be to wrap the original 'onChange' handler in one of the parents?

```
function parentInput (props: { onChange }) {
 
  function customOnChange(event) {
     // do stuff
     onChange(event)
  }
   return <NumberInput onChange={customOnChange} />
}


```

lib/blockchain/getNFTs.ts
```diff
@@ -0,0 +1,47 @@
+import { prisma } from 'db';
+import { getNFTs } from 'lib/blockchain/nfts';
```
mattcasey, line 2: this is confusing: 'getNFTs' imports another method 'getNFTs' from a sibling folder called 'nfts'. This method just belongs inside 'nfts' or inside a folder with them grouped together.

mattcasey: On second though, this probably belongs in lib/profile since it's based on the user profile, and u could rename it "getUserNFTs.ts"

mattcasey (changes requested):
Left some comments...

design-wise, maybe for another iteration, I think the member properties should just be editable and not inside a popup when editing in settings:

image

components/common/Modal/SettingsDialog.tsx
valentinludu: There were some cases where I saw the whole address. Will bring it up in today's meeting.

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopup.tsx
valentinludu: Done!

components/common/form/fields/FieldWrapper.tsx
valentinludu: Removed.
Checked on mobile and desktop and looks fine without the width and maxWidth.

components/profile/components/SpacesMemberDetails/components/MemberPropertiesPopup.tsx
valentinludu: yes, it would be easy.
It took me some time to get my head around it.

Now that you mention it, I removed this pattern and kept only what we needed in a more readable form.

components/common/form/interfaces.ts
valentinludu: Removed customOnChange and kept only onChange

components/common/form/fields/NumberInputField.tsx
valentinludu: Added onChange logic outside the Controller. When I send the onChange to the controller I also await for the change

lib/blockchain/getNFTs.ts
valentinludu: Moved to lib/profile/getUserNFTs.ts

components/profile/components/UserDetails/utils.ts
```diff
@@ -0,0 +1,5 @@
+import type { LoggedInUser } from 'models';
+import type { PublicUser } from 'pages/api/public/profile/[userId]';
+
+export const isPublicUser = (user: PublicUser | LoggedInUser, currentUser: null | LoggedInUser): user is PublicUser =>
```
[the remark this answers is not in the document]
valentinludu: I agree that this function is not the best solution and name. Also this function was here before and I just moved it to a separate location because it was used in multiple places.

I added the verification to be sure the object structure is indeed a public profile.

If I return a boolean without a type this will make PublicProfile and UserDetails full of ts errors because they are relying on that type (Public or not). 

The problem is that PublicProfile is used from /u/ public page (Public user) or in profile settings (LoggedIn user). 
The solution is to decouple these 2 and make separate flows and reuse components that they have in common. But I think I will leave this for another PR.

pages/api/spaces/[id]/members/index.ts
Comment (valentinludu) [reply pages/api/spaces/[id]/members/index.ts]: Right. It was more of a proof of concept.

Moved this login inside the user creation.
Now, if a user is created with a wallet and has nfts, then I will populate his profile.

mattcasey (approved):

PR #1740, opened 2023-02-16
Add space template

motechFR:
@mattcasey Took liberty of sorting this one since I knew the stack. Took opportunity to do some small cleanup

Ignore the big diff, just some json :)

here is the attached support script
https://app.charmverse.io/charmverse/page-996109324018648

Last todo, we might want different logos for each template


lib/spaces/createWorkspace.ts
```diff
@@ -140,10 +140,10 @@ export async function createWorkspace({
       ...seedPagesTransactionInput.blocksToCreate.map((input) => prisma.block.create({ data: input })),
       ...seedPagesTransactionInput.pagesToCreate.map((input) => createPage({ data: { ...input, autoGenerated: true } }))
     ]);
-  } else if (createSpaceOption === 'templateNftCommunity') {
+  } else if (createSpaceOption.match('template')) {
```
mattcasey, line 143: would it be safer to just check if the option exists in `spaceContentTemplates`? seems a little arbitrary to expect template names always include the name 'template'

PR #1741, opened 2023-02-16
Drag and drop page and image blocks

Devorein:
It works for both page and image blocks (we can extend that to other non-text blocks)
But this error comes up at times. Interestingly it work gets triggered when I move anything upwards not downwards.

image


mattcasey:
@Devorein can u see why it was reaching the code inside ReactDNDProvider? maybe we are not ignoring the element properly.

mattcasey (approved):
Works great. I added a check for parentNode when dragging - which means the DOM element is not embedded in the page, so should be ok to ignore. Drag/and drop cards and card properties seems to still work fine in focalboard.

I'm wondering if this could be simplified further, so that we don't have to add listener on drag-start (thats not a prosemirror requriement, so its something with one of our other plugins)

PR #1745, opened 2023-02-17
Proposal permissions

motechFR:
Provide proposal permissions engine

Testing
1: Create a new workspace
- You should have default categories all set to full access for the space

2: Create a new category
- The category should have full access for the space

3: Filtering permissions
- Remove permissions from a category
- Create a proposal
- Move it to discussion stage
- Log in with separate user
- Category should be invisible
- Proposal should be invisible

4: Restricting interactions
- As admin: Add space / view comment
- Create discussion stage proposal in category
- Switch to other user
- User should be able to comment
- Switch back to admin and move proposal to vote active stage
- Switch back to user
- User should not be able to vote
- Switch back to admin; upgrade category permissions to view comment vote
- Switch back to user; they should now be able to vote

5: Make page public
- Make any proposal public via share settings
- Open the page. User should be able to view the proposal


lib/permissions/proposals/deleteProposalCategoryPermission.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { PermissionToDelete } from '../interfaces';
+
+export async function deleteProposalCategoryPermission({ permissionId }: PermissionToDelete) {
+  const permission = await prisma.proposalCategoryPermission.findUnique({
+    where: {
+      id: permissionId
+    }
+  });
+
+  if (!permission) {
```
mattcasey, line 12: should log a warning at least or use findUniqueOrThrow so it thorws an error if the user is expecting a successful change

lib/permissions/proposals/computeProposalPermissions/pfpStatusDiscussionEditableCommentable.ts
```diff
@@ -0,0 +1,40 @@
+import type { ProposalOperation } from '@prisma/client';
```
mattcasey, line 1: Please don't use acronyms in the file names... I don't know what 'pfp' means, so you might as well leave it out or put them in a folder

lib/proposal/errors.ts
```diff
@@ -0,0 +1,30 @@
+import { SystemError } from 'lib/utilities/errors';
```
mattcasey, line 1: lib/proposals should be plural

lib/permissions/buildComputePermissionsWithPermissionFilteringPolicies.ts
```diff
@@ -49,20 +53,50 @@ export function buildComputePermissionsWithPermissionFilteringPolicies<R, F>({
...
+        if (flagValue === true && applicableFlags[key] !== true) {
           throw new InsecureOperationError(
             `Permission filtering policy ${pfp.name} attempted to add a new permission flag: ${key.toString()}`
           );
         }
       });
 
       applicableFlags = newFlags;
+
+      // Perform an early return if a PFP results in fully empty permissions
```
mattcasey, line 96: we don't want to use acronyms to refer to parts of our system. it makes onboarding new devs hard and confusing for non-tech people when we talk about it in Discord

lib/permissions/proposals/hasSpaceWideProposalReviewerPermission.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (!userSpaceRole) {
+    return false;
+  }
+
+  if (userSpaceRole.isAdmin) {
+    return true;
+  }
+  const permissions = await prisma.spacePermission.findMany({
+    where: {
+      forSpaceId: spaceId,
```
mattcasey, line 43: what's the difference between forSpaceId and spaceId in this schema?

lib/permissions/proposals/listProposalCategoryPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+    throw new ProposalCategoryNotFoundError(resourceId);
+  }
+
+  const { error } = await hasAccessToSpace({
+    spaceId: proposalCategory.spaceId,
+    userId
+  });
+
+  if (error) {
+    return [];
```
mattcasey, line 43: we should probably throw an error here instead

lib/permissions/proposals/__tests__/listProposalCategoryPermissions.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+      userId: user.id
+    });
+
+    expect(foundPermissions.length).toBe(permissions.length);
+    foundPermissions.forEach((foundPermission) => {
+      expect(permissions).toContainEqual(expect.objectContaining(foundPermission));
+    });
+  });
+
+  it('should return an empty list for non-space members', async () => {
```
mattcasey, line 60: since these are lib methods, i think they should throw an error if they are given garbage, just as you are throwing an error if the id is invalid.

lib/permissions/proposals/listProposalCategoryPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+import type { PermissionCompute } from '../interfaces';
+
+import type { AssignedProposalCategoryPermission } from './interfaces';
+import { mapProposalCategoryPermissionToAssignee } from './mapProposalCategoryPermissionToAssignee';
+
+export async function listProposalCategoryPermissions({
+  resourceId,
+  userId
+}: PermissionCompute): Promise<AssignedProposalCategoryPermission[]> {
+  if (!userId) {
```
mattcasey, line 16: i think this is a bad pattern, where a function silently handles some edge case. Why not make userId required?

lib/permissions/proposals/listProposalCategoryPermissions.ts
```diff
@@ -0,0 +1,64 @@
...
+import { InvalidInputError } from 'lib/utilities/errors';
+import { isUUID } from 'lib/utilities/strings';
+
+import type { PermissionCompute } from '../interfaces';
+
+import type { AssignedProposalCategoryPermission } from './interfaces';
+import { mapProposalCategoryPermissionToAssignee } from './mapProposalCategoryPermissionToAssignee';
+
+export async function listProposalCategoryPermissions({
+  resourceId,
```
mattcasey, line 13: suggestion: just call this 'categoryId'. it's much easier to know what this function is asking for. 'resourceId' is a vague variable name, its' like using "data" for the return of a page or block model.

lib/permissions/proposals/upsertProposalCategoryPermission.ts
```diff
@@ -0,0 +1,115 @@
...
+    where: {
+      id: proposalCategoryId
+    },
+    select: {
+      spaceId: true
+    }
+  });
+
+  if (!proposalCategory) {
+    throw new DataNotFoundError(`Proposal category with id ${proposalCategoryId} not found`);
```
mattcasey, line 54: should this be ProposalCategoryNotFoundError?

lib/permissions/proposals/computeProposalPermissions/pfpStatusPrivateDraftVisibleOnlyByAuthor.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { typedKeys } from 'lib/utilities/objects';
+
+import { AvailableProposalPermissions } from '../availableProposalPermissions.class';
+import type { AvailableProposalPermissionFlags } from '../interfaces';
+
+import type { ProposalPfpInput } from './interfaces';
+import { isProposalAuthor } from './isProposalAuthor';
+
+export async function pfpStatusPrivateDraftVisibleOnlyByAuthor({
```
mattcasey, line 11: i dont know how these policy functions will be used but seems pretty easy to follow :)

lib/proposal/errors.ts
mattcasey: ah nvm i see thats how it's been for a while.. probably do it in a separate PR

mattcasey (changes requested):
Lots of code, but looks good... I assume you know what all these methods will be used for :P

One request is please find a better prefix than "pfp" for function and file names, or maybe use a folder as a namespace so the filenames don't get too long. Just removing pfp would be fine too, imo since all this code is already inside a folder called /permissions/proposals

lib/permissions/proposals/deleteProposalCategoryPermission.ts
motechFR: Added log.warn

lib/permissions/proposals/computeProposalPermissions/pfpStatusDiscussionEditableCommentable.ts
motechFR: Suggestion: I can move these to permissionFilteringPolicies subfolder

lib/permissions/buildComputePermissionsWithPermissionFilteringPolicies.ts
motechFR: Noted, will use the long form

lib/permissions/proposals/hasSpaceWideProposalReviewerPermission.ts
motechFR: Potential idea; space-wide: true

lib/permissions/proposals/__tests__/listProposalCategoryPermissions.spec.ts
motechFR: We'll need to change this test :)

lib/permissions/proposals/listProposalCategoryPermissions.ts
motechFR: We're using resourceId for cross-permissions subsystem compatibility

motechFR:
Note on PFP builder
- Don't throw an error if PFP grants extra permissions, just refuse to assign the new true permissions instad

prisma/migrations/20230223203450_proposal_permissions/migration.sql
```diff
@@ -0,0 +1,69 @@
...
+    "categoryOperations" "ProposalCategoryOperation"[],
+    "proposalOperations" "ProposalOperation"[],
+    "roleId" UUID,
+    "spaceId" UUID,
+    "public" BOOLEAN,
+
+    CONSTRAINT "ProposalCategoryPermission_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
```
mattcasey, line 38: @motechFR make sure to uncomment this before deploying to production

scripts/migrations/2023_02_23_migrateProposals.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  }, {} as Record<string, Pick<Proposal, 'id' | 'spaceId'>[]>)
+
+  const uniqueSpaces = Object.keys(bySpace);
+
+  for (let i = 0; i < uniqueSpaces.length; i+= concurrent) {
+    const generalCategory = await prisma.proposalCategory.findFirst({
+      where: {
+        spaceId: uniqueSpaces[i],
+        title: 'General'
```
mattcasey, line 128: It looks like we need a "defaultForSpace" property here? If we have a default category, I think users would expect it to behave the same as forum default category

scripts/migrations/2023_02_23_migrateProposals.ts
```diff
@@ -0,0 +1,205 @@
...
+}
+
+
+export function migrateProposals() {
+  // Ran this once, no duplicates detected
+  //detectSpacesWithDuplicateCategories();
+  
+  provisionGeneralProposalCategory();
+  assignProposalsToDefaultCategory();
+  disconnectProposalsFromChildren();
```
mattcasey, line 200: The only migration that seems necessary right now is assignSpaceDefaultPermissions(). disconnectProposalsFromChildren() is not necessary after we confirmed proposals always have type === 'proposal' and hence their children would not be visible in the sidebar anyway. This might actually surprise someone with pages they didn't know existed :P

scripts/migrations/2023_02_23_migrateProposals.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  assignSpaceDefaultPermissions();
```
mattcasey, line 202: make sure to await on these methods too!

mattcasey (approved):

PR #1746, opened 2023-02-17
set pagePermissions to null until user is loaded

mattcasey:
While working on a page with access via my user role, I would see the "Access denied" error page briefly sometimes on page load. I traced it to the fact that userSpaceRole was undefined inside getPagePermissions(), which is being called inside of <EditorPage /> as soon as the page is loaded. If we just return null, than the components can choose to wait to display errors or whatnot until real permissions are defined.

Alternative solution could be that the EditorPage component should wait til user is loaded itself, but I feel like it's the responsibility of getPagePermissions() since it is retrieving the user internally.

### DIFF
--- a/hooks/usePages.tsx
+++ b/hooks/usePages.tsx
@@ -34,7 +34,7 @@ export type PagesContext = {
   mutatePage: (updates: PageUpdates, revalidate?: boolean) => void;
   mutatePagesRemove: (pageIds: string[], revalidate?: boolean) => void;
   deletePage: (data: { pageId: string; board?: Block }) => Promise<PageMeta | null | undefined>;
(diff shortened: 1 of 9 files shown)


hooks/usePages.tsx
```diff
@@ -100,15 +100,18 @@ export function PagesProvider({ children }: { children: ReactNode }) {
...
     const computedPermissions = new AllowedPagePermissions();
 
     const targetPage = pages[pageId] ?? deletedPage;
 
     // Return empty permission set so this silently fails
-    if (!targetPage) {
+    if (!targetPage || !isUserLoaded) {
       return computedPermissions;
     }
+    if (!isUserLoaded) {
```
mattcasey, line 112: added this check that user exists, or return null

PR #1748, opened 2023-02-17
Fix sidebar not opening

valentinludu:
The problem was we were syncing a state with another state and on mobile the logic for that was different.


Devorein (approved):

hooks/useSettingsDialog.tsx
```diff
@@ -6,39 +7,47 @@ import type { TasksPageProps } from 'components/nexus/TasksPage';
...
 export function SettingsDialogProvider({ children }: { children: ReactNode }) {
+  const settingsModalState = usePopupState({ variant: 'dialog', popupId: 'settings-dialog' });
   const [activePath, setActivePath] = useState('');
   const [pathProps, setPathProps] = useState<PathProps | undefined>();
 
   const onClick = (_path?: string, props?: PathProps) => {
     setActivePath(_path ?? '');
-    setPathProps(props);
+    settingsModalState.open();
+    if (props) {
```
mattcasey, line 32: it was intentional to set pathProps to undefined when setting activePath, if no props are defined

valentinludu: resolved

mattcasey (approved):

PR #1749, opened 2023-02-17
Fixed forum post breadcrumb after closing

Devorein:


mattcasey (changes requested):
Please just remove the setting of page title when a post in a modal. You can think of the modal as a sub-state of the page you're on, which should be the forum category/database board, so it shouldn't be setting metadata for the top page. This also keeps it simple and that way the modal doesn't have to know what the previous page title was.

components/forum/components/PostPage/PostPage.tsx
```diff
@@ -96,7 +97,8 @@ export function PostPage({
   contentUpdated,
   setContentUpdated,
   showOtherCategoryPosts,
-  newPostCategory
+  newPostCategory,
+  insideModal = false
```
mattcasey, line 101: can u put this logix in [pagePath].ts? why do we need a new boolean logic to set page title here?

PR #1750, opened 2023-02-17
support pasting markdown into CharmEditor

mattcasey:
Includes a cleanup of some markdown code as well by putting it in `/lib/markdown` - it was in a mix of lib/pages and CharmEditor folders before, but I think going forward it makes sense to just leave React-stuff in CharmEditor since we use markdown in other places.

Used example from https://markdown-it.github.io/ for testing


components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
```diff
@@ -203,14 +200,10 @@ export default function PublishingForm({ onSubmit, page }: Props) {
...
-      const content = await generateMarkdown(
-        {
-          content: pageWithDetails.content,
-          title: pageWithDetails.title
-        },
-        false,
-        { members }
-      );
+      const content = await generateMarkdown({
+        content: pageWithDetails.content,
```
Devorein, line 204: Why are we removing the title of the page?

Devorein (approved):

Devorein:
Writing a few issues I encountered after copy/pasting md

1. Paragraph node after code is included inside code block
image
2. Clicking on links causes two tabs to open
3. Subscript and superscript doesn't work
image
4. Emojis and images aren't working

components/common/PageLayout/components/Header/components/Snapshot/PublishingForm.tsx
mattcasey: the 2nd argument is "withTitle" which means we were not even using the title. :shrug: I'm not changing that, just making it more explicit. This is why I don't like so many arguments :)

mattcasey:
Thanks for the QA @Devorein I am creating a card in backlog for now, so i can get this merged and focus on other stuff. I think this is at least better than what happens now

PR #1756, opened 2023-02-20
Invalid token error on discord oauth

Devorein:
There are still two issues with our wallet identity provider.
1. Wallet provider shows `Connect wallet` even though a wallet is already connected. Its due to `{account: undefined}` from `useWeb3React`. Since its undefined `storedAccount` is not updated from `null`. I was not able to figure out the reason though

2. "cancelled the wallet signature" error still shows up if a discord only user tries to connect their wallet that is already attached with another account (due to `Unique constraint failed on the fields: (`address`)`). We should lock the wallet after a user has logged out or show a better error message @mattcasey

### DIFF
--- a/components/login/WalletConnect.tsx
+++ b/components/login/WalletConnect.tsx
@@ -25,12 +25,11 @@ type Props = {
  */
 export function WalletConnect({ onSuccess }: Props) {
   const { account, connectWallet } = useWeb3AuthSig();
-  const { updateUser } = useUser();
+  const { updateUser, user: currentUser } = useUser();
 
   async function signSuccess(signature: AuthSig) {
     const updatedUser = await charmClient.addUserWallets([signature]);
-
-    updateUser(updatedUser);
+    updateUser({ ...currentUser, ...updatedUser });
     onSuccess();
   }
 
--- a/components/login/WalletConnect.tsx
+++ b/components/login/WalletConnect.tsx
@@ -25,11 +25,11 @@ type Props = {
  */
 export function WalletConnect({ onSuccess }: Props) {
   const { account, connectWallet } = useWeb3AuthSig();
-  const { updateUser, user: currentUser } = useUser();
+  const { updateUser } = useUser();
 
   async function signSuccess(signature: AuthSig) {
     const updatedUser = await charmClient.addUserWallets([signature]);
-    updateUser({ ...currentUser, ...updatedUser });
+    updateUser(updatedUser);
     onSuccess();
   }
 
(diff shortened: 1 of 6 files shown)


components/login/WalletConnect.tsx
```diff
@@ -25,12 +25,11 @@ type Props = {
...
 export function WalletConnect({ onSuccess }: Props) {
   const { account, connectWallet } = useWeb3AuthSig();
-  const { updateUser } = useUser();
+  const { updateUser, user: currentUser } = useUser();
 
   async function signSuccess(signature: AuthSig) {
     const updatedUser = await charmClient.addUserWallets([signature]);
-
-    updateUser(updatedUser);
+    updateUser({ ...currentUser, ...updatedUser });
```
msadura, line 32: if you check `updateUser` method, it does internally the same logic you added, so I do not see a point in adding that.

```
  const updateUser = useCallback((updatedUser: Partial<LoggedInUser>) => {
    setUser((u) => (u ? { ...u, ...updatedUser } : null));
  }, []);
```

pages/_app.tsx
```diff
@@ -263,32 +264,34 @@ function DataProviders({ children }: { children: ReactNode }) {
...
-                        </BountiesProvider>
-                      </MembersProvider>
-                    </WebSocketClientProvider>
-                  </IsSpaceMemberProvider>
-                </CurrentSpaceProvider>
-              </SpacesProvider>
-            </Web3AccountProvider>
-          </Web3ConnectionManager>
-        </Web3ReactProvider>
+        <DiscordProvider>
```
msadura, line 267: can you please explain this? I do not see using `useDiscordConnection` anywhere outside `IdentityProviders.tsx` file so I miss the point of moving it to be global provider

msadura (changes requested):

components/login/WalletConnect.tsx
Devorein: Thank you for catching that. It was a mistake on my end.

pages/_app.tsx
Devorein: Yes absolutely. After connecting discord we attach the auth token in the cookie. All the logic for connecting your discord account with charmverse was located inside `IdentityProvider` thus unless that component gets mounted discord connection wasn't triggered. Thats why it was relocated outside and moved inside `_app`.

mattcasey (approved):

pages/_app.tsx
msadura: thx, now it makes sense :D

msadura (approved):

PR #1758, opened 2023-02-21
Fix duplicating daylight abilities

msadura:
We've got a report from daylight that some of our abilities are duplicated. This pr fixes two issues:

1) Turned out you can click "done" on add token gate modal quickly, and it will create duplicated token gate in ou system. This is fixed by debouning button click

2) 1st point should fix duplicates in daylight, but to make it 100% sure I changed daylight create ability method (POST) to updasert ability (PUT) - this way ability will be updated rather than duplicated if we do anything wrong in the future.

### DIFF
--- a/lib/token-gates/daylight.ts
+++ b/lib/token-gates/daylight.ts
@@ -37,7 +37,7 @@ export async function addDaylightAbility(tokenGate: TokenGate) {
   }
 
   const params = {
-    method: 'POST',
+    method: 'PUT',
     headers: HEADERS,
     body: JSON.stringify({
       requirements: requirementsData.requirements,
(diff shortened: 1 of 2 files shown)


lib/token-gates/daylight.ts
```diff
@@ -37,7 +37,7 @@ export async function addDaylightAbility(tokenGate: TokenGate) {
   }
 
   const params = {
-    method: 'POST',
+    method: 'PUT',
```
mattcasey, line 40: i'm not sure how this actually changes anything on the server side. PUT/POST doesn't change behavior of the server on its own?

mattcasey (approved):
I think this is helpful,  I wonder if it is the issue they were having. (Like maybe someone was just creating multiple token gates?) Still, let's merge this and I'll see if Alex can notify them with our findings.

msadura:
> I think this is helpful, I wonder if it is the issue they were having. (Like maybe someone was just creating multiple token gates?) Still, let's merge this and I'll see if Alex can notify them with our findings.

according to their screenshots, "duplicated" token gates had the same time created, so I am quite sure it was because of duplicated token gate. I can actually try to check in db if they still have duplicated gates

mattcasey:
merging @msadura since this is a bug fix. I noticed you got some weird errors during test that DATABASE_URI is not defined, and posted about it in #devops. Please let me know if u see them again

PR #1763, opened 2023-02-22
Proposal comments 

msadura:
Proposal page comments.
NOTE: introduced model is in general "page comment" because it can be attached to any page type. We show them for proposals for now only.
I refactored and reused some parts of post comments.

NOTE2: Permissions are not implemented due to upcoming changes:
- server CRUD and vote methods are not secured
- ~~frontend uses hardcoded set of permission (same as post permission model)
Once proposal permissions PR will be ready I can update those parts.~~ - this one is handled by page permissions.

Screenshot 2023-02-23 at 10 36 30


components/common/comments/Comment.tsx
```diff
@@ -0,0 +1,289 @@
+import { useTheme } from '@emotion/react';
```
msadura, line 1: This is just copied old `PostComment` component with removed "Post" context - it will just work with any sort of comment, and it does need to know about comment type.

components/common/comments/CommentForm.tsx
```diff
@@ -0,0 +1,78 @@
+import { Stack, Box } from '@mui/material';
```
msadura, line 1: This is copy of `CreatePostForm` component with generic method to create comment

components/common/comments/utils.ts
```diff
@@ -0,0 +1,72 @@
+import type { CommentSortType } from 'components/common/comments/CommentSort';
+import type { GenericCommentWithVote, CommentWithChildren, GenericCommentVote } from 'lib/comments';
+
+export function processComments<T>(postComments: GenericCommentWithVote<T>[]): CommentWithChildren<T>[] {
```
msadura, line 4: Moved some common comment methods from `PostComment` and made types to be generic.

lib/pages/comments/__tests__/deletePostComment.spec.ts
```diff
@@ -0,0 +1,34 @@
+import type { Space, User } from '@prisma/client';
```
mattcasey, line 1: is this file supposed to be 'deletePageComment.spec.ts'?

lib/pages/comments/__tests__/updatePostComment.spec.ts
```diff
@@ -0,0 +1,41 @@
+import type { Space, User } from '@prisma/client';
```
mattcasey, line 1: updatePageComment.spec.ts

mattcasey (changes requested):
looks good.. one request: can we hide the comment input when creating a proposal? Or perhaps just hide it when it's in 'draft' state, because i don't know if we have a way to check if it was just created in the modal

msadura:
> looks good.. one request: can we hide the comment input when creating a proposal? Or perhaps just hide it when it's in 'draft' state, because i don't know if we have a way to check if it was just created in the modal

yeah there is no clear way to tell if it was "just created". I will do it for draft state. Thx for the idea!

components/common/comments/Comment.tsx
```diff
@@ -0,0 +1,288 @@
...
+    menuState.close();
+  }
+
+  async function onClickDeleteComment() {
+    menuState.close();
+    await handleDeleteComment(comment.id);
+  }
+
+  const isCommentAuthor = comment.createdBy === user?.id;
+  const canDeleteComment = (permissions?.delete_comments || isCommentAuthor) && !deletingDisabled;
```
motechFR, line 107: Looks good, this is our pattern for comments

components/common/comments/Comment.tsx
```diff
@@ -0,0 +1,288 @@
...
+      <Menu
+        {...bindMenu(menuState)}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'right' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'right' }}
+        onClick={(e) => e.stopPropagation()}
+      >
+        <Tooltip title={!canDeleteComment ? "You cannot edit another user's comment" : ''}>
+          <div>
+            <MenuItem
+              disabled={!canDeleteComment}
```
motechFR, line 258: Can we split this logic to also have canEditComment ?

Only the author should be able to edit their own comment

components/forum/components/PostPage/components/PostComment.tsx
```diff
@@ -133,197 +60,23 @@ export function PostComment({ post, comment, setPostComments, permissions }: Pro
...
+    <Comment
+      permissions={permissions}
+      comment={comment}
+      key={comment.id}
+      handleCreateComment={addComment}
+      handleUpdateComment={updateComment}
+      handleDeleteComment={deleteComment}
+      handleVoteComment={voteComment}
+      deletingDisabled={!!post?.proposalId}
+    />
```
motechFR, line 80: Cool refactor, I like this!

pages/api/pages/[id]/comments/[commentId]/index.ts
```diff
@@ -0,0 +1,94 @@
...
+    throw new DataNotFoundError(pageId);
+  }
+
+  const spaceRole = await hasAccessToSpace({
+    spaceId: page.spaceId,
+    userId
+  });
+
+  if (!spaceRole.success) {
+    throw new UserIsNotSpaceMemberError();
```
Comment (motechFR) [inline pages/api/pages/[id]/comments/[commentId]/index.ts:40]: We can delete these checks, and just use the comment.createdBy === userId staement below

pages/api/pages/[id]/comments/[commentId]/index.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+      }
+    }
+  });
+
+  if (!pageComment) {
+    throw new DataNotFoundError(`Comment with id ${commentId} not found`);
+  }
+
+  if (pageComment.createdBy === userId) {
```
Comment (motechFR) [inline pages/api/pages/[id]/comments/[commentId]/index.ts:85]: For deleting pattern, we can use isAdmin or comment === createdBy (make sure this is reflected in the UI)

lib/pages/comments/createPageComment.ts
```diff
@@ -0,0 +1,37 @@
...
+  contentText,
+  parentId,
+  pageId,
+  userId
+}: CreateCommentInput & {
+  pageId: string;
+  userId: string;
+}) {
+  const page = await prisma.page.findUniqueOrThrow({
+    where: { id: pageId }
```
motechFR, line 15: Performance improvement; you can use the select prop in prisma to only return page.id and reduce database and memory load (since we won't need the page comtent for example

where: { id: pageId },
select: {
 id: true
}

PR #1768, opened 2023-02-24
Get ENS details

valentinludu:
Populate a new user with details from his ens profile.


lib/users/prepopulateUserProfile.ts
```diff
@@ -1,39 +1,101 @@
...
+          id: user.id,
+          ...(ensDetails?.description ? { description: ensDetails?.description } : undefined),
+          social: {
+            ...(ensDetails?.discord ? { discordUsername: ensDetails?.discord } : undefined),
+            ...(ensDetails?.github ? { githubURL: ensDetails?.github } : undefined),
+            ...(ensDetails?.twitter ? { twitterURL: ensDetails?.twitter } : undefined)
+          }
+        }
       });
+    } catch (error) {
```
mattcasey, line 50: why do we need a try/catch around db update? This is a pattern i want to avoid unless it's clear there's a reason - ie if it's a unique field that is expected to fail sometimes

lib/users/prepopulateUserProfile.ts
```diff
@@ -1,39 +1,101 @@
...
+ * 4. First 5 NFTs as pinned items
+ * @param user we need the user id and user avatar to populate many
+ * @param ens the ENS name
+ */
+export async function prepopulateUserProfile(user: User, ens: string | null) {
+  const ensDetails = await getENSDetails(ens);
+
+  if (!user.avatar && ensDetails?.avatar && acceptedImageFormats.some((ext) => ensDetails?.avatar?.endsWith(ext))) {
+    try {
+      await updateProfileAvatar({
```
mattcasey, line 25: this method is super-heavy, seems it's designed for updating _NFT avatars_. we should just make a call to update the db directly (and remove try/catch)

lib/blockchain/getENSName.ts
```diff
@@ -19,3 +19,29 @@ export function getENSName(address: string) {
     return null;
   });
 }
+
+export async function getENSDetails(address?: string | null) {
```
mattcasey, line 23: please update 'address' since it's not actually an address. I think it'd be fine if you first called getENSName() inside of getENSDetails though, since most of the time we are working with addresses.

lib/users/prepopulateUserProfile.ts
```diff
@@ -1,39 +1,101 @@
...
+              id: nft.id,
+              userId: user.id,
+              isHidden: true,
+              isPinned: true,
+              type: 'nft'
+            }
+          });
+        })
+      );
+    } catch (error) {
```
mattcasey, line 97: remove try/catch around db update

mattcasey (approved):
seems fine - had a couple suggestions

lib/users/prepopulateUserProfile.ts
valentinludu: Removed!
Thanks for the advice.

lib/users/prepopulateUserProfile.ts
valentinludu: Removed!

lib/blockchain/getENSName.ts
valentinludu: Good point.

Because the details are returned only if the user has a valid ens name I will rename the address into ensName so it is less confusing.

valentinludu: I also added a validation so we know for sure that the string we send to this function is indeed a valid name.

PR #1773, opened 2023-02-27
Trigger notion import automatically

Devorein:
Currently you have to click on settings to trigger notion import, this PR addresses that issue.

### DIFF
--- a/hooks/useNotionImport.tsx
+++ b/hooks/useNotionImport.tsx
@@ -1,0 +1,184 @@
+import CompleteIcon from '@mui/icons-material/CheckCircle';
+import ErrorIcon from '@mui/icons-material/Dangerous';
+import WarningIcon from '@mui/icons-material/HourglassBottom';
+import Alert from '@mui/material/Alert';
+import Box from '@mui/material/Box';
+import CircularProgress from '@mui/material/CircularProgress';
+import Typography from '@mui/material/Typography';
+import { createContext, useContext, useEffect, useMemo, useState } from 'react';
+import { useSWRConfig } from 'swr';
+
+import charmClient from 'charmClient';
+import { useAppDispatch } from 'components/common/BoardEditor/focalboard/src/store/hooks';
+import { initialLoad } from 'components/common/BoardEditor/focalboard/src/store/initialLoad';
+import Modal from 'components/common/Modal';
+import { AUTH_CODE_COOKIE, AUTH_ERROR_COOKIE } from 'lib/notion/constants';
+import type { FailedImportsError } from 'lib/notion/types';
+import { deleteCookie, getCookie } from 'lib/utilities/browser';
+
+import { useCurrentSpace } from './useCurrentSpace';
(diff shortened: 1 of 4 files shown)


hooks/useNotionImport.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const space = useCurrentSpace();
+  const dispatch = useAppDispatch();
+  const notionCode = getCookie(AUTH_CODE_COOKIE);
+  const notionError = getCookie(AUTH_ERROR_COOKIE);
+
+  function closeModal() {
+    setModalOpen(false);
+  }
+
+  useEffect(() => {
```
valentinludu, line 54: I know you just copied this useEffect from another file, but I have a suggestion here to reduce code and use more from swr out of the box options:

You can useSWRMutation and keep the state of the promise there like this
`const {
    trigger,
    data,
    error: importError,
    isMutating
  } = useSWRMutation(
    '/notion/import',
    (_url, { arg }: Readonly<{ arg: { code: string; spaceId: string } }>) => charmClient.importFromNotion(arg),
    {
      onSuccess(_data) {
        if (space?.id) {
          mutate(`pages/${space.id}`);
          dispatch(initialLoad({ spaceId: space.id }));
        }
        if (_data.failedImports.length === 0) {
          showMessage('Notion workspace successfully imported');
          closeModal();
        }
        return _data?.failedImports ?? [];
      }
    }
  );`

The useEffect will have just the minimum requirement to trigger the mutation

`useEffect(() => {
    if (space?.id && notionCode) {
      deleteCookie(AUTH_CODE_COOKIE);
      setModalOpen(true);
      trigger({
        code: notionCode,
        spaceId: space.id
      });
    }
  }, [space?.id, notionCode]);`

For error messages you can have only one place where you handle the messages.
This way you will not need the useEffect to handle notion coockie error and another useEffect to handle the mutation error.
`const errorMessage = useMemo(() => {
    if (importError?.status === 504) {
      return {
        warning:
          'It can take up to an hour to import large Notion spaces. Your data will appear on the left navigation when the import is completed.'
      };
    }
    if (importError) {
      return {
        error:
          (typeof importError.message === 'string' ? importError.message : null) ||
          (typeof importError.error === 'string' ? importError.error : null) ||
          'Something went wrong. Please try again'
      };
    }
    if (notionError) {
      return {
        error: notionError
      };
    }
  }, [notionError, importError]);`

This preview of the code looks really weird, but if you put it inside the IDE it will make perfect sence.

hooks/useNotionImport.tsx
```diff
@@ -0,0 +1,184 @@
...
+            });
+          } else {
+            setNotionState({
+              loading: false,
+              error: err.message || err.error || 'Something went wrong. Please try again'
+            });
+          }
+        });
+    }
+  }, [space]);
```
valentinludu, line 90: You can use `space?.id` as a dependency to run it only if the id changes and not the whole space object

PR #1774, opened 2023-02-27
Feat/reevaluate token gate roles

msadura:
https://app.charmverse.io/charmverse/page-34945048868459705

Re-evaluating token gates and adding missing roles based on token gate conditions.

NOTE: There is an issue which is unrelated to this pr and I did not want to mix everything together. Looks like after page refresh reac-web3 does not store current connected account.


lib/token-gates/reevaluateRoles.ts
```diff
@@ -0,0 +1,58 @@
...
+        })) ?? []
+    });
+
+    const updatedSpaceMembership = await getSpaceMembershipWithRoles({ spaceId, userId });
+    const updatedUserRoles = updatedSpaceMembership?.spaceRoleToRole.map((r) => r.roleId) ?? [];
+    const newRoles = updatedUserRoles.filter((r) => !userRoles.includes(r)) || [];
+
+    return newRoles;
+  } catch (e) {
+    log.warn('Error reevaluating roles', e);
```
mattcasey, line 55: please log the userId and spaceId too so we have some chance of finding the issue.. :)

`log.warn('Erorr reevaluating roles', { error: e, userId, spaceId })`

components/_app/hooks/useSpaceGatesReevaluate.ts
```diff
@@ -0,0 +1,33 @@
...
+        const authSig = getStoredSignature();
+        if (!authSig) return;
+
+        const newRoles = await charmClient.tokenGates.reevaluateRoles({
+          spaceId: currentSpaceId,
+          userId: user.id,
+          authSig
+        });
+
+        if (newRoles.length) {
```
mattcasey, line 25: why only new roles? what if the user lost some roles?

mattcasey: Answering myself.. I think the backend will still correctly handle permissions, in the worst case the user might be able to access some pages but as soon as the page reloads, They won't see them anymore. So I think it's OK

mattcasey (approved):
Works! nice cleanup + tests :)

components/_app/hooks/useSpaceGatesReevaluate.ts
msadura: yeah, I think we wanted to leave "removing" as it was previously (via cron). This one only adds roles

PR #1775, opened 2023-02-27
Convert page to proposal

Devorein:
Create a new proposal based on the page's content and title, redirect to that proposal, make the page read-only and show proposal banner. @motechFR might need some help on how to disable changing the permissions for a page; currently its making use of a custom pfp for page permissions.


__e2e__/po/document.po.ts
```diff
@@ -29,6 +31,7 @@ export class DocumentPage {
     this.deletePermanentlyButton = this.page.locator('data-test=banner--permanently-delete');
     this.restoreArchivedButton = this.page.locator('data-test=banner--restore-archived-page');
     this.trashModal = this.page.locator('data-test=trash-modal');
+    this.charmEditor = this.page.locator('data-test=page-charmeditor').locator('div[contenteditable]').first();
```
mattcasey, line 34: i dont think it makes sense to call .locator() twice.. why not just `data-test=page-charmeditor div[contenteditable]`

__e2e__/convertPageToProposal.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  await documentPage.goToPage({
+    domain: space.domain,
+    path: generatedPage.path
+  });
+
+  const isEditable = await documentPage.isPageEditable();
+
+  expect(isEditable).toBe(false);
+
+  const postProposalBanner = page.locator('data-test=proposal-banner');
```
mattcasey, line 60: .locator('data-test=proposal-banner') and .locator('data-test=page-convert-proposal-action') should be an attribute of the documentPage. If you're using a `po` then you really shouldn't ever need `page` as well

pages/api/pages/[id]/convert-to-proposal.ts
```diff
@@ -34,7 +34,8 @@ async function convertToProposal(req: NextApiRequest, res: NextApiResponse<IPage
       type: true,
       spaceId: true,
       content: true,
-      title: true
+      title: true,
+      contentText: true
```
Comment (mattcasey) [inline pages/api/pages/[id]/convert-to-proposal.ts:38]: i dont think u need contentText? you are just returning PageMeta right

mattcasey (approved):

pages/api/pages/[id]/convert-to-proposal.ts
Comment (Devorein) [reply pages/api/pages/[id]/convert-to-proposal.ts:38]: `contentText` is required when creating the proposal page

image

Comment (mattcasey) [reply pages/api/pages/[id]/convert-to-proposal.ts:38]: sounds like we have room to do cleanup then - it's fine not to fix here but we should always question why the types are a certain way. There is no reason to be sending page content to the frontend anymore or return it in page creation methods. We always explicitly ask

Comment (mattcasey) [reply pages/api/pages/[id]/convert-to-proposal.ts:38]: ah, my bad. you do need it to duplicate of course.. :)

mattcasey (approved):

lib/permissions/pages/page-permission-compute.ts
```diff
@@ -141,3 +186,12 @@ export async function computeUserPagePermissions({
 
   return computedPermissions.operationFlags;
 }
+
+export const computeUserPagePermissions = buildComputePermissionsWithPermissionFilteringPolicies<
```
motechFR, line 190: great job Safwan! This is exactly how we should use this method

motechFR:
@Devorein Great job updating the compute method! Exactly how we should be using this.

One small detail: Can you please double check that a user is the only one who can edit their comment (in front and back end) ?

motechFR (approved):

Devorein:
Yes I've tested it out and the comments are working as expected both on fe and be

PR #1781, opened 2023-02-28
Move bounty related discussions inside bounty nexus task

Devorein:
Chris mentioned to have two separate sections one for bounty and the other for discussions and they should have a single scrollbar.
image


mattcasey (approved):
I guess this is fine for now - it's weird we have to filter out certain tasks on the frontend, right? But I seem to recall saying that we will completely get rid of Discussions in the future

Devorein:
I've asked the team for comment, if we get rid of the discussion tasks completely then I will have to make some more changes to the code.

PR #1782, opened 2023-02-28
Fix page title not updating

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-4890602409201261?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565)

### DIFF
--- a/components/[pageId]/EditorPage/EditorPage.tsx
+++ b/components/[pageId]/EditorPage/EditorPage.tsx
@@ -107,6 +107,10 @@ export default function EditorPage({ pageId: pageIdOrPath }: { pageId: string })
     [currentPage]
   );
 
+  useEffect(() => {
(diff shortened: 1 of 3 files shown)


components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -107,6 +107,10 @@ export default function EditorPage({ pageId: pageIdOrPath }: { pageId: string })
     [currentPage]
   );
 
+  useEffect(() => {
+    setTitleState(currentPage?.title || 'Untitled');
```
Comment (mattcasey) [inline components/[pageId]/EditorPage/EditorPage.tsx:111]: 'Untitled' is a default value - the components displaying title should have it set instaed. sometimes it is greyed out a little bit because it's untitled vs actually set

mattcasey (approved):

PR #1785, opened 2023-02-28
Gallery and board view for bounties

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=bfde0976-1489-4911-830f-d6373d80a195)
image


mattcasey (changes requested):
- "Ongoing" -> Open
- replace focalboard icons with bounty icons for Open / All tabs . Maybe use the bounty icon in the sidebar for "All"?
- completely remove whitespace for images. Ideally the cards should look the same as All status. We didn't intend to be quite so literal when we said use 'gallery view' - it was meant to reference the layout of cards that just sort horizontally and down and are not grouped by column

PR #1786, opened 2023-02-28
add xps engine id to user table

mattcasey:
We should have another table for XPS-related info, similar to what we do for Discord, Gmail, and wallets. But I'm waiting to work on the story for profiles so we know what we actually need, just want to get the token gating stuff working properly. In the meantime I'll just get the info by calling their api

### DIFF
--- a/lib/summon/addUserToSpace.ts
+++ b/lib/summon/addUserToSpace.ts
@@ -5,9 +5,10 @@ import { prisma } from 'db';
 type Props = {
   spaceId: string;
   userId: string;
+  userXpsEngineId: string;
 };
 
-export async function addUserToSpace({ spaceId, userId }: Props): Promise<Space | null> {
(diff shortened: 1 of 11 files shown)


lib/summon/addUserToSpace.ts
```diff
@@ -34,6 +35,14 @@ export async function addUserToSpace({ spaceId, userId }: Props): Promise<Space
       }
     });
   }
+  await prisma.user.update({
+    where: {
+      id: userId
+    },
+    data: {
+      xpsEngineId: userXpsEngineId
```
mattcasey, line 43: i'm using xpsEngineId to mirror the column in 'space' settings. It's technically accurate, even if they're called Summon I think the engine is called "XPS"... anyway, not 100% and am willing to migrate this later once we figure it out. But I think it's better than starting with a new name like "summonId" for users than what we have on spaces

PR #1787, opened 2023-03-01
Hide features

motechFR:
Allow users to hide features for the space from the sidebar

### DIFF
--- a/__e2e__/settings/hideFeatures.spec.ts
+++ b/__e2e__/settings/hideFeatures.spec.ts
@@ -1,5 +1,4 @@
-import { test as base, expect } from '@playwright/test';
-import { GlobalPage } from '__e2e__/po/global.po';
+import { expect, test as base } from '@playwright/test';
 import { v4 } from 'uuid';
 
 import { SpaceProfileSettings } from '../po/settings/spaceProfileSettings.po';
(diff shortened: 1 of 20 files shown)


msadura (commented):
Overall pr looks really good, and ou did an awesome e2e test, kudos!

I would consider changing one thing here - terminology 😅 For some reason "features blacklist" feels like some sort of "destructive and dramatic" action 😆 and does not reflect what it really does which is just "hiding features from the sidebar". It is not even disabling them, you can still access those via browser URL. This is why I would rename it to "hiddenFeatures" or something like that

components/settings/workspace/SpaceFeatureSettings.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Typography variant='caption'>
+        Admins can turn on and off the visibility of the following modules in the sidebar. The functionality will still
+        exist, but it won't be visible in the sidebar.
+      </Typography>
+      <Stack>
+        {typedKeys(featureLabels).map((feature) => (
+          <FormControlLabel
+            key={feature}
+            control={
+              <Switch
```
valentinludu, line 88: Can you look into this error you have in the console when you switch?

MUI: A component is changing the default checked state of an uncontrolled SwitchBase after being initialized.

valentinludu (commented):
Tested and works perfect.

Added a small comment.

__e2e__/settings/hideFeatures.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import { login } from '../utils/session';
+
+type Fixtures = {
+  spaceSettings: SpaceProfileSettings;
+  globalPage: GlobalPage;
+};
+
+const test = base.extend<Fixtures>({
+  spaceSettings: ({ page }, use) => use(new SpaceProfileSettings(page)),
+  globalPage: ({ page }, use) => use(new GlobalPage(page))
```
mattcasey, line 16: we could just have SpaceProfileSettings inherit from GlobalPage.. there's no reason to import multiple page objects really

__e2e__/settings/hideFeatures.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+});
+
+test('Space settings - toggle feature visibility', async ({ page, spaceSettings, globalPage }) => {
```
mattcasey, line 19: `page` is attached to the other page objects already, you don't have to import them separately. not a big deal how u do it but i think its less confusing to have a single page object

mattcasey (approved):
looks good, i agree w/marek the column name could be more specific

components/settings/workspace/SpaceFeatureSettings.tsx
motechFR: Thanks for the tip Valentin! I changed it to use checked attribute instead of defaultValue and the error went away

__e2e__/settings/hideFeatures.spec.ts
motechFR: Thanks, I wanted to do this yesterday, but it was already inheriting from SpaceSettingsModal. I've gone ahead and made SpaceSettingsModal inherit from GlobalPage

__e2e__/settings/hideFeatures.spec.ts
motechFR: Cleaned up and removed the page ref

components/common/PageLayout/components/Sidebar/Sidebar.tsx
```diff
@@ -264,35 +264,48 @@ export default function Sidebar({ closeSidebar, favorites, navAction }: SidebarP
...
-              <SidebarLink
-                href={`/${space.domain}/forum`}
-                data-test='sidebar-link-forum'
-                active={router.pathname.startsWith('/[domain]/forum')}
-                icon={<MessageOutlinedIcon fontSize='small' />}
-                label='Forum'
-                onClick={navAction}
-              />
+
+              {!space?.featureBlacklist.includes('member_directory') && (
```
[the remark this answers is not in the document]
motechFR: Cleaned this up, thanks for spotting

motechFR:
All good @msadura @mattcasey I've updated the naming convention to **hiddenFeatures**

Agree with you all, it's alot more explicit

PR #1789, opened 2023-03-01
import web3 react without ssr

msadura:
Disabling SSR locally for react web3 provider.

NOTE: This will have a side effect of all child of that component (which is most of the app) to be not server side rendered. It is not a perfect solution because it will work a bit differently than in production. Everyone should remember that, in case of debugging ssr issues locally, `ssr` flag should be set to true


motechFR (approved):
Tested locally and works, let's see what happens in staging

components/_app/Web3ReactProvider.ts
```diff
@@ -0,0 +1,10 @@
+import { Web3ReactProvider as DefaultWeb3ReactProvider } from '@web3-react/core';
+import dynamic from 'next/dynamic';
+
+import { isDevEnv } from 'config/constants';
+
+export const Web3ReactProvider = isDevEnv
```
mattcasey, line 6: I think i was actually running into this issue in production as well.. i'm also very nervous any time we do something special in dev vs production. Something may work fine for you and then fail in production

mattcasey (changes requested):
Look good but request that we don't do anything special for dev

PR #1790, opened 2023-03-01
Feat/proposal draft

msadura:
Merge "private_draft" and "draft" proposal status into single "draft" status with same functionality as "private_draft"


components/[pageId]/DocumentPage/components/ProposalProperties.tsx
```diff
@@ -86,8 +86,7 @@ export default function ProposalProperties({ proposalId, readOnly, isTemplate }:
     });
 
   const canUpdateProposalProperties =
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/components/ProposalProperties.tsx:88]: It's a bit unrelated but shouldn't we be using proposal permissions for this 🤔

lib/proposal/proposalStatusTransition.ts
```diff
@@ -1,9 +1,8 @@
 import type { ProposalStatus } from '@prisma/client';
 
 export const proposalStatusTransitionRecord: Record<ProposalStatus, ProposalStatus[]> = {
-  private_draft: ['draft', 'discussion'],
-  draft: ['private_draft', 'discussion'],
-  discussion: ['private_draft', 'draft', 'review'],
+  draft: ['draft', 'discussion'],
```
Devorein, line 4: Should we include draft here? Its already in `draft`

lib/proposal/proposalStatusTransition.ts
```diff
@@ -1,9 +1,8 @@
...
+  discussion: ['draft', 'draft', 'review'],
```
Devorein, line 5: There should be one `draft` here

Devorein (approved):

components/[pageId]/DocumentPage/components/ProposalProperties.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/ProposalProperties.tsx:88]: Good spot @Devorein 

@msadura We can use pagePermissions from usePagePermissions hook

components/proposals/components/ProposalStatusBadge.tsx
```diff
@@ -15,8 +14,7 @@ import { PROPOSAL_STATUS_LABELS } from 'lib/proposal/proposalStatusTransition';
 import type { BrandColor } from 'theme/colors';
 
 const PROPOSAL_STATUS_ICONS: Record<ProposalStatus, ReactNode> = {
-  private_draft: <LockOutlinedIcon />,
-  draft: <ModeEditOutlineOutlinedIcon />,
+  draft: <LockOutlinedIcon />,
```
motechFR, line 17: Suggestion; How about using the Edit icon that was previously on the draft level?

motechFR (approved):
@msadura looks good

I committed a quick cleanup on the permissions.

I see you've removed private_draft everywhere. This means we'll need to run the update script before the new app is deployed

msadura:
closing this one in favor of https://github.com/charmverse/app.charmverse.io/pull/1797 
(same code but fixed git history)

PR #1793, opened 2023-03-01
Show bounty discussions in bounty section for emails

Devorein:
image

### DIFF
--- a/lib/emails/templates/PendingTasks.tsx
+++ b/lib/emails/templates/PendingTasks.tsx
@@ -32,6 +32,7 @@ const buttonStyle = {
   background: '#009Fb7'
 };
 const h2Style = { lineHeight: '1.2em', fontSize: '24px', fontWeight: 'bold', marginTop: '10px' };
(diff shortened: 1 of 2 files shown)


lib/emails/templates/PendingTasks.tsx
```diff
@@ -161,6 +166,14 @@ export default function PendingTasks(props: PendingTasksProps) {
         {props.bountyTasks.slice(0, MAX_ITEMS_PER_TASK).map((proposalTask) => (
           <BountyTaskMjml key={proposalTask.id} task={proposalTask} />
         ))}
+        {bountyDiscussions.length > 0 && (
+          <MjmlText>
+            <div style={{ ...h3Style, fontWeight: 'bold', marginBottom: 5 }}>Bounty Discussions</div>
```
msadura, line 171: those additional styles should go to h3style const, otherwise there is no point in defining it

Devorein: Yeah that makes sense, I've updated the styles

PR #1805, opened 2023-03-03
Add “Duplicate” to context menu for all pages

Devorein:


pages/api/pages/[id]/duplicate.ts
```diff
@@ -27,16 +27,28 @@ async function duplicatePageRoute(req: NextApiRequest, res: NextApiResponse<IPag
...
+
+  const pagesMap: Record<string, PageMeta> = {};
+  Object.entries(pages).forEach(([_pageId, page]) => {
+    pagesMap[_pageId] = page;
+  });
+
+  rootPageIds.forEach((rootPageId) => {
+    const page = pagesMap[rootPageId];
+    if (page) {
+      trackUserAction('duplicate_page', {
```
Comment (mattcasey) [inline pages/api/pages/[id]/duplicate.ts:42]: there should never be a chance we track more than one user action event for a single user action, so this should not be inside of a map.

lib/pages/duplicatePage.ts
```diff
@@ -0,0 +1,64 @@
+import { prisma } from 'db';
+import type { BountyWithDetails } from 'lib/bounties';
+import { exportWorkspacePages } from 'lib/templates/exportWorkspacePages';
+import { importWorkspacePages } from 'lib/templates/importWorkspacePages';
+
+import type { DuplicatePageResponse } from './server';
+import { includePagePermissions, PageNotFoundError } from './server';
+
+export async function duplicatePage({
```
mattcasey, line 9: It is confusing that duplicatePage accepts a single pageId but returns a list of rootPageIds. I think this is because of the return signature of importWorkspacePages, but duplicatePage should have its own response type then.

lib/pages/duplicatePage.ts
```diff
@@ -0,0 +1,64 @@
...
+      page: {
+        include: includePagePermissions()
+      }
+    }
+  });
+
+  return {
+    pages,
+    rootPageIds,
+    bounties: bounties as BountyWithDetails[]
```
mattcasey, line 62: we don't have nested bounties, this should just be 'bounty'

pages/api/pages/[id]/duplicate.ts
```diff
@@ -27,16 +27,28 @@ async function duplicatePageRoute(req: NextApiRequest, res: NextApiResponse<IPag
...
+        userId,
+        spaceId: page.spaceId,
+        pageId: page.id,
+        type: page.type
+      });
+    }
   });
 
-  return res.status(200).json(pageWithPermissions);
```
Comment (mattcasey) [inline pages/api/pages/[id]/duplicate.ts:39]: this method needs to broadcast the new pages it creates.

Please remember; we should rely on websockets to update the page and not rely on sending the new page to the front-end. I will take 1 extra second of UI delay for a single flow of data.

lib/templates/exportWorkspacePages.ts
```diff
@@ -96,6 +115,27 @@ export async function exportWorkspacePages({
...
+          where: {
+            id: node.bountyId
+          },
+          include: {
+            permissions: true
+          }
+        })) ?? undefined;
+    } else if (node.proposalId && node.type === 'proposal' && !skipProposals) {
+      node.proposal =
+        (await prisma.proposal.findUnique({
```
mattcasey, line 130: why not just let proposal be null? I think collapsing the `await` inside `()` is pretty hard to read, and all you're doing is falling back to 'undefined'

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
```diff
@@ -436,6 +436,17 @@ function PageActionsMenu({ closeMenu, pageId, pagePath }: { closeMenu: () => voi
           </PageMenuItem>
         </div>
       </Tooltip>
+      {page && (
+        <DuplicatePageAction
+          page={page}
+          sx={{
+            '.MuiTypography-root': {
```
mattcasey, line 443: Looking at Notion, they don't bold the actions in the sidebar. Might be a bit out of scope, but we should just remove the bold effect in the page navigation.

mattcasey (changes requested):
left comments

lib/pages/duplicatePage.ts
Devorein: Yeah makes sense, but the function works for duplicating a single page, an array of pages or even a space, so there is a chance it might return multiple `rootPageIds`.

components/forum/components/PostDialog/PostDialog.tsx
```diff
@@ -105,8 +105,9 @@ export function PostDialog({ post, spaceId, onClose, open, newPostCategory }: Pr
       toolsMenu={
         post && (
           <PageActions
-            page={{ ...post, relativePath }}
+            page={{ ...post, relativePath, type: 'proposal', parentId: null }}
```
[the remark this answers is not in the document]
Devorein: 😬 Thank you for catching that. It should be disabled for forum posts

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
Devorein: Yeah that makes sense, updated styles

PR #1807, opened 2023-03-03
small ux changes for duplicate page

Devorein:


mattcasey (approved):

lib/pages/duplicatePage.ts
```diff
@@ -40,6 +41,14 @@ export async function duplicatePage({
     updateTitle: true
   });
 
+  if (rootPageIds.length > 1) {
+    log.info(`[duplicate]: Found multiple rootPageIds for a single page duplication`, {
```
mattcasey, line 45: i would make this log.error since it should never happen, that way we will see it in the logs

PR #1808, opened 2023-03-03
call page permissions less often

mattcasey:
1) pass down permissions instead of calling usePagePermissions at various levels. It seems like the way SWR's cache works is that it will use the previous response if one exsits, but it still makes another request to refresh the data anyway.
2) don't get page permissions for action menu until user opens it

As a user, you wouldn't notice anything because it doesn't affect page rendering, but it makes sense to me that this could be contributing to us running out of db connections

### DIFF
--- a/components/common/PageActionsMenu.tsx
+++ b/components/common/PageActionsMenu.tsx
@@ -53,7 +53,7 @@ export function PageActionsMenu({
   const charmversePage = members.find((member) => member.id === page.createdBy);
   const open = Boolean(anchorEl);
   const { formatDateTime } = useDateFormatter();
-  const { permissions: pagePermissions } = usePagePermissions({ pageIdOrPath: page.id });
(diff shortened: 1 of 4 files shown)


components/common/PageActionsMenu.tsx
```diff
@@ -53,7 +53,7 @@ export function PageActionsMenu({
   const charmversePage = members.find((member) => member.id === page.createdBy);
   const open = Boolean(anchorEl);
   const { formatDateTime } = useDateFormatter();
-  const { permissions: pagePermissions } = usePagePermissions({ pageIdOrPath: page.id });
+  const { permissions: pagePermissions } = usePagePermissions({ pageIdOrPath: open ? page.id : null });
```
motechFR, line 56: Smart optimisation! 👍

motechFR (approved):
Works great, generated 10 proposals, but didn't fire off a bunch of compute requests.

I think this might even help solve the timeouts issue

PR #1811, opened 2023-03-07
dont delete messages that fail from collabland

mattcasey:
With SQS, if you delete the message it's gone forever. Instead, we should let the process retry and AWS will eventually send it to a dead letter queue which I added to AWS already, called "prd-webhook-collabland-dlq.fifo".


mattcasey:
I'm merging this now because the delete message fails and is crashing the app

PR #1813, opened 2023-03-07
Dont let errors fail silently in collab worker

mattcasey:
Currently, we are "handling" all incoming messages even though our API was incorrect. This PR makes it so that the error can propagate. With SQS, we only want to remove a message from queue if it succeeds. SQS will re-send it a preset number of times and then put it in a dead letter queue so that we can debug and retry them manually later (and we can put an alarm on the size of this queue as well). The visibility timeout determines how often the message will appear as a available to consumers of the queue. (So if the time out is 30s, it will appear available again 30s after we first request it).

### DIFF
--- a/lib/collabland/collablandClient.ts
+++ b/lib/collabland/collablandClient.ts
@@ -143,16 +143,10 @@ export async function canJoinSpaceViaDiscord({
 }
 
 export async function getGuildRoles(discordServerId: string) {
-  try {
-    const allRoles = await fetch<ExternalRole[]>(`${COLLABLAND_API_URL}/discord/${discordServerId}/roles`, {
-      headers: getHeaders()
-    });
+  const allRoles = await fetch<ExternalRole[]>(`${COLLABLAND_API_URL}/discord/${discordServerId}/roles`, {
+    headers: getHeaders()
+  });
 
-    // filter out irrelevant roles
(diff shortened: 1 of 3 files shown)


lib/collabland/collablandClient.ts
```diff
@@ -143,16 +143,10 @@ export async function canJoinSpaceViaDiscord({
...
-  try {
-    const allRoles = await fetch<ExternalRole[]>(`${COLLABLAND_API_URL}/discord/${discordServerId}/roles`, {
-      headers: getHeaders()
-    });
-
-    // filter out irrelevant roles
-    return allRoles.filter((role) => role.name !== '@everyone' && !role.managed);
-  } catch (error) {
-    log.error(`Failed to fetch roles for server ${discordServerId}`, { error, apiDomain: COLLABLAND_API_URL });
+  const allRoles = await fetch<ExternalRole[]>(`${COLLABLAND_API_URL}/discord/${discordServerId}/roles`, {
```
mattcasey, line 146: removed the try/catch here - this is just wrapping the fetch call, i think it's better to let callers of this determine whether to swallow the error or propagate it. It might be a missing API key, it could also be that collab land's API is down.

motechFR: +1, good devX improvement 👍

PR #1816, opened 2023-03-07
wait until permissions are loaded to display CharmEditor in DocumentPage

mattcasey:
Page permissions are now loaded async. Once they load, the editor is re-mounted in editable state. This sort of thrashes the UI, but also triggers an error when the initial readonly flow tries to update the original view, which no longer exists:

image

### DIFF
--- a/components/[pageId]/Comments/PageComments.tsx
+++ b/components/[pageId]/Comments/PageComments.tsx
@@ -15,7 +15,7 @@ import type { IPagePermissionFlags } from 'lib/permissions/pages';
 
 type Props = {
   page: PageMeta;
-  permissions?: IPagePermissionFlags;
+  permissions: IPagePermissionFlags;
 };
 
 export function PageComments({ page, permissions }: Props) {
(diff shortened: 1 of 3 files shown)


components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -189,8 +189,10 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
             doc,
             plugins: _editor.view.state.plugins
           };
-          // Set document in prosemirror
-          _editor.view?.setProps({ state: EditorState.create(stateConfig) });
+          if (_editor.view && !_editor.view.isDestroyed) {
+            // Set document in prosemirror
+            _editor.view.setProps({ state: EditorState.create(stateConfig) });
+          }
```
motechFR, line 195: Nice, thanks for spotting solution to this CharmEditor one

components/[pageId]/Comments/PageComments.tsx
```diff
@@ -15,7 +15,7 @@ import type { IPagePermissionFlags } from 'lib/permissions/pages';
 
 type Props = {
   page: PageMeta;
-  permissions?: IPagePermissionFlags;
+  permissions: IPagePermissionFlags;
```
Comment (motechFR) [inline components/[pageId]/Comments/PageComments.tsx:18]: alot cleaner this way

motechFR (approved):

PR #1820, opened 2023-03-07
Feat/robust collabland role assign

msadura:
Improved flow for collabland roels assign. Previously it fetched roles from CB's api every time. Now it will call CB's api only when some roles are not present in our db. This should make most of user role updates without need to request external api once roles are synced.


mattcasey (approved):

__e2e__/utils/mocks.ts
```diff
@@ -125,16 +125,6 @@ export async function createUserAndSpace({
   };
 }
 
-export async function createDiscordUser({ userId, discordUserId }: { userId: string; discordUserId: string }) {
```
mattcasey, line 128: this is still used by some e2e tests, but i think it's fine if u just fix the import?

PR #1822, opened 2023-03-07
Add support for multi-select properties when adding card to a filtered table

mattcasey:


lib/focalboard/cardFilter.ts
```diff
@@ -136,7 +135,7 @@ class CardFilter {
         return { id: filterClause.propertyId };
       }
       case 'isNotEmpty': {
-        if (template.type === 'select') {
+        if (template.type === 'select' || template.type === 'multiSelect') {
```
mattcasey, line 138: actually a pretty simple fix :)

valentinludu (approved):

PR #1824, opened 2023-03-08
Table of contents

mattcasey:


Devorein:
Awesome work 🥳 

Just had a few questions

1. Page and user mentions inside headings are not transferred

CharmVerse
image


Notion
image

2. Notion shows Untitled for empty headings
3. We have disabled markdown export. Is that out of scope for now?

mattcasey:
Thanks @Devorein 

- regarding mentions, etc. i'm just using "textContent" so i'm also ignoring links and possibly other funny things. I think this limitation is OK? I wonder if we can extend mentions somehow to include the correct value in textContent in the future.

- markdown: ugh you caught me :) i should probably support this.

PR #1825, opened 2023-03-08
Fix save avatar failure

valentinludu:
Bug from Errors channel:
updateProfileAvatar() fails because the path of the image contains `?` symbol.

### DIFF
--- a/lib/users/prepopulateUserProfile.ts
+++ b/lib/users/prepopulateUserProfile.ts
@@ -20,7 +20,12 @@ const acceptedImageFormats = ['.jpg', '.jpeg', '.png', '.webp'];
 export async function prepopulateUserProfile(user: User, ens: string | null) {
   const ensDetails = await getENSDetails(ens);
 
-  if (!user.avatar && ensDetails?.avatar && acceptedImageFormats.some((ext) => ensDetails?.avatar?.endsWith(ext))) {
+  if (
+    !user.avatar &&
+    ensDetails?.avatar &&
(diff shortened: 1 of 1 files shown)


lib/users/prepopulateUserProfile.ts
```diff
@@ -20,7 +20,12 @@ const acceptedImageFormats = ['.jpg', '.jpeg', '.png', '.webp'];
 export async function prepopulateUserProfile(user: User, ens: string | null) {
   const ensDetails = await getENSDetails(ens);
 
-  if (!user.avatar && ensDetails?.avatar && acceptedImageFormats.some((ext) => ensDetails?.avatar?.endsWith(ext))) {
+  if (
+    !user.avatar &&
+    ensDetails?.avatar &&
+    acceptedImageFormats.some((ext) => ensDetails?.avatar?.endsWith(ext)) &&
```
mattcasey, line 26: To be generous, we could probably lowercase this right? since uppercase extensions should be valid

lib/users/prepopulateUserProfile.ts
```diff
@@ -20,7 +20,12 @@ const acceptedImageFormats = ['.jpg', '.jpeg', '.png', '.webp'];
...
+    !ensDetails.avatar.includes('?')
```
mattcasey, line 27: why is the "?" invalid? I could see a URL that includes image dimensions in the URL. What if we stripped the part afterwards instead?

mattcasey (changes requested):
left a question

PR #1826, opened 2023-03-08
Non token custom bounty rewards

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=508e1944-c747-4c11-a8f4-54e91c9fa32b)

image


motechFR:
Found a bug: As a bounty creator, when I click "accept application" nothing happens in the UI.

When I reload the page, I can see that the application was accepted.

__integration-tests__/server/pages/api/v1/bounties/index.spec.ts
```diff
@@ -71,7 +71,8 @@ describe('GET /api/v1/bounties', () => {
         reward: {
           amount: bountyWithPaidApplication.rewardAmount,
           chain: bountyWithPaidApplication.chainId,
-          token: bountyWithPaidApplication.rewardToken
+          token: bountyWithPaidApplication.rewardToken,
+          custom: null
```
motechFR, line 75: Do we need to add custom here?

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
```diff
@@ -63,8 +68,17 @@ export default function BountyProperties(props: {
     [isAmountInputEmpty, currentBounty]
   );
 
+  const [rewardType, setRewardType] = useState<RewardType>(isTruthy(currentBounty?.customReward) ? 'Custom' : 'Token');
+  const rewardTypeTabChangeRef = useRef<number>(0);
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:72]: Why are we using this ref? What are you trying to achieve?

__integration-tests__/server/pages/api/v1/bounties/index.spec.ts
Devorein: Yep otherwise the test fails

components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/BountyProperties/BountyProperties.tsx:72]: This is to make sure on load we switch to `Custom` tab if there is a custom reward. Ref is used to make sure we don't do it more than once. You can try the following:-
1. Add a custom reward
2. Reload the bounty, it should redirect to custom tab
3. Try to switch to Token tab, it will keep redirecting you back to custom one

mattcasey (approved):

PR #1828, opened 2023-03-08
Better link behavior

msadura:
- "hover" preview will now not dissapear when it is hovered
- slight preview UI update to look better
- moved "edit" view to use popover to act more stable

Screenshot 2023-03-08 at 21 13 49
Screenshot 2023-03-08 at 21 13 43


mattcasey (approved):
seems to be much better

PR #1832, opened 2023-03-10
Multi person property

Devorein:


scripts/migrations/2022_03_10_multiPersonProperty.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  })
+
+  for (const card of cards) {
+    const boardPersonProperty = boardPersonPropertyRecord[card.parentId];
+    if (boardPersonProperty) {
+      const cardFields = (card.fields as CardFields);
+      const currentPersonPropertyValue = cardFields.properties[boardPersonProperty.personPropertyId];
+      if (typeof currentPersonPropertyValue === "string") {
+        cardFields.properties[boardPersonProperty.personPropertyId] = currentPersonPropertyValue ? [currentPersonPropertyValue as string] : []
```
mattcasey, line 43: If this script was run more than once, I think it could end up converting each person property to a list of lists? I would change this to check typeof === 'string' just to be safe. No need to use type casting then, either.

mattcasey, line 43: i was gonna suggest checking typeof currentPersonPropertyValue, but then i saw you're doing it on line 42. i don't think u needed the type cast here

mattcasey (changes requested):
Works great, and the migration script looks correct. Let me know if u want me to run it for you...

One request: on focus, the input seems to appear below the name, but it should replace it. Also, there is a weird animation when closing it, where "No options available" floats to the top left as it disappears. 

In my loom, i try the PR first, then production, then Notion:
https://www.loom.com/share/4b833ff958284235a2aba03da02a1dd5

PR #1833, opened 2023-03-10
Fix sorting proposals

msadura:
Fix sorting proposals. https://discord.com/channels/894960387743698944/1083515846552469514

NOTE: There might be an issue with queried proposals, because I am not sure if proposals without page should be returned to the user? cc: @motechFR

### DIFF
--- a/components/proposals/hooks/useProposalSortAndFilters.ts
+++ b/components/proposals/hooks/useProposalSortAndFilters.ts
@@ -20,12 +20,15 @@ export function useProposalSortAndFilters(proposals: ProposalWithUsers[]) {
     filteredProposals = filteredProposals.filter((proposal) => proposal.categoryId === categoryIdFilter);
   }
 
-  filteredProposals = filteredProposals.sort((p1, p2) => {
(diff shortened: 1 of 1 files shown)


components/proposals/hooks/useProposalSortAndFilters.ts
```diff
@@ -20,12 +20,15 @@ export function useProposalSortAndFilters(proposals: ProposalWithUsers[]) {
...
 
-  filteredProposals = filteredProposals.sort((p1, p2) => {
-    const page1 = pages[p1.id];
-    const page2 = pages[p2.id];
-
-    return (page1?.createdAt ?? 0) > (page2?.createdAt ?? 0) ? -1 : 1;
-  });
+  filteredProposals = filteredProposals
+    // filter proposals with no according page as they might break sorting
+    .filter((p) => !!pages[p.id])
```
mattcasey, line 25: seems like this should be done in the query right, since we're paginating?

motechFR (approved):
Looks good, thanks for fixing that

mattcasey (approved):

PR #1841, opened 2023-03-13
Add a date property when adding a calendar view, if one does not exist

mattcasey:


Devorein (approved):

Devorein:
Things to point out:-
1. Notion doesn't even let you delete the date property for calendar view, it just removes the value (for the page if you are deleting from a card modal). 
2. They also don't show any pages with no value for that date property.

PR #1849, opened 2023-03-14
Tracking space Origin in mixpanel

msadura:
Screenshot 2023-03-14 at 09 49 42

Tracking origin (super api key name) in mixpanel space profile

### DIFF
--- a/lib/metrics/mixpanel/updateTrackGroupProfile.ts
+++ b/lib/metrics/mixpanel/updateTrackGroupProfile.ts
@@ -22,7 +22,7 @@ export async function updateTrackGroupProfile(space: Space, spaceOrigin?: string
 
 export function getTrackGroupProfile(space: Space, spaceOrigin?: string) {
   const spaceProfile: MixPanelSpaceProfile = {
-    $created: String(space.createdAt),
+    $created: space.createdAt,
     $name: space.name,
     'Space Created By': space.createdBy,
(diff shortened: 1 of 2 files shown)


lib/metrics/mixpanel/updateTrackGroupProfile.ts
```diff
@@ -4,19 +4,33 @@ import log from 'lib/log';
...
-export function getTrackGroupProfile(space: Space) {
-  return {
-    $created: space.createdAt,
+export function getTrackGroupProfile(space: Space, spaceOrigin?: string) {
+  const spaceProfile: MixPanelSpaceProfile = {
+    $created: String(space.createdAt),
     $name: space.name,
     'Space Created By': space.createdBy,
-    'Space Updated At': space.updatedAt
+    'Space Updated At': String(space.updatedAt)
```
mattcasey, line 28: dates are turned into string automatically when converted to JSON. If you're going to cast to a string your self you should use toISOString() to generate the proper UTC-agnostic format

mattcasey (changes requested):
date strings

lib/metrics/mixpanel/updateTrackGroupProfile.ts
msadura: it was just to satisfy typescript, so I simply fixed types instead

PR #1850, opened 2023-03-14
Hide empty property and tooltip

Devorein:
Before:
image

After:
image


mattcasey (approved):
we might be able to delete the Focalboard tooltip now :)

PR #1851, opened 2023-03-14
Fix/focus editor first line

msadura:
https://app.charmverse.io/charmverse/page-14046344584328407

https://app.charmverse.io/charmverse/page-7618451838156732

### DIFF
--- a/components/common/CharmEditor/utils.ts
+++ b/components/common/CharmEditor/utils.ts
@@ -1,6 +1,8 @@
+import type { EditorView } from '@bangle.dev/pm';
 import { safeInsert } from '@bangle.dev/utils';
 import type { Node } from 'prosemirror-model';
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/utils.ts
```diff
@@ -39,3 +41,18 @@ export const safeRequestAnimationFrame =
         (window as any).lastTime = currTime + timeToCall;
         return id;
       };
+
+export function insertAndFocusFirstLine(view: EditorView) {
+  const { tr, doc, schema } = view.state;
+  const firstNode = doc.child(0);
+
+  if (!firstNode?.isText && !firstNode?.isTextblock) {
```
mattcasey, line 49: Notion always creates a new line, even if it is text. I think we could keep it simple and do the same

mattcasey (approved):

PR #1852, opened 2023-03-14
Email ux improvements

motechFR:
Improve the modal UX when connecting via email

Allow user to disconnect email account

### DIFF
--- a/components/login/CollectEmail.tsx
+++ b/components/login/CollectEmail.tsx
@@ -15,12 +15,15 @@ type FormValues = yup.InferType<typeof schema>;
 
 type Props = {
   handleSubmit: (email: string) => void;
+};
+
+type DialogProps = Props & {
   isOpen: boolean;
   onClose: () => void;
 };
 
-export function CollectEmailDialog({ handleSubmit, isOpen, onClose }: Props) {
-  const { register, getValues, getFieldState, reset, watch } = useForm<FormValues>({
+export function CollectEmail({ handleSubmit }: Props) {
+  const { register, getValues, getFieldState, watch } = useForm<FormValues>({
     mode: 'onChange',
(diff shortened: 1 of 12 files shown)


components/login/CollectEmail.tsx
```diff
@@ -49,12 +53,41 @@ export function CollectEmailDialog({ handleSubmit, isOpen, onClose }: Props) {
...
+      {description && (
+        <Typography pl={0.2} mb={2}>
+          {description}
+        </Typography>
+      )}
+      <TextField {...register('email')} placeholder='email@charmverse.io' type='text' fullWidth sx={{ mb: 2 }} />
+      <Button disabled={!validEmail() || loading} onClick={submitEmail}>
         Submit
+        {loading && (
+          <span style={{ marginLeft: '6px' }}>
```
mattcasey, line 68: you can just use the `loading` prop on Button

mattcasey (approved):

lib/users/updateUsedIdentity.ts
```diff
@@ -69,7 +69,7 @@ export async function updateUsedIdentity(userId: string, identityUpdate?: Identi
 
   // Priority of identities: [wallet, discord, unstoppable domain, google account]
   if (user.wallets.length) {
-    updateContent.username = shortWalletAddress(user.wallets[0].address);
+    updateContent.username = shortWalletAddress(user.wallets[0].ensname ?? user.wallets[0].address);
```
valentinludu, line 72: I like this simple fix.

I would add something to it. If the user does not have an ensname the fallback is the full address.

I suggest to wrap it inside `shortWalletAddress()` so you don't make the full address the username.

valentinludu (approved):
Tested and looks good.

PR #1853, opened 2023-03-14
Fixed multi-person property issues in board view

Devorein:


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -112,12 +111,13 @@ function PropertyValueElement(props: Props): JSX.Element {
   } else if (propertyTemplate.type === 'person') {
     return (
       <UserProperty
-        memberIds={typeof propertyValue === 'string' ? [propertyValue] : propertyValue ?? []}
+        memberIds={propertyValue as string[]}
```
mattcasey, line 114: this looks like it will introduce an  error if the value is actually a string.

components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
```diff
@@ -61,9 +70,13 @@ function UserProperty(props: Props): JSX.Element | null {
   return (
     <StyledUserPropertyContainer
       onClick={() => {
-        setIsOpen(true);
-        setClicked(true);
+        // Only register click if display type is details or table
+        if (!props.readOnly && (props.displayType === 'details' || props.displayType === 'table')) {
```
mattcasey, line 74: Suggestion: make this another prop?

PR #1855, opened 2023-03-14
add templates and update nft hackathon template

mattcasey:
also adds scrollbars since we have so many now:
image


components/common/CreateSpaceForm/SelectNewSpaceTemplate.tsx
```diff
@@ -20,56 +19,79 @@ type SelectNewSpaceTemplateProps = {
...
+              onClick={() => onSelect(template)}
+              label={spaceContentTemplates[template]}
+              icon={templateIcon[template]}
+            />
+          </Grid>
+        ))}
+        <Grid item>
+          <Divider flexItem sx={{ mb: 2 }} />
+          <Typography variant='caption' color='secondary' textTransform='uppercase' fontWeight='bold'>
+            Transfer from another platform
```
motechFR, line 74: This is a nice demarcation

motechFR:
For the hackathon template, we have this "Page Not Found" showing up

Screenshot 2023-03-14 at 21 11 50

lib/templates/interfaces.ts
```diff
@@ -19,7 +19,11 @@ export interface PageWithBlocks {
     card?: Block;
   };
   votes?: (Vote & { voteOptions: VoteOptions[] })[];
-  proposal?: (Proposal & { authors: ProposalAuthor[]; reviewers: ProposalReviewer[] }) | null;
+  proposal?:
+    | (Proposal & {
+        category: null | (ProposalCategory & { proposalCategoryPermissions: ProposalCategoryPermission[] });
```
motechFR, line 24: Why are we importing category permissions?

I think we should make import / exports focused on content only

lib/templates/importWorkspacePages.ts
```diff
@@ -314,27 +308,47 @@ export async function generateImportWorkspacePages({
...
           bountyId: oldNewPageIdHashMap[node.id]
         });
       });
     } else if (node.type === 'proposal' && node.proposal) {
       // TODO: Handle cross space reviewers and authors
-      const { createdBy: proposalCreatedBy, authors, reviewers, ...proposal } = node.proposal;
+      const { category, ...proposal } = node.proposal;
+      let categoryId: string | undefined;
+      if (category) {
+        categoryId = space.proposalCategories.find((cat) => cat.title === category.title)?.id;
```
motechFR, line 327: I was wondering what would be the best way to handle this.

- Agree with matching the proposal categories by title.

- Disagree with creating new proposal categories. We could maybe do that thing of having a readonly general category and defaulting proposals to that one if no match found?

lib/templates/exportWorkspacePages.ts
```diff
@@ -130,8 +130,11 @@ export async function exportWorkspacePages({
           id: node.proposalId
         },
         include: {
-          authors: true,
-          reviewers: true
+          category: {
+            include: {
+              proposalCategoryPermissions: true
```
motechFR, line 135: I think we should ignore permissions in context of import/exports

mattcasey: done

motechFR (approved):
Looks good 👍

PR #1856, opened 2023-03-14
fix github regex

mattcasey:
Github names can include "-" which the regex doesn't match. I just tried to make it more relaxed by checking non-whitespace - we dont have to be so perfect imo

### DIFF
--- a/components/profile/components/SocialInputs.tsx
+++ b/components/profile/components/SocialInputs.tsx
@@ -23,7 +23,7 @@ export const schema = yup.object({
     .notRequired()
     .ensure()
     .trim()
-    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?github\.([a-z])+\/([A-Za-z0-9]{1,})+\/?$/i, 'Invalid GitHub link'),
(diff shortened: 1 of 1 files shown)


components/profile/components/SocialInputs.tsx
```diff
@@ -23,7 +23,7 @@ export const schema = yup.object({
     .notRequired()
     .ensure()
     .trim()
-    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?github\.([a-z])+\/([A-Za-z0-9]{1,})+\/?$/i, 'Invalid GitHub link'),
+    .matches(/^$|^http(?:s)?:\/\/(?:www\.)?github\.([a-z])+\/([^\s\\]{1,})+\/?$/i, 'Invalid GitHub link'),
```
Devorein, line 26: According to one library, the correct regex is this `/^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,38}$/i`

https://github.com/shinnn/github-username-regex/blob/master/module.js

Devorein (approved):

PR #1857, opened 2023-03-14
Fix proposal titles from DeepDAO and smoother loading on profiles

mattcasey:
Bug story: https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=c664e335-b495-4d38-8309-49af7c3a186d

I started out just hiding some random timezone UI while the data was still loading, but then also enabled some of the headings to appear while data is loading

image


valentinludu (approved):
Looks good.

Just something I noticed - When you access the page for the first time or force refresh it, the social media icons, description and timezone are appearing, disappearing and appearing again. Basically that part is flickering.
Here is a video of the behaviour.
https://www.loom.com/share/2e8b2040df3f4876bf516d92ac507713

Other then that, approved, since it's a small thing that we can deal with later.

mattcasey:
thanks @valentinludu I fixed it.. we had some logic to determine if user is 'public' or not which was going back and forth as data got loaded. I just removed this completely, it seemed like an optimization to avoid calling the backend sometimes

PR #1861, opened 2023-03-14
use member name as username inside of a space

mattcasey:


valentinludu (approved):
Tested and works perfect.

PR #1863, opened 2023-03-15
Add Proposal Templates to all space Templates

Devorein:
[Task Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=d15c9b90-8918-44da-bbd8-db8712faa723&cardId=5e32761e-6e21-4bae-bc4d-0a673a90fa45)

image

image

image


lib/templates/exportWorkspacePages.ts
```diff
@@ -53,10 +55,10 @@ export async function exportWorkspacePages({
   const rootPages = await prisma.page.findMany({
     where: {
       ...(rootPageIds ? { id: { in: rootPageIds } } : { spaceId: space.id, parentId: null }),
-      deletedAt: null,
-      type: {
+      deletedAt: null
+      /* type: {
```
mattcasey, line 59: can you just delete this code?

mattcasey (changes requested):
Please make the proposal templates _separate_ from the templates we're using. That would be very hard to maintain - we'd have to know exactly what to add/update any time we resync a template or add a new one. I think we have some approach for "Getting started" already where it is grabbed from seeddata and added to any tempalte that gets used

mattcasey (approved):

PR #1865, opened 2023-03-15
Invite guests

motechFR:
Allow inviting users to a space via their email


lib/google/firebaseApp.ts
```diff
@@ -1,12 +1,13 @@
...
+import { v4 } from 'uuid';
 
 import { googleWebClientConfig } from 'config/constants';
 
 class FirebaseApp {
   app: firebase.app.App;
 
   constructor() {
-    this.app = firebase.initializeApp(googleWebClientConfig, 'google-verify');
+    this.app = firebase.initializeApp(googleWebClientConfig, v4());
```
mattcasey, line 10: name is optional according to types... do we even need a random string?

mattcasey (approved):
code look sgood :) didnt test the whole e2e because we ran out of emails in staging

lib/google/firebaseApp.ts
motechFR: Added a comment, we'd get an error inside localhost due to hot module reload.

We could potentially also use the prisma client approach?

PR #1868, opened 2023-03-16
E2e for bounty comments

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-06584245854638437)

Only when you edit a comment you can click save with an empty string inside the text area.

The first case was not reproducable anymore, but I added an e2e test just to be sure.


mattcasey (approved):
This is a case where a unit test would be better for testing. You should be able to just load the input component and check whether the button is clickable based on input

PR #1871, opened 2023-03-16
Fix space template columns

valentinludu:
__title should be put only inside visiblePropertyIds if the view is a gellery type.


mattcasey (approved):

mattcasey:
@valentinludu I recently synced the hack one, so it means the data is probably wrong in production. Can you fix the production as well? Might be able to fix it using prisma studio

PR #1877, opened 2023-03-17
Combine title and description editing in the forum

motechFR:


Devorein:
Everything works apart from the description value not being stored after removing it.

Devorein (approved):

PR #1878, opened 2023-03-17
Close settings dialog on url change

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-08668865101642687)

### DIFF
--- a/hooks/useSettingsDialog.tsx
+++ b/hooks/useSettingsDialog.tsx
@@ -1,8 +1,7 @@
-import path from 'path';
-
 import { usePopupState } from 'material-ui-popup-state/hooks';
+import { useRouter } from 'next/router';
 import type { ReactNode } from 'react';
-import { useMemo, createContext, useContext, useState } from 'react';
+import { useEffect, useMemo, createContext, useContext, useState } from 'react';
 
 import type { TasksPageProps } from 'components/nexus/TasksPage';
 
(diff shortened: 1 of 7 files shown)


hooks/useSettingsDialog.tsx
```diff
@@ -54,6 +53,21 @@ export function SettingsDialogProvider({ children }: { children: ReactNode }) {
     settingsModalState.close();
     setActivePath('');
   };
+
+  useEffect(() => {
+    const close = () => {
+      if (!router.route.endsWith('/[domain]')) {
```
mattcasey, line 59: what is this case where we wouldn't want to  close the modal?

mattcasey: we should also not close it if task is set as a query param, for example "/nexus?task=discussion"

valentinludu: Sooo... 
1. The logic for opening the modal on ?task=something works without needing any extra logic from my end. The logic is inside LoginPage and it works great.

3. The explanation for my logic:
If the user has visited us before and has al last visited page in localStorage he will be redirected like this
app.charmverse.io/nexus?task=something -> /domain/page
And the settings modal will open.

If the user is a first timer, that means he does not have a last visited page, the app will take him from app.charmverse.io/nexus?task=something -> /domain -> /domain/page
This means open modal and close modal before the user gets to the last link.
It is an edge case but it can happen to users that switch browsers and are redirected through 2 links.
This is why I don't close the modal if the link changed from /[domain] to /[domain]/[page]

mattcasey (approved):
works well

PR #1879, opened 2023-03-17
Fix sort&filter items in database

valentinludu:
+ Redone sort&filter dropdown
+ fix a warning in styled prop

To test: databases and inline databases & other types if exist
*The behaviour should be the same.

I just added mui because of the popover instead of a dropdown that can't be on top of an overflowed table


mattcasey (approved):
works great

PR #1882, opened 2023-03-20
Do not refresh posts list when new posts banner is opened

msadura:
https://app.charmverse.io/charmverse/page-46571373385070425

Issue was caused by refreshing data on focus


motechFR (approved):
Tested and works

PR #1883, opened 2023-03-20
Show email settings checkboxes in onboarding modal

Devorein:
image

### DIFF
--- a/components/members/MemberEmailForm.tsx
+++ b/components/members/MemberEmailForm.tsx
@@ -2,7 +2,6 @@ import { yupResolver } from '@hookform/resolvers/yup';
 import { Checkbox, FormControlLabel, FormGroup, TextField, Typography } from '@mui/material';
 import { Stack } from '@mui/system';
 import type { ChangeEvent } from 'react';
-import { useCallback } from 'react';
 import { useForm } from 'react-hook-form';
 import * as yup from 'yup';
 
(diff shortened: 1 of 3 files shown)


components/members/MemberEmailForm.tsx
```diff
@@ -27,27 +29,36 @@ export function MemberEmailForm({ onClick, readOnly = false }: { onClick: VoidFu
...
-        }
-      }
-    }, 300),
-    [readOnly]
-  );
+  const onChange = useCallback(async (event: ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.type === 'checkbox' ? event.target.checked : event.target.value;
+    setValue(event.target.name as keyof FormValues, value);
+    await trigger();
+  }, []);
```
mattcasey, line 48: shouldn't trigger and setValue be added to list of dependencies here? i would suggest just don't use useCallback tho

components/members/MemberEmailForm.tsx
```diff
@@ -57,25 +68,34 @@ export function MemberEmailForm({ onClick, readOnly = false }: { onClick: VoidFu
...
-          const val = event.target.value;
-          setValue('email', val);
-        }}
+        onChange={onChange}
       />
+      <FormGroup>
+        <FormControlLabel
+          disabled={!!errors.email || email.length === 0}
+          control={<Checkbox {...register('emailNotifications')} checked={emailNotifications} onChange={onChange} />}
+          label='Receive email updates on mentions, comments, post and other things in CharmVerse.'
```
mattcasey, line 82: i think we should shorten this by one or two words so it all fits on one line

PR #1884, opened 2023-03-20
Custom reward bounty improvements

Devorein:
- [x] add mark as paid to context menu (converts all submissions to `paid`)
- [x] after approving application, hide "Approve application" button
- [x] Remove address to get paid for custom reward bounties
- [x] Fix ux issue of adding custom token and switching to token tab
- [x] Fix wrapping of custom bounty reward
- [x] Show mark paid button for each submission
- [x] Disable `mark paid` for bounty context when atleast one bounty is not `complete | paid`
- [x] Bounty page menu actions in full page view (mark completed, mark paid) 
- [x] edit and view my submission when it is in review
- [x] reject when the user is in "awaiting submission" (ApplicationStatus / inProgress) stage
- [x] enforce wallet address requirement only if the bounty has token reward


motechFR:
So far, some UX items

- We should be able to reject when the user is in "awaiting submission" stage. This doesn't need to pop open the submission, just allow user to confirm

- I should be able to edit and view my submission when it is in review

motechFR:
We need to ensure the bounty interaction options show in the case of full page view ie. "mark complete, mark paid" etc

lib/bounties/__tests__/markBountyAsPaid.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      userId: member.id
+    });
+
+    const updatedBounty = await markBountyAsPaid(bounty.id);
+
+    expect(updatedBounty.status).toBe('paid');
+    expect(updatedBounty.applications.every((application) => application.status === 'paid')).toBeTruthy();
+  });
+
+  it('should fail if the atleast one bounty submission is neither complete nor paid', async () => {
```
motechFR, line 49: We should test the case where we have more states of applications


Case 1
- Paid
- Complete
- In progress
- Rejected

// Here all the applications are in a "terminal" state
Case 2
- Paid
- Complete
- Rejected

// What happens if there are still some open applications?
Case 3
- Paid
- Complete
- Rejected
- Applied

PR #1885, opened 2023-03-20
Comment form browser UT

valentinludu:
+ Fixed csvExport browser tests


mattcasey (approved):

testing/mocks/user.ts
```diff
@@ -0,0 +1,34 @@
+import { v4 as uuid } from 'uuid';
+
+import type { LoggedInUser } from 'models';
+
+export const mockedUser: LoggedInUser = {
```
mattcasey, line 5: suggestion: this should be a method, so that the id, paths, and dates are generated new each time. might not be useful now but could be in the future.

testing/customRender.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { render } from '@testing-library/react';
+import type { ReactNode } from 'react';
+
+import type { IContext } from 'hooks/useUser';
+import { UserContext } from 'hooks/useUser';
+import { createThemeLightSensitive } from 'theme';
+
+import { mockedUser } from './mocks/user';
+
+const theme = createThemeLightSensitive('light');
```
mattcasey, line 11: i think it would be better to call this inside customRenderWithContext. That way each theme instance is unique, allowing tests to have custom theme settings. I think it's good practice to not call functions immediately when a file is loaded, even if it's to generate a config.

testing/mocks/user.ts
valentinludu: Done

testing/customRender.tsx
valentinludu: Done

mattcasey:
Finally got the test around CharmEditor working @ValentinL#3950 ... 
- added support to mock out .svg files - this is why ErrorBoundary was hiding the errors
- had to patch skiff-org/prosemirror-tables. It was trying to append to a DOM element that doesn't exist. (not sure why this was not happening only in tests!)

PR #1886, opened 2023-03-20
Fix Discord roles import

motechFR:
This PR fixes an issue where Discord Role import was not happening.

Since we have moved the settings to a modal and don't have a dedicated page, the workflow wasn't getting triggered.

2 key changes
- Hoisted the useDiscordImport to top of app
- Passed the redirect from currentSpace instead of the browser url


mattcasey (approved):
approved but getting an auth error when i test locally -- the settings modal appears, and we make a request so i assume it[s working, and we can test in prod

PR #1889, opened 2023-03-21
Clarify the label for linked pages

motechFR:


mattcasey (approved):
we originally had a way to create a new child page, I think that is the reason for the placeholder. but AFAICT this is just link to other pages now :)

motechFR:
> we originally had a way to create a new child page, I think that is the reason for the placeholder and also why it's called 'nested page'. but AFAICT this is just link to existing pages now :)

Thanks for that, good to know!

PR #1890, opened 2023-03-21
Drag and drop board and card properties

Devorein:
Properties submenu in board sidebar
image

New menu for card and board properties
image

CharmVerse style confirm dialog when deleting/renaming properties
image

Drag and drop for card properties
image


components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -185,70 +281,78 @@ function CardDetailProperties(props: Props) {
 
   return (
     <div className='octo-propertylist'>
-      {board.fields.cardProperties.map((propertyTemplate: IPropertyTemplate) => {
-        const propertyValue = card.fields.properties[propertyTemplate.id];
+      {board.fields.visibleCardPropertyIds.map((cardPropertyId) => {
```
mattcasey, line 284: `cardProperties` is already a sorted array, why do we need another array of just the ids? That would save a lot of data and overhead of updating the two fields together.

mattcasey (changes requested):
noticed some bugs:
- I can't drag and drop a Text type property
- In card view, can we include the drag handlebar on the far left of properties on hover? I think without them, there's no way to know that you can sort the properties
- instead of a new field, can we just use the existing order in `fields.cardProperties`? That also means we won't need to keep them in sync or run a migration.

PR #1896, opened 2023-03-22
Added static links to Page

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-6394982161729259)

I also added forum categories in the dropdown list since it was related and very easy to add.

Refactored a bit the sidebar static links.


components/common/PageLayout/components/Sidebar/Sidebar.tsx
```diff
@@ -271,50 +269,28 @@ export default function Sidebar({ closeSidebar, favorites, navAction }: SidebarP
...
+                ) {
+                  return (
+                    <SidebarLink
+                      key={page.path}
+                      href={`/${space.domain}/${page.path}`}
+                      active={router.pathname.startsWith(`/[domain]/${page.path}`)}
+                      icon={<PageIcon icon={null} pageType={page.path} />}
+                      label={page.title}
+                      onClick={navAction}
+                      data-test={`sidebar-link-${page.path}`}
```
valentinludu, line 287: I have no idea how did the e2e test pass before, because this button did not have a data-test attr.

components/common/PageLayout/components/Sidebar/utils/staticPages.tsx
```diff
@@ -0,0 +1,23 @@
+import type { Feature } from '@prisma/client';
+
+export enum StaticPagesPath {
+  members = 'members',
+  forum = 'forum',
+  bounties = 'bounties',
+  proposals = 'proposals'
+}
+
+export type StaticPagesType = `${StaticPagesPath}`;
```
motechFR, line 10: An alternative syntax we often use: type X = keyof typeof Y

motechFR:
UX issue: Downwards scrolling only applies to page, not the static pages

components/common/PageLayout/components/Sidebar/utils/staticPages.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export type StaticPagesList = {
```
motechFR, line 12: One small suggestion, since we are using this in interoperable fashion with pages for the nested page feature, we could define this interface as

Pick<Page, 'title' | 'path'> & {feature: Feature} to make this coupling a bit more explicit

components/common/PageLayout/components/Sidebar/Sidebar.tsx
motechFR: If you wanted, we could wrap the e2e test body from l:27 (checking link is visible) till the end in a for loop and test all 4 of these menu options to make sure they work

motechFR (approved):
Approving PR as it works, double check with product team if we want to scroll between page types, but sounds like desirable behaviour

components/common/PageLayout/components/Sidebar/utils/staticPages.tsx
valentinludu: Niceeeee. I did not know that it's the same thing with what I did. Thanks.

components/common/PageLayout/components/Sidebar/Sidebar.tsx
valentinludu: Good idea. Just pushed the improved e2e test

valentinludu:
> UX issue: Downwards selection with keyboard only applies to page, not the static pages (I return to the top option after)
> 
> Screenshot 2023-03-22 at 16 15 40

Fixed!

Thanks for finding that.

__e2e__/settings/hideFeatures.spec.ts
```diff
@@ -34,9 +38,7 @@ test('Space settings - toggle feature visibility', async ({ spaceSettings }) =>
 
   expect(!!classes?.match('Mui-checked')).toBe(true);
   await spaceSettings.proposalsVisibilityToggle.click();
-  await spaceSettings.page.waitForTimeout(500);
-
-  await spaceSettings.page.waitForTimeout(200);
+  await spaceSettings.page.waitForTimeout(2000);
```
mattcasey, line 41: 2000 is too long for a timeout. Instead, it is better to use waitForFunction and declare what you are waiting for.

components/common/PageLayout/components/Sidebar/utils/staticPages.tsx
mattcasey, line 1: this file could just go in lib/pages or something

components/common/CharmEditor/components/PageList.tsx
```diff
@@ -19,52 +31,80 @@ export default function PagesList({
...
+      title: page.title,
+      type: page.type,
+      icon: page.icon
+    }));
+
+    const memoForumCategories: AllPagesProp[] = (forumCategories || []).map((page) => ({
+      id: page.id,
+      path: page.path || '',
+      hasContent: true,
+      title: page.name,
```
mattcasey, line 57: I think we should say `Forum > {page.name}` ? otherwise it's hard to find all your forum categories, and not obvious that this isn't just a regular page

mattcasey (changes requested):
Works well, using 'id' for everything feels pretty hacky but I can't come up with any real issue. It works well only because we happen to have all the categories and pages in memory, but if you were trying to read this someplace else you would have no way to know if it was a category or a page, and you'd have to guard against non-uuid values before running a query. Maybe add a 'type' attribute, even if we aren't reading it?

Please fix the test - increasing timeouts is not really addressing the issue, it could always take longer than 2 seconds.

__e2e__/settings/hideFeatures.spec.ts
valentinludu: Instead of that complicated check of classes and waiting for classes to change I simplified the logic and I am just waiting for the input to be checked or not.
e.g: `await expect(spaceSettings.proposalsVisibilityToggle).toBeChecked();`

components/common/CharmEditor/components/PageList.tsx
valentinludu: Yep. It will be a better UX for the user.

Thanks for that. I added it.

__e2e__/settings/hideFeatures.spec.ts
valentinludu: I deleted all the waitFor stuff.

PR #1898, opened 2023-03-22
Reduce pages size

motechFR:


charmClient/apis/pagesApi.ts
```diff
@@ -15,6 +15,11 @@ export class PagesApi {
     return http.GET<PageMeta[]>(`/api/spaces/${spaceId}/pages`, { meta: true });
   }
 
+  getDeletablePages(spaceId: string) {
+    // meta=true - TEMP param to keep backward compatibility with old clients
+    return http.GET<PageMeta[]>(`/api/spaces/${spaceId}/pages`, { meta: true, deletable: true });
```
mattcasey, line 20: lets remove 'meta' - we dont have these 'old clients' anymore

charmClient/apis/pagesApi.ts
```diff
@@ -15,6 +15,11 @@ export class PagesApi {
     return http.GET<PageMeta[]>(`/api/spaces/${spaceId}/pages`, { meta: true });
   }
 
+  getDeletablePages(spaceId: string) {
```
mattcasey, line 18: deletablePages is just another way to say 'archivedPages'. i think we should stick with the latter

lib/pages/server/getAccessiblePages.ts
```diff
@@ -96,130 +97,44 @@ function selectPageFields(meta: boolean) {
...
-      where: {
-        spaceId,
-        permissions: {
-          some: {
-            public: true
-          }
-        }
-      }
-    };
+// Assumes the user is a space member - External caller should check the user is a space member and null out userId if not
```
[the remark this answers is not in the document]
mattcasey: maybe we  should have a required field "isSpaceMember"? similar to isGuest, isAdmin, etc. To reduce the chance of this being called improperly

charmClient/apis/pagesApi.ts
motechFR: I've updated the method to go for archived pages.

We were only consuming this in the Trash Modal, so I've replicated previous UI logic where archived pages would only show archived pages user can delete

lib/pages/server/getAccessiblePages.ts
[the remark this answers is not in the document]
motechFR: Agree this could be easier. Since we already get the spaceRole in getAccessiblePages, I updated this method to receive the spaceRole, and it can then do validation

mattcasey (approved):

PR #1899, opened 2023-03-22
Remove bots from member directory, make nft template default for coll…

msadura:
Updates requested in discord:
- use NFT template for spaces created via integration
- do not show bots in member directory


pages/api/spaces/[id]/members/index.ts
```diff
@@ -25,7 +25,10 @@ async function getMembers(req: NextApiRequest, res: NextApiResponse<Member[]>) {
   if (error) {
     const spaceRoles = await prisma.spaceRole.findMany({
       where: {
-        spaceId
+        spaceId,
+        user: {
```
Comment (mattcasey) [inline pages/api/spaces/[id]/members/index.ts:29]: we want bots to appear in the list of members. they should just be filtered out where they should not appear (roles page, member directory, etc). this is because we need the context for showing metadata like createdBy, updatedBy, etc.

PR #1904, opened 2023-03-22
add alias for prosemirror-tables so we dont have to patch dependencies

mattcasey:
This was an error blocking tests that imported remirror - we are going to try writing them w/o remirror, but this also removes the need to patch bangle.dev.


components/integrations/components/IdentityProviders.tsx
```diff
@@ -199,7 +199,7 @@ export function IdentityProviders() {
             loading={isConnectingToTelegram || isDisconnectingTelegram}
             disabled={cannotDisconnect}
             connected={true}
-            actions={<MenuItem onClick={disconnectFromTelegram}>Disconnect</MenuItem>}
+            actions={<MenuItem onClick={() => disconnectFromTelegram()}>Disconnect</MenuItem>}
```
mattcasey, line 202: Typescript caught this, some libs got updated because i reinstalled package-lock.json

PR #1910, opened 2023-03-24
Improve Quick Find search

mattcasey:
- fixed debounce/throttle on front-end so that we always show the latest result
- sort matches by title before matches by content
- removed deprecated 'meta' attribute from pages endpoint


lib/pages/server/getAccessiblePages.ts
```diff
@@ -259,28 +169,51 @@ export async function getAccessiblePages(input: PagesRequest): Promise<IPageWith
...
 
+  // ref: https://www.postgresql.org/docs/12/functions-textsearch.html
+  // ref: https://www.postgresql.org/docs/10/textsearch-controls.html
+  // prisma refs: https://github.com/prisma/prisma/issues/8950
   const formattedSearch = input.search
     ? `${input.search
         .split(/\s/)
         .filter((s) => s)
-        .join(' & ')}:*`
+        .join(' <-> ')}:*`
```
mattcasey, line 179: this just enforces the order of words, which i think makes sense for the type of search users are doing

Devorein (approved):

components/common/PageLayout/components/SearchInWorkspaceModal.tsx
```diff
@@ -227,34 +222,38 @@ function getSearchResults(params: {
...
-      .map((page) => ({
-        name: page.title || 'Untitled',
-        path: getPagePath(page, params.allPages),
-        link: `/${params.spaceDomain}/${page.path}`,
-        type: ResultType.page,
-        id: page.id
-      }))
-      .sort((item1, item2) => (item1.name > item2.name ? 1 : -1))
-  );
+  return charmClient.pages.searchPages(params.spaceId, params.query).then((pages) => {
```
valentinludu, line 227: An idea:

Instead of slice(0,50)
Why not adding a third parameter to charmClient.pages.searchPages as a number (e.g. 50)
Prisma will return maximum 50 results.
Also I think this will be good for performance as well, both BE and FE.

valentinludu: Another idea:

I noticed that while the user types for the first time, the user can see only: No results. Then, when the response comes back, the list is populated.

I would use useSWR to get the isLoading variable and show a loading spinner to the user. 
In addition you can add the keepPreviousData option to  not make multiple requests for a previous search term.
https://swr.vercel.app/docs/advanced/understanding#return-previous-data-for-better-ux

Probably there is a way to debounce that, but not seeing it in the docs.

valentinludu (approved):
Tested and works!

Added a few ideas on the table for UX and performance.

PR #1914, opened 2023-03-24
Update hackathon template

motechFR:
Fixes 2 issues with imports

- "Page not found" for certain links
_We didn't yet have a new reference for some pages_

- "Empty inline databases not showing"
_We were treating inline databases as a special cases. Simplified this to treat them as normal children_
_Also fixed an edge case where we weren't actually creating pages if there were no views (which prevented having an empty inline database)_


lib/templates/exportWorkspacePages.ts
```diff
@@ -148,20 +136,6 @@ export async function exportWorkspacePages({
       })
     );
     const pollIds: string[] = [];
-    const inlineDatabasePageIds: string[] = [];
-    recurse(node.content as PageContent, (_node) => {
```
Devorein, line 152: Please restore the code to extract the inline `poll`, this makes sure we duplicate the inline polls.

Devorein (approved):

lib/templates/exportWorkspacePages.ts
motechFR: Done, thanks for spotting this bug!

PR #1909, opened 2023-03-23
Feat/mobile views update

msadura:
- fixed "add property" popup on mobile
- fixed database source sidebar on mobile
- fixed page embed (and other plugins) popup

https://app.charmverse.io/charmverse/page-6639566872795888
https://app.charmverse.io/charmverse/page-08728974767596265
https://app.charmverse.io/charmverse/page-23324326338262136

Screenshot 2023-03-23 at 19 20 47
Screenshot 2023-03-23 at 19 20 13
ss1

PR #1919, opened 2023-03-24
Fix the large popup for editing forum categories

motechFR:


mattcasey (changes requested):
this looks like a bug introduced by @msadura in this PR: https://github.com/charmverse/app.charmverse.io/pull/1909/files#diff-e45bfc88466d6c960d04e86d6b9efe64492813c7309aa2ec681ee2838cb16621. Rather than add more customization on this component, I wonder why not just use MUI's component directly. This seems to just be adding a bunch of defaults, but they don't fit every case so we're making our jobs even harder. The bug also exists on the proposal category popups, for example

motechFR:
> https://github.com/charmverse/app.charmverse.io/pull/1909/files#diff-e45bfc88466d6c960d04e86d6b9efe64492813c7309aa2ec681ee2838cb16621

I'm counting 7 locations where we use this component, so i think it would be good to continue using it.

I did some checking, and this issue pops up in the member directory menu where you update the options (see with / without) screenshots below.

Went for the solution of making the popper simpler, and passing the style props from the outside.

I created a common setting for CharmEditor components that we can pass down.

Replicated intent of @msadura PR by passing these props down to the PDF, Media and Image selectors


Screenshot 2023-03-25 at 00 23 10

Screenshot 2023-03-25 at 00 23 42

components/common/CharmEditor/components/common/selectorPopupSizeConfig.ts
```diff
@@ -0,0 +1,7 @@
+import type { Theme } from '@mui/material';
+import type { SxProps } from '@mui/system';
+
+// This is consumed by popup widgets in charm editor such as pdf or file selector to provide sufficient width
+export const selectorPopupSizeConfig: SxProps<Theme> = {
```
msadura, line 5: I would move this directly  to `[MediaSelectionPopup.tsx]` - later on (not in a scope of this issue)we should refactor PDF and Image selector components to use this one as well, and this style will only be needed there.

components/common/PopperPopup.tsx
```diff
@@ -13,6 +13,7 @@ interface PopperPopupProps {
   onClose?: () => void;
   onOpen?: () => void;
   onClick?: () => void;
+  sx?: SxProps<Theme>;
```
msadura, line 16: tiny stuff - lets call it `paperSx` to explicitly tell where those styles will be applied

msadura (approved):

components/common/PopperPopup.tsx
motechFR: Sounds good, made the update

mattcasey (approved):

components/common/PdfSelector/PdfSelector.tsx
```diff
@@ -5,6 +5,8 @@ import MultiTabs from 'components/common/MultiTabs';
 import PopperPopup from 'components/common/PopperPopup';
 import { uploadToS3 } from 'lib/aws/uploadToS3Browser';
 
+import { selectorPopupSizeConfig } from '../CharmEditor/components/common/selectorPopupSizeConfig';
```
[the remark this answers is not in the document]
mattcasey: looks like PDFSelector should be inside CharmeDitor actually :P

PR #1920, opened 2023-03-24
Clean up how we handle making a proposal public so it works across the lifecycle

motechFR:
I noticed a bug where edit_isPublic wasn't available after the proposal was in vote, so I cleaned up how we handle this


motechFR:
@ccarella Can you please confirm something?

Currently, admin can edit proposal during discussion and review stages. Should we extend this to "reviewed" stage?

I can see this being desirable if there's an error in the reviewed proposal before we go ahead and create a vote

mattcasey (approved):

PR #1926, opened 2023-03-27
Fix user details form

msadura:
Fix "about me" input in user popup. Inputs will not be rendered while data is being loaded. 
Added sskeleteon loader for a bit better ux

Screenshot 2023-03-27 at 20 15 49


0xepicode:
@motechFR  thx for finding this funny edge case :D Turned out when you create a new user your "details" are not yet defined and  endpoint basically does not return anything. Case fixed.

PR #1927, opened 2023-03-27
minor improvements to creating page

mattcasey:
I am not able to create pages in my test space - this whole process is timing out. BUt it is because resolvePageTree takes up to 5 seconds when you have 8000 pages, and the subsequent queries are using the same transaction which times out.

I do'nt have a fix for that case, but while debugging I cleaned up some of the code

### DIFF
--- a/lib/permissions/pages/actions/upsert-permission.ts
+++ b/lib/permissions/pages/actions/upsert-permission.ts
@@ -176,14 +176,14 @@ async function validatePermissionToCreate(pageId: string, permission: IPagePermi
   }
 
   // Load the page space ID
-  const pageSpaceId = (await tx.page.findUnique({
+  const pageSpaceId = await tx.page.findUnique({
     where: {
       id: pageId
     },
     select: {
       spaceId: true
     }
-  })) as Pick<Page, 'spaceId'>;
+  });
 
   if (permission.spaceId && permission.spaceId !== pageSpaceId?.spaceId) {
     throw new InsecureOperationError(
@@ -249,7 +249,6 @@ export async function upsertPermission(
           flatChildren: flattenTree(tree.targetPage)
         };
       }))) as TargetPageTreeWithFlatChildren<PageNodeWithPermissions>;
-
     // Get the source permission we are inheriting from
     let permissionData: Prisma.PagePermissionUpsertArgs;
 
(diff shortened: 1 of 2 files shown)


lib/permissions/pages/triggers/page-created.ts
```diff
@@ -12,73 +12,65 @@ import { upsertPermission } from '../actions/upsert-permission';
...
+        createdBy: true
+      }
+    });
     if (!page) {
       throw new PageNotFoundError(pageId);
     }
 
     // This is a root page, so we can go ahead
     if (!page.parentId) {
-      // Space id could be null
```
mattcasey, line 23: "Space id could be null" should never be null, according to our db. even stranger, though, the 'else' condition was still using page.spaceId in the call to upsertPermission()

lib/permissions/pages/triggers/page-created.ts
```diff
@@ -12,73 +12,65 @@ import { upsertPermission } from '../actions/upsert-permission';
...
-            {
-              permissionLevel: 'view',
-              public: true
-            },
-            undefined,
-            tx
-          );
-        }
-      } else {
+      const permissionLevel = space?.defaultPagePermissionGroup || 'full_access';
```
mattcasey, line 42: removed an if/else by just defining this variable conditionally

lib/permissions/pages/actions/upsert-permission.ts
```diff
@@ -340,17 +339,15 @@ export async function upsertPermission(
...
-      });
+      const childrenIds = resolvedPageTree.flatChildren.map((page) => page.id);
 
       await tx.pagePermission.updateMany({
         where: {
           AND: [
             {
-              OR: childrenIds
+              pageId: {
+                in: childrenIds
```
motechFR, line 349: good db optimisation there 👍

lib/permissions/pages/triggers/page-created.ts
```diff
@@ -12,73 +12,65 @@ import { upsertPermission } from '../actions/upsert-permission';
 export async function setupPermissionsAfterPageCreated(pageId: string): Promise<IPageWithPermissions> {
   // connect to page node from use pages (Removing iPageWithPermissions)
   return prisma.$transaction(async (tx) => {
-    const page = await getPage(pageId, undefined, tx);
-
+    const page = await tx.page.findUnique({
+      where: {
+        id: pageId
+      },
+      select: {
```
motechFR, line 19: Another good one, we get to drop the page content and permissions join query

lib/permissions/pages/triggers/page-created.ts
motechFR: Probably a leftover from earliest iteration where we didn't yet have the spaces enforced I think. Glad we could get rid of this

motechFR (approved):
Tested locally and works fine. Both a root page and a child page took between 1.3s and 1.7s in a space with 10 000 pages

PR #1932, opened 2023-03-28
Consolidate discord connect inside connect button

Devorein:


components/login/Login.tsx
```diff
@@ -36,10 +39,10 @@ export interface DialogProps {
 function LoginHandler(props: DialogProps) {
   const { onClose, isOpen } = props;
   const { loginFromWeb3Account } = useWeb3AuthSig();
-
   // Governs whether we should auto-request a signature. Should only happen on first login.
   const [enableAutosign, setEnableAutoSign] = useState(true);
-
+  const returnUrl = new URLSearchParams(decodeURIComponent(window.location.search)).get('returnUrl');
```
mattcasey, line 44: since you're importing router, i believe you can just do router.query.returnUrl. That's what we're doing in pages/index.tsx

components/login/Login.tsx
```diff
@@ -164,6 +167,22 @@ function LoginHandler(props: DialogProps) {
                 isLoading={false}
               />
             </ListItem>
+
+            <ListItem>
+              <ConnectorButton
+                onClick={() => router.push(`/api/discord/oauth?type=login&redirect=${returnUrl ?? '/'}`)}
+                data-test='connect-discord'
+                name='Connect with discord'
```
mattcasey, line 175: "Discord" should be capitalized, and I think this should appear before Gmail and email

mattcasey (changes requested):
see comments

components/login/Login.tsx
Devorein: Thank you. Didn't know about that. Definitely much better if it works on both runtime

mattcasey (approved):

PR #1933, opened 2023-03-28
Disconnect user wallet

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-8430206861166976)


lib/users/disconnectWallet.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!address || !userId) {
+    throw new InvalidInputError(`Address and userId are required to disconnect your wallet from the user account`);
+  }
+
+  const wallet = await prisma.userWallet.findUnique({
+    where: {
+      userId,
+      address
+    },
+    include: {
```
Devorein, line 21: Including the user might be unnecessary

Devorein (approved):
Works great. Had a few questions though:-
1. What will happen to users gnosis wallets/safes?
2. What will happen to their gnosis tasks?

components/integrations/components/IdentityProviders.tsx
```diff
@@ -167,19 +172,41 @@ export function IdentityProviders() {
...
+            <ConfirmDeleteModal
+              title='Disconnect wallet'
+              question={
+                <>
+                  <Typography mb={1}>
+                    Are you sure you want to Disconnect your {wallet.ensname || shortWalletAddress(wallet.address)}{' '}
+                    wallet?
+                  </Typography>
+                  <Typography variant='body2'>
+                    This action will delete your wallet, nfts, poaps and will lose roles and permissions in spaces where
```
Devorein, line 199: Might need to include organizations (deepdao) as well, as mentioned in the card

lib/users/disconnectWallet.ts
valentinludu: Yep. I added that before, when I was doing more logic, but then I removed it so no ned for it.

valentinludu:
> Works great. Had a few questions though:-
> 
> 1. What will happen to users gnosis wallets/safes?
> 2. What will happen to their gnosis tasks?

After a discussion with Chris, I understood that gnosis safe will disappear and I will not concentrate on gnosis at all in this PR.

components/integrations/components/IdentityProviders.tsx
valentinludu: Added!

charmClient/charmClient.ts
```diff
@@ -177,6 +178,10 @@ class CharmClient {
     return http.POST<User>('/api/profile/add-wallets', { addressesToAdd: data });
   }
 
+  removeUserWallet(address: Pick<UserWallet, 'address'>) {
```
msadura, line 181: this could take array of wallets to support multiple wallets, but I believe it is not necessary right now

hooks/useWeb3AuthSig.tsx
```diff
@@ -232,12 +237,34 @@ export function Web3AccountProvider({ children }: { children: ReactNode }) {
...
-
-  function disconnectWallet() {
+  const logoutWallet = useCallback(() => {
     if (account) {
       window.localStorage.removeItem(`${PREFIX}.wallet-auth-sig-${account}`);
       setWalletAuthSignature(null);
     }
+  }, [account]);
+
+  const { trigger: triggerDisconnectWallet, isMutating: isDisconnectingWallet } = useSWRMutation(
```
msadura, line 247: nice stuff, we do not use useSWRMutation often enough :D

pages/api/profile/remove-wallet.ts
```diff
@@ -0,0 +1,36 @@
...
+  .use(requireUser)
+  .post(removeWalletController);
+
+async function removeWalletController(req: NextApiRequest, res: NextApiResponse<LoggedInUser | { error: string }>) {
+  const userId = req.session.user.id;
+  const address = req.body.address as UserWallet['address'];
+
+  const loggedInUser = await disconnectWallet({ address, userId });
+
+  await prisma.profileItem.deleteMany({
```
msadura, line 27: I am not sure why we need it, but if it is coupled with wallet disconnect then maybe move it to `disconnectWallet` in 1 transaction?

msadura (approved):

hooks/useWeb3AuthSig.tsx
valentinludu: Thx.

Totally agree.

pages/api/profile/remove-wallet.ts
valentinludu: We need this to remove the items related to the user address he just removed.

Ok

charmClient/charmClient.ts
valentinludu: I think this does not need an array of wallets because you disconnect the wallets one by one. And you will send one request per wallet.

There will not be an option to disconnect all with one click.

PR #1937, opened 2023-03-28
Roles UI: Add admins/members from a button

mattcasey:
You can now switch members between admin/member by clicking "Add member" on their respective rows.


motechFR:
For the forums; I think we should handle these 2 cases

**Case 1**
The space already has some level of access to a forum category

Permission levels are view, view + comment, full access

Space has View + comment. Role has nothing explicitly

Solution; Show the view+ comment as inherited from space, but allow you to upgrade it inline. Don't show view, since this access level will be irrelevant


Screenshot 2023-03-29 at 23 47 49

**Case 2**
The role has forum-wide moderator. This is a shortcut to moderator level abilities across the entire forum (full access + moderation)

In this case, I suggest replacing all the dropdowns by a readonly state. I have enclosed below a screenshot of normal page permissions vs. proposal page permissions showcasing this idea.

Screenshot 2023-03-29 at 23 51 42

Screenshot 2023-03-29 at 23 50 44

motechFR:
I think we should remove the "all eligible members have been added to this role" in members row (possibly admin and guest too, and just have this for normal roles)

Screenshot 2023-03-30 at 01 50 08

motechFR:
Some personalisation for built-in roles (Admin, Guest, Member);

Suggest have the label as "Add admins", and add guests.

In the members row, we could instead provide a link to the token gate, or the available invite links

For guests, it would make sense if the dialog supports invite via email


Screenshot 2023-03-30 at 01 53 17

motechFR (approved):
I tried a few combos and works great

- Guests are filtered out from the list of people I can add a role to
- In Members row, I only see guests (we might want to drop the feature on this row)

One extra consideration; we might want to prevent promoting from guest straight to admin, requiring you to make the user a member first. Slack has this paradigm

It protects against accidental escalation of privileges of someone who was a guest moments ago. Since adding a new admin is an infrequent act, I think the added friction is worth is vs. the security tradeoff

mattcasey:
Thanks @motechFR 
> I think we should remove the "all eligible members have been added to this role" in members row (possibly admin and guest too, and just have this for normal roles)
I decided to just move this message into the popup for all roles. Instead of showing a dropdown we'll let you know there are no more users to select.

> For guests, it would make sense if the dialog supports invite via email (could be a fast follower). The "Add members" labelling here is a bit confusing
That's why i didn't think we should bother letting admins move guests around here. It also wouldn't make sense to let you invite someone by email without also selecting the page. At that point, you should use our primary UX to add a guest

PR #1940, opened 2023-03-29
Show only the current space in SettingsModal

valentinludu:


mattcasey (changes requested):
The collapsible space icon is still being used:
image


It should just be a clean static list of links, like under "User Settings"

msadura (approved):

PR #1946, opened 2023-03-29
update lit sdk to v2

mattcasey:
Lit has a whole new SDK which is supposed to be mostly backwards-compatible.

I fixed the TS types, which are now provided by Lit, and followed this migration guide: https://developer.litprotocol.com/sdk/explanation/migration/

Still needs fixing:

When testing a token gate, I get `Unsupported encoding "base64url"`

image


math77:
I'm getting this issue in my nextjs project, any tips on how I can solve it?

PR #1947, opened 2023-03-29
Feat/notifications preview

msadura:
Notification preview popup.

Screenshot 2023-03-29 at 20 57 50


components/common/PageLayout/components/Header/components/NotificationPreview/useNotificationPreview.ts
```diff
@@ -0,0 +1,239 @@
...
+      type,
+      groupType
+    }: {
+      taskId: string;
+      groupType: NotificationGroupType;
+      type: NotificationType;
+    }) => {
+      await charmClient.tasks.markTasks([{ id: taskId, type }]);
+
+      mutateTasks(
```
valentinludu, line 56: In theory you only need:
mutateTasks()

or

import { mutate } from 'swr';
mutate(/tasks/list/${user.id})

Of course, this will trigger a new GET request to revalidate the data. If you want it faster your approach is better. 
I was thinking more on the logic part because this needs to be maintained for all 5 types of notifications and maybe this logic will change in the future.
If you use mutateTasks() you are 100% sure the data is synced with the db. 

Let me know if this helps.

components/common/PageLayout/components/Header/components/NotificationPreview/useNotificationPreview.ts
```diff
@@ -0,0 +1,239 @@
...
+        (_tasks) => {
+          if (!_tasks) {
+            return;
+          }
+
+          const taskIndex = _tasks?.[groupType].unmarked.findIndex((t) => t.taskId === taskId);
+          if (typeof taskIndex === 'number' && taskIndex > -1) {
+            const marked = [_tasks?.[groupType].unmarked[taskIndex], ..._tasks.forum.marked];
```
valentinludu, line 64: If you want to keep all the logic:

Are you sure this is correct? `..._tasks.forum.marked` ?

Shouldn't this be:
`const marked = [_tasks?.[groupType].unmarked[taskIndex], ..._tasks.[groupType].marked];`

?

components/common/PageLayout/components/Header/components/NotificationPreview/NotificationButton.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const popupState = usePopupState({ variant: 'popover', popupId: 'share-menu' });
+  const theme = useTheme();
+  const isLargeScreen = useMediaQuery(theme.breakpoints.up('md'));
+  const handleSeeAllClick = () => {
+    onSeeAllClick();
+    popupState.close();
+  };
+  return (
+    <>
+      <Box onClick={popupState.open} sx={{ cursor: 'pointer' }}>
```
valentinludu, line 21: You can put the `onClick` and `sx` directly on the NotificationsBadge and delete the Box
So you will have one less div to render.

components/common/PageLayout/components/Header/components/NotificationPreview/NotificationPreview.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const icon = useMemo(() => getIcon(groupType), [groupType]);
+
+  return (
+    <Link color='inherit' href={href} onClick={onClose}>
+      <Box
+        sx={{
+          '&:hover': {
+            cursor: 'pointer',
+            background: theme.palette.background.light
```
valentinludu, line 37: For shortening the imports, you can directly say 
background : (theme) => theme.palette.background.light
And remove useTheme

components/common/PageLayout/components/Header/components/NotificationPreview/NotificationPreviewPopover.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Card>
+        <Typography fontWeight={600} p={2}>
+          Notifications
+        </Typography>
+      </Card>
+      <Divider />
+      <Box height={400} sx={{ overflowY: 'auto', overflowX: 'hidden' }}>
+        {notificationPreviews.length > 0 ? (
+          notificationPreviews.map((notification) => (
+            <>
```
valentinludu, line 28: You need to put the key on the Fragment
That is why you have an error in the console.

lib/proposal/getProposalTasks.ts
```diff
@@ -17,22 +19,26 @@ export type ProposalTaskAction = 'start_discussion' | 'start_vote' | 'review' |
...
   action: ProposalTaskAction | null;
   eventDate: Date;
+  createdAt: Date;
   spaceDomain: string;
   spaceName: string;
   pageId: string;
   pageTitle: string;
   pagePath: string;
   status: ProposalStatus;
+  createdBy: NotificationActor | null;
```
valentinludu, line 32: Why not just an userId for createdBy?
And use useMembers to find the user info in the FE.

With this approach you simplify the ProposalTask interface and also get the member username in case the user has a specific username in that space.

valentinludu: I also noticed you only need { id, username, avatar } from createdBy.

msadura: This will not work as you mentioned, because notifications come from all your spaces, so you might not have al the needed members.

components/common/PageLayout/components/Header/components/NotificationPreview/useNotificationPreview.ts
msadura: tasks reload is rather resource-consuming thing, this is why I think it is not worth to reload it from the server

valentinludu (approved):
Tested on Desktop Firefox, Chrome, Brave
Tested on mobile: iPhone Safari and responsive Chrome

PR #1948, opened 2023-03-29
Add Forum category permissions to settings

mattcasey:


lib/permissions/forum/listGroupPostCategoryPermissions.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import type { AssignedPermissionsQuery } from '../interfaces';
+import { permissionGroupIsValid } from '../utils';
+
+import type { AssignedPostCategoryPermission } from './interfaces';
+import { mapPostCategoryPermissionToAssignee } from './mapPostCategoryPermissionToAssignee';
+
+export type PermissionsGroupQuery = Pick<AssignedPermissionsQuery, 'id' | 'group'>;
+
+export async function listGroupPostCategoryPermissions({
```
motechFR, line 15: Suggestion; we can instead provide all the assigned permissions to all categories in the space

Small extra improvement; we could also only return the subset of permissions the user has access to

motechFR:
For the forums; I think we should handle these 2 cases

**Case 1**
The space already has some level of access to a forum category

Permission levels are view, view + comment, full access

Space has View + comment. Role has nothing explicitly

Solution; Show the view+ comment as inherited from space, but allow you to upgrade it inline. Don't show view, since this access level will be irrelevant


Screenshot 2023-03-29 at 23 47 49

**Case 2**
The role has forum-wide moderator. This is a shortcut to moderator level abilities across the entire forum (full access + moderation)

In this case, I suggest replacing all the dropdowns by a readonly state as it doesn't make sense to assign individual category permissions. I have enclosed below a screenshot of normal page permissions vs. proposal page permissions showcasing this idea.

Screenshot 2023-03-29 at 23 51 42

Screenshot 2023-03-29 at 23 50 44

motechFR:
QA Notes

**Case 1:** Forum category permissions not showing for members, only roles

I can see the forum category permissions for a role, but not for members.

The role does show "inherited from space", but I can't see this in members tab

Screenshot 2023-03-30 at 23 19 09
Screenshot 2023-03-30 at 23 19 34


**Case 2:** Inheritance UX inside the role row

When permissions exist for the space, we show the "inherited" for the role.

There are 2 sub-issues
- The individual access level is "full access."
_This is the highest individual level, so I don't think we should even offer a dropdown_

- The individual access level is full access. We are still showing the "view" and "remove" options.
Since this is inherited, we can't subtract, so we should ideally only provide access levels above the existing one. In this case, "view" and "remove" options would preferably not show

In this screenshot, I was able to assign "view" permission even though there is already full access inherited from the space

Screenshot 2023-03-30 at 23 26 04


**Case 3: (Suggestion)** 
When the role is assigned "space-wide moderator" and we click save, the dropdowns go to "readonly mode" and the tooltip changes. This reflects what we discussed

I would like to suggest introducing a new label for moderation level which is something like "Full Access & Moderator" (or maybe just moderator as this is an established concept in various products mods generally have full access to a forum category / subreddit / discord channel"

components/common/form/InputEnumToOptions.tsx
```diff
@@ -20,28 +22,26 @@ export default function InputEnumToOptions<T extends string>({
...
-  useEffect(() => {
-    if (defaultValue && !value) {
-      setValue(defaultValue);
-    }
-  }, [defaultValue]);
+  // useEffect(() => {
+  //   if (defaultValue && !value) {
+  //     setValue(defaultValue);
+  //   }
+  // }, [defaultValue]);
```
motechFR, line 33: Let's delete this code if unused

components/forum/components/permissions/PostCategoryPermissionRow.tsx
```diff
@@ -50,28 +58,30 @@ export function PostCategoryRolePermissionRow({
     }
   }
 
+  const tooltip = !canEdit
```
motechFR, line 61: Re: my comment about space-wide moderator role, we could add the more explanatory tooltip here

components/proposals/components/permissions/ProposalCategoryPermissionRow.tsx
motechFR, line 1: A bug maybe in this file; When I tried changing the value of a permission for a proposal category inline from this area, the save button did not activate after changing a value.

This bug happened in the members area

mattcasey: yep, thats gonna be in a follow-up PR

mattcasey:
>The individual access level is full access. We are still showing the "view" and "remove" options.
Since this is inherited, we can't subtract, so we should ideally only provide access levels above the existing one. In this case, "view" and "remove" options would preferably not show

This is getting very complicated, and it doesn't seem like we reduce the options in the existing forms. Plus, you might set the role before you change the space-level permissions. maybe we should have some sort of warning like "this rule is negated by member role"

PR #1949, opened 2023-03-30
Feat/improved member search

msadura:
https://app.charmverse.io/charmverse/page-3260419489530417

- reorganized search member code a bit
- added `searchValue` string to member with all relevant values to search
- I also found an opportunity to optimize our frontend code when it comes to members. We sometimes used Members array to build membersRecord, and we oftes searched member in array by id. I added memorized `membersRecord` to members context, and additional `getMemberById` convenience method while will pick user from that record. 

NOTE: Seems like we have all the data we need in Members Directory search to just filter users on the frontend without additional server search. Was there any other reason to search for members via API or I can switch to frontend filtering? @mattcasey @Devorein


components/members/hooks/useFilteredMembers.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!query) {
+      return members;
+    }
+
+    return members.filter((member) => {
+      if (member.searchValue?.includes(query)) {
+        return true;
+      }
+
+      if (member.roles.some((role) => role.name.toLowerCase().includes(query))) {
```
Devorein, line 19: There seems to be a duplication issue here.

Devorein (approved):

PR #1962, opened 2023-04-03
Fix gnosis integration + refactor

msadura:
Gnosis issue reported by the user: https://discord.com/channels/894960387743698944/1092352184483581982

This pr:
- fixes retrieving user signature (user needs to sign tx manually with wallet)
- adds datadog logs in case of failure
- adds snackar with information if user does not have any safe to connect
- small refactor - just removed `default` export, moved one component to separate file

### DIFF
--- a/hooks/useGnosisPayment.ts
+++ b/hooks/useGnosisPayment.ts
@@ -3,6 +3,7 @@ import EthersAdapter from '@safe-global/safe-ethers-lib';
 import SafeServiceClient from '@safe-global/safe-service-client';
 import { getChainById } from 'connectors';
 import { ethers, utils } from 'ethers';
+import log from 'loglevel';
 
 import type { MultiPaymentResult } from 'components/bounties/components/MultiPaymentButton';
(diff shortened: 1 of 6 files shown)


hooks/useGnosisPayment.ts
```diff
@@ -42,22 +43,31 @@ export function useGnosisPayment({ chainId, safeAddress, transactions, onSuccess
...
-      safeTxHash: txHash,
-      senderAddress: utils.getAddress(account),
-      senderSignature: [...safeTransaction.signatures][0][0],
-      origin
-    });
-    onSuccess({ safeAddress, transactions, txHash });
+
+    try {
+      const txHash = await safe.getTransactionHash(safeTransaction);
+      const senderSignature = await safe.signTransactionHash(txHash);
```
msadura, line 49: this is actual fix for the issue. Docs: https://github.com/safe-global/safe-core-sdk/blob/main/guides/integrating-the-safe-core-sdk.md#5-propose-the-transaction-to-the-service

Devorein:
Tested and it works 🚀

PR #1964, opened 2023-04-03
Add no access for hidden mention and nested pages

Devorein:
image

[Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=8bae196c-0542-407e-9393-23e99cc3df85)


mattcasey (changes requested):
This PR seems to break an existing use case, when page is not found (ie. was deleted). We need a way to know the user simply doesn't have access to the page, but that one does exist. I thought one way to do this is to return inaccessible pages in usePages() hook but dont return any content (title, etc). we might want to add a flag on PageMeta such as "hasAccess = false" or "hidden = true"

mattcasey (approved):

PR #1965, opened 2023-04-03
Improve unique key for nft profile items

mattcasey:
Occasionally we're seeing an error about unique constraints for the "id" field on ProfileItem. We use the contract address + token id to generate it, but this is not necessarily unique. In the case of ERC1155, even a single user may own multiple copies of the same token id.

Link to logs: https://app.datadoghq.com/logs?query=service%3Awebapp%20Unique%20constraint%20failed%20on%20the%20fields&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1680458712848&to_ts=1680545112848&live=true

Includes migration to prepend <userId> to the id's of all nft items


pages/api/nft/index.ts
mattcasey, line 1: from what i could tell, looking for "api/nft", we don't use this route

msadura (approved):

PR #1967, opened 2023-04-04
Remove extra step in signup flow

mattcasey:
Replace /signup and send user to /createSpace immediately. This is a WIP, and open to suggestions especially on copy:

image

WARNING: I temporarily disabled redirect on /createSpace so it's easier to preview. Will bring it back before deploying


msadura:
this is 🔥 !!! I hate attacking people with modals, this gives much better UX. Huge love from me (and future users :P) for this improvement

PR #1968, opened 2023-04-04
Support for space template in api

Devorein:


lib/public-api/interfaces.ts
```diff
@@ -285,8 +287,6 @@ export interface Workspace {
  *    CreateWorkspaceRequestBody:
  *      required:
  *        - name
- *        - discordServerId
```
Devorein, line 288: According to the typescript type, neither `discordServerId` nor `adminDiscordUserId` is required, but the user must provide either `adminDiscordUserId` or `adminWalletAddress` as an identifier. So a bit confused what to put here as required

mattcasey: Since we have to mark them either all required or all optional, I would just make them all optional here. The API will tell you why your request is invalid.

lib/public-api/createWorkspaceApi.ts
```diff
@@ -70,7 +71,7 @@ export async function createWorkspaceApi({
     webhookUrl,
     extraAdmins: [botUser.id],
     // Create new spaces with NFT template by default
-    createSpaceOption: 'templateNftCommunity'
+    createSpaceOption: template
```
mattcasey, line 74: we still want to continue using 'templateNftCommunity' as the default template when creating from the API

motechFR:
Some feedback:
- We mentioned yesterday using snake case (ie. a_b) for the template options. I think we should add these in.
Also, you can add an 

- Can we get a test that the API creates the space properly given a specific template?
For this, you could just hardcode the page names of a given template into the test.

lib/public-api/interfaces.ts
```diff
@@ -322,6 +326,7 @@ export interface CreateWorkspaceRequestBody {
   adminUsername?: string;
   webhookUrl?: string;
   avatar?: string;
+  template?: SpaceTemplateType;
```
mattcasey, line 329: just to be clear: we should NOT use SpaceTemplateType. it should be its own mapping of clean versions for the API. we could migrate our internal mapping to be snake case as well but that is out of scope

PR #1972, opened 2023-04-04
Person property mention notification

Devorein:
[Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=2c9d8701-33ff-4c3a-942a-6d403a638565&cardId=02c71c8a-7b1c-4429-9011-1594aae097ec)


lib/discussion/getDiscussionTasks.ts
```diff
@@ -455,6 +463,94 @@ async function getPageCommentMentions({
...
+    }
+  });
+
+  const cards = (await prisma.block.findMany({
+    where: {
+      parentId: {
+        in: Object.keys(boardBlocksPersonPropertyRecord)
+      },
+      type: 'card',
+      deletedAt: null
```
mattcasey, line 497: we should probably add a 'greater than' filter on createdAt here, so that i dont get notifications for all past blocks?

mattcasey (changes requested):
see comments

lib/discussion/getDiscussionTasks.ts
```diff
@@ -455,6 +463,94 @@ async function getPageCommentMentions({
...
+        pageTitle: card.page.title || 'Untitled',
+        bountyId: card.page.bountyId,
+        bountyTitle: card.page.title,
+        type: card.page.bountyId ? 'bounty' : 'page',
+        mentionId: null,
+        taskId: `${card.id}.${personPropertyId}`,
+        // Fake value
+        createdAt: new Date().toString(),
+        userId: card.createdBy,
+        text: `${card.user.username} mentioned you in the card ${card.page.title || 'Untitled'}`,
```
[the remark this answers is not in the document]
Devorein: Notion uses mentioned but `assigned` works much better

lib/discussion/getDiscussionTasks.ts
Devorein: Ideally you shouldn't get any new notifications as after sending notifications we track them as sent so that we don't send them again, Secondly, what if a page was created 10 days ago but the person property was changed 10 mins ago, filter by `createdAt` greater than yesterday wouldn't work.

Devorein: We have decided to go with `updatedAt` filter

PR #1976, opened 2023-04-05
clear cache when user logs out

mattcasey:
Currently when I log out, sometimes i see a lightning bolt appear for a minute as the view tries to catch up with the new state. This is an experiment to see if clearing the cache will make it better or worse. 

Something else we could try is redirecting the user to "/" immediately, instead of waiting for RouteGuard

### DIFF
--- a/hooks/useUser.tsx
+++ b/hooks/useUser.tsx
@@ -1,5 +1,6 @@
 import type { ReactNode } from 'react';
 import { createContext, useCallback, useContext, useEffect, useMemo, useState } from 'react';
+import { mutate } from 'swr';
 
 import charmClient from 'charmClient';
 import type { LoggedInUser } from 'models';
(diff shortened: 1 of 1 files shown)


motechFR (approved):
Worked for me!

Tested as follows
- Log in
- Arrive on document page
- Log out

On the main branch, I saw the error message on second login / logout cycle

After 5 attempts on this branch, I did not receive a lightning bolt

hooks/useUser.tsx
```diff
@@ -80,4 +82,9 @@ export function UserProvider({ children }: { children: ReactNode }) {
   return <UserContext.Provider value={value}>{children}</UserContext.Provider>;
 }
 
+// clear cache from SWR - https://swr.vercel.app/docs/mutation
+function clearSWRCache() {
+  return mutate(() => true, undefined, { revalidate: false });
```
motechFR, line 87: Pretty handy technique, just learnt you can make use of filter function in global SWR key

PR #1978, opened 2023-04-05
Confirm csv import with 2 actions

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-08447284233799945)

Added confirm modal with 2 options: merge / delete.

Additional:
Fixed an error when a page is deleted and the useEffect in EditorPage component is fetching again that page.
Fixed a FE error when the card has a boolean as an html attribute.


pages/api/pages/index.ts
```diff
@@ -113,4 +115,58 @@ async function createPageHandler(req: NextApiRequest, res: NextApiResponse<IPage
...
+  }
+
+  log.info('User deleted pages', {
+    userId,
+    pageIds,
+    childPageIds: modifiedChildPageIds,
+    spaceId
+  });
+
+  return res.status(200).json({ pageIds: modifiedChildPageIds });
```
mattcasey, line 169: doesn't look like we use this list of pageids, and it could be a decent size list.. i would just do res.status(200).end()

pages/api/blocks/index.ts
```diff
@@ -286,4 +292,45 @@ async function updateBlocks(req: NextApiRequest, res: NextApiResponse<Block[]>)
...
+
+  const spaceIds = [...new Set(blocks.map((p) => p.spaceId).filter(isTruthy))];
+
+  // A user can delete only a batch of blocks from a single space
+  if (spaceIds.length > 1) {
+    throw new ActionNotPermittedError("You can't delete blocks from multiple spaces at once");
+  }
+
+  for (const blockId of blockIds) {
+    const permissions = await computeUserPagePermissions({
```
mattcasey, line 315: this op is extremely slow for spaces with a lot of pages. can we find a way to verify a list of page ids?

mattcasey (approved):
looks good once u address my comment on permissions, thanks!

pages/api/pages/index.ts
valentinludu: Yes, we are not using it.

Done!

pages/api/blocks/index.ts
valentinludu: Done. Will break the loop if at least one has access.

PR #1980, opened 2023-04-05
Track wallet address for profile item

Devorein:
- [x] remove them when a user no longer has that wallet
- [x] Keep track of wallet address used to gather profile items


prisma/schema.prisma
```diff
@@ -1025,13 +1026,15 @@ model ProposalReviewer {
...
-  type     ProfileItemType
-  isPinned Boolean?
+  id         String          @id
+  userId     String          @db.Uuid
+  user       User            @relation(fields: [userId], references: [id], onDelete: Cascade)
+  isHidden   Boolean?
+  metadata   Json?
+  type       ProfileItemType
+  isPinned   Boolean?
+  address    String?
```
mattcasey, line 1036: change this to "walletAddress"? 'address' might be confused with 'contractAddress' too

mattcasey, line 1036: rename this to 'walletAddress'? 'address' could be confused with 'contractAddress'

scripts/migrations/2023_03_30_add_default_path.ts
mattcasey, line 1: this migration should not be deleted

scripts/migrations/2023_04_05_add_profile_item_wallet.ts
```diff
@@ -0,0 +1,31 @@
...
+    }
+  })
+
+  for (const profileItem of profileItems) {
+    await prisma.profileItem.update({
+      where: {
+        id: profileItem.id
+      },
+      data: {
+        address: profileItem.user.wallets[0].address,
```
mattcasey, line 25: we should check if a user has more than one wallet. i know of at least one user besides myself with multiple wallets. if they have more than one, i think it's safe to just skip them and not tie the profile items to a wallet

pages/api/profile/items.ts
```diff
@@ -13,9 +13,23 @@ handler.use(requireUser).put(updateUserProfileItems);
...
+    select: {
+      wallets: {
+        select: {
+          address: true
+        }
+      }
+    }
+  });
+
+  const firstWalletAddress = userWallets?.wallets[0].address;
```
mattcasey, line 29: the wallet address needs to be passed in, let's not assume which address it is. that defeats the purpose of tracking wallet address :P

pages/api/profile/items.ts
```diff
@@ -54,7 +68,8 @@ async function updateUserProfileItems(req: NextApiRequest, res: NextApiResponse<
             userId: req.session.user.id,
             metadata: profileItem.metadata === null ? undefined : profileItem.metadata,
             isHidden: true,
-            type: profileItem.type
+            type: profileItem.type,
+            address: firstWalletAddress
```
mattcasey, line 72: I guess this should actually just be profileItem.address

mattcasey (changes requested):
It looks like we send profileItems to the frontend for users that don't own the profile items. To be safe, we should add a new unique key to wallets that we can link the profile items to, instead of using wallet addresss as the foreign key

scripts/migrations/2023_03_30_add_default_path.ts
```diff
@@ -1,26 +0,0 @@
-import { prisma } from 'db';
-import { uid } from 'lib/utilities/strings';
-
-async function init() {
-  // const users = await prisma.user.findMany({
```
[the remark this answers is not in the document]
Devorein: As far as I remember ts was complaining, but I will remove the comments and see if it still does the same or not.

lib/users/prepopulateUserProfile.ts
```diff
@@ -18,7 +18,10 @@ const acceptedImageFormats = ['.jpg', '.jpeg', '.png', '.webp'];
  * @param user we need the user id and user avatar to populate many
  * @param ens the ENS name
  */
-export async function prepopulateUserProfile(user: User, ens: string | null) {
+export async function prepopulateUserProfile(
+  user: User & { wallets: Pick<UserWallet, 'address' | 'ensname' | 'id'>[] },
```
mattcasey, line 22: it looks like we only need the 'id' of wallet

pages/api/profile/items.ts
```diff
@@ -14,6 +15,24 @@ handler.use(requireUser).put(updateUserProfileItems);
...
+      wallets: {
+        select: {
+          id: true
+        }
+      }
+    }
+  });
+
+  const userWalletIds = userWallets.wallets.map((wallet) => wallet.id);
+  if (profileItems.some((profileItem) => !profileItem.walletId || !userWalletIds.includes(profileItem.walletId))) {
```
mattcasey, line 32: i'm not sure this logic is correct... shouldn't it be "profileItem.walletId && !userWallets.includes(profileItem.walletId)"?

mattcasey (approved):

PR #1988, opened 2023-04-07
Updated person property display

Devorein:
### <samp>🤖 Generated by Copilot at 77f7140</samp>

Improved the user interface and performance of the user property component in the board editor. Fixed a layout bug in the `propertyValueElement` component. Updated `user.tsx` and `propertyValueElement.tsx` files.

### WHY

1. All `person` properties in board view were not shown
2. Removed clearable icon for table view (notion)

### <samp>🤖 Generated by Copilot at 77f7140</samp>

* Optimize the rendering of the members display component in the user property value element by using `useMemo` hook and extracting the component from the `renderTags` prop of the `Autocomplete` component ([link](https://github.com/charmverse/app.charmverse.io/pull/1988/files?diff=unified&w=0#diff-7e7ec4a92a624e245c3d93562f7a2abe8bf71b9959ed50343f9ff73fda81a7fcL5-R5), [link](https://github.com/charmverse/app.charmverse.io/pull/1988/files?diff=unified&w=0#diff-7e7ec4a92a624e245c3d93562f7a2abe8bf71b9959ed50343f9ff73fda81a7fcL67-R99), [link](https://github.com/charmverse/app.charmverse.io/pull/1988/files?diff=unified&w=0#diff-7e7ec4a92a624e245c3d93562f7a2abe8bf71b9959ed50343f9ff73fda81a7fcL106-R138)) in `user.tsx`

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
@@ -80,6 +80,7 @@ function UserProperty(props: Props): JSX.Element | null {
       hideInput={props.readOnly || (props.displayType !== 'details' && !clicked)}
     >
       <InputSearchMemberMultiple
+        disableClearable
         open={isOpen}
         disableCloseOnSelect
         defaultValue={memberIds}
--- a/components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
@@ -2,7 +2,8 @@ import styled from '@emotion/styled';
 import CloseIcon from '@mui/icons-material/Close';
 import { IconButton } from '@mui/material';
 import { Box, Stack } from '@mui/system';
-import { useEffect, useMemo, useState } from 'react';
+import type { Dispatch, SetStateAction } from 'react';
+import { useEffect, useState } from 'react';
 
 import type { PropertyValueDisplayType } from 'components/common/BoardEditor/interfaces';
 import { InputSearchMemberMultiple } from 'components/common/form/InputSearchMember';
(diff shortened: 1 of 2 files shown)


components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
```diff
@@ -64,8 +64,39 @@ function UserProperty(props: Props): JSX.Element | null {
 
   const [isOpen, setIsOpen] = useState(false);
 
-  if (props.readOnly && memberIds.length === 0) {
-    return null;
+  const membersDisplayComponent = useMemo(() => {
```
mattcasey, line 67: what's the point of this memo? i find our templates are very hard to follow when using variables to capture a react component. make this a separate component or just do it inline

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -210,7 +210,7 @@ function PropertyValueElement(props: Props) {
   if (props.showTooltip) {
     return (
       <Tooltip title={props.propertyTemplate.name}>
-        <div>{propertyValueElement}</div>
+        <div style={{ width: '100%' }}>{propertyValueElement}</div>
```
mattcasey, line 213: is this certain not to break any other properties?

components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
```diff
@@ -80,6 +111,7 @@ function UserProperty(props: Props): JSX.Element | null {
       hideInput={props.readOnly || (props.displayType !== 'details' && !clicked)}
     >
       <InputSearchMemberMultiple
+        disableClearable
```
mattcasey, line 114: why dont we want to let users clear the person input?

mattcasey (changes requested):
left comments

components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
Devorein: The original reason was to both avoid duplication of inline and more verbose version of using another component (since there are quite a few props we need to pass). Making it a separate component seems to be better than just duplicating it twice.

components/common/BoardEditor/focalboard/src/components/properties/user/user.tsx
Devorein: We do allow that. If you click on the cell we show a cross icon beside each of the values, but there is no global clear icon in nothing either. It complicates styling as there is always a padding right when this is on (which causes the values to wrap to the next line even though there is spaces left)

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Yep I've tested it out for all views, both public and private, modal and everything looks alright.

mattcasey (approved):

PR #1994, opened 2023-04-10
Typeform webhook implementation

valentinludu:
[Task item](https://app.charmverse.io/charmverse/page-0990227035007396)

### WHAT
copilot:summary

### HOW
copilot:walkthrough


package.json
```diff
@@ -119,6 +119,7 @@
     "@skiff-org/prosemirror-tables": "^3.5.1",
     "@snapshot-labs/snapshot.js": "^0.4.60",
     "@socket.io/redis-adapter": "^7.2.0",
+    "@typeform/api-client": "^1.7.0",
```
valentinludu, line 122: I am using this official lib from typeform just to know the form response types.

prisma/schema.prisma
```diff
@@ -521,6 +526,7 @@ model User {
   GoogleCredential          GoogleCredential[]
   PageComment               PageComment[]
   verifiedEmails            VerifiedEmail[]
+  webhooks                  ApiPageKeys[]
```
mattcasey, line 529: 'webhooks' is confusing since we have other webhook settings on the space. Maybe 'apiPageKeys'?

lib/pages/createFormResponseCard.ts
```diff
@@ -52,6 +53,14 @@ export async function createFormResponseCard({
         }
       }
     });
+
+    relay?.broadcast(
```
mattcasey, line 57: you need to import relay. I'm not sure why it exists without the import

lib/apiPageKey/utilities.ts
```diff
@@ -0,0 +1,73 @@
+import type { Typeform } from '@typeform/api-client';
```
mattcasey, line 1: use 'utils' instead of utilities

pages/api/v1/databases/[id]/[key]/index.ts
```diff
@@ -0,0 +1,87 @@
...
+ *     responses:
+ *       200:
+ *         description: Summary of the database
+ *         content:
+ *            application/json:
+ *              schema:
+ *                $ref: '#/components/schemas/Page'
+ */
+export async function createFormResponse(req: NextApiRequest, res: NextApiResponse) {
+  const pageId = req.query.id as string;
```
Comment (mattcasey) [inline pages/api/v1/databases/[id]/[key]/index.ts:42]: Let's not require pageId in the URL, since it is already defined in the api page key. Using it in the URL also introduces the possibility that it could be incorrect

lib/apiPageKey/utilities.ts
```diff
@@ -0,0 +1,73 @@
...
+
+import { isTruthy } from 'lib/utilities/types';
+
+function simplifyTypeformAnswers(payload: Typeform.Response['answers']) {
```
mattcasey, line 5: Request: rename this folder to "lib/typeform"? this doesn't seem to have anything to do w/the api page key model.

prisma/schema.prisma
```diff
@@ -1364,3 +1370,19 @@ model PostCategoryPermission {
   @@index([postCategoryId, spaceId])
   @@index([postCategoryId, public])
 }
+
+model ApiPageKeys {
```
mattcasey, line 1374: Model names should be singular, not plural

mattcasey (changes requested):
comments

lib/apiPageKey/utilities.ts
valentinludu: Done

pages/api/v1/databases/[id]/[key]/index.ts
Comment (valentinludu) [reply pages/api/v1/databases/[id]/[key]/index.ts:42]: Removed pageId and moved it to `api/v1/webhooks/[key]`

I think this can be very flexible since I am getting the apiPageKey and then depending on what type it is I transform the forrm response into `AddFormResponseInput` type.

It's like this:

```
...general logic

if(type === 'typeform') {
   typeform logic
}

...validate

... create card

return
```

In the future we can have:

```
if(type === 'typeform') {
...custom logic
}

if(type === 'surveyMonkey') {
...custom logic
}

if(type === 'mailChimp') {
...custom logic
}
```

prisma/schema.prisma
valentinludu: Done

mattcasey (approved):

PR #1997, opened 2023-04-10
fix missing permissions for default role

mattcasey:
### <samp>🤖 Generated by Copilot at 67f9090</samp>

This pull request improves the role creation and management feature by allowing roles without users, avoiding empty relations, and ensuring default member permissions for spaces. It modifies the `CreateRoleInput` type and the `userIds` check in `pages/api/roles/index.ts`, and adds a default member permissions object in `lib/permissions/spaces/listPermissions.ts`.

### <samp>🤖 Generated by Copilot at 67f9090</samp>

*  Add default member permissions for spaces ([link](https://github.com/charmverse/app.charmverse.io/pull/1997/files?diff=unified&w=0#diff-ada17ca7ee1336082793c11b475cfea4ec9f5ff22a3910adf01a020122c5a319R46-R62))
*  Make `userIds` optional for creating roles ([link](https://github.com/charmverse/app.charmverse.io/pull/1997/files?diff=unified&w=0#diff-da5eeb2b032c79f526bc12377dc0a0bad5144be77e1121bbea2e5f9167af17b0L16-R16), [link](https://github.com/charmverse/app.charmverse.io/pull/1997/files?diff=unified&w=0#diff-da5eeb2b032c79f526bc12377dc0a0bad5144be77e1121bbea2e5f9167af17b0L71-R84))

### DIFF
--- a/lib/permissions/spaces/listPermissions.ts
+++ b/lib/permissions/spaces/listPermissions.ts
@@ -54,6 +54,7 @@ export async function listPermissions({ spaceId }: { spaceId: string }): Promise
         createPage: true,
         createBounty: true,
         createForumCategory: true,
+        moderateForums: false,
         reviewProposals: true
       }
     });
--- a/lib/permissions/spaces/listPermissions.ts
+++ b/lib/permissions/spaces/listPermissions.ts
@@ -43,5 +43,22 @@ export async function listPermissions({ spaceId }: { spaceId: string }): Promise
       .then((permissions) => permissions.map(mapSpacePermissionToAssignee))
   ]);
 
+  // add default member permissions if not defined
+  if (space.filter((s) => s.assignee.group === 'space').length === 0) {
+    space.push({
+      assignee: {
+        group: 'space',
+        id: spaceId
+      },
+      operations: {
+        createPage: true,
+        createBounty: true,
+        createForumCategory: true,
+        moderateForums: true,
+        reviewProposals: true
+      }
+    });
+  }
+
   return { proposalCategories, forumCategories, space };
 }
--- a/lib/permissions/spaces/listPermissions.ts
+++ b/lib/permissions/spaces/listPermissions.ts
@@ -51,11 +51,11 @@ export async function listPermissions({ spaceId }: { spaceId: string }): Promise
         id: spaceId
       },
       operations: {
-        createPage: true,
-        createBounty: true,
-        createForumCategory: true,
+        createPage: false,
+        createBounty: false,
+        createForumCategory: false,
         moderateForums: false,
-        reviewProposals: true
+        reviewProposals: false
       }
     });
   }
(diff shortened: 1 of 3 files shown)


pages/api/roles/index.ts
```diff
@@ -68,18 +68,20 @@ async function createRole(req: NextApiRequest, res: NextApiResponse<Role>) {
...
       }
     },
-    spaceRolesToRole: {
-      create: data.userIds.map((userId) => ({
-        spaceRole: {
-          connect: {
-            spaceUser: {
-              userId,
-              spaceId: data.spaceId
+    spaceRolesToRole: data.userIds
```
motechFR, line 71: Are we consuming this logic anywhere?

lib/permissions/spaces/listPermissions.ts
```diff
@@ -43,5 +43,22 @@ export async function listPermissions({ spaceId }: { spaceId: string }): Promise
       .then((permissions) => permissions.map(mapSpacePermissionToAssignee))
   ]);
 
+  // add default member permissions if not defined
+  if (space.filter((s) => s.assignee.group === 'space').length === 0) {
```
motechFR, line 47: I think it would be better to write this to the database so it's taken into account in other workflows

If for some reason there isn't a space permission, then the compute operation would fail.

We should probably also add a unit test for this

lib/permissions/spaces/listPermissions.ts
```diff
@@ -43,5 +43,22 @@ export async function listPermissions({ spaceId }: { spaceId: string }): Promise
...
+    space.push({
+      assignee: {
+        group: 'space',
+        id: spaceId
+      },
+      operations: {
+        createPage: true,
+        createBounty: true,
+        createForumCategory: true,
+        moderateForums: true,
```
motechFR, line 57: moderateForums should only be assigned to specific roles, as otherwise the whole space would be forum moderators

motechFR (approved):
Some feedback re: injecting the permission for list permissions

motechFR (changes requested):
meant to request changes

lib/permissions/spaces/listPermissions.ts
mattcasey: @motechFR I'm ok adding to the db - but how were permissions working before then? I think it would be better to just add a default value for any compute method, that way we don't need to make sure there's always one space permission?

lib/permissions/spaces/listPermissions.ts
mattcasey: OK changing to false, thanks!

lib/permissions/spaces/listPermissions.ts
mattcasey: @motechFR i checked with Chris and the space permissions were supposed to be "off" in his space, so I've updated the code so that everything is false by default. It looks like the other compute methods also assume false by default

mattcasey: let me know if there is a class or constructor i could use instead to generate the operations flag

PR #2013, opened 2023-04-13
Fix submitting header using google ime

msadura:
### WHAT
copilot:summary

### WHY
https://app.charmverse.io/charmverse/page-5320148491924774

### HOW
copilot:walkthrough


components/[pageId]/DocumentPage/components/PageTitleInput.tsx
```diff
@@ -72,23 +72,21 @@ export function PageTitleInput({ value, updatedAt: updatedAtExternal, onChange,
...
-      variant='standard'
-      onKeyDown={(e) => {
-        if (e.code === 'Enter') {
-          // prevent inserting a new line in title input
-          e.preventDefault();
-
-          insertAndFocusFirstLine(view);
-        }
+    <form
+      onSubmit={(e) => {
```
Comment (msadura) [inline components/[pageId]/DocumentPage/components/PageTitleInput.tsx:76]: This is actually interesting find for the future - `onKeydown` is triggered ALWAYS even if IME (extension) is opened, while "submit" is triggered only when we need it

PR #2024, opened 2023-04-14
remove user data from result of inline comments

mattcasey:
### <samp>🤖 Generated by Copilot at 9974c22</samp>

Refactored the types and queries of comments and threads in the app to simplify the data structure and reduce unnecessary data fetching. Removed the user details from the `Comment` type and the comments API, and added them to the threads hook using `useMembers`. Updated the integration tests, the CharmEditor component, and the threads API to reflect these changes.

My next step will be to change the interface of useThreads, and have compnents get member info from useMembers() instead. I'd also like to consider renaming that to "useInliineComments" cc @Devorein

### DIFF
--- a/charmClient/apis/commentsApi.ts
+++ b/charmClient/apis/commentsApi.ts
@@ -1,15 +1,22 @@
+import type { Comment } from '@prisma/client';
+
 import * as http from 'adapters/http';
 import type { CommentCreate, CommentWithUser } from 'lib/comments/interfaces';
 import type { PageContent } from 'lib/prosemirror/interfaces';
-import type { MultipleThreadsInput, ThreadCreate, ThreadWithCommentsAndAuthors } from 'lib/threads/interfaces';
+import type {
+  MultipleThreadsInput,
+  ThreadCreate,
+  ThreadWithComments,
+  ThreadWithCommentsAndAuthors
+} from 'lib/threads/interfaces';
 import type { ResolveThreadRequest } from 'pages/api/threads/[id]/resolve';
 
 export class CommentsApi {
-  addComment(request: Omit<CommentCreate, 'userId'>): Promise<CommentWithUser> {
+  addComment(request: Omit<CommentCreate, 'userId'>): Promise<Comment | CommentWithUser> {
     return http.POST('/api/comments', request);
   }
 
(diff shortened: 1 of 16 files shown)


hooks/useThreads.tsx
```diff
@@ -32,17 +33,38 @@ export const ThreadsContext = createContext<Readonly<IContext>>({
...
     { revalidateOnFocus: false }
   );
+
+  function populateThreads(_threads: ThreadWithComments[]): CommentThreadsMap {
+    const threadsandAuthors = _threads.reduce<CommentThreadsMap>((acc, thread) => {
+      const newThread: ThreadWithCommentsAndAuthors = {
+        ...thread,
+        comments: thread.comments.map((comment) => ({
+          ...comment,
+          user: members.find((m) => m.id === comment.userId) || ({} as any)
```
Devorein, line 55: Making user `User | null` would guarantee type safety

charmClient/apis/commentsApi.ts
```diff
@@ -1,27 +1,34 @@
...
 import type { ResolveThreadRequest } from 'pages/api/threads/[id]/resolve';
 
 export class CommentsApi {
-  addComment(request: Omit<CommentCreate, 'userId'>): Promise<CommentWithUser> {
+  addComment(request: Omit<CommentCreate, 'userId'>): Promise<Comment | CommentWithUser> {
     return http.POST('/api/comments', request);
   }
 
-  editComment(commentId: string, content: PageContent): Promise<CommentWithUser> {
+  editComment(commentId: string, content: PageContent): Promise<Comment | CommentWithUser> {
```
Devorein, line 19: Do we need to support `CommentWithUser` if the user is attached in fe?

Devorein (approved):
Rather than keeping the type `ThreadWithCommentsAndAuthors`, where ever the user is required for comment and thread we can attach the user from that component. Otherwise we need to keep track of two different type `ThreadWithCommentsAndAuthors` on the frontend and `ThreadWithComments` for the backend

charmClient/apis/commentsApi.ts
mattcasey: I am being mindful of the previous build, to make sure that the frontend still works with the previous response type. I plan to do a follow-up PR and remove this

PR #2026, opened 2023-04-16
Fixed clicking inline comments

Devorein:
### <samp>🤖 Generated by Copilot at e9194ba</samp>

This pull request improves the charm editor functionality by fixing a bug with the `inline comment spec`. It changes the `components/common/CharmEditor/specRegistry.ts` file to prevent the spec from interfering with other charm editing features.

### <samp>🤖 Generated by Copilot at e9194ba</samp>

*  Adjust the inline comment spec to avoid conflicts with other specs ([link](https://github.com/charmverse/app.charmverse.io/pull/2026/files?diff=unified&w=0#diff-3707d3aea700881fc2e310324b90f45209fa7496c3082f20607ef217b984eb10L57), [link](https://github.com/charmverse/app.charmverse.io/pull/2026/files?diff=unified&w=0#diff-3707d3aea700881fc2e310324b90f45209fa7496c3082f20607ef217b984eb10R90-R91))

### DIFF
--- a/components/common/CharmEditor/specRegistry.ts
+++ b/components/common/CharmEditor/specRegistry.ts
@@ -54,7 +54,6 @@ export const specRegistry = new SpecRegistry([
   // MAKE SURE THIS IS ALWAYS AT THE TOP! Or deleting all contents will leave the wrong component in the editor
   paragraph.spec(), // OK
   mentionSpecs(), // NO
-  inlineComment.spec(),
   inlineVote.spec(),
   bold.spec(), // OK
   bulletList.spec(), // OK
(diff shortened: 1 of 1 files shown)


msadura (approved):
Looks ok, but what was the issue and why it needs to be below text plugins?

Devorein:
Two points I forgot to mention:-
1. The order in the spec determines the order of application. Previously the inline comment spec was at the top, before the text format and track spec. Thus it was applied before those specs. This resulted in a document like this `div.inline-comment -> b/em (bold or italic formatting) -> div.track-plugin`
2. The logic for showing the inline comment ui is located in the inline comment plugin. We check whether the clicked dom node or its parent has a specific class or not. Since the format and track dom nodes were surrounding the inline comment it didn't get triggered. Its located over here. Moving the spec was a more practical solution than traversing all the parent nodes and concatenating their class names, as there is no definite upper limit of nesting (the inline comment can be surrounded by multiple text formats each with their own nodes) `components/common/CharmEditor/components/inlineComment/inlineComment.plugins.tsx:63`

PR #2028, opened 2023-04-17
Fix inline comment box not showing for bounty and proposal cards

Devorein:
### <samp>🤖 Generated by Copilot at 1d1341e</samp>

The pull request refactors the logic for getting the `cardId` parameter from the URL in several components and functions related to inline commenting and highlighting in the editor. It also moves the `SuggestionsPopup` component to the same file as the `InlineCommentSubMenu` component for better cohesion.

### <samp>🤖 Generated by Copilot at 1d1341e</samp>

*  Extract `cardId` parameter from URL query string and add fallbacks for `id` and `bountyId` in `inlineComment.components.tsx` and `highlightMarkedElement.ts` ([link](https://github.com/charmverse/app.charmverse.io/pull/2028/files?diff=unified&w=0#diff-275e846564ebd2c6850d6c25f2b00a47870c6bf633abdbaf9c53a4005326cd62L51-R52), [link](https://github.com/charmverse/app.charmverse.io/pull/2028/files?diff=unified&w=0#diff-275e846564ebd2c6850d6c25f2b00a47870c6bf633abdbaf9c53a4005326cd62L114-R115), [link](https://github.com/charmverse/app.charmverse.io/pull/2028/files?diff=unified&w=0#diff-005518828a991147f9f230bb147cb316d26eaae08981cbc7aafee05b31989babL33-R33), [link](https://github.com/charmverse/app.charmverse.io/pull/2028/files?diff=unified&w=0#diff-303290dcca61f52b43238d613694b1caebbb7f87bc1586ffc1852b0fa0e77bf4L55-R55))
* Move `SuggestionsPopup` component from `SuggestionPopup.tsx` to `inlineComment.components.tsx` and update `isInPageDialog` variable computation ([link](https://github.com/charmverse/app.charmverse.io/pull/2028/files?diff=unified&w=0#diff-005518828a991147f9f230bb147cb316d26eaae08981cbc7aafee05b31989babL33-R33))

### DIFF
--- a/components/common/CharmEditor/components/inlineComment/inlineComment.components.tsx
+++ b/components/common/CharmEditor/components/inlineComment/inlineComment.components.tsx
@@ -48,7 +48,8 @@ export default function InlineCommentThread({ pluginKey }: { pluginKey: PluginKe
   const { tooltipContentDOM, show: isVisible, ids } = usePluginState(pluginKey) as InlineCommentPluginState;
   const { threads } = useThreads();
 
-  const cardId = new URLSearchParams(window.location.href).get('cardId');
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const cardId = urlSearchParams.get('cardId') || urlSearchParams.get('id') || urlSearchParams.get('bountyId');
 
   const { currentPageActionDisplay } = usePageActionDisplay();
   // Find unresolved threads in the thread ids and sort them based on desc order of createdAt
@@ -111,7 +112,7 @@ export function InlineCommentSubMenu({ pluginKey }: { pluginKey: PluginKey }) {
 
   const handleSubmit = async (e: React.KeyboardEvent<HTMLElement> | React.MouseEvent<HTMLElement, MouseEvent>) => {
     if (!isEmpty) {
-      const cardId = typeof window !== 'undefined' ? new URLSearchParams(window.location.href).get('cardId') : null;
+      const cardId = typeof window !== 'undefined' ? new URLSearchParams(window.location.search).get('cardId') : null;
       e.preventDefault();
       const threadWithComment = await charmClient.comments.startThread({
         comment: commentContent,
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/components/inlineComment/inlineComment.components.tsx
```diff
@@ -48,7 +48,8 @@ export default function InlineCommentThread({ pluginKey }: { pluginKey: PluginKe
   const { tooltipContentDOM, show: isVisible, ids } = usePluginState(pluginKey) as InlineCommentPluginState;
   const { threads } = useThreads();
 
-  const cardId = new URLSearchParams(window.location.href).get('cardId');
+  const urlSearchParams = new URLSearchParams(window.location.search);
+  const cardId = urlSearchParams.get('cardId') || urlSearchParams.get('id') || urlSearchParams.get('bountyId');
```
mattcasey, line 52: Can you try replacing this with `useCurrentPage()`? That is a more direct place to get current state, it also will prevent further bugs if we use a new id in the future.

lib/prosemirror/highlightMarkedElement.ts
```diff
@@ -52,7 +52,7 @@ export function highlightElement({
   // Page action list node might not be present
   const isShowingActionList = !!pageActionListNode && pageActionListNode.style.visibility !== 'hidden';
   // Check if we are inside a card page modal
-  const cardId = new URLSearchParams(window.location.href).get('cardId');
+  const cardId = new URLSearchParams(window.location.search).get('cardId');
```
mattcasey, line 55: maybe we don't need this at all? we don't use the value below, just checking for its existence

mattcasey (changes requested):
see comments

mattcasey (approved):
Approved since it at least fixes the bug, but had some suggestions to make it better

components/common/CharmEditor/components/inlineComment/inlineComment.components.tsx
Devorein: Right. We did decide not to rely on the url to provide the current state. I will make use of `useCurrentPage` and see if it works in all cases or not

lib/prosemirror/highlightMarkedElement.ts
Devorein: Agree. Rather than making use of dom for the state, I passed `insideModal` to to components and functions that needs it. Its working correctly now without any sort of reliance of url or dom for tracking state.

PR #2046, opened 2023-04-20
Fixed person property disappearing in small screens

Devorein:
### <samp>🤖 Generated by Copilot at e4293c7</samp>

Fixed a layout bug in the card detail modal of the board editor component. Added a `width` property to the `div` element that wraps the card detail properties in `cardDetailProperties.tsx`.

### WHY
Before
image

Now
image

### <samp>🤖 Generated by Copilot at e4293c7</samp>

*  Added `width` property to the style object of the `div` element that wraps the card detail properties to prevent overflow on window resize ([link](https://github.com/charmverse/app.charmverse.io/pull/2046/files?diff=unified&w=0#diff-a07279e14123e2d30e0ff8c44c042ca57fa8558dad05c1a5ff3166b69b3fadddL95-R96))


mattcasey (changes requested):
We should avoid calculating CSS, especially with page width and arbitrary values - on mobile the div is much larger than the actual screen width, this could cause other bugs in the future or not with other property types

It looks like the real issue is we have "overflow: hidden" on the StyledUserPropertyContainer. Why is that needed?

Devorein:
Its for the column wrap property, we want to hide the overflown portion when its not toggled

PR #2050, opened 2023-04-20
Typeform Db page - add banner and custom title

valentinludu:
Added a banner for db pages to show the webhook type and link + split the banner components.
Overwrite the title coming from a csv, if the db page has an api page key generated for typeform + test.


components/common/BoardEditor/focalboard/src/components/viewSidebar/viewSourceOptions.tsx
```diff
@@ -145,7 +145,7 @@ export function ViewSourceOptions(props: ViewSourceOptionsProps) {
             <br />
             Paste the following URL:
             <br />
-            <i>{`${webhookBaseUrl}/${webhookApi?.apiKey}`}</i>
+            <i>{`${Constants.webhookBaseUrl}/${webhookApi?.apiKey}`}</i>
```
mattcasey, line 148: this is "ok" but it would be better to put all our constants in /config/constants

mattcasey (approved):

PR #2070, opened 2023-04-25
apply font styles to sidebar

mattcasey:
### <samp>🤖 Generated by Copilot at ce9dce3</samp>

Add font switcher feature to `Header` component. This allows the user to change the font of the app content between `monoFont` and `serifFont`.

### WHY
It didn't work because Popover creates itself as a child outside the span tag in _app which assigns the font variable. Next.js requires us to use it in _app and not _document, so to add it to the 'body' tag I'd have to do it in a useEffect()

image

### <samp>🤖 Generated by Copilot at ce9dce3</samp>

*  Import and apply font variables from `theme/fonts` module to enable font switching feature ([link](https://github.com/charmverse/app.charmverse.io/pull/2070/files?diff=unified&w=0#diff-97400afdebf25587839de062948405ef1a7258a6ede291863ed9dd8b2ce2ba5eR44), [link](https://github.com/charmverse/app.charmverse.io/pull/2070/files?diff=unified&w=0#diff-97400afdebf25587839de062948405ef1a7258a6ede291863ed9dd8b2ce2ba5eL386-R387))

### DIFF
--- a/components/common/PageLayout/components/Header/Header.tsx
+++ b/components/common/PageLayout/components/Header/Header.tsx
@@ -41,6 +41,7 @@ import { useSnackbar } from 'hooks/useSnackbar';
 import { useToggleFavorite } from 'hooks/useToggleFavorite';
 import { useUser } from 'hooks/useUser';
 import type { PageUpdates } from 'lib/pages';
+import { monoFont, serifFont } from 'theme/fonts';
 
 import { BountyActions } from './components/BountyActions';
 import BountyShareButton from './components/BountyShareButton/BountyShareButton';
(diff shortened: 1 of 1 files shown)


motechFR (approved):
Tested and works well for example characters.

Should we also extend this to the option names? (ie. styling Default, Serif, Mono labels)

Screenshot 2023-04-26 at 00 57 53

mattcasey:
> Should we also extend this to the option names? (
I think of the "Aa" are more like icons here, where as the labels are normal UI text. But also, I am following Notion's design

PR #2074, opened 2023-04-26
Fix create new user on disconnect wallet

valentinludu:
Fixes:
- the modal was opening for the same first wallet to disconnect
- on wallet disconnect, a new user was created because the signature and the account connected were different
   * I am creating a new signature if another wallet is in the user account.
- mui error for a button
- redirect to createSpace instead of signup that does not exist anymore

To test
- when disconnecting a wallet, don't logout the user from his space


components/common/RouteGuard.tsx
```diff
@@ -104,7 +104,7 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
         return {
           authorized: false,
           redirect: {
-            pathname: spaces.length !== 0 ? `/${spaces[0].domain}` : '/signup'
+            pathname: spaces.length !== 0 ? `/${spaces[0].domain}` : '/createSpace'
```
mattcasey, line 107: nice cleanup - i think the redirect in next.config was 'fixing' this ?

components/integrations/components/IdentityProviders.tsx
```diff
@@ -198,15 +205,16 @@ export function IdentityProviders() {
                     Are you sure you want to Disconnect your {wallet.ensname || wallet.address} wallet?
                   </Typography>
                   <Typography variant='body2'>
-                    This action will remove your wallet, NFTs, POAPs, Organizations from CharmVerse. It will also remove
-                    roles and permissions if you joined the Space via a token gate.
+                    This action will {lowerCaseEqual(wallet.address, account) ? 'log you out,' : ''} remove your wallet,
```
mattcasey, line 208: good idea to warn the user

hooks/useWeb3AuthSig.tsx
```diff
@@ -253,14 +253,10 @@ export function Web3AccountProvider({ children }: { children: ReactNode }) {
     {
       async onSuccess(updatedUser) {
         logoutWallet();
-        setUser(updatedUser);
```
mattcasey, line 256: @motechFR please review that not setting these to null is ok, not sure what the intended effect is

mattcasey (approved):

motechFR (approved):
Tested and worked!

PR #2076, opened 2023-04-26
add retry to server-side fetch

mattcasey:
### WHAT
copilot:summary

### HOW
copilot:walkthrough


adapters/http/__test__/fetch.server.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('Http retry client tests', () => {
+  afterAll(async () => {
+    await mockPool.close();
+  });
+
+  it('Should retry on 5xx', async () => {
+    mockPool.intercept({ path: '/' }).reply(500);
+    mockPool.intercept({ path: '/' }).reply(200, { success: true }, responseOptions);
+
+    const request = fetch(DUMMY_BASE_URL);
```
msadura, line 26: Just note: I was wondering what is detault behavior of `fetch-retry`, according to their docs: `default to 3 retries, with a 1000ms retry delay`

package.json
```diff
@@ -226,6 +227,7 @@
     "tslib": "^2.4.0",
     "twemoji": "^14.0.1",
     "typescript": "^5.0.2",
+    "undici": "^5.22.0",
```
msadura, line 230: out of curiosity - why this package? I am asking because I did not hear about it before, and I am wondering why this  over `fetch`

msadura (approved):

adapters/http/__test__/fetch.server.spec.ts
mattcasey: i've overridden the defaults

package.json
mattcasey: fetch uses undici under the hood!  It took me some time to figure that out, here's the actual commit: https://github.com/nodejs/node/commit/6ec225392675c92b102d3caad02ee3a157c9d1b7 

I wanted to use their mocking utils, but if we are keeping the fetch paradigm I like the jest-fetch-mock lib instead. I just didn't realize we already had a solution until the test failed heh

That said, I will still use undici - jest-fetch-mock does not work when I apply fetch-retry for some reason, but I'm using undici to be able to mock out the fetch method that gets used

PR #2078, opened 2023-04-27
use next in development mode

mattcasey:
### WHAT
copilot:summary

### HOW
copilot:walkthrough


msadura:
this looks good, the only question is - can we trust "custom" server on prod will work the same way as `next dev` locally?

msadura (approved):

mattcasey:
> this looks good, the only question is - can we trust "custom" server on prod will work the same way as next dev locally?
I feel comfortable because our tests are using the production build w/server.js. Also, I am going to test removing the custom server maybe tomorrow or this weekend.

PR #2097, opened 2023-05-02
Restore optimism in token gates

msadura:
### WHAT
- remove patch that was deleting optimism from chain list
- add a prop to lit modal which can be used to set list of allowed chains which should be a proper way of disabling some of them https://github.com/LIT-Protocol/lit-share-modal-v3#optional


mattcasey (approved):
my bad for forgetting to update this when we brought Optimism back!

PR #2100, opened 2023-05-03
Typeform - map the types & CSV - Transform Options

valentinludu:
### <samp>🤖 Generated by Copilot at ecb7af3</samp>

This pull request adds new functionality to handle typeform webhooks and create or update database cards with the form responses. It refactors some existing logic to use common functions and modules, and adds new types, fields, and tests. It also allows multi-select properties in database card pages and updates the board and its views accordingly. The main files affected are `pages/api/v1/webhooks/addToDatabase/[key]/index.ts`, `lib/pages/utils.ts`, `lib/typeform/simplifyTypeformResponse.ts`, and `lib/public-api/updateDatabaseBlocks.ts`.

### WHY
Typeform should add the types for each question as a card property
CSV import should:
- remove columns if their name is equal to an option from a card property
- add the values as options if the case above occurs


lib/pages/utils.ts
```diff
@@ -10,3 +16,47 @@ export function canReceiveManualPermissionUpdates({ pageType }: { pageType: Page
   }
   return true;
 }
+
+export function transformWebhookBody(data: BodyFormResponse, properties: FormResponseProperty[]) {
```
msadura, line 20: I am not entirely sure about name (and placement) for this function, looks like `transformWebhookBody` feels too generic, because it does very specific oparation.

msadura (approved):

lib/pages/utils.ts
valentinludu: Yeah, you're right. I moved it inside lib/typeform and renamed to be more aligned with what it does.

PR #2118, opened 2023-05-05
Fix deletion for inline dbs

valentinludu:
### <samp>🤖 Generated by Copilot at b3b44b8</samp>

This pull request improves the functionality and reliability of the board editor component and the `modifyChildPages` function. It refactors the `centerPanel.tsx` component to use a custom hook and handle errors for linked boards. It adds unit tests for the `modifyChildPages` function in the `modifyChildPages.spec.ts` file. It also fixes a bug in the `modifyChildPages` function by changing the `where` clause of the database query.

### WHY
An inline database that was deleted was being seen in the document. Also after permanen deletion it was still shown but couldn't be interacted with.

Also cleaned up the centerPage of the old code that was messing up the visibiloity of google forms responses. 

FIY: 
- 50% of this PR is me reverting this PR https://github.com/charmverse/app.charmverse.io/pull/2080
- And then improving the deletion process not to affect the user but hide the table.
- The approach i've taken is 100% exactly like Notion does it


lib/pages/modifyChildPages.ts
```diff
@@ -46,7 +46,9 @@ export async function modifyChildPages(parentId: string, userId: string, action:
             {
               id: {
                 in: modifiedChildPageIds
-              },
+              }
```
msadura, line 49: just making sure - isn't there any mistake? Just asking because it is hard to understand this diff ;p

msadura (approved):

lib/pages/modifyChildPages.ts
valentinludu: The mistake was before:
It was 
```
prisma.block.deleteMany({
        where: {
          OR: [
            {
              id: {
                in: modifiedChildPageIds
              },
              parentId: {
                in: modifiedChildPageIds
              }
            },
          ]
        }
      })
```

and not it is 
```
prisma.block.deleteMany({
        where: {
          OR: [
            {
              id: {
                in: modifiedChildPageIds
              }
            },
            {
              parentId: {
                in: modifiedChildPageIds
              }
            }
          ]
        }
      })
```

valentinludu: With this in place now we also delete the view block of a board

PR #2128, opened 2023-05-10
Feat/save safe tx hash multisig

msadura:
### WHAT
copilot:summary

### WHY
Add url to gnosis wallet when multisig tx is in processing state


hooks/useGnosisTx.tsx
mattcasey, line 1: it would be better to call this "useGnosisTransaction" abbreviations make it hard to know what something is if you are not familiar with the domain space

PR #2132, opened 2023-05-10
Stripe integration

Devorein:
Local Stripe Setup:
1. Log in to your Stripe account.
2. Navigate to the Developers section of the Stripe dashboard.
3. Click on the API keys tab.
4. You will see two types of API keys: Test and Live. For local development, you should use the Test API keys.
5. Copy the Test API keys and configure them in your development environment (`NEXT_PUBLIC_STRIPE_PUBLIC_KEY` & `STRIPE_SECRET_KEY`).
6. Run `scripts/createStripeProducts.ts` script

@mattcasey you were right, I couldn't find any specific information on why its a good practice to send the stripe public key from your backend.

### <samp>🤖 Generated by Copilot at 32ff600</samp>

This pull request adds subscription management features to the app. It introduces new frontend components, such as `CheckoutForm`, `PaymentForm`, and `SubscriptionSettings`, to render the subscription UI and interact with the Stripe elements and API. It also adds new backend functions and API routes, such as `createSubscription`, `getSpaceSubscription`, and `/api/subscription/*`, to create and retrieve the subscription data from the Stripe API and the database. It updates the `CharmClient` class, the settings dialog component, and the `package.json` file to support the new features. It also includes some minor formatting and cleanup changes in some files.


lib/subscription/utils.ts
```diff
@@ -0,0 +1,53 @@
+export type SubscriptionUsage = 1 | 2 | 3 | 4 | 5;
+
+export type SubscriptionPeriod = 'monthly' | 'annual';
+
+export const SubscriptionUsageRecord: Record<
```
mattcasey, line 5: you can use THIS_CASE or camelCase.. capitalizing the first letter should be reserved for types/classes

components/subscription/CheckoutForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+            aria-label='usage'
+            valueLabelDisplay='off'
+            value={usage}
+            marks={Object.keys(SubscriptionUsageRecord).map((_usage) => ({
+              value: Number(_usage),
+              label: `$${SubscriptionUsageRecord[parseInt(_usage) as SubscriptionUsage].pricing[period]}/${
+                period === 'annual' ? 'yr' : 'mo'
+              }`
+            }))}
+            min={Object.keys(SubscriptionUsageRecord).map(Number)[0]}
```
mattcasey, line 139: please replace "SubscriptionUsage" with an array of numbers. Then u can just take the first and last in the list

lib/subscription/utils.ts
mattcasey, line 1: rename utils.ts => config.ts or constants.ts. There's no utils in here :)

lib/subscription/interfaces.ts
```diff
@@ -0,0 +1,9 @@
+import type { SubscriptionTier } from '@charmverse/core/dist/cjs/prisma';
+
+import type { SubscriptionUsage, SubscriptionPeriod } from './utils';
+
+export type SpaceSubscription = {
```
mattcasey, line 5: I suggest just define this in SpaceSubscription.ts and we don't need an interfaces.ts file

lib/subscription/getSpaceSubscription.ts
```diff
@@ -0,0 +1,33 @@
+import { prisma } from '@charmverse/core';
+import Stripe from 'stripe';
+
+import { NotFoundError } from 'lib/middleware';
+
+import type { SpaceSubscription } from './interfaces';
+
+const stripe = new Stripe(process.env.STRIPE_SECRET_KEY as string, {
```
mattcasey, line 8: this should be moved to a "stripe.ts" file so it's only set up once

lib/subscription/getSpaceSubscription.ts
```diff
@@ -0,0 +1,33 @@
...
+    where: {
+      id: spaceId
+    },
+    select: {
+      subscriptionId: true
+    }
+  });
+
+  if (!space) {
+    throw new NotFoundError('Space not found');
```
mattcasey, line 23: i recommend just using findUniqueOrThrow() - it's less code

components/subscription/SubscriptionSettings.tsx
```diff
@@ -0,0 +1,102 @@
...
+import LoadingComponent from 'components/common/LoadingComponent';
+import Legend from 'components/settings/Legend';
+import { useMembers } from 'hooks/useMembers';
+import { SubscriptionUsageRecord } from 'lib/subscription/utils';
+
+import { CheckoutForm } from './CheckoutForm';
+
+const stripePublicKey = process.env.NEXT_PUBLIC_STRIPE_PUBLIC_KEY as string;
+
+const stripePromise = loadStripe(stripePublicKey);
```
mattcasey, line 21: this should only be loaded when the component is rendered

charmClient/apis/subscriptionApi.ts
```diff
@@ -0,0 +1,16 @@
...
+  CreatePaymentSubscriptionResponse
+} from 'pages/api/subscription/subscribe';
+
+export class SubscriptionApi {
+  createSubscription(payload: CreatePaymentSubscriptionRequest) {
+    return http.POST<CreatePaymentSubscriptionResponse>('/api/subscription/subscribe', payload);
+  }
+
+  getSpaceSubscription({ spaceId }: { spaceId: string }) {
+    return http.GET<SpaceSubscription>(`/api/spaces/${spaceId}/subscription`);
```
mattcasey, line 14: i believe this can also be null

components/subscription/CheckoutForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+  onCancel: VoidFunction;
+  refetch: KeyedMutator<SpaceSubscription>;
+}) {
+  const stripe = useStripe();
+  const elements = useElements();
+
+  const space = useCurrentSpace();
+  const [isProcessing, setIsProcessing] = useState(false);
+  const { showMessage } = useSnackbar();
+  const [period, setPeriod] = useState<SubscriptionPeriod>(spaceSubscription?.period ?? 'monthly');
```
mattcasey, line 44: Since `spaceSubscription` is always undefined/null to begin with, there's no need to pass it in as a default. You should just set it to "monthly" and usage to 1

components/subscription/CheckoutForm.tsx
```diff
@@ -0,0 +1,259 @@
...
+            mx: 2
+          }}
+        >
+          <Slider
+            disabled={isProcessing}
+            size='small'
+            aria-label='usage'
+            valueLabelDisplay='off'
+            value={usage}
+            marks={Object.keys(SubscriptionUsageRecord).map((_usage) => ({
```
mattcasey, line 133: it'd be cleaner if this was just an array of numbers, then no need to use parseInt and Number below

components/subscription/CheckoutForm.tsx
mattcasey, line 1: might be a good candidate to add to Storybook?

components/settings/config.tsx
```diff
@@ -20,8 +21,10 @@ export const SETTINGS_TABS: SpaceSettingsTab[] = [
   { icon: <SettingsIcon fontSize='small' />, path: 'space', label: 'Overview' },
   { icon: <UserRoleIcon fontSize='small' />, path: 'roles', label: 'Roles & Permissions' },
   { icon: <GroupAddOutlinedIcon fontSize='small' />, path: 'invites', label: 'Invites' },
-  { icon: <CloudSyncOutlinedIcon fontSize='small' />, path: 'api', label: 'API' }
+  { icon: <CloudSyncOutlinedIcon fontSize='small' />, path: 'api', label: 'API' },
+  { icon: <AttachMoneyIcon fontSize='small' />, path: 'subscription', label: 'Subscription' }
```
mattcasey, line 25: if you do deploy to production, be sure to hide this unless the space tier is 'free'

mattcasey (commented):
left some comments, some r just suggesitons

lib/subscription/createSubscription.ts
```diff
@@ -0,0 +1,80 @@
...
+  apiVersion: '2022-11-15'
+});
+
+export async function createSubscription({
+  paymentMethodId,
+  spaceId,
+  spaceDomain,
+  period,
+  usage,
+  tier = 'pro'
```
motechFR, line 18: I suggest we rename this method to createProSubscription without a tier.

Reasoning:
- free doesn't need this method
- Enterprise would be per company workflow

pages/api/subscription/subscribe.ts
```diff
@@ -0,0 +1,65 @@
...
+  spaceId: string;
+  paymentMethodId: string;
+  usage: SubscriptionUsage;
+  period: SubscriptionPeriod;
+};
+
+export type CreatePaymentSubscriptionResponse = {
+  clientSecret: string | null;
+  paymentIntentStatus: stripe.PaymentIntent.Status | null;
+};
```
motechFR, line 21: How about moving these interfaces to the lib method? Looks cleaner and more in line with our patterns

pages/api/subscription/subscribe.ts
```diff
@@ -0,0 +1,65 @@
...
+    throw new UnauthorisedActionError();
+  }
+
+  if (!space) {
+    throw new NotFoundError('Space not found');
+  }
+
+  if (space.subscriptionId) {
+    throw new InvalidStateError('Space already has a subscription');
+  }
```
motechFR, line 49: This code would be better suited inside the lib method, where we can do input validation and testing

pages/api/spaces/[id]/subscription.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireUser).get(getSpaceSubscriptionController);
+
+async function getSpaceSubscriptionController(req: NextApiRequest, res: NextApiResponse<SpaceSubscription | null>) {
+  const { id: spaceId } = req.query as { id: string };
+  const userId = req.session.user.id;
+
+  const spaceAccess = await hasAccessToSpace({ spaceId, userId, adminOnly: false });
```
Comment (motechFR) [inline pages/api/spaces/[id]/subscription.ts:18]: Purely informative, we have a requireSpaceMembership middleware you can use in line 12, instead of inside the function

scripts/createStripeProducts.ts
```diff
@@ -0,0 +1,19 @@
+import { SubscriptionUsageRecord } from 'lib/subscription/utils';
+import Stripe from 'stripe';
+
+const stripe = new Stripe(process.env.STRIPE_SECRET_KEY as string, {
+  apiVersion: '2022-11-15'
+});
+
+async function createStripeProducts() {
```
motechFR, line 8: Awesome!

components/settings/config.tsx
motechFR: One reason we might want this to always show; managing your current subscription

lib/subscription/utils.ts
Devorein: got it, going for CONSTANT_CASE

lib/subscription/getSpaceSubscription.ts
Devorein: Good point. Thanks.

charmClient/apis/subscriptionApi.ts
Devorein: Correct

components/subscription/CheckoutForm.tsx
Devorein: Done

lib/subscription/createSubscription.ts
Devorein: Gotcha

pages/api/subscription/subscribe.ts
Devorein: Lets do it. The import statement isn't the most visually pleasing either `import type {} from api/pages/[id]/[spaceId] ....`

components/subscription/SubscriptionSettings.tsx
Devorein: Not really, since it should only be loaded one time, it was relocated outside the component. But I can move it inside the component and use a ref to initialize it.

lib/subscription/createSubscription.ts
```diff
@@ -0,0 +1,77 @@
+import { prisma } from '@charmverse/core';
+import type { SubscriptionTier } from '@charmverse/core/prisma';
+import type Stripe from 'stripe';
+
+import { stripeClient } from './stripe';
+import type { SubscriptionPeriod, SubscriptionUsage } from './utils';
```
valentinludu, line 6: I presume this should be ./constants
Here and in all the places.

Devorein: Correct

lib/subscription/createProSubscription.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (space.subscriptionId) {
+    throw new InvalidStateError('Space already has a subscription');
+  }
+
+  // Create a customer
+  const customer = await stripeClient.customers.create({
+    name: fullName,
+    payment_method: paymentMethodId,
+    invoice_settings: { default_payment_method: paymentMethodId },
+    email: billingEmail
```
motechFR, line 64: What if this customer is paying for multiple spaces with same billing email? Wouldn't stripe throw an error creating  a new client with same email?

lib/subscription/createProSubscription.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const invoice = await stripeClient.invoices.create({
+    customer: customer.id,
+    description: `Invoice for ${capitalize(period)} Community Subscription`,
+    currency: 'usd',
+    subscription: subscription.id,
+    days_until_due: period === 'monthly' ? 30 : 365
+  });
+
+  await stripeClient.invoices.sendInvoice(invoice.id);
```
motechFR, line 108: is the invoice for the current or the next payment?

I also suggest extracting invoice creation to its own method

lib/subscription/getSpaceSubscription.ts
```diff
@@ -0,0 +1,30 @@
...
+    select: {
+      subscriptionId: true
+    }
+  });
+
+  if (!space.subscriptionId) {
+    return null;
+  }
+
+  const subscription = await stripeClient.subscriptions.retrieve(space.subscriptionId);
```
valentinludu, line 27: What about if it's a coinbase id instead of a stripe payment id? 
Some options for this:
1. Have also a `subscriptionPaymenyType: 'coinbase' | 'stripe'` => to know if we need to query for info coinbaseClient or stripeClient.
2. `coinbasePaymentId` and `stripePaymentId`

I am also open to any suggestions.

lib/subscription/createProSubscription.ts
Devorein: I've removed this code portion as Stripe automatically creates an invoice when creating a subscription

lib/subscription/createProSubscription.ts
Devorein: It won't throw any error, but it's definitely good practice to reuse the existing customer based on the email address.

components/settings/config.tsx
Devorein: We have decided to only show subscription for charmverse related spaces for now

mattcasey (approved):

PR #2138, opened 2023-05-11
Fixed inline palette suggestions

Devorein:
### <samp>🤖 Generated by Copilot at 988def1</samp>

Added a new option to customize the tooltip trigger characters for different marks in the `CharmEditor` component. This prevents the tooltips from interfering with the placeholders for the inline palette and the image selector.

### DIFF
--- a/components/common/CharmEditor/components/ResizableImage.tsx
+++ b/components/common/CharmEditor/components/ResizableImage.tsx
@@ -239,7 +239,7 @@ function getFileBinary(src: string): File | null {
 
 export function spec() {
   // this is a dummy marker to let us know to show the image selector
-  const tooltipSpec = suggestTooltip.spec({ markName: 'tooltip-marker', trigger: 'image' });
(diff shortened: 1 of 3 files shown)


components/common/CharmEditor/components/ResizableImage.tsx
```diff
@@ -239,7 +239,7 @@ function getFileBinary(src: string): File | null {
 
 export function spec() {
   // this is a dummy marker to let us know to show the image selector
-  const tooltipSpec = suggestTooltip.spec({ markName: 'tooltip-marker', trigger: 'image' });
+  const tooltipSpec = suggestTooltip.spec({ markName: 'tooltip-marker', trigger: 'image', excludes: '_' });
```
msadura, line 242: I would appreciate some explanation, why this is needed and how it worked previously without that?

msadura:
I am not sure if that is related, but inline palette covers text you type:
Screenshot 2023-05-11 at 09 35 33

Devorein:
Its not related, I've checkout to a commit before we added support for emojis and mentions and its still the same issue there.

PR #2140, opened 2023-05-11
Handle system errors in sqs processing

msadura:
### <samp>🤖 Generated by Copilot at 6425610</samp>

Improved error handling for Discord integration. Added a test case for `createAndAssignRolesDiscord` and handled `SystemError` in `processWebhookMessage`.

### WHY
Handle system errors (User Not found, Space not found etc.) in sqs processing and do not retry such messages.


mattcasey (changes requested):
i think this is swinging the pendulum too far backward - where we will just swallow messages that need to be processed, or at least notify us. SystemError is really just a 'clean way to throw errors' - it doesn't necessarily mean it's not an issue we shouldn't fix. I think it would be safer to add specific errors.

- should look into why the DLQ is not sufficient. Last message was from April 23rd.
- we should log a warning at least

mattcasey (approved):

PR #2142, opened 2023-05-11
Remove null values before sending to frontend

mattcasey:
### <samp>🤖 Generated by Copilot at 1a88dd3</samp>

Refactored `getAccessiblePages` to use a helper function `getPageMeta` for cleaner code. Updated `lib/pages/server/getAccessiblePages.ts`.

### WHY
This shaves off almost a whole MB from the pages endpoint (before gzip). It was not as drastic when I tried blocks (only .1MB), since the only null value is deletedAt.


lib/pages/server/getAccessiblePages.ts
```diff
@@ -235,10 +232,17 @@ export async function getAccessiblePages(input: PagesRequest): Promise<PageMeta[
...
-    });
+    .map(getPageMeta);
 
   return filteredPages;
 }
+
+function getPageMeta(page: IPageWithPermissions): PageMeta {
+  delete (page as any).permissions;
+  // eslint-disable-next-line guard-for-in
+  for (const propName in page) {
```
motechFR, line 243: This would be a nice little utility for improving all our API responses

motechFR (approved):

PR #2144, opened 2023-05-11
dont refresh page meta when permissions change

mattcasey:
### WHAT
copilot:summary

### WHY
I think this may no longer be necessary since we dont get permissiosn from usePages?


components/common/PageLayout/components/Header/components/ShareButton/ShareButton.tsx
```diff
@@ -17,18 +15,12 @@ import { ProposalPagePermissions } from './components/PagePermissions/ProposalPa
...
 
 function ShareButton({ headerHeight, pageId }: { headerHeight: number; pageId: string }) {
-  const { refreshPage, pages } = usePages();
+  const { pages } = usePages();
   const popupState = usePopupState({ variant: 'popover', popupId: 'share-menu' });
   const { pagePermissions, refreshPermissions } = usePagePermissionsList({
     pageId
   });
   const isLargeScreen = useMediaQuery((theme: Theme) => theme.breakpoints.up('md'));
-  // watch changes to the page in case permissions get updated
```
motechFR, line 26: All good makes sense, since we're getting permissions from page permissions list hook

PR #2145, opened 2023-05-11
remove more fields from page meta

mattcasey:
### <samp>🤖 Generated by Copilot at 6fd71de</samp>

Simplified the `PageMeta` type and the `selectPageFields` function to remove unused or redundant properties. This improves the data model and the performance of the `usePages` hook.

### WHY
reduces pages payload from 4.33MB to 3.58MB!


motechFR (approved):
Liking this cleanup wave!

PR #2148, opened 2023-05-12
fix searching members by custom attribute

mattcasey:
### WHAT
copilot:summary

### WHY
we are using the uuid of properties, not the actual name of the option.

https://app.charmverse.io/charmverse/page-49366552253645923?cardId=17e1202a-3191-4a87-bd8c-a9d2989d52a6


valentinludu (approved):
Tested and it works

PR #2156, opened 2023-05-13
Bug/discussion tasks

motechFR:
### <samp>🤖 Generated by Copilot at 1b17d47</samp>

This pull request improves the test data generation and the discussion task fetching features. It refactors the `seedTestPages` function to accept more input options and create more realistic page diffs. It also adds a `take` parameter to the `getPageMentions` function to limit the database query size and avoid potential errors. These changes affect the files `scripts/seedTestPages.ts` and `lib/discussion/getDiscussionTasks.ts`.

### WHY
Mention tasks fails for users with alot of pages

### DIFF
--- a/lib/discussion/getDiscussionTasks.ts
+++ b/lib/discussion/getDiscussionTasks.ts
@@ -1,5 +1,5 @@
 import { prisma } from '@charmverse/core';
-import type { Space, User } from '@charmverse/core/prisma';
+import type { Page, Space, User } from '@charmverse/core/prisma';
 
 import { prismaToBlock } from 'lib/focalboard/block';
 import type { Board } from 'lib/focalboard/board';
@@ -549,6 +549,8 @@ async function getBoardPersonPropertyMentions({
   };
 }
 
+type PageToExtractMentionsFrom = Pick<Page, 'bountyId' | 'content' | 'id' | 'path' | 'title' | 'createdBy' | 'spaceId'>;
+
 async function getPageMentions({
   userId,
   username,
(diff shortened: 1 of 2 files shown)


lib/discussion/getDiscussionTasks.ts
```diff
@@ -563,6 +563,8 @@ async function getPageMentions({
       },
       deletedAt: null
     },
+    // This query will return a huge amount of content is user is part of alot of spaces, we need to split it up
+    take: 1000,
```
motechFR, line 567: @mattcasey I looked into getDiscussionTasks. The issue is not coming from Permissions API, but from this method.

If you remove the take parameter, you can reproduce the 500 error 
(Need to use the seedTestPages first to get the error to pop up)

Screenshot 2023-05-13 at 18 15 24

https://github.com/prisma/prisma/issues/13864

motechFR:
@mattcasey I looked into getDiscussionTasks. The issue is not coming from Permissions API, but from this method.

If you remove the take parameter, you can reproduce the 500 error
(Need to use the seedTestPages first to get enough pages in spaces user belongs to for the error to pop up)

Screenshot 2023-05-13 at 18 15 24

Related prisma issue
https://github.com/prisma/prisma/issues/13864

lib/discussion/getDiscussionTasks.ts
```diff
@@ -602,6 +586,30 @@ async function getPageMentions({
...
+        path: true,
+        title: true,
+        createdBy: true,
+        spaceId: true
+      }
+    });
+    for (let page of pages) {
+      extractMentionsFromPage(page);
+      // Make page eligible for garbage collection
+      page = null as any;
```
mattcasey, line 609: is this actually doing anything? we are not capturing the result inside of a closure. why not just set pages to null afterwards instead?

mattcasey (approved):

lib/discussion/getDiscussionTasks.ts
motechFR: Yes you’re right, I also did a second check, and this would only set the local ref to null.

we’d have to do pages[i] = null instead 

will update now

PR #2158, opened 2023-05-15
add grant recipient template

mattcasey:
### <samp>🤖 Generated by Copilot at e06736e</samp>

The pull request updates the codebase to use a new model for space templates, which are predefined layouts and settings for creating spaces. It replaces the old `SpaceCreateTemplate` object with a new `SpaceTemplateType` enum and a static array of templates in `lib/spaces/config.ts`. It also updates the components, public API, metrics, and scripts that depend on space templates to use the new model. This improves the code quality, consistency, and flexibility of space creation.

### WHY
I also cleaned up the types -- we should *never* use the "api" name version of templates internally. These are just for the public api. I noticed we had them mixed in with the original template types "templateHackathon" vs "hackathon". This was causing a lot of complexity that we don't need, mapping between them in the UI, etc.


msadura (approved):
nice cleanup

PR #2159, opened 2023-05-15
fix page path

mattcasey:
### <samp>🤖 Generated by Copilot at 9e095b6</samp>

Refactored page path generation in `lib/pages/untitledPage.ts` to use a custom function instead of `uuid`. This enhances the clarity and uniformity of the page paths.

### WHY
We have fixed a downstream issue where we assume page paths that are UID must be an id, but we should always use the some algo for page paths. This is the only place I found (it's a default page that gets created when u delete all pages in ur space)


lib/pages/server/createPage.ts
```diff
@@ -8,23 +8,25 @@ import type { PageContent } from 'lib/prosemirror/interfaces';
...
-            })
-          }
-        : undefined
+      hasContent,
+      galleryImage: extractPreviewImage(data.content as PageContent)
     }
   };
 
+  if (hasContent) {
+    createArgs.data.diffs = {
```
mattcasey, line 21: Not sure why this was happening but i was getting an error from Prisma for mixing up the input types - eg. `spaceId` isntead of `{ space: { id: spaceId } }`. Adding populated data in this method is probably dangerous..

valentinludu (approved):

PR #2161, opened 2023-05-16
adding in permissions api to run alongside webapp

yiwork:
### <samp>🤖 Generated by Copilot at 9fad1e1</samp>

Added a new `permissions-api` service to `docker-compose.yml` for staging the web application. The service runs a node application from a private AWS ECR image and uses the host `.env` file.

### WHY
adding permissions api as sidecar to the staging webapp environment

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,7 @@ services:
       ]
     environment:
       - NEXT_PUBLIC_APP_ENV=staging
+      - PERMISSIONS_API_URL="http://permissions-api"
     profiles:
       - stg-webapp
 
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -56,6 +56,7 @@ services:
       ]
     environment:
       - NEXT_PUBLIC_APP_ENV=staging
+      - PERMISSIONS_API_URL="http://permissions-api:3001"
```
motechFR, line 59: we just need this to be http://127.0.0.1:3001 and we should be good!

yiwork: 127.0.0.1 refers to the webapp container itself. The ultimate resolution is to remove the double quotes. When you have time, I can jump on a meet to demonstrate why.

PR #2164, opened 2023-05-16
Subdomains with middleware POC

msadura:
### <samp>🤖 Generated by Copilot at 0e11d9a</samp>

Added custom middleware to handle subdomain and domain rewriting and redirection for spaces. Refactored and simplified the logic for getting and storing the space domain in the `RouteGuard` component. Added utility functions to extract and validate subdomains and domains from URLs. Fixed a minor issue with the default workspace URL in the `pages/index.tsx` file.

### WHY
Subdomains POC


lib/spaces/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const DOMAIN_BLACKLIST = [
```
msadura, line 1: I had to move this const to separate file, because  `config.ts` was dependent on `lodash` import and for some reason it failed to build middleware

components/login/LoginPage.tsx
```diff
@@ -0,0 +1,110 @@
+import { log } from '@charmverse/core/log';
```
msadura, line 1: This is just login page moved to external component

PR #2168, opened 2023-05-17
Feat/subdomain cookies

msadura:
### <samp>🤖 Generated by Copilot at 499e1d3</samp>

Refactored the session handling code to use a dynamic function for iron session options instead of a static object. This allows the session configuration to adapt to different cookie names, subdomains, and environments. Updated the integration test and the websocket authentication module to use the new function. Added utility functions for subdomain validation and localhost alias detection.

### WHY
We will need cross domain cookies for subdomain support. For now domain for each env is hardcoded

### DIFF
--- a/lib/session/getIronOptions.ts
+++ b/lib/session/getIronOptions.ts
@@ -1,0 +1,43 @@
+import type { IncomingMessage } from 'http';
+
+import type { IronSessionOptions } from 'iron-session';
+
+import { baseUrl, cookieName, isDevEnv, isProdEnv, isStagingEnv } from 'config/constants';
+import { authSecret } from 'lib/session/config';
+import { isLocalhostAlias } from 'lib/utilities/getValidSubdomain';
+
+// TODO - domains based on request / env config
+const cookieDomains = {
+  dev: 'local.io',
+  staging: 'charmverse.co',
+  prod: 'charmverse.io'
+};
+
+export function getIronOptions(req?: IncomingMessage): IronSessionOptions {
+  const ironOptions: IronSessionOptions = {
+    cookieName,
+    password: authSecret,
+    cookieOptions: {
+      sameSite: 'strict' as const,
+      // secure: true should be used in production (HTTPS) but can't be used in development (HTTP)
+      secure: typeof baseUrl === 'string' && baseUrl.includes('https'),
+      // domain: cookieDomain TODO: change domain to subdomain without logging people out, so we can use them across subdomains
+      domain: isLocalhostAlias(req?.headers.host) ? undefined : getCookieDomain()
+    }
+  };
+  return ironOptions;
+}
+
+function getCookieDomain() {
+  if (isDevEnv) {
+    return cookieDomains.dev;
+  }
+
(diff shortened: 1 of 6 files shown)


lib/utilities/getValidSubdomain.ts
```diff
@@ -0,0 +1,26 @@
+export function getValidSubdomain(host?: string | null) {
```
msadura, line 1: this helper is not used right now, it is just picked from subdomains code

lib/session/getIronOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+export function getIronOptions(req?: IncomingMessage): IronSessionOptions {
+  const ironOptions: IronSessionOptions = {
+    cookieName,
+    password: authSecret,
+    cookieOptions: {
+      sameSite: 'strict' as const,
+      // secure: true should be used in production (HTTPS) but can't be used in development (HTTP)
+      secure: typeof baseUrl === 'string' && baseUrl.includes('https'),
+      // domain: cookieDomain TODO: change domain to subdomain without logging people out, so we can use them across subdomains
+      domain: isLocalhostAlias(req?.headers.host) ? undefined : getCookieDomain()
```
msadura, line 25: we do not want to set cookie domain in dev mode when using 'localhost' or just 127.0.0.1

lib/session/getIronOptions.ts
```diff
@@ -0,0 +1,41 @@
+import type { IncomingMessage } from 'http';
+
+import type { IronSessionOptions } from 'iron-session';
+
+import { baseUrl, cookieName, isDevEnv, isProdEnv, isStagingEnv } from 'config/constants';
+import { authSecret } from 'lib/session/config';
+import { isLocalhostAlias } from 'lib/utilities/getValidSubdomain';
+
+const cookieDomains = {
```
mattcasey, line 9: i think this could still go into the 'config' file?

lib/session/getIronOptions.ts
```diff
@@ -0,0 +1,41 @@
+import type { IncomingMessage } from 'http';
+
+import type { IronSessionOptions } from 'iron-session';
+
+import { baseUrl, cookieName, isDevEnv, isProdEnv, isStagingEnv } from 'config/constants';
+import { authSecret } from 'lib/session/config';
+import { isLocalhostAlias } from 'lib/utilities/getValidSubdomain';
```
mattcasey, line 7: i know it's a WIP but it feels weird that you have a file named specifically after one function but then export a second function. you could just call it utilities/server.ts

mattcasey (approved):
Works! I even moved the wildcard Route 53 domain to your staging env and was able to access the space without having to log in again :) (I am switching it back so you can continue testing custom subdomains)

lib/session/getIronOptions.ts
msadura: I will merge this one as it is and try to find a better place for it in follow up pr for subdomains. The problem is that it can be used on both server and browser so I would not like to put it into server.ts

PR #2172, opened 2023-05-19
fix page link icon for forums

mattcasey:
### <samp>🤖 Generated by Copilot at 2458a5b</samp>

Refactored and renamed some components and types related to the charm editor and the page layout, to improve the reusability, clarity, and consistency of the code. Extracted a `PageList` component from the `NestedPage` component, and added a `PageIcon` component to render icons based on the page type. Removed an unused prop from the `PageIconProps` type.

### WHY
I rewrote the code. to ensure we only ever return one icon
[](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=50d940bb-fdbd-40ba-9816-616d6138a663&cardId=628e407a-c334-4b87-b0eb-3ea34a45a761)

### DIFF
--- a/components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
+++ b/components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
@@ -4,10 +4,12 @@ import { Typography } from '@mui/material';
 
 import Link from 'components/common/Link';
 import { NoAccessPageIcon, PageIcon } from 'components/common/PageLayout/components/PageIcon';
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
```diff
@@ -72,21 +74,43 @@ export default function NestedPage({ node, currentPageId }: NodeViewProps & { cu
       data-type={node.attrs.type}
     >
       <div>
-        {nestedStaticPage ? (
```
mattcasey, line 75: when we see this many ternary conditions, i think it's best to rewrite the logic with if/else

PR #2174, opened 2023-05-19
always convert values to string before filtering

mattcasey:
### <samp>🤖 Generated by Copilot at 8c63e03</samp>

Fixed a bug in `cardFilter.ts` that caused the app to crash when filtering by a number property. Added `toString()` to the `filterValue` variable to handle non-string values.

### WHY
It looks like when you use "number" input type in our UI, numbers get saved as strings for some reason. But there is a table in production whose cards are all numbers. They look like they may have been imported from somewhere.

### DIFF
--- a/lib/focalboard/cardFilter.ts
+++ b/lib/focalboard/cardFilter.ts
@@ -1,3 +1,5 @@
+import { log } from '@charmverse/core/log';
+
 import { Utils } from 'components/common/BoardEditor/focalboard/src/utils';
 import type { IPropertyTemplate } from 'lib/focalboard/board';
 import type { Card } from 'lib/focalboard/card';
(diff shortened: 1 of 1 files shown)


lib/focalboard/cardFilter.ts
```diff
@@ -53,13 +55,14 @@ class CardFilter {
   static isClauseMet(filter: FilterClause, templates: readonly IPropertyTemplate[], card: Card): boolean {
     const value = card.fields.properties[filter.propertyId] ?? [];
     const filterProperty = templates.find((o) => o.id === filter.propertyId);
-    const filterValue = filter.values[0]?.toLowerCase() ?? '';
+    const filterValue = filter.values[0]?.toString()?.toLowerCase() ?? '';
```
mattcasey, line 58: this is the fix.. making sure the values are always strings

PR #2182, opened 2023-05-22
Fixed list item delete issues

Devorein:
### <samp>🤖 Generated by Copilot at 3d23cfa</samp>

Fix backspace bug and improve nested list support in `CharmEditor`. Refactor `commands.ts` to simplify list item logic.


mattcasey:
In my first test, hitting "delete" deletes the entire row, which is not what we want. Also my cursor somehow dropped to 2-3 lines below: https://www.loom.com/share/17236b2420b748bab60bdc3cc37c1d39?from_recorder=1&focus_title=1

PR #2183, opened 2023-05-22
Fix user created another user account when logged in through google after having a verified email

valentinludu:
### <samp>🤖 Generated by Copilot at 863aa15</samp>

This pull request adds support for logging in with Google when the user has an existing verified email in the database. It modifies the `loginWithGoogle` function in `lib/google/loginWithGoogle.ts` and adds a test case in `lib/google/__tests__/loginWithGoogle.spec.ts`.

### WHY
An user lost access to their account by doing:
Create user and space with just their email (verifiedEmail) -> Logout -> Login with Google. 
The app created another user with no space.

[Task item](https://app.charmverse.io/charmverse/page-9248424773658832)

### DIFF
--- a/lib/google/loginWithGoogle.ts
+++ b/lib/google/loginWithGoogle.ts
@@ -1,3 +1,4 @@
+import { log } from '@charmverse/core/log';
 import { prisma } from '@charmverse/core/prisma-client';
 
 import type { SignupAnalytics } from 'lib/metrics/mixpanel/interfaces/UserEvent';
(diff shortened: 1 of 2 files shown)


lib/google/loginWithGoogle.ts
```diff
@@ -69,6 +82,30 @@ export async function loginWithGoogle({
       }
     }
 
+    if (emailAccount) {
+      if (emailAccount.user.deletedAt) {
+        throw new DisabledAccountError(`This account has been disabled`);
```
mattcasey, line 87: i dont think this should be an error....? could we just ignore the user

mattcasey (changes requested):
i don't think a previously deleted user should start failing this case.. maybe just log a warning?

lib/google/loginWithGoogle.ts
valentinludu: OK. I can ignore that. 

I was just taking the example from above (google account)

valentinludu: Pushed update!

mattcasey (approved):

PR #2185, opened 2023-05-22
Remove bg for inline dbs with apikeys

valentinludu:
### <samp>🤖 Generated by Copilot at 645be19</samp>

Added a `noBg` prop to the `Banner` and `PageWebhookBanner` components to control their background and border styles. Used this prop in the `CenterPanel` component to adjust the banner appearance based on the `isEmbedded` prop.

### WHY
Leftover for Typeform implementation:
After feedback from demo day we decided to remove the background and add a simple border for databases that have a banner.


mattcasey (approved):

mattcasey (changes requested):
please avoid shortening names especially on properties/interfaces of things. 'noBackground' is not much longer and takes less guess-work when reading

valentinludu:
> please avoid shortening names especially on properties/interfaces of things. 'noBackground' is not much longer and takes less guess-work when reading

I pushed a change where I simplify even more the logic and use just 'sx' prop.

PR #2187, opened 2023-05-22
Fix member search

mattcasey:
### <samp>🤖 Generated by Copilot at 3e2f03b</samp>

Enhance the member search input component and the space members page to support username search. Refactor and clean up some code related to the member search functionality.

### DIFF
--- a/components/common/form/InputSearchMember.tsx
+++ b/components/common/form/InputSearchMember.tsx
@@ -87,6 +87,7 @@ export function InputSearchMemberBase({
         <UserDisplay
           {...(_props as any)}
           user={user}
+          avatarSize='small'
           avatarIcon={user.id === 'email' ? <EmailIcon fontSize='large' /> : undefined}
         />
       )}
@@ -105,7 +106,6 @@ export function InputSearchMemberBase({
           inputRef={inputRef}
         />
       )}
-      filterOptions={filterOptions}
       {...props}
     />
   );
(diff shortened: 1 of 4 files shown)


components/common/form/InputSearchMember.tsx
```diff
@@ -87,6 +87,7 @@ export function InputSearchMemberBase({
         <UserDisplay
           {...(_props as any)}
           user={user}
+          avatarSize='small'
```
mattcasey, line 90: more room for names in the dropdown and mirrors what we do for proposal reviewers

components/common/form/InputSearchMember.tsx
```diff
@@ -105,7 +106,6 @@ export function InputSearchMemberBase({
           inputRef={inputRef}
         />
       )}
-      filterOptions={filterOptions}
```
mattcasey, line 108: this was causing the bug. i'm not sure why, filterOptions always had something in the list but also not sure why we had this set in the first place. I think it's for when u do filtering on your own

mattcasey: I tested the Share dropdown, mentions in the editor, and author for proposals and filter seems to be working better

lib/members/getSpaceMembers.ts
```diff
@@ -92,7 +92,7 @@ export async function getSpaceMembers({
           joinDate: spaceRole.createdAt.toISOString(),
           hasNftAvatar: hasNftAvatar(spaceRole.user),
           properties: getPropertiesWithValues(visibleProperties, memberPropertyValues),
-          searchValue: getMemberSearchValue(spaceRole.user, visiblePropertiesMap),
+          searchValue: getMemberSearchValue(spaceRole.user, visiblePropertiesMap, username),
```
mattcasey, line 95: we were not including the username for the space since it's not a 'visible property'  and not part of user object

Devorein (approved):

PR #2189, opened 2023-05-22
Fix/properly remove session cookie

msadura:
### <samp>🤖 Generated by Copilot at e80f405</samp>

This pull request improves the user login and profile integration with Discord by updating the `getUser` and `logout` functions and removing the old non cross-domain cookie. It also simplifies the code and the app performance by refactoring some logic and removing unused imports.

### WHY
Properly remove old session cookie + add it to /profile endpoint


pages/api/session/removeOldCookie.ts
```diff
@@ -0,0 +1,19 @@
...
+    return;
+  }
+
+  const cookiesHeader = res.getHeader('Set-Cookie') || [];
+  const setCookiesArray = Array.isArray(cookiesHeader) ? cookiesHeader : [String(cookiesHeader)];
+
+  res.setHeader('Set-Cookie', [
+    ...setCookiesArray,
+    // remove old non cross-domain cookie
+    `${cookieName}=; Max-Age=0; Path=/; HttpOnly;`
```
msadura, line 17: the trick was to not set `Domain=app.charmverse.io` here, because initially cookie was set without domain flag

mattcasey: all files in pages/ should be kebab-case. it doesn't look like we actually call this endpoint though?

pages/index.tsx
```diff
@@ -82,8 +82,6 @@ export default function LoginPage() {
         redirectToDefaultPage();
       } else {
         setShowLogin(true);
-        // make sure to remove user cookies
-        logoutUser();
```
mattcasey, line 86: i'm glad we are removing this.. seems like it might trigger issues when someone is going thru oauth redirects, etc. if we are not careful

pages/api/session/removeOldCookie.ts
msadura: lol thx, good catch, it should land in /lib/session

mattcasey: i suggest just leave this method in logout.ts

mattcasey: and you don't need `SameSite=Strict;${secureCookie}`?

msadura: nope, the only things that are needed to tell that cookie is "the same" is  name, domain and httpOnly

mattcasey (approved):

pages/api/session/removeOldCookie.ts
mattcasey: ok thanks for cleaning it up

PR #2190, opened 2023-05-22
Trigger logout from web3AuthSig only if user can only connect via wallet

motechFR:
### <samp>🤖 Generated by Copilot at c6448d7</samp>

The pull request enhances the web3 authentication feature by adding a function to `hooks/useWeb3AuthSig.tsx` that counts the user's connectable identities and by fixing the logout logic to prevent session loss.

### WHY
More conservative approach to auto-logout


mattcasey (changes requested):
doesn't result in a consistent UX for switching wallets. i think we can do better, maybe keep track of how you logged in originally

PR #2191, opened 2023-05-22
Simplify rules to switch web3 accounts

mattcasey:
### WHAT
The change is that we will only respond to changes in the connected address if you logged in thru Metamask.

### WHY

We've had a couple cases in the past including this one where we inadvertently log a user out at the wrong time. We can  simplify the logic by reducing the variables in play when we allow users to switch wallets.  

- assumes that the existence of wallet signature means a user logged in via wallet (we remove it when you log out). 

Flows that may no longer work:
- Log in thru Oauth, then connect thru Metamask expecting to be logged out and go to the login page

### DIFF
--- a/components/_app/hooks/useSpaceGatesReevaluate.ts
+++ b/components/_app/hooks/useSpaceGatesReevaluate.ts
@@ -13,7 +13,7 @@ export function useSpaceGatesReevaluate() {
   useEffect(() => {
     const reevaluateRoles = async () => {
       if (user?.id && currentSpace?.id && account) {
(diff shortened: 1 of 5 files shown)


motechFR (approved):
This worked well

Flow 1: Google login
_Account has just google connected_
Logged in via Google, switched wallets, nothing happened

Flow 2: Wallet login
_Account has just wallet connected_
Logged in via Wallet, switched wallets, got logged out

Flow 3: Wallet login
_Account has google + wallet connected_
Logged in via google, switched wallets, got logged out

PR #2200, opened 2023-05-24
Fix flaky e2e tests

mattcasey:
## signup test
I noticed that sometimes, when opening the member profile form or running the test locally, my console would show a  "maximum depth exceeded" error inside this component. With logging, I Could see that the useEffect() that watches `defaultValues` was being called over and over. My hypothesis is that the default array `[]` supplied for `properties` was causing the issue.

## token gates test
This must have broken in my update to simplify token gates, we were not picking up the signature from local storage because I was mixing up the casing of wallet address :/

### DIFF
--- a/pages/join.tsx
+++ b/pages/join.tsx
@@ -2,7 +2,7 @@ import NavigateNextIcon from '@mui/icons-material/ArrowRightAlt';
 import { Alert, Box, Card, Divider } from '@mui/material';
 import { useRouter } from 'next/router';
 import type { ReactNode } from 'react';
-import { useEffect } from 'react';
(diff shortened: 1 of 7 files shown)


pages/join.tsx
```diff
@@ -48,6 +49,13 @@ export default function JoinWorkspace() {
     }
   }, [spaces]);
 
+  useEffect(() => {
+    // isReady should only be used conditionally inside a useEffect()
+    if (router.isReady) {
+      setRouterReady(true);
```
msadura, line 55: what does using `isRouterReady` state instead of `router.ready` gives? Did it need like extra render cycle to work?

PR #2201, opened 2023-05-24
Remove experimental "esmExternals" from next.config

mattcasey:
### <samp>🤖 Generated by Copilot at ec2fb58</samp>

This pull request updates the @bangle.dev/pm and @bangle.dev/markdown packages to use the latest prosemirror-utils dependency and fix an import bug. This improves the performance and stability of the app.charmverse.io editor.

### WHY
This setting was supposedly added by default in next 12, so we shouldn't need it: https://nextjs.org/blog/next-12#es-modules-support-and-url-imports.
However, I found I needed to do more patching of @bangle.dev to get it to work. . I'm hoping that less overrides of next.js config might help us improve performance, or utilize other things like turbo sooner.


mattcasey:
Closing.. I learned that we are probably compiling both ESM and CommonJS versions of various libraries, particularly prosemirror. And this is due in part because of libs like "prosemirror-utils" which only exports common-js. An alternative is to use prosemiror-utils-bangle, but it only exports ESM.

PR #2208, opened 2023-05-24
Pass specific branches to build & install steps

motechFR:
### <samp>🤖 Generated by Copilot at 5369f6c</samp>

This pull request improves the GitHub actions and workflow for testing and deploying features that involve changes in the `core` and `permissions_api` repositories. It adds input parameters and environment variables to allow specifying custom branch names for these dependencies, and adjusts the cache keys and installation steps accordingly. It also adds a `custom-branch` label to trigger the workflow with custom branches.

### WHY
Enforcing specific branches for our core repo and permissions API

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -52,6 +52,10 @@ inputs:
   NEXT_PUBLIC_STRIPE_PUBLIC_KEY:
     description: "..."
     required: false
+  CORE_BRANCH:
+    description: "..."
+    required: false
+    default: "main"
 
 runs:
   using: "composite"
@@ -64,6 +68,16 @@ runs:
 
     - run: echo '${{ env.GIT_DIFF }}'
       shell: bash
+
+    - name: Set cache key name
+      shell: bash
+      env:
+        CORE_BRANCH: ${{ inputs.CORE_BRANCH }}
+      run:  |
+        echo -e "Received for build: $CORE_BRANCH"
+        echo "CACHE_KEY=$(echo $CORE_BRANCH | sed 's/\///g')" >> $GITHUB_ENV
+      
+
     - name: Cache build
       id: cache_build
       uses: actions/cache@v3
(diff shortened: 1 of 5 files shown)


.github/actions/build/action.yml
```diff
@@ -52,6 +52,10 @@ inputs:
   NEXT_PUBLIC_STRIPE_PUBLIC_KEY:
     description: "..."
     required: false
+  CORE_BRANCH:
+    description: "..."
+    required: false
+    default: "main"
```
mattcasey, line 58: these can just be undefined

mattcasey, line 58: i think these should just be undefined by default -- no one should ever have to do "main", leaving it empty should just mean you get what's in package.json/package-lock

.github/actions/build_permissions_api/action.yml
```diff
@@ -28,14 +44,21 @@ runs:
       uses: actions/cache@v3
       with:
         path: ./${{inputs.API_FOLDER}}/node_modules
-        key: nodemodules-permissions-cache-${{ hashFiles(format('{0}/package-lock.json', inputs.API_FOLDER)) }}
+        key: nodemodules-permissions-cache-${{ hashFiles(format('{0}/package-lock.json', inputs.API_FOLDER)) }}-${{ env.CORE_CACHE_KEY}}-${{ env.API_CACHE_KEY }}
```
mattcasey, line 47: the cache key being added is really just the names of branches, right? I think it is necessary so that main doesn't accidentally pull the cache in from a branch that installed a different version of core. But trying to cache the custom installs in-between builds on the same branch may be overkill.. it also doesn't consider the commit sha right now, so i think if i kept changing my branch of core, it would not notice

.github/actions/install/action.yml
```diff
@@ -18,9 +29,15 @@ runs:
...
+
 
     - name: Install npm dependencies
       shell: bash
       if: steps.cache_node_modules.outputs.cache-hit != 'true'
       run: npm ci --no-audit --no-fund
+
+    - name: Install custom core branch
+      shell: bash
+      if: inputs.CORE_BRANCH != 'main'
```
mattcasey, line 42: devs should set the branch that they want to trigger, instead of this triggering for all branches or relying on a tag. Downsides of tag: it's configuration in two separate places. Also you might change the tag before changing yml and it will try to install whatever branch was left in code last.. so long as we need to change the yml i think it's best to just have a single place to configure

PR #2214, opened 2023-05-26
disable long-polling in sockets client + other cleanup

mattcasey:
Long-polling is not a great experience for typing, and socket.io always starts with it by default. This might help reduce latency that people experience.

AFAICT, long-polling is just needed for really old versions of IE.

### DIFF
--- a/lib/focalboard/cardFilter.ts
+++ b/lib/focalboard/cardFilter.ts
@@ -95,7 +95,7 @@ class CardFilter {
             return sourceValue !== '';
           }
           default: {
-            Utils.assertFailure(`Invalid filter condition ${filter.condition}`);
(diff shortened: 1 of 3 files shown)


lib/focalboard/cardFilter.ts
```diff
@@ -95,7 +95,7 @@ class CardFilter {
             return sourceValue !== '';
           }
           default: {
-            Utils.assertFailure(`Invalid filter condition ${filter.condition}`);
+            Utils.assertFailure(`Invalid filter condition: ${filter.condition} for type ${filterPropertyDataType}`);
```
mattcasey, line 98: making these logs more useful

PR #2222, opened 2023-05-29
Ux/no permission to comment

motechFR:
### <samp>🤖 Generated by Copilot at f2f01d5</samp>

This pull request improves the comment feature of the forum by adding permissions checks and feedback messages to the `CommentForm` and its related components. It also fixes a typo in the `CommentVote` component and refactors the `PostPage` component to pass props to the `PostCommentForm` component.

### WHY
Add UX to specify user cannot comment on current post.

Also added tooltip on "Reply" button

Screenshot 2023-05-29 at 14 17 14


mattcasey (changes requested):
Can you use `placeholderText` on the charmeditor instead? It should use the slightly faded out style instead of looking like text was typed in the input, and then you don't need to create a custom prosemirror node.

mattcasey (approved):

PR #2228, opened 2023-05-30
Show paid announcement banner

Devorein:
### WHAT
copilot:summary

### DIFF
--- a/components/_app/GlobalComponents.tsx
+++ b/components/_app/GlobalComponents.tsx
@@ -1,3 +1,4 @@
+import { PaidAnnouncementBanner } from 'components/common/Banners/PaidAnnouncementBanner';
 import HexagonalAvatarMask from 'components/common/HexagonalAvatarMask';
 import Snackbar from 'components/common/Snackbar';
 import { UserProfileDialogGlobal } from 'components/common/UserProfile/UserProfileDialogGlobal';
--- a/components/common/Banners/PaidAnnouncementBanner.tsx
+++ b/components/common/Banners/PaidAnnouncementBanner.tsx
@@ -1,0 +1,29 @@
+import EastIcon from '@mui/icons-material/East';
+import { Typography, Stack } from '@mui/material';
+import { useEffect, useState } from 'react';
+
+import Link from 'components/common/Link';
+
+import { StyledBanner } from './Banner';
+
+export function PaidAnnouncementBanner() {
+  const [component, setComponent] = useState<JSX.Element | null>(null);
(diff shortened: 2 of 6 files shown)


components/_app/GlobalComponents.tsx
```diff
@@ -1,3 +1,4 @@
+import { PaidAnnouncementBanner } from 'components/common/Banners/PaidAnnouncementBanner';
```
mattcasey, line 1: this doesnt actually get used?

components/common/Banners/PaidAnnouncementBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function PaidAnnouncementBanner() {
+  const [showPaidAnnouncementBar, setShowPaidAnnouncementBar] = useLocalStorage('show-paid-banner', true);
+  const [component, setComponent] = useState<JSX.Element | null>(null);
+  // Using use-effect to render the component only client side, otherwise next.js throws rehydration mismatch error
+  useEffect(() => {
+    setComponent(
+      showPaidAnnouncementBar ? (
+        <StyledBanner top={50} data-test='paid-announcement-banner'>
+          <Typography>
+            Community Edition coming on June 30th. Include complex role-based access control, API, & custom domain.{' '}
```
mattcasey, line 20: "Include" -> "Includes"

components/common/Banners/PaidAnnouncementBanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { useEffect, useState } from 'react';
+
+import Link from 'components/common/Link';
+import { useLocalStorage } from 'hooks/useLocalStorage';
+
+import { StyledBanner } from './Banner';
+
+export function PaidAnnouncementBanner() {
+  const [showPaidAnnouncementBar, setShowPaidAnnouncementBar] = useLocalStorage('show-paid-banner', true);
+  const [component, setComponent] = useState<JSX.Element | null>(null);
```
mattcasey, line 13: ths is a weird pattern, can you just change it to useState(true/false) and then return the component or null otherwise

mattcasey: actually, are you sure we even need two states?

components/_app/GlobalComponents.tsx
Devorein: I wanted to put i in GlobalComponents first but then saw it overlaps the sidebar

PR #2240, opened 2023-05-31
Bug/wrap columns

motechFR:
### <samp>🤖 Generated by Copilot at 94c50ff</samp>

This pull request enhances the board editor and the user display components in the `app.charmverse.io` repository. It adds features to wrap the labels of date range and user properties to improve the layout and readability. It also improves the user interface and functionality of the user property input. It adds a debugging component and a prop for date range wrapping to the `PropertyValueElement` component.

### WHY
Fix wrapping for User and Date columns

Wrapping implemented for user and date columns.

_Notes:
-- Implemented forced wrapping when selecting values, as otherwise values were off screen
-- Made values deleteable by clicking anywhere on that value (once in input mode). This handles the case where the column might be too small to display the X
_ 
Screenshot 2023-06-01 at 00 14 48

When wrapped, the user column does not go beyond the column anymore

Screenshot 2023-06-01 at 00 14 48

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/properties/dateRange/dateRange.tsx
@@ -21,6 +21,7 @@ type Props = {
   value: string;
   showEmptyPlaceholder?: boolean;
   onChange: (value: string) => void;
+  wrapColumn?: boolean;
 };
 
 export type DateProperty = {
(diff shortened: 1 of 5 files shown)


Devorein:
1. There seems to be an extra scrollbar since the menu list goes beyond the screen.
image
2. Notion shows the person property in a single row (even if the wrap column is toggled on) if there are spaces horizontally.
image
3. Notion truncates the display if the "Wrap column" is toggled off, but we seem to be squeezing the individual items
image
image

motechFR:
@Devorein Merging for now, we can add these tweaks later, I've logged your feedback to a new task card

PR #2250, opened 2023-06-03
Setup merge script

motechFR:
### <samp>🤖 Generated by Copilot at 2260d7b</samp>

Added a script `scripts/mergeWallet.ts` to merge wallets from duplicate user accounts to original user accounts. The script can take different types of arguments to identify the accounts and updates the userWallet table accordingly. The script also has some safety features to prevent accidental data loss.

### WHY
Script to merge wallets to another account

### DIFF
--- a/scripts/mergeWallet.ts
+++ b/scripts/mergeWallet.ts
@@ -1,0 +1,125 @@
+import { stringUtils } from "@charmverse/core";
+import { Prisma, prisma } from "@charmverse/core/prisma-client";
+import { utils} from 'ethers'
+import fs from 'node:fs/promises'
+
+import * as readline from "node:readline";
+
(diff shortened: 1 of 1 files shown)


mattcasey (changes requested):
- when is this script needed vs having the user fix it themselves?
- this is not really a 'merge' you should at least rename it to something like "reassignUserWallet'. "Merge" implies some sort of data loss, i think, or updating of many records. all we're doing is reassigning the user.
- I appreciate the time you spend adding all the checks, and it's fine as-is, but i think it could be simpler - just the 5 lines to update the user id. We should expect that anyone running this script is going to read it first and make sure they know what it's going to do.

mattcasey (approved):

PR #2261, opened 2023-06-06
update logic to ignore proposal tasks based on space settings

mattcasey:
### WHAT
copilot:summary

### DIFF
--- a/lib/proposal/getProposalTasks.ts
+++ b/lib/proposal/getProposalTasks.ts
@@ -50,12 +50,6 @@ export async function getProposalTasks(userId: string): Promise<{
   marked: ProposalTask[];
   unmarked: ProposalTask[];
 }> {
-  const userNotifications = await prisma.userNotification.findMany({
-    where: {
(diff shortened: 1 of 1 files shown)


lib/proposal/getProposalTasks.ts
```diff
@@ -169,15 +171,28 @@ export async function getProposalTasks(userId: string): Promise<{
...
-          proposalsRecord.unmarked.push(proposalTask);
-        } else {
-          proposalsRecord.marked.push(proposalTask);
-        }
+        tasks.push(proposalTask);
       }
     }
   });
 
+  const userNotifications = await prisma.userNotification.findMany({
```
mattcasey, line 179: slight optimization: instead of querying all notifications a user ever received, just get the ones we are looking for

components/settings/space/SpaceSettings.tsx
mattcasey, line 1: this is the updated space settings form

PR #2264, opened 2023-06-07
Cleanup: remove permissions from default response when creating a page

mattcasey:
### <samp>🤖 Generated by Copilot at 8b90e7e</samp>

Refactored and improved some functions and types related to page creation, metrics, and proposals. Moved `createPage` to a separate module and added an `include` parameter for related data. Simplified type annotations and imports in `postToDiscord.ts`, `pages/api/pages/index.ts`, and `testing/setupDatabase.ts`. Used generic types in `testing/utils/proposals.ts` to specify the return type of `createPageDb`.

### WHY
less queries.. altho it doesn't seem to have made much difference in the test run times :/


lib/pages/server/createPage.ts
```diff
@@ -27,18 +28,5 @@ export function createPage<T>({ data, include }: Prisma.PageCreateArgs): PrismaP
     };
   }
 
-  const includeData =
```
mattcasey, line 30: this was the main change

PR #2268, opened 2023-06-07
Fixed member property value search

Devorein:
### <samp>🤖 Generated by Copilot at 3529fbb</samp>

Refactored the code for getting and displaying member properties in `lib/members`. Removed unnecessary parameters and variables and used database values directly.

### DIFF
--- a/lib/members/getMemberSearchValue.ts
+++ b/lib/members/getMemberSearchValue.ts
@@ -1,5 +1,4 @@
 import type { DiscordUser, MemberPropertyValue, TelegramUser, User, UserDetails } from '@charmverse/core/prisma';
-import { stringUtils } from '@charmverse/core/utilities';
 
 type UserData = User & {
   memberPropertyValues: MemberPropertyValue[];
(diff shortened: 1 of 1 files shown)


lib/members/getMemberSearchValue.ts
```diff
@@ -29,9 +25,9 @@ export function getMemberSearchValue(
   const propertyValuesString = memberPropertyValues
     .map((prop) => {
       if (Array.isArray(prop.value)) {
-        return prop.value.map((val) => memberProperties[val as string] || val).join(' ');
+        return prop.value.join(' ');
       }
-      return typeof prop.value === 'string' ? memberProperties[prop.value] : prop.value;
```
Devorein, line 34: memberProperties was a record of memberPropertyId and memberPropertyName ie

`{
	68a21e87-0369-48f3-9e90-19549e06ee24: 'Discord'
}`

`prop.value` is the actual value of the member property, so it can't actually be used as key.

mattcasey: i just fixed this recently as prop.value was in fact an object id. something weird is going on

PR #2289, opened 2023-06-12
Feat/count blocks

motechFR:
### <samp>🤖 Generated by Copilot at 41f37a9</samp>

This pull request implements a new feature to count and store the number of blocks and inline comments in each space. It also adds support for creating multiple views for board pages in the testing database. It updates the `@charmverse/core` dependency and renames the `countData` function to `countSpaceBlocks` in several files. It also modifies and adds some tests for the new features.


lib/spaces/countSpaceBlocks.ts
```diff
@@ -144,6 +152,16 @@ export async function countData({ spaceId }: { spaceId: string }) {
     views
   };
 
+  const total = getTotal(counts);
+
+  await prisma.blockCount.create({
```
mattcasey, line 157: i'd like to keep the 'retrieving' of data separate from saving of data, can u make a separate method to put this in? you can make a 'countAndSave' function if u want

package.json
```diff
@@ -84,7 +84,7 @@
     "@bangle.dev/tooltip": "^0.31.6",
     "@bangle.dev/utils": "^0.31.6",
     "@beam-australia/react-env": "^3.1.1",
-    "@charmverse/core": "github:charmverse/core#main",
+    "@charmverse/core": "github:charmverse/core#feat/count-blocks",
```
mattcasey, line 87: remove this before merge - remember u can just update the code in actions/install/action.yml still

lib/spaces/getSpaceBlockCount.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (blockCount) {
+    return {
+      count: blockCount.count,
+      createdAt: blockCount.createdAt,
+      details: blockCount.details
+    };
+  }
+
+  const { counts, total } = await countSpaceBlocks({
```
mattcasey, line 26: seems like we should log this case as an error?

motechFR: Added a log to the method, and also throwing an error if space count not found

lib/spaces/countSpaceBlocks.ts
motechFR: Sounds good, split that into separate countSpaceBlocksAndSave method

PR #2295, opened 2023-06-12
Fix: multiple tabs opening when a link is clicked

mattcasey:
### <samp>🤖 Generated by Copilot at a60b815</samp>

Remove link opening behavior in charm editor. This allows users to edit links without leaving the editor or opening a new window.

### WHY

Source: https://discord.com/channels/894960387743698944/1116744433799536750/1116744433799536750

Two issues here:
1) the handleClickOn inside inline comment is always called even if there are not inline comments, so it would open a second tab
2) inside a bullet list, even after the first fix, we'd get 3 tabs open instead of 1 because handleClickOn() was being called three times (with the same inputs afaict). I fixed this by returning true.. but am curious if @Devorein has any idea around the source of this behavior? Seems like we might be doing something funky with lists.

### DIFF
--- a/components/common/CharmEditor/components/link/link.specs.ts
+++ b/components/common/CharmEditor/components/link/link.specs.ts
@@ -3,7 +3,7 @@ import type { BaseRawMarkSpec } from '@bangle.dev/core';
 
 export function spec() {
   const linkSpec = link.spec({
-    openOnClick: true
+    openOnClick: false
   }) as BaseRawMarkSpec;
 
   linkSpec.schema.toDOM = (node) => {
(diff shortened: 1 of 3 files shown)


components/common/CharmEditor/components/link/link.specs.ts
```diff
@@ -3,7 +3,7 @@ import type { BaseRawMarkSpec } from '@bangle.dev/core';
 
 export function spec() {
   const linkSpec = link.spec({
-    openOnClick: true
+    openOnClick: false
```
mattcasey, line 6: this wasn't affecting anything but i removed the additional listener inside bangle.dev

PR #2299, opened 2023-06-13
Test/page public toggle

motechFR:
### <samp>🤖 Generated by Copilot at 070e167</samp>

This pull request adds unit tests for the `ShareToWeb` component that allows users to share a page to the web. It also updates some testing configurations and refactors some testing utilities to improve the code quality and consistency.

### WHY
Regression test for public page toggle that ensures we can use when necessary, and also account for public proposals

### DIFF
--- a/components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
+++ b/components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
@@ -1,9 +1,17 @@
 import { AvailablePagePermissions } from '@charmverse/core/permissions/flags';
-import type { Space } from '@charmverse/core/prisma';
+import type { Space } from '@charmverse/core/prisma-client';
 import { render } from '@testing-library/react';
 import { v4 as uuid } from 'uuid';
 
+// Import hooks to mock
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import { usePagePermissions } from 'hooks/usePagePermissions';
+import { usePages } from 'hooks/usePages';
+// import { usePages } from 'hooks/usePages';
+import { useProposal } from 'hooks/useProposal';
 import { createMockSpace } from 'testing/mocks/space';
+
+import ShareToWeb from '../ShareToWeb';
 
 jest.mock('next/router', () => ({
   useRouter: jest.fn(() => ({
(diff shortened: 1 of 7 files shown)


components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
```diff
@@ -0,0 +1,232 @@
...
+jest.mock('hooks/useCurrentSpace');
+jest.mock('hooks/usePagePermissions');
+jest.mock('hooks/usePages', () => ({
+  usePages: jest.fn(() => ({
+    pages: {}
+  }))
+}));
+jest.mock('hooks/useCurrentSpace', () => ({
+  useCurrentSpace: jest.fn(() => createMockSpace())
+}));
```
motechFR, line 40: We mock all hook modules so we can easily modify their behaviour later.

Jest hoists these top-level mocks before the imports, so our components will work with the mocked versions, meaning we only need to import them once

components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
```diff
@@ -0,0 +1,232 @@
...
+    query: {},
+    route: '/',
+    pathname: '',
+    path: '',
+    asPath: ''
+  }))
+}));
+jest.mock('hooks/useProposal', () => ({
+  useProposal: jest.fn(() => ({
+    proposal: null
```
motechFR, line 27: Jest automocks the exports, but provides no values. In cases where the component does some destructuring, we need to ensure an initial object value is provided. (This can also be done locally)

components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+afterAll(() => {
+  jest.resetModules();
+});
+
+describe('shareToWeb', () => {
+  it('should render the toggle as checked if no public permission exists or as unchecked if a public permission exists', async () => {
+    const pageId = uuid();
+
+    (usePagePermissions as jest.Mock<ReturnType<typeof usePagePermissions>>).mockReturnValueOnce({
```
motechFR, line 50: We are using mockReturnValueOnce. This value will be consumed the next time we call the render method.

If we use mockReturnValue, this has the side effect of affecting the mock return for subsequent tests.

package.json
```diff
@@ -49,7 +49,7 @@
     "db-tool:test": "dotenv -e .env.test.local -- npx prisma studio",
     "typecheck": "node_modules/typescript/bin/tsc --project tsconfig.json --noEmit",
     "test:setup-db": "./scripts/configure-db.sh",
-    "test:browser": "TZ=UTC jest --config='./jest.config.browser.ts'",
+    "test:browser": "TZ=UTC jest components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.ts --config='./jest.config.browser.ts'",
```
mattcasey, line 52: please remove the hard-coded file. I'm also wondering if you can just pass it when you call it? This is how I run e2e test for example: `npm run debug:e2e __e2e__/tests....spec.ts` and it works fine

testing/mocks/space.ts
mattcasey, line 1: thanks for splitting this out

mattcasey (approved):
Approved but please cleanup package json

PR #2307, opened 2023-06-14
More robust cusotm domain check

msadura:
### <samp>🤖 Generated by Copilot at c8b0b27</samp>

Improved custom domain validation for spaces. Added and updated some checks in `getValidCustomDomain.ts` to avoid invalid or reserved domains.

### WHY
Do not try to rewrite URLs or consider them as custom domain for blacklisted domains (amazonaws.com - direct ec2 URL) or IP address

### DIFF
--- a/lib/utilities/domains/getValidCustomDomain.ts
+++ b/lib/utilities/domains/getValidCustomDomain.ts
@@ -1,6 +1,8 @@
 import { getValidDefaultHost } from 'lib/utilities/domains/getValidDefaultHost';
 import { isLocalhostAlias } from 'lib/utilities/domains/isLocalhostAlias';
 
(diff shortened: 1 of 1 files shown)


lib/utilities/domains/getValidCustomDomain.ts
```diff
@@ -1,6 +1,8 @@
 import { getValidDefaultHost } from 'lib/utilities/domains/getValidDefaultHost';
 import { isLocalhostAlias } from 'lib/utilities/domains/isLocalhostAlias';
 
+const BLACKLISTED_DOMAIN_NAMES = ['amazonaws.com'];
```
mattcasey, line 4: why do we need to blacklist these? What is the harm in rewriting URLs?

PR #2309, opened 2023-06-14
Subdomain oauth login redirect

msadura:
### <samp>🤖 Generated by Copilot at 8db0456</samp>

Added subdomain detection and redirection logic to `RouteGuard` component. This ensures that users are redirected to the main app login page on the `app` subdomain if they are not logged in or have no wallet address on a subdomain URL.

### WHY
PR to make oauth logins work with subdomains.

current approach - assuming you are trying to visit subdomain.chrmverse.io and you are not logged in, it will reridect you to app.charmverse.io for login with proper back redirect url.


components/common/RouteGuard.tsx
```diff
@@ -83,15 +86,10 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
...
-      return { authorized: true };
-    }
-
     // condition: no user session and no wallet address
     else if (!user) {
-      // condition: space subdomain / custom domain main path
-      // do not redirect - it will display login
-      if (isSpaceSubdomainPath) {
+      // if app is running on a subdomain, redirect to main app login
+      if (isSubdomainUrl && redirectToAppLogin()) {
```
motechFR, line 92: I see in the body of redirectToAppLogin, we are triggering a redirect to a specific URL anyways.

Should we return authorized: false, since we are performing a redirect? This would allow the RouteGuard to return null, and avoid any UI flashes

msadura: ok, I was worried that route guard will try redirecting on its own and there will be some sort of race between redirects. But if it can just return null without redirect then it should be fine

PR #2310, opened 2023-06-14
show some of the sidebar while rest is loading on page load

mattcasey:
### <samp>🤖 Generated by Copilot at b513e09</samp>

This pull request enhances the sidebar component of the app by refactoring the code and optimizing the rendering. It also moves the `useHasMemberLevel` hook to the `Sidebar.tsx` file, where it is exclusively used, and removes unnecessary variables and imports.

### WHY
We can start rendering the sidebar, its background and top links while we're waiting to load the setting for member features.

The more we can render before stuff loads, the more we can render on server-side as well.

Currently, the sidebar is a white space at first:
image

### DIFF
--- a/components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
+++ b/components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
@@ -1,5 +1,4 @@
 import { AvailablePagePermissions } from '@charmverse/core/permissions/flags';
-import type { Space } from '@charmverse/core/prisma-client';
 import { render } from '@testing-library/react';
 import { v4 as uuid, v4 } from 'uuid';
 
@@ -9,7 +8,7 @@ import { usePagePermissions } from 'hooks/usePagePermissions';
 import { usePages } from 'hooks/usePages';
 // import { usePages } from 'hooks/usePages';
 import { useProposal } from 'hooks/useProposal';
-import { createMockSpace } from 'testing/mocks/space';
+import { mockCurrentSpaceContext } from 'testing/mocks/useCurrentSpace';
 
 import ShareToWeb from '../ShareToWeb';
 
@@ -36,7 +35,7 @@ jest.mock('hooks/usePages', () => ({
   }))
 }));
 jest.mock('hooks/useCurrentSpace', () => ({
-  useCurrentSpace: jest.fn(() => createMockSpace())
+  useCurrentSpace: jest.fn(() => mockCurrentSpaceContext())
 }));
 
 afterAll(() => {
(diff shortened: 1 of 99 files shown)


hooks/useHasMemberLevel.tsx
```diff
@@ -11,12 +11,10 @@ type MemberLevelOutput = {
 
 export function useHasMemberLevel(level: MemberLevel): MemberLevelOutput {
   const space = useCurrentSpace();
-  const { isLoaded: isSpacesLoaded, spaces } = useSpaces();
+  const { spaces } = useSpaces();
   const { user, isLoaded: isUserLoaded } = useUser();
-
   if (
     !isUserLoaded ||
-    !isSpacesLoaded ||
```
mattcasey, line 19: @motechFR what i'm seeing most of the time is that `space` gets loaded before `spaces`. This is because the `publicSpace` returns a space from a special endpoint. And my assumption is that if space exists, that should be enough to verify access.

I don't completely understand the case below, but it's the only place we use `spaces` in this hook. Would it be more accurate if instead we had a `isLoading` check from useCurrentSpace()?

components/common/PageLayout/components/Header/components/ShareButton/components/__tests__/ShareToWeb.spec.tsx
```diff
@@ -9,7 +8,7 @@ import { usePagePermissions } from 'hooks/usePagePermissions';
 import { usePages } from 'hooks/usePages';
 // import { usePages } from 'hooks/usePages';
 import { useProposal } from 'hooks/useProposal';
-import { createMockSpace } from 'testing/mocks/space';
+import { mockCurrentSpaceContext } from 'testing/mocks/useCurrentSpace';
```
mattcasey, line 11: i created a mock for the entire result from the hook, so it's easy to make changes in the future

testing/mocks/useCurrentSpace.ts
```diff
@@ -0,0 +1,12 @@
+import type { Space } from '@charmverse/core/prisma-client';
```
mattcasey, line 1: maybe we'll want to put these in a hooks subfolder eventualy

testing/mocks/useCurrentSpace.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import type { useCurrentSpace } from 'hooks/useCurrentSpace';
+
+import { createMockSpace } from './space';
+
+export const mockCurrentSpaceContext = (space?: Partial<Space>): ReturnType<typeof useCurrentSpace> => {
```
motechFR, line 7: Naming convention suggestion;

Renaming this to mockUseCurrentSpace

motechFR (approved):

testing/mocks/useCurrentSpace.ts
mattcasey: I had it that way at first but it's actually mocking the _return_ of use current space, which is the context

PR #2313, opened 2023-06-15
Fix person property not closing on its own

mattcasey:
- i removed some unecessary state and tried to reuse styles from the 'select property' component so that the UI doesn't jump around as much. It's still not perfect! :/
- i think the fix here was not controlling the 'open' state, and instead using 'openOnFocus' which was not being passed down to the autocomplete component.


motechFR (approved):
Nice cleanup, I can see that we'd want to collapse the InputSearchMembers on select with this PR.

I guess most cases we only assign 1 user at a time, so should work well

mattcasey:
> I can see that we'd want to collapse the InputSearchMembers on select with this PR.
Actually @motechFR i didn't want it to close on select! But can't figure it out. I want to update the backend each time you select someone, and it seems to reset the internal state of the dropdown each time. The multi-select has this same issue.

PR #2323, opened 2023-06-16
Fixed reorder page in sidebar

Devorein:
### <samp>🤖 Generated by Copilot at d695892</samp>

Refactor page navigation logic and UI. Fix `parentId` type issue in `PageNavigationItem`.

### WHY
Fix for: https://www.loom.com/share/a1fd81df39ad4c9e9a73b244140d2b87

### DIFF
--- a/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
+++ b/components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
@@ -141,7 +141,7 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
           charmClient.pages.updatePage({
             id: page.id,
             index: _index,
-            parentId
+            parentId: parentId ?? null
           });
         });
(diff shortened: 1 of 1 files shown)


components/common/PageLayout/components/PageNavigation/PageNavigation.tsx
```diff
@@ -141,7 +141,7 @@ function PageNavigation({ deletePage, isFavorites, rootPageIds, onClick }: PageN
           charmClient.pages.updatePage({
             id: page.id,
             index: _index,
-            parentId
+            parentId: parentId ?? null
```
motechFR, line 144: When we use "null" we are setting the page to be a root level page.

For more explicit explanation of this behaviour in future I suggest moving this statement to l. 123 with the comment below

```typescript
// If there is no parentId, the page was dropped at the root level
const parentId = containerItem.parentId ?? null;
```

motechFR (approved):

PR #2325, opened 2023-06-16
Use log.info when there is an error counting prosemirror blocks

motechFR:
### <samp>🤖 Generated by Copilot at 04c94c2</samp>

Changed log level of `countBlocks` error from `error` to `info` in `lib/prosemirror/countBlocks.ts`. This reduces the noise in the error logs and improves the billing system.

### WHY
We are often having an error counting prosemirror blocks. This pollutes the datadog logs, and our current counts are within the correct magnitude.


mattcasey (changes requested):
if it's in a try/catch it should probably at least be a warning and  please report this as a bug. it's not OK that we just fail to count blocks

PR #2327, opened 2023-06-16
Resize text field popup size based on wrap status and cell dimensions

Devorein:
### <samp>🤖 Generated by Copilot at a899c86</samp>

This pull request enhances the `BoardEditor` module by improving the popover positioning and appearance for editing property values in a table. It adds a new `columnRef` prop to the `PropertyValueElement` and `TextInput` components, and refactors the `TableRow` component to use a constant for the column reference map.

### WHY
For a small text-cell it opens the popup component as full-height (when wrapped column is toggled on)

### DIFF
--- a/components/common/BoardEditor/focalboard/src/widgets/TextInput.tsx
+++ b/components/common/BoardEditor/focalboard/src/widgets/TextInput.tsx
@@ -32,6 +32,13 @@ function Editable(
   const elementRef = useRef<HTMLTextAreaElement>(null);
   const { className, ...props } = useEditable(_props, ref, elementRef);
 
+  const memoizedHeight = React.useMemo(() => {
+    if (wrapColumn && columnRef?.current) {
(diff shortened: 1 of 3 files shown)


components/common/BoardEditor/focalboard/src/widgets/TextInput.tsx
```diff
@@ -51,15 +52,15 @@ function Editable(
       paperSx={{
         width: 350,
         p: 2,
-        height: wrapColumn ? 'calc(100vh - 32px)' : 'fit-content'
+        height: wrapColumn && columnRef?.current ? `${columnRef?.current?.clientHeight}px` : 'fit-content'
```
mattcasey, line 55: Can u put this in a memo, or have it claculate when the popup opens? - it is expensive to query DOM every render

mattcasey (changes requested):
don't query DOM measurements on renders

PR #2332, opened 2023-06-19
Fixed bookmarked page not being removed

Devorein:
### <samp>🤖 Generated by Copilot at a042de9</samp>

Simplify favorites API by removing a route and adding a delete handler. This makes it easier to add or remove pages from the favorites list with a single query parameter.


mattcasey (approved):
thanks

PR #2341, opened 2023-06-20
Webhook events for space subscription activation and updates

Devorein:
### <samp>🤖 Generated by Copilot at a8e88a4</samp>

Updated the app to use WebSockets for real-time updates on space subscription status. Added new WebSocket message types and a `relay` module to send and receive subscription events. Modified the `SubscriptionSettings` component to use the `useWebSocketClient` hook and refetch subscription data.

### DIFF
--- a/components/settings/subscription/SubscriptionSettings.tsx
+++ b/components/settings/subscription/SubscriptionSettings.tsx
@@ -45,22 +45,12 @@ export function SubscriptionSettings({ space }: { space: Space }) {
   );
 
   useEffect(() => {
-    const unsubscribeFromSpaceSubscriptionActivation = subscribe('space_subscription_activated', () => {
-      refetchSpaceSubscription();
-    });
-
-    const unsubscribeFromSpaceSubscriptionUpdates = subscribe('space_subscription_updated', () => {
-      refetchSpaceSubscription();
-    });
-
-    const unsubscribeFromSpaceSubscriptionCancellation = subscribe('space_subscription_cancelled', () => {
+    const unsubscribeFromSpaceSubscriptionUpdates = subscribe('space_subscription', () => {
       refetchSpaceSubscription();
     });
 
     return () => {
-      unsubscribeFromSpaceSubscriptionActivation();
       unsubscribeFromSpaceSubscriptionUpdates();
-      unsubscribeFromSpaceSubscriptionCancellation();
     };
   }, []);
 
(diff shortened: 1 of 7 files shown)


components/settings/subscription/SubscriptionSettings.tsx
```diff
@@ -38,13 +40,30 @@ export function SubscriptionSettings({ space }: { space: Space }) {
     {
       onError() {
         showMessage('Updating failed! Please try again', 'error');
-      },
-      async onSuccess() {
```
Devorein, line 42: Since the websocket event listeners refetch space subscription, this will cause an additional request

lib/websockets/interfaces.ts
```diff
@@ -101,6 +101,21 @@ type SubscribeToWorkspace = {
   } & SocketAuthReponse;
 };
 
+type SpaceSubscriptionUpdated = {
+  type: 'space_subscription_updated';
+  payload: object;
```
Devorein, line 106: I was not sure what to send as payload to the client. The listeners would just refetch the data rather than update the cached data

components/settings/subscription/SubscriptionSettings.tsx
```diff
@@ -38,13 +40,30 @@ export function SubscriptionSettings({ space }: { space: Space }) {
...
-        await refetchSpaceSubscription();
       }
     }
   );
 
+  useEffect(() => {
```
valentinludu, line 47: Can you move this and the getter (line 27) into a useSpaceSubscription hook?
Should return subscription, refetch, isLoading.

Maybe a context would also make sense? To know in all the app when a subscription has been created/removed.

valentinludu:
Please think about the case where the space.tier is updated. I think this should trigger a websocket event

components/settings/subscription/SubscriptionSettings.tsx
Devorein: Yeah I think a global context would make much more sense. I don't see any useSpaceSubscription hook anywhere though :/

Devorein:
@valentinludu The only place where we are updating the `space.paidTier` is inside the webhook. The only other place is when we create a space and delete a subscription (but this will be removed later on).

The space upgrade/downgrade would be triggered in the `customer.subscription.updated` webhook event and we are sending the websocket event for that as well.

components/settings/subscription/SubscriptionSettings.tsx
valentinludu: it was a suggestion to create it. Or whatever name works for you.

Devorein: Ah got it. Lets go with that

valentinludu (approved):

PR #2347, opened 2023-06-20
add tests to counting blocks script

mattcasey:
### <samp>🤖 Generated by Copilot at 697dfd7</samp>

Improved the `countBlocks` function in `lib/prosemirror/countBlocks.ts` by using a more precise upper bound, skipping over irrelevant nodes, and avoiding counting undefined nodes.


lib/prosemirror/countBlocks.ts
```diff
@@ -39,8 +39,11 @@ export function countBlocks(pageContent: any | null, spaceId?: string) {
     try {
       const doc = getNodeFromJson(pageContent);
       if (doc) {
-        doc.nodesBetween(0, doc.nodeSize, (node) => {
-          if (node.type && !INLINE_NODES_TO_IGNORE.includes(node.type.name)) {
+        doc.nodesBetween(0, doc.content.size, (node) => {
+          if (INLINE_NODES_TO_IGNORE.includes(node.type.name)) {
```
motechFR, line 43: Suggestion for even more conciseness (since node.type at 46 would always exist, since we use it on l:43)

```typescript
doc.nodesBetween(0, doc.content.size, (node) => {
  if (node.type && !INLINE_NODES_TO_IGNORE.includes(node.type.name)) {
   count += 1;
  }
})
```

motechFR (approved):

lib/prosemirror/countBlocks.ts
mattcasey: it's intentional to return false

mattcasey: i reverted it anyway,  as I 'm no longer sure it matters to return false - the error seems to have gone away :)

PR #2354, opened 2023-06-21
Feat/allow custom origins

msadura:
### <samp>🤖 Generated by Copilot at 568670c</samp>

This pull request introduces a new module `allowedOriginsCache` that caches the custom domains of the spaces and allows the websocket server to use them as allowed origins. This improves the security and performance of the websocket communication.

### WHY
We need a way to allow access to websocket server for custom domains. It is a matter of loading list of custom domains and using it for cors settings


lib/middleware/allowedOriginsCache.ts
```diff
@@ -0,0 +1,33 @@
...
+const DEFAULT_ORIGINS = ['.charmverse.co', '.charmverse.io'];
+
+class AllowedOriginsCache {
+  allowedOrigins: string[] | null = null;
+
+  constructor() {
+    this.refreshOrigins();
+  }
+
+  async refreshOrigins() {
```
msadura, line 15: in the future, this method can be called whenever we need to update custom origins list (i.e. when user saves custom domain)

background/websockets.ts
```diff
@@ -22,10 +23,14 @@ const io = new Server(server, {
...
     credentials: true,
     origin: (requestOrigin, callback) => {
-      // support any subdomain for staging
-      if (requestOrigin?.endsWith('.charmverse.co') || requestOrigin?.endsWith('.charmverse.io')) {
+      if (isDevEnv) {
         callback(null, requestOrigin);
-      } else if (isDevEnv) {
+      } else if (
+        // custom origins did not load yet, allow all origins temporarily
+        !allowedOriginsCache.allowedOrigins ||
```
mattcasey, line 30: can we actually just check for .charmverse.co/.io first? no need to wait for a cache for 99% of spaces

lib/middleware/allowedOriginsCache.ts
```diff
@@ -0,0 +1,33 @@
...
+    const spaceCustomDomains = await prisma.space.findMany({
```
mattcasey, line 16: i think this should cache results one at a time - a map of promises, i guess. Loading all spaces will be a memory issue at some point

msadura: but I query only space with custom domains, we care only about those:
`customDomain: { not: null }`

msadura:
> please change the logic to something like:
> 
> 1. check dev
> 2. check whitelist of charmverse domains
> 3. check if specific domain exists, and cache the result by domain
> 
> In the future we might add something like a TTL for each cache key, but since its just being whitelisted its probably not an issue

@mattcasey I ended up having a 5min cache - origin method is called on every page change (editor connect), so it feels better to have some sort of caching. LMK if you think we should still ignore caching.

background/websockets.ts
```diff
@@ -21,14 +22,21 @@ const io = new Server(server, {
...
-    origin: (requestOrigin, callback) => {
+    origin: async (requestOrigin, callback) => {
       // support any subdomain for staging
       if (requestOrigin?.endsWith('.charmverse.co') || requestOrigin?.endsWith('.charmverse.io')) {
         callback(null, requestOrigin);
       } else if (isDevEnv) {
         callback(null, requestOrigin);
       } else {
-        callback(new Error('Not allowed by CORS'));
+        const isCustomOriginAllowed = await verifyCustomOrigin(requestOrigin);
```
mattcasey, line 32: need to add a catch since i dont think origin is designed to handle a promise

PR #2359, opened 2023-06-21
Fix issue and update the tests

motechFR:
### <samp>🤖 Generated by Copilot at 257a924</samp>

Added permissions service integration to form and webhook API handlers for creating database cards. Updated and renamed integration tests for cards API endpoint. Removed unused `permissions` field from `createDatabaseCardPage` function.

### WHY
Create cards endpoint was broken. I'm wondering how this wasn't picked up by the tests


pages/api/v1/databases/[id]/form.ts
```diff
@@ -46,6 +47,11 @@ export async function createFormResponse(req: NextApiRequest, res: NextApiRespon
     userId: req.botUser.id
   });
 
+  premiumPermissionsApiClient.pages.setupPagePermissionsAfterEvent({
```
Comment (mattcasey) [inline pages/api/v1/databases/[id]/form.ts:50]: this is async so we need to await don't we?

mattcasey (changes requested):
await the promises. if permissions fails, the caller should still get an error. it also lets us capture the errors if they happen

pages/api/v1/databases/[id]/form.ts
Comment (motechFR) [reply pages/api/v1/databases/[id]/form.ts:50]: It wasn't really necessary for the job to complete (since the card gets created anyways) but updated them to await

PR #2365, opened 2023-06-22
Fixed proposal-source db being replicated in non-proposal source db

Devorein:
### <samp>🤖 Generated by Copilot at 28b037e</samp>

This pull request improves the type safety and error handling of the `updateCardsFromProposals` function and its related components. It also fixes a bug that caused unnecessary updates of the proposal source when switching views.

### WHY
[Conversation](https://discord.com/channels/894960387743698944/1118245084085878895)

### DIFF
--- a/lib/pages/updateCardsFromProposals.ts
+++ b/lib/pages/updateCardsFromProposals.ts
@@ -4,6 +4,8 @@ import { prisma } from '@charmverse/core/prisma-client';
 
 import { prismaToBlock } from 'lib/focalboard/block';
 import type { Board } from 'lib/focalboard/board';
+import type { BoardView } from 'lib/focalboard/boardView';
+import { InvalidStateError } from 'lib/middleware';
 import { isTruthy } from 'lib/utilities/types';
 import { relay } from 'lib/websockets/relay';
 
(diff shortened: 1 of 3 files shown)


mattcasey (approved):
looks good thanks for the test

lib/pages/updateCardsFromProposals.ts
```diff
@@ -30,6 +32,20 @@ export async function updateCardsFromProposals({
...
+  const views = (
+    await prisma.block.findMany({
+      where: {
+        type: 'view',
+        parentId: boardId
+      }
+    })
+  ).map(prismaToBlock) as BoardView[];
+
+  // Ideally all the views should have sourceType proposal when created, but there are views which doesn't have sourceType proposal even though they are created from proposal source
```
mattcasey, line 44: @Devorein this sounds like a bug we should fix?

mattcasey (approved):

lib/pages/updateCardsFromProposals.ts
Devorein: Yeah this is a bug, I had a proposal source table view (with sourceType: `proposals`), created another table and it had the same sourceType: `proposals`, but then I added a board view and it didn't have one.

PR #2369, opened 2023-06-22
downgrade next.js

mattcasey:
### <samp>🤖 Generated by Copilot at d9e26f1</samp>

Updated `next` dependency to use caret range for better compatibility and security. This fixes a dependabot alert for `package.json`.

### WHY
Dev server restarting every 10 minutes is really painful. I have another PR to try upgrading next.js but they have broken integration with socket.io, so we have no way to upgrade until that is also fixed. In the meantime, I don't recall a need to be on 13.2.4, and some users in Next's issues said that 13.2.2 was the last 'good' version.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -127,7 +127,7 @@
         "nanoevents": "^6.0.2",
         "nanoid": "^3.3.6",
         "nanoid-dictionary": "^3.0.0",
-        "next": "13.2",
+        "next": "^13.2.2",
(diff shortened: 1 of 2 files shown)


msadura (approved):
nice find, lets try it out :D

Devorein:
@mattcasey Thank you. This seems to have fixed the issue 🥳 🔥 🚀 💯

PR #2375, opened 2023-06-24
Prevent infinite redirect loop

motechFR:
### <samp>🤖 Generated by Copilot at 24ced43</samp>

This pull request enhances the `RouteGuard` component to handle authentication and redirection more efficiently. It fixes the `authCheckAndRedirect` function and respects the `returnUrl` query parameter.

### WHY
Don't return a redirect in the RouteGuard if we already have a returnUrl

### DIFF
--- a/components/common/RouteGuard.tsx
+++ b/components/common/RouteGuard.tsx
@@ -66,7 +66,7 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
     return () => {
       router.events.off('routeChangeComplete', authCheckAndRedirect);
     };
-  }, [isLoading, user, spaces]);
+  }, [isLoading, user, spaces, router.query]);
 
   // authCheck runs before each page load and redirects to login if user is not logged in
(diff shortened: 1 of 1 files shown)


components/common/RouteGuard.tsx
```diff
@@ -66,7 +66,7 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
     return () => {
       router.events.off('routeChangeComplete', authCheckAndRedirect);
     };
-  }, [isLoading, user, spaces]);
+  }, [isLoading, user, spaces, router.query]);
```
motechFR, line 69: This router.query bit is necessary to rerun RouteGuard with the new value of the returnUrl and return a simple authorized: true

mattcasey: can we just watch explicitly on router.query.returnURl then? this seems like iw ill trigger maybe more tan we want - whenever the query changes?

motechFR: Yep sounds fair, fixing now

mattcasey (approved):

PR #2377, opened 2023-06-26
Track mixpanel events for billing and subscription actions

Devorein:
### <samp>🤖 Generated by Copilot at 04a8710</samp>

This pull request implements various improvements and updates to the subscription feature of the app. It enhances the checkout form, the subscription information, and the subscription settings components, and tracks more user events with Mixpanel. It also handles and tracks Stripe webhook events more accurately, and updates the types for the subscription events. It also reorders some imports in a minor change.


lib/metrics/mixpanel/interfaces/SubscriptionEvent.ts
```diff
@@ -1,21 +1,47 @@
...
-
 import type { SubscriptionPeriod } from 'lib/subscription/constants';
+import type { PaymentMethodType } from 'lib/subscription/mapStripeFields';
 
 import type { BaseEvent } from './BaseEvent';
 
-export type ViewSubscriptionEvent = BaseEvent;
-export type InitiateSubscriptionEvent = BaseEvent;
+export type CreateSubscriptionEvent = BaseEvent & {
+  paymentMethod: PaymentMethodType;
```
motechFR, line 7: I think we should add subscription ID here

pages/api/v1/webhooks/stripe/index.ts
```diff
@@ -141,21 +144,30 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse):
...
-        }
 
-        log.info(
-          `The invoice number ${invoice.id} for the subscription ${stripeSubscription.id} was paid for the spaceId ${spaceId}`
-        );
+          trackUserAction('create_subscription', {
+            blockQuota,
+            period,
+            spaceId,
+            paymentMethod: 'card',
```
motechFR, line 152: We probably should update this to account for different payment methods, and not hardcode it

motechFR (approved):

components/settings/subscription/CheckoutForm.tsx
```diff
@@ -63,12 +64,19 @@ export function CheckoutForm({
     resolver: yupResolver(schema())
   });
 
+  useEffect(() => {
+    if (space) {
+      charmClient.track.trackAction('view_checkout_screen', {
```
mattcasey, line 69: I think this should still be page_view, with a specific type for checkout

mattcasey (changes requested):
We should stick with 'page_view' when tracking views. The type could be 'billing/marketing', 'billing/checkout', 'billing/settings', etc

lib/metrics/mixpanel/interfaces/SubscriptionEvent.ts
Devorein: Nice one. I will add it

pages/api/v1/webhooks/stripe/index.ts
Devorein: Correct. I have contacted valentin and we will get this fixed.

components/settings/subscription/CheckoutForm.tsx
Devorein: Lets go with that then

mattcasey (approved):

pages/api/v1/webhooks/stripe/index.ts
```diff
@@ -225,6 +247,19 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse):
...
+          const activeSubscription = await getActiveSpaceSubscription({ spaceId });
+          if (activeSubscription) {
+            trackUserAction('update_subscription', {
+              blockQuota: subscription.items.data[0].quantity as number,
+              period: subscription.items.data[0].price.recurring?.interval === 'month' ? 'monthly' : 'annual',
+              previousBlockQuota: activeSubscription.blockQuota,
+              previousPeriod: activeSubscription.period,
+              subscriptionId: activeSubscription.id,
+              spaceId,
+              userId: ''
```
mattcasey, line 260: why do we ened to pass in an empty string here? are the types wrong?

pages/api/v1/webhooks/stripe/index.ts
```diff
@@ -301,6 +343,52 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse):
...
+              ? 'crypto'
+              : charge.payment_method_details?.type?.startsWith('ach')
+              ? 'ach'
+              : 'card',
+            userId: ''
+          });
+        }
+
+        const lastFinalizationError = invoice.last_finalization_error;
+        log.error(`Invoice payment failed for invoice ${invoice.id} for the subscription ${stripeSubscription.id}`, {
```
mattcasey, line 376: this should be a warning, since it's not an error from our application

Devorein: Thank you. Updated

pages/api/v1/webhooks/stripe/index.ts
Devorein: Right. userId is a required field and used as the distinct_id when creating the mixpanel event. I will just go with space.createdBy for now

PR #2386, opened 2023-06-27
log error with stripe subscription

mattcasey:
### <samp>🤖 Generated by Copilot at 698a497</samp>

This pull request improves the error logging and handling for the workspace and subscription creation functions in `lib/spaces/createSpace.ts` and `lib/subscription/createProSubscription.ts`. It adds the `error` property to the log object in `createSpace.ts` and removes the redundant `try-catch` block and variable declaration in `createProSubscription.ts`.

### DIFF
--- a/lib/subscription/createProSubscription.ts
+++ b/lib/subscription/createProSubscription.ts
@@ -110,47 +110,33 @@ export async function createProSubscription({
     throw new InvalidStateError(`No price for product ${productId} and space ${spaceId}`);
   }
 
-  let subscription: Stripe.Subscription | undefined;
-  try {
-    // Case when the user is updating his subscription in checkout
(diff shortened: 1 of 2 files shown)


lib/subscription/createProSubscription.ts
```diff
@@ -110,47 +110,33 @@ export async function createProSubscription({
...
-      period,
-      billingEmail,
-      error
-    });
+  // Case when the user is updating his subscription in checkout
+  if (existingStripeSubscription && existingStripeSubscription.status === 'incomplete') {
+    await stripeClient.subscriptions.del(existingStripeSubscription.id);
   }
 
-  if (!subscription) {
```
mattcasey, line 151: we are already logging the error, no need to catch and throw a second error. this actually makes things confusing because the 2nd thrown error has less information about what happened

PR #2406, opened 2023-06-28
Mini & public profile re-design

Devorein:
### <samp>🤖 Generated by Copilot at c5275ec</samp>

Refactored and redesigned the user profile components to use a tabbed interface, a grid of widgets, and a consistent style. Added new components for displaying user's collections, social accounts, and space properties. Fixed a typo in an error message.


charmClient/apis/profileApi.ts
```diff
@@ -31,4 +33,12 @@ export class ProfileApi {
   setForumCategoryNotification(input: Omit<SetForumCategoryNotificationInput, 'userId'>) {
     return http.PUT('/api/profile/space-notifications/set-forum-category', input);
   }
+
+  getLensDefaultProfile(userId: string) {
+    return http.GET<ProfileFragment | null>(`/api/lens/profile/${userId}`);
+  }
+
+  getEnsProfile(userId: string) {
+    return http.GET<EnsProfile | null>(`/api/profile/ens/${userId}`);
```
mattcasey, line 42: please make the endpoints the same: IE if it's profile/ens/userId, then it should be profile/lens/userId (I think that makes sense) and the folders in lib can probably follow the same structure

mattcasey: Actually i think these should go into /api/public/profile/<userId/lens, etc. since that is where we get the profile for public pages?

pages/_app.tsx
```diff
@@ -92,11 +90,15 @@ import 'components/common/BoardEditor/focalboard/src/widgets/menu/subMenuOption.
...
-import 'theme/focalboard/focalboard.main.scss';
 import 'lit-share-modal-v3/dist/ShareModal.css';
+import 'prosemirror-menu/style/menu.css';
 import 'react-resizable/css/styles.css';
+import 'theme/@bangle.dev/styles.scss';
+import 'theme/focalboard/focalboard.button.scss';
+import 'theme/focalboard/focalboard.main.scss';
 import 'theme/lit-protocol/lit-protocol.scss';
+import 'theme/print.scss';
+import 'theme/prosemirror-tables/prosemirror-tables.scss';
```
mattcasey, line 101: why were all these CSS files reordered? Ordering of CSS can affect the UI

pages/_app.tsx
```diff
@@ -208,9 +210,11 @@ function DataProviders({ children }: { children: ReactNode }) {
                           <BountiesProvider>
                             <PaymentMethodsProvider>
                               <PagesProvider>
-                                <UserProfileProvider>
-                                  <PageTitleProvider>{children}</PageTitleProvider>
-                                </UserProfileProvider>
+                                <MemberPropertiesProvider>
```
mattcasey, line 213: i think we should keep this OUT Of _app. just add it to what components you need it for?

pages/api/lens/profile/[userId].tsx
```diff
@@ -0,0 +1,21 @@
...
+
+async function getLensProfileController(req: NextApiRequest, res: NextApiResponse<ProfileFragment | null>) {
+  const userId = req.query.userId as string;
+
+  const defaultLensProfile = await getDefaultLensProfile(userId);
+
+  res.status(200).json(defaultLensProfile);
+}
+
+export default withSessionRoute(handler);
```
Comment (mattcasey) [inline pages/api/lens/profile/[userId].tsx:21]: you don't need the session, don't need to add withSessionRoute here

pages/api/profile/ens/[userId].ts
```diff
@@ -0,0 +1,21 @@
...
+
+async function getEnsProfileController(req: NextApiRequest, res: NextApiResponse<EnsProfile | null>) {
+  const userId = req.query.userId as string;
+
+  const ensProfile = await getEnsProfile({ userId });
+
+  res.status(200).json(ensProfile);
+}
+
+export default withSessionRoute(handler);
```
Comment (mattcasey) [inline pages/api/profile/ens/[userId].ts:21]: also don't need withSessionRoute here

public/images/ens_logo.svg
mattcasey, line 1: we should avoid dumping anything at the top-level of images. I think these images could go into images/logos

mattcasey (approved):

charmClient/apis/profileApi.ts
Devorein: Yeah, that makes sense. I will move the code and also make `publicProfileApi`

pages/api/lens/profile/[userId].tsx
Comment (Devorein) [reply pages/api/lens/profile/[userId].tsx:21]: Good point. Removing it for ens as well.

pages/api/profile/ens/[userId].ts
Comment (Devorein) [reply pages/api/profile/ens/[userId].ts:21]: Correct. Thanks for catching.

pages/_app.tsx
Devorein: I think I know what caused this. When I moved the `MemberPropertiesProvider` it was imported in the wrong order, and thus eslint complained. I used eslint to re-order the imports and it most likely alphabetically ordered the CSS imports

Devorein: I've reverted the css imports back to the original order

public/images/ens_logo.svg
Devorein: Let's do it

lib/profile/getDefaultLensProfile.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (walletAddresses.length === 0) {
+    return null;
+  }
+
+  // test wallet: 0x2b3DaEB14f069dB301cEAD63338a56d27A982CED
+  let ownedProfile: ProfileFragment | null = null;
+
+  for (const walletAddress of walletAddresses) {
+    const ownedProfiles = await lensClient.profile.fetchAll({
+      ownedBy: [walletAddress],
```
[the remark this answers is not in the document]
Devorein: Tried, and it worked. Thanks for the suggestion but I should've figured it out 🤦🏼‍♂️

pages/_app.tsx
Devorein: I should've pointed it out beforehand.
It was because the `UserProfileProvider`, the provider that is responsible for displaying the user profile in a modal (when clicking on the member directory card) did not have access to the member properties data (provided by `MemberPropertiesProvider`). That provider was used inside `MemberDirectoryPageWithContext` but the user profile modal required the provider to be much higher in the component tree. It was extremely difficult to find the bug. The member properties card wasn't showing up at all (that means there was no data).

Here are some images to visualize the issue and solution
Space Member properties
image

With `MemberPropertiesProvider` not relocated. No member properties card.
image

This was actually an existing bug. See this image taken from our prod app
image
image

And after the fix
image

mattcasey: thanks for figuring this out @Devorein !

PR #2416, opened 2023-07-01
Workflows to test specific version of core and deploy if there was a new version.

yiwork:
### <samp>🤖 Generated by Copilot at a85286c</samp>

This pull request improves the GitHub Actions workflows for testing and deploying the app. It removes unnecessary inputs and logic, fixes bugs, and adds debugging information. It also updates the `install` action to handle Core package versions more flexibly.

### WHY
We want to install npm package instead of pulling the code in via github


.github/actions/install/action.yml
```diff
@@ -35,19 +34,14 @@ runs:
       if: steps.cache_node_modules.outputs.cache-hit != 'true'
       run: npm ci --no-audit --no-fund
 
-    # - name: Install specific version of Core
```
mattcasey, line 38: i'd like to keep this logic here, will it still work though?

mattcasey (approved):

PR #2417, opened 2023-07-03
Fixed multi row bullet selection tab issue

Devorein:
### <samp>🤖 Generated by Copilot at 933a12a</samp>

Improve list item indentation in `CharmEditor`. Allow multi-line selections to be indented or outdented regardless of nesting level and list start. Remove unused import in `commands.ts`.

### WHY
[Card](https://app.charmverse.io/charmverse/page-49366552253645923?cardId=aa7972a2-211b-48c9-81d0-100fe4302bfc&viewId=50d940bb-fdbd-40ba-9816-616d6138a663)

### DIFF
--- a/components/common/CharmEditor/components/listItem/__tests__/commands.spec.ts
+++ b/components/common/CharmEditor/components/listItem/__tests__/commands.spec.ts
@@ -1,0 +1,81 @@
+import { Fragment, Node } from 'prosemirror-model';
+import { EditorState, TextSelection } from 'prosemirror-state';
+
+import { specRegistry } from 'components/common/CharmEditor/specRegistry';
+
+import { indentListItem } from '../listItem.component';
+
+const specRegistrySchema = specRegistry.schema;
+
+function doc(...items: (string | Node)[]): any {
+  const content = items.map((item) => {
+    if (typeof item === 'string') {
+      const textNode = specRegistrySchema.text(item);
+      const paragraphNode = specRegistrySchema.nodes.paragraph.create({}, Fragment.from(textNode));
+      return paragraphNode;
+    } else {
+      return item;
+    }
+  });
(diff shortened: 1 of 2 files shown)


mattcasey (changes requested):
Can you think of a way we can write a test for this?

I think i'd even be fine if we just started out with a browser test that doesn't call anything low-level.. you should be able to send keys to the editor. you could look at CommentForm.spec.tsx for inspiration?

Here's a script I was using to automatically insert text while testing multiplayer:
```
function insertTextAtCursor(text)
{
    let selection = window.getSelection();
    let range = selection.getRangeAt(0);
    range.deleteContents();
    let node = document.createTextNode(text);
    range.insertNode(node);

    for(let position = 0; position != text.length; position++)
    {
        selection.modify("move", "right", "character");
    };
}
var intvl = setInterval(() =>{
  insertTextAtCursor('\nbar')
}, 500);

```

components/common/CharmEditor/components/listItem/__tests__/commands.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { Fragment, Node } from 'prosemirror-model';
+import { EditorState, TextSelection } from 'prosemirror-state';
+
+import { specRegistry } from 'components/common/CharmEditor/specRegistry';
+
+import { indentListItem } from '../listItem.component';
+
+const specRegistrySchema = specRegistry.schema;
+
+function doc(...items: (string | Node)[]): any {
```
mattcasey, line 10: next time please use the utility we have for generating docs in testing/prosemirror

Devorein: Thank you for pointing this out. Will definitely use those utilities next time.

PR #2431, opened 2023-07-05
Game7 profile cards

Devorein:
### <samp>🤖 Generated by Copilot at 946aa93</samp>

This pull request adds a new feature to the app.charmverse.io website, which is to display the Game7 profile data of a user in their user profile page. It adds a new Game7 profile widget component, a new API route to fetch the Game7 profile data, and several helper functions and types to interact with the Game7 API. It also updates the existing components and modules to support the new feature.


lib/summon/interfaces.ts
```diff
@@ -43,3 +44,11 @@ export type XPSUserInventory = {
   materials: any[];
   gear: any[];
 };
+
+export type XPSUserProfile = {
```
mattcasey, line 48: Let's stick with "Summon" when possible...? rename to "SummonUserProfile" to be consistent for example you renamed a method to getUserSummonProfile()

mattcasey (approved):

lib/summon/interfaces.ts
Devorein: I think it will be best if we go with "summon", as that's the common term. I've not found a single reference where Game7 mentioned xps

PR #2433, opened 2023-07-06
Enable loop payments

valentinludu:
### <samp>🤖 Generated by Copilot at 38ef4bb</samp>

Added coupon support for crypto subscriptions in the app. Updated the `createCryptoSubscription` function and the `CheckoutForm` component to handle coupon codes and apply them to the Stripe subscription. Fixed the Stripe webhook response format in `index.ts`.

### WHY
Enable loop payments
Added a return statement for webhook in order to fix some errors.

[Test steps](https://app.charmverse.io/charmverse/page-36428412842339175)

### DIFF
--- a/components/settings/subscription/CheckoutForm.tsx
+++ b/components/settings/subscription/CheckoutForm.tsx
@@ -22,7 +22,7 @@ import type { PaymentType } from './PaymentTabs';
 import PaymentTabs, { PaymentTabPanel } from './PaymentTabs';
 
 export function CheckoutForm({
-  onCancel,
+  onCloseCheckout,
   handleCoupon,
   registerCoupon,
   errors,
@@ -46,7 +46,7 @@ export function CheckoutForm({
   }>;
   validating: boolean;
   registerCoupon: UseFormRegisterReturn<'coupon'>;
-  onCancel: VoidFunction;
+  onCloseCheckout: VoidFunction;
   handleCoupon: (coupon: string | undefined) => Promise<void>;
 }) {
   const stripe = useStripe();
@@ -113,7 +113,7 @@ export function CheckoutForm({
           'Payment successful! Please revisit this page in a few minutes to see your new account details.',
           'success'
         );
-        onCancel();
+        onCloseCheckout();
       }
     };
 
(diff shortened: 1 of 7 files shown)


components/settings/subscription/CheckoutForm.tsx
```diff
@@ -112,7 +109,11 @@ export function CheckoutForm({
       }
 
       if (event?.type === 'message' && event?.data === 'CheckoutComplete') {
-        setPendingPayment(true);
+        showMessage(
+          'Payment successful! Please revisit this page in a few minutes to see your new account details.',
+          'success'
+        );
+        onCancel();
```
motechFR, line 116: Readability suggestion; Rename this to onClose, since cancelling is a bit weird after a payment successful

lib/subscription/createCryptoSubscription.ts
```diff
@@ -2,20 +2,50 @@ import { getLoopProducts } from 'lib/loop/loop';
...
 
 export async function createCryptoSubscription({
   subscriptionId,
-  email
+  email,
+  coupon
 }: CreateCryptoSubscriptionRequest): Promise<CreateCryptoSubscriptionResponse> {
   const encodedEmail = encodeURIComponent(email);
   const subscription = await stripeClient.subscriptions.retrieve(subscriptionId);
+  await stripeClient.subscriptions.del(subscriptionId);
```
motechFR, line 16: Is this supposed to be here?

lib/subscription/createCryptoSubscription.ts
```diff
@@ -2,20 +2,50 @@ import { getLoopProducts } from 'lib/loop/loop';
...
-  await stripeClient.customers.update(subscription.customer as string, {
-    email
+  let promoCodeData;
+  if (coupon) {
+    promoCodeData = await getCouponDetails(coupon);
+  }
+
+  const newSubscription = await stripeClient.subscriptions.create({
+    metadata: {
+      ...(subscription.metadata || {})
```
motechFR, line 31: If we don't initially have a subscription, wouldn't this fail?

valentinludu: You will always have an initial subscription with metadata

lib/subscription/createCryptoSubscription.ts
valentinludu: Yes.

Unfortunately I can't update the subscription to make loop work. I need a new subscription.

valentinludu: After reviewing more closely I noticed this doesn't work that good when you close the drawer and start again the process because the old subscription was deleted.

valentinludu: I updated this so we delete only overdue subscriptions that are created only if a user will retry to checkout and pay with loop.

PR #2444, opened 2023-07-10
Transfer lit shared modal in lib

Devorein:
### <samp>🤖 Generated by Copilot at 5ecc0d7</samp>

This pull request adds the `lit-protocol-modal` package to the project, which is a user interface component that allows users to create and share access control conditions based on blockchain tokens or NFTs. It updates the `TokenGates` component to use the new package instead of the old one, and adds the `.eslintignore` file to avoid linting errors for the package files. It also adds several files that define the configuration, properties, and components for the different types of access control conditions that are supported by the package for the Ethereum and Solana networks. It adds some reusable components and styles for the package as well.

### WHY
1. Transfer lit shared modal inside `lib/lit-protocol-modal` directory
2. Add mantle network


mattcasey (approved):

.eslintignore
```diff
@@ -1,2 +1,2 @@
-
+lib/lit-protocol-modal/**/*
```
mattcasey, line 1: since we're taking over the code, we should format it too

tsconfig.json
```diff
@@ -119,6 +119,7 @@
   "exclude": [
     "node_modules",
     "**/focalboard/src/**/*.spec.ts",
+    "**/lib/lit-protocol-modal/**/*.js",
```
mattcasey, line 122: were there too many TS errors?

Devorein: Yeah there were quite a few, but I think its best to add types right away.

Devorein:
On second thoughts, in order to make all the type checks pass, we need to update the logic in certain cases. And totally overwhelmed by the sheer number of changes we needed to make to satisfy eslint and typescript. I think it's best to do it on a separate PR.

lib/lit-protocol-modal/chainComponents/chainConfig.js
```diff
@@ -0,0 +1,178 @@
+import { ethereumTypesConfig } from "./ethereum/ethereumTypesConfig";
+import { solanaTypesConfig } from "./solana/solanaTypesConfig";
+
+import { utils } from 'ethers';
+
+const chainConfig = {
```
motechFR, line 6: Suggestion; Let's refactor this file as follows

- Import RPC connectors
- Define an array of supported connectors in the modal using EIP-155 shortnames
https://github.com/ethereum-lists/chains/blob/master/_data/chains/eip155-137.json


Map the RPCs to the desired Lit Config

motechFR (approved):

PR #2446, opened 2023-07-10
Sanitise data from API and add tests for logging

motechFR:
### <samp>🤖 Generated by Copilot at f2fe3c4</samp>

This pull request introduces a new feature for creating and using api page keys, which are api keys that are linked to a specific page and require a paid space. It adds a new helper function `apiPageKeyHandler` for creating api handlers that use api page keys, and a new middleware function `requireApiPageKey` for validating and authorizing api page key requests. It also refactors the existing `addToDatabase` endpoint to use middleware functions, and adds test files for the new and existing functions.

### WHY
Fix mixpanel datalogging for calls to add response card

Prevent using the inbound database-specific webhook endpoint if user has no subscription


lib/middleware/requireApiPageKey.ts
```diff
@@ -0,0 +1,52 @@
...
+import type { ApiPageKey } from '@charmverse/core/prisma-client';
+import { prisma } from '@charmverse/core/prisma-client';
+import type { NextApiRequest } from 'next';
+import type { NextHandler } from 'next-connect';
+
+import { getPermissionsClient } from 'lib/permissions/api';
+
+import { InvalidApiKeyError } from './errors';
+
+declare module 'http' {
```
mattcasey, line 11: probably should not update the global http namespace, use a type assertion instead?

pages/api/v1/webhooks/addToDatabase/[apiPageKey]/index.ts
```diff
@@ -43,36 +42,16 @@ handler.use(requireKeys(['key'], 'query')).post(createFormResponse);
 //  *                $ref: '#/components/schemas/Page'
 //  */
 export async function createFormResponse(req: NextApiRequest, res: NextApiResponse) {
```
Comment (mattcasey) [inline pages/api/v1/webhooks/addToDatabase/[apiPageKey]/index.ts:44]: You could use a custom NextApiRequest that has the apiPageKey type attached

pages/api/v1/webhooks/addToDatabase/[apiPageKey]/index.ts
```diff
@@ -1,19 +1,18 @@
-import { prisma } from '@charmverse/core/prisma-client';
 import type { NextApiRequest, NextApiResponse } from 'next';
```
Comment (mattcasey) [inline pages/api/v1/webhooks/addToDatabase/[apiPageKey]/index.ts:1]: this is not the webhook we're exposing to users right?

Comment (mattcasey) [reply pages/api/v1/webhooks/addToDatabase/[apiPageKey]/index.ts:1]: i'm assuming its just part of cleanup... :)

lib/middleware/requireApiPageKey.ts
```diff
@@ -0,0 +1,52 @@
...
+   */
+  interface IncomingMessage {
+    apiPageKey: ApiPageKey & { page: { spaceId: string } };
+  }
+}
+
+/**
+ * Used for databases receiving data from external services.
+ */
+export async function requireApiPageKey(req: NextApiRequest, res: NextApiRequest, next: NextHandler) {
```
mattcasey, line 24: this file could export a custom NextApiRequest type that has the apiKey added to it

mattcasey (changes requested):
left comments

lib/middleware/requireApiPageKey.ts
motechFR: We've been using this pattern for alot of middlewares so far, hence why I took this approach.

That said, I've updated this to the suggested pattern

PR #2449, opened 2023-07-10
Resizable columns!!!!

mattcasey:
Here's an over view of the column feature:

Uses the existing Prosemirror node types: columnLayout, columnBlock

Plugins:
- Row: Creates a node view that is responsible for instantiating an instance of ColumnResizer
- Column: Creates a node view to prevent re-rendering by Prosemirror, since it is controlled by ColumnResizer. It also keeps the HTML attributes updated so that the column sizes are correct when we re-init the CR instance in the Row plugin. Both Row and Column use ignoreMutation() to hide the changes to style attribute that CR applies.
- BarResizer: Injects HTML as 'Prosemirror decorations' which is used by ColumnResizer. Main problem w/this approach is that Prosemirror re-renders the decorations whenever the document changes. Re-rendering breaks the binding with the ColumnResizer instance, so I have to re-init the resizer instance in the Row plugin when the columns are resized (I think this is also why multi-player currently works). This is almost invisible to the user, but an alternatie would be to find a way to create unique keys for each column that we can pass to the widgets. However I think that would require generating and storing unique ids in the column nodes.


Devorein:
Some small issues I noticed
1. Column resized inside column resets the parent column width
2. Two column resize bars are shown for nested columns
3. The scrollbar for tables doesn't allow scrolling across the full table
4. I Tried to update the emoji of a callout inside a column, and it threw the following error
image
5. A resizer bar seems to be appearing on the left side (even though there are only 2 columns)
image

Devorein (approved):

PR #2455, opened 2023-07-13
Mantle testnet bounty payment + login support

Devorein:
…eat/mantle-login-bounty### WHAT

### <samp>🤖 Generated by Copilot at bd55481</samp>

Added support for Mantle testnet and updated ZKSYNC icon in `./connectors/index.ts`. Moved unsupported chains to `unsupportedChains` in `./lib/lit-protocol-modal/chainConfig.ts` to improve UX.

### WHY
[Card](https://app.charmverse.io/charmverse/page-08760149565574182)


mattcasey (approved):
approved but remember to add the Mantle default payment method. and once we add mainnet we should probably do a migration so existing spaces will see it

PR #2496, opened 2023-07-19
better redirects after deleting a space

mattcasey:
### <samp>🤖 Generated by Copilot at 210680a</samp>

This pull request enhances the space creation and deletion features by refactoring the redirection logic and fixing a rendering issue. It affects the `SpaceSettings` and `createSpace` components.

### WHY
Currently, when you delete a space we jsut send you to the token gate trying to join that space... i think we lost a redirect at some point.
Also, allow someone to go to createSpace and not be redirected to someplace else

### DIFF
--- a/__e2e__/po/settings/spaceProfileSettings.po.ts
+++ b/__e2e__/po/settings/spaceProfileSettings.po.ts
@@ -11,6 +11,10 @@ export class SpaceProfileSettings extends SettingsModal {
 
   readonly submitSpaceUpdateButton: Locator;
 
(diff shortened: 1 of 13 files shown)


Devorein:
If I have a single space and I delete it, it still takes me to that space's token gate page. It should take the user to `/createSpace` route
image

Devorein (changes requested):
If I have a single space and I delete it, it still takes me to that space's token gate page. It should take the user to /createSpace route.

PR #2517, opened 2023-07-26
Send receipt to customer after payment succeeded

valentinludu:
### <samp>🤖 Generated by Copilot at 0aaddc0</samp>

Added email and URL parameters to Stripe checkout session request in `CheckoutForm.tsx`. This improves the user experience and payment tracking for subscription payments.

### WHY
Send receipt to customer after payment succeeded
Add return url as a fail safe if user has 3DS enabled

### DIFF
--- a/components/settings/subscription/CheckoutForm.tsx
+++ b/components/settings/subscription/CheckoutForm.tsx
@@ -209,7 +209,9 @@ export function CheckoutForm({
             name: space.name,
             email: emailField
           }
-        }
+        },
+        receipt_email: emailField,
+        return_url: `${window?.location.origin}?settingTab=subscription`
       });
 
       if (confirmCardError) {


components/settings/subscription/CheckoutForm.tsx
```diff
@@ -209,7 +209,9 @@ export function CheckoutForm({
             name: space.name,
             email: emailField
           }
-        }
+        },
+        receipt_email: emailField,
+        return_url: `${window?.location.origin}?settingTab=subscription`
```
msadura, line 214: out of curiosity - does origin of `return_url` matters? I am asking because on custom domain it will not be app.charmverse.io and at least for authentication providers it was a problem

msadura (approved):

components/settings/subscription/CheckoutForm.tsx
valentinludu: It matters for some rare cases when the user can be redirected to another page to confirm the payment and then come back to our app. 
But it's fine to return the user to: https://custom.charmverse.io?settingsTab=subscription

PR #2518, opened 2023-07-26
Fix update subscription failure

valentinludu:
### WHAT
copilot:summary

### WHY
There are some cases where the old trial subscription was not deleted and we we're not retrieving the latest subscription.

### DIFF
--- a/lib/subscription/updateProSubscription.ts
+++ b/lib/subscription/updateProSubscription.ts
@@ -1,10 +1,9 @@
-import { prisma } from '@charmverse/core/prisma-client';
 import type Stripe from 'stripe';
 
 import { InvalidStateError } from 'lib/middleware';
 
 import type { SubscriptionStatusType } from './constants';
-import type { SpaceSubscriptionWithStripeData } from './getActiveSpaceSubscription';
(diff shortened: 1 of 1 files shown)


lib/subscription/updateProSubscription.ts
```diff
@@ -21,12 +20,7 @@ export async function updateProSubscription({
 }) {
   const { billingEmail, status } = payload;
 
-  const spaceSubscription = await prisma.stripeSubscription.findFirst({
-    where: {
-      spaceId,
-      deletedAt: null
-    }
-  });
+  const spaceSubscription = await getActiveSpaceSubscription({ spaceId });
```
msadura, line 23: is it just a refactor, or `getActiveSpaceSubscription` has some additional  needed logic?

msadura (approved):

lib/subscription/updateProSubscription.ts
valentinludu: 1. It standardises the fetching of active subscription so we have the same getter everywhere
2. It takes the latest subscription created in case there are more then 1.

PR #2519, opened 2023-07-26
Compute proposal permissions for vote

msadura:
### <samp>🤖 Generated by Copilot at fe6e1b0</samp>

Modified `updateVote` function to support proposal pages. Added `page` field to query and checked proposal permissions.

### WHY
This fixes both paid and free plan permissions

### DIFF
--- a/pages/api/votes/[id]/index.ts
+++ b/pages/api/votes/[id]/index.ts
@@ -51,7 +51,11 @@ async function updateVote(req: NextApiRequest, res: NextApiResponse<Vote | { err
       createdBy: true,
       pageId: true,
       postId: true,
-      page: true
+      page: {
+        select: {
+          proposalId: true
+        }
+      }
     }
   });
 
--- a/pages/api/votes/[id]/index.ts
+++ b/pages/api/votes/[id]/index.ts
@@ -50,7 +50,8 @@ async function updateVote(req: NextApiRequest, res: NextApiResponse<Vote | { err
       spaceId: true,
       createdBy: true,
       pageId: true,
-      postId: true
+      postId: true,
+      page: true
     }
   });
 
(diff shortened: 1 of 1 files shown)


pages/api/votes/[id]/index.ts
```diff
@@ -50,22 +50,34 @@ async function updateVote(req: NextApiRequest, res: NextApiResponse<Vote | { err
       spaceId: true,
       createdBy: true,
       pageId: true,
-      postId: true
+      postId: true,
+      page: true
```
Comment (motechFR) [inline pages/api/votes/[id]/index.ts:54]: Can we update the page statement to only return the proposalID? otherwise we will be loading the full page content

```typescript
page: {
  select: {
    proposalId: true
  }
}
```

pages/api/votes/[id]/index.ts
```diff
@@ -50,22 +50,34 @@ async function updateVote(req: NextApiRequest, res: NextApiResponse<Vote | { err
...
-    const pagePermissions = await req.basePermissionsClient.pages.computePagePermissions({
-      userId,
-      resourceId: vote.pageId
-    });
-
-    if (!pagePermissions.create_poll) {
-      throw new UnauthorisedActionError('You do not have permissions to update the vote.');
+  if (vote.pageId && vote.page) {
+    if (vote.page.proposalId) {
+      const proposalPermissiong = await req.basePermissionsClient.proposals.computeProposalPermissions({
```
Comment (motechFR) [inline pages/api/votes/[id]/index.ts:64]: proposalPermission**s**

pages/api/votes/[id]/index.ts
Comment (msadura) [reply pages/api/votes/[id]/index.ts]: good point

pages/api/votes/[id]/index.ts
```diff
@@ -50,22 +50,34 @@ async function updateVote(req: NextApiRequest, res: NextApiResponse<Vote | { err
...
 
-  if (vote.pageId) {
-    const pagePermissions = await req.basePermissionsClient.pages.computePagePermissions({
-      userId,
-      resourceId: vote.pageId
-    });
-
-    if (!pagePermissions.create_poll) {
-      throw new UnauthorisedActionError('You do not have permissions to update the vote.');
+  if (vote.pageId && vote.page) {
```
Comment (motechFR) [inline pages/api/votes/[id]/index.ts:66]: If the page linked to the vote does not exist, should we allow any updates? With current setup, we'd bypass the permission checks if vote page is deleted

If we don't want this behaviour, we can remove the vote.page check

motechFR (approved):

pages/api/votes/[id]/index.ts
Comment (msadura) [reply pages/api/votes/[id]/index.ts:66]: good question. I believe right now it will be fine if we `throw new DataNotFoundError` when there is not page related to the vote. This might change if in the future we will allow some sort of votes not attached to page :P But for now I think I will simplify this

PR #2522, opened 2023-07-26
Improved Notion Import UX

Devorein:
### <samp>🤖 Generated by Copilot at 4394c67</samp>

Added a feature to import boards from Notion and show the import status and errors to the user. Refactored the code to use web socket broadcasting and relay module to communicate the notion import events and results. Fixed some types and removed some unused imports.

### WHY
[Card](https://app.charmverse.io/charmverse/page-49366552253645923?viewId=50d940bb-fdbd-40ba-9816-616d6138a663&cardId=aa14ccea-2a10-4e71-9db4-179f6bff417f)

### DIFF
--- a/hooks/useSnackbar.tsx
+++ b/hooks/useSnackbar.tsx
@@ -14,9 +14,13 @@ type IContext = {
   setActions: Dispatch<SetStateAction<ReactNode[]>>;
   showMessage: (msg: string, newSeverity?: AlertColor) => void;
   handleClose: SnackbarProps['onClose'];
+  autoHideDuration?: number | null;
+  setAutoHideDuration?: Dispatch<SetStateAction<number | null>>;
+  resetState: () => void;
 };
 
 export const SnackbarContext = createContext<Readonly<IContext>>({
+  resetState: () => {},
   handleClose: () => {},
   isOpen: false,
   actions: [],
@@ -27,7 +31,9 @@ export const SnackbarContext = createContext<Readonly<IContext>>({
   setSeverity: () => {},
   severity: 'info',
   showMessage: () => {},
-  setActions: () => {}
+  setActions: () => {},
+  autoHideDuration: 5000,
+  setAutoHideDuration: () => {}
 });
 
 export function SnackbarProvider({ children }: { children: ReactNode }) {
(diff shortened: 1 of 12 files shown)


hooks/useSnackbar.tsx
```diff
@@ -27,7 +29,9 @@ export const SnackbarContext = createContext<Readonly<IContext>>({
   setSeverity: () => {},
   severity: 'info',
   showMessage: () => {},
-  setActions: () => {}
+  setActions: () => {},
+  autoHideDuration: 5000,
```
mattcasey, line 33: a better default would be null, since we already have 5000 as the default written elsewhere

hooks/useSnackbar.tsx
```diff
@@ -36,20 +40,32 @@ export function SnackbarProvider({ children }: { children: ReactNode }) {
...
     setIsOpen(true);
   };
 
+  const resetState = () => {
+    setIsOpen(false);
+    setActions([]);
+    setOrigin({ vertical: 'bottom', horizontal: 'left' });
+    setSeverity('info');
+    setMessage(null);
+    setAutoHideDuration(5000);
```
mattcasey, line 55: another reason to use null as default.. 5000 is repeated "3" times. that's a lot, for a default.

hooks/useSnackbar.tsx
mattcasey: @Devorein i'm looking at the code and i can't see how we are using setAutoHideDuration or even overriding autoHideDuration ... was this something you added but decided not to use? the logic to manage the state's gotten very complicated. If you can't revert, I think it's worth talking thru how we might simplify it

mattcasey (changes requested):
We'd like the 'loading' message to stay up until everything is complete. I only tested in dev, maybe it was the hot-reloading that closed it?

hooks/useSnackbar.tsx
Devorein: sorry I forgot to mention in comments, `autoHideDuration` is used to set how long in milliseconds the snackbar will be shown. It's set `5000` as default. We set it once when we open the snackbar in `useNotionImport` and reset it once it closes.

PR #2525, opened 2023-07-26
Add support for comments in multi-player

mattcasey:
### <samp>🤖 Generated by Copilot at aac4035</samp>

Added comment update handling for `BangleEditor` and `FidusEditor` components. Fixed a syntax error in `fiduseditor.ts`.


components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -184,6 +185,9 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
           onDocLoaded: () => {
             isLoadingRef.current = false;
           },
+          onCommentUpdate: () => {
+            mutate(`pages/${pageId}/threads`);
```
msadura, line 189: I believe that react editor should not build the key for mutation, f you change it in `useThreads` noone will realize to update it also here. I would create a separate `const getThreadsKey => (pageId) => `pages/${pageId}/threads` function and reuse it in both places.

In perfect scenario even `onCommentUpdate` should be passed in props to ReactEditor because we should not couple editor with swr imo

PR #2527, opened 2023-07-26
Feat/proposals datasource privacy

motechFR:
### <samp>🤖 Generated by Copilot at bd13eb8</samp>

This pull request adds features and logic to support different view source types and property types in the app, especially for the proposal data source. It also updates the test and deployment workflows to use feature branches of the permission API service and the `github_core` library. The changes affect various components in the `BoardEditor` module, such as `Table`, `Kanban`, `Gallery`, and `Calendar`, as well as the `PageHeader` and `DocumentPage` components. The changes also add new types and translations to the `shared` and `i18n` modules.

### WHY
Upgrade board UI to handle proposal properties

- Readonly proposal property, status and URL as well as title
- Auto-update category and status when proposal changes
- User cannot add cards manually
- Users only see the subset of cards where they have access to the proposal, and the cardPage itself

This PR also ensures that cards cannot be added to a board with proposal properties, so that the only datasource is the proposals themselves

How I'm handling custom proposal props:
- Copy over all categories as props, where the optionID matches the category ID
- Copy over all proposal statuses (except draft) as options, with the option ID being a random UUID, and the value being the enum key from prisma. In the UI, we then take the value and map it to our display value that we use in proposal props

Also, I noticed an issue where we add a cardId to a block even if the card isn't created in DB yet. I didn't want to mess with our mutator, so just added some cleanup in the blocks API

In the UI, we detect the propertyType in various locations in the board code, and make the property readonly.

We disable adding new cards if the current view is of type proposal. To avoid a whole class of bugs, I've restricted adding new datasources on board pages of type board or inline_board (which I believe we'll be applying across all source types in future)


lib/focalboard/board.ts
```diff
@@ -6,6 +6,9 @@ import type { PageContent } from 'lib/prosemirror/interfaces';
 
 import type { Card, CardPage } from './card';
 
+export const proposalPropertyTypesList = ['proposalUrl', 'proposalStatus', 'proposalCategory'] as const;
```
mattcasey, line 9: +1 just making them new field types

mattcasey (approved):

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -224,7 +224,7 @@ function DocumentPage({ page, refreshPage, savePage, insideModal, readOnly = fal
                   }
                   key={editorKey}
                   content={page.content as PageContent}
-                  readOnly={readOnly}
+                  readOnly={readOnly || !!page.syncWithPageId}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:227]: @motechFR readonly charmeditor state needs to be defined in the `editorMode`

Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:227]: This feels like a hack, but i think it was done because card properties are still editable, while content is not. The 'editMode' is 'editing', so i wonder if users will be confused

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -252,6 +252,7 @@ function DocumentPage({ page, refreshPage, savePage, insideModal, readOnly = fal
                     updatedAt={page.updatedAt.toString()}
                     readOnly={readOnly || !!enableSuggestingMode}
                     setPage={savePage}
+                    readonlyTitle={!!page.syncWithPageId}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:255]: readOnly should be camelcased

components/common/BoardEditor/focalboard/src/components/shared.ts
```diff
@@ -0,0 +1,9 @@
+export type ReadonlyTitleProp = {
```
mattcasey, line 1: this kind of type-sharing doesn't really add any value - it just tightly couples us to using the same interface, and you can't see the fields from the call-site.

components/common/BoardEditor/focalboard/src/components/kanban/kanban.tsx
```diff
@@ -270,6 +277,11 @@ function Kanban(props: Props) {
   };
   const menuTriggerProps = !props.readOnly ? { ...restBindings, onClick: addNewGroupHandler } : {};
 
+  // eslint-disable-next-line no-console
+  console.log({
```
mattcasey, line 281: console.log

lib/pages/createCardPage.ts
```diff
@@ -5,10 +5,11 @@ import { v4 } from 'uuid';
 
 import { createPage } from 'lib/pages/server/createPage';
 import { getPagePath } from 'lib/pages/utils';
+import type { BoardPropertyValue } from 'lib/public-api';
```
mattcasey, line 8: * we should never be using public api interfaces in our lib code. Should be the other way around

lib/pages/createCardsFromProposals.ts
```diff
@@ -19,6 +22,20 @@ export async function createCardsFromProposals({
   spaceId: string;
   userId: string;
 }) {
+  if (!stringUtils.isUUID(spaceId)) {
+    throw new InvalidInputError('Invalid spaceId');
+  } else if (!stringUtils.isUUID(boardId)) {
+    throw new InvalidInputError('Invalid boardId');
+  }
+  await prisma.space.findUniqueOrThrow({
```
mattcasey, line 30: thanks :)

lib/pages/setDatabaseProposalProperties.ts
mattcasey, line 1: this file feels like it belongs in lib/focalboard

PR #2532, opened 2023-07-27
Fix/create space duplicates

msadura:
### <samp>🤖 Generated by Copilot at 3284906</samp>

This pull request improves the appearance and functionality of various button components in the app. It refactors the `CreateSpaceForm` component to use the new `CharmedButton` component, which accepts a `disabled` prop and has no box-shadow. It also adds the `disabled` and `disableElevation` props to the `StyledButton`, `DeprecatedButton`, and `ButtonGroup` components, to allow for more control over the button state and style.

### WHY
Fix issue with creating duplicated spaces by fixing loading / disabled button states


components/common/Button.tsx
```diff
@@ -29,10 +29,10 @@ export type InputProps<C extends ElementType = ElementType> = ButtonProps &
   };
 
 export const CharmedButton = forwardRef<HTMLButtonElement, InputProps<ElementType>>((_props, ref) => {
-  const { children, loading, loadingMessage, disabledTooltip, ...props } = _props;
+  const { children, loading, loadingMessage, disabledTooltip, disabled, ...props } = _props;
 
   const buttonComponent = (
-    <StyledButton ref={ref} disabled={loading} {...props}>
+    <StyledButton ref={ref} {...props} disabled={loading || disabled} disableElevation>
```
mattcasey, line 35: +1 This is good for now... I recall originally having the button disable itself if it's loading. Altho i could see maybe the UI would be slightly different, like disabled should be faded out, loading should just be unclickable. and we can probably put 'disableElevation' in the default theme props

mattcasey (approved):
The more responsive button is really nice!

components/common/__tests__/Button.spec.tsx
```diff
@@ -0,0 +1,59 @@
+import { render } from '@testing-library/react';
+
+import { Button } from 'components/common/Button';
+import { mockCurrentSpaceContext } from 'testing/mocks/useCurrentSpace';
+
+jest.mock('hooks/useCurrentSpace', () => ({
+  useCurrentSpace: jest.fn(() => mockCurrentSpaceContext())
+}));
+
+describe('Atomic Button component', () => {
```
mattcasey, line 10: :100:

components/votes/components/NewProposalButton.tsx
```diff
@@ -97,7 +97,7 @@ export function NewProposalButton({ mutateProposals }: { mutateProposals: KeyedM
     <>
       <Tooltip title={!canCreateProposal ? 'You do not have the permission to create a proposal.' : ''}>
         <Box>
-          <ButtonGroup variant='contained' ref={buttonRef}>
+          <ButtonGroup variant='contained' ref={buttonRef} disableElevation>
```
mattcasey, line 100: we could put disableElevation on the default props in the theme setting too... i don't think we ever intentionally use elevation except for maybe the signup screen

msadura: oh I like it! Forgot about default props in theme. I will add it

PR #2533, opened 2023-07-27
add support for nfts via Ankr (support Mantle)

mattcasey:
Adds support for Fantom, Avalanche, Binance, and Mantle via Ankr APIs.

Example contract address: 0x22fa785f3bd88b54ac076ad253b1b68332c27d58, chain: 5000. token: 1

image


Devorein (approved):

msadura:
Looks like image in picker excels preview area and does not fit the modal
Adidtionally right now when you type id of token it fetches data on each keypress -would be nice to debounce this as we most likely pay for requests 

Screenshot 2023-08-01 at 12 29 35

components/common/CharmEditor/components/nft/NFTNodeView.tsx
```diff
@@ -37,7 +36,7 @@ const StyledCard = styled(Card)`
   }
 `;
 
-const blockchains = [1, 10, 137, 42161];
+const blockchains = [1, 10, 56, 137, 250, 5000, 42161, 43114];
```
msadura, line 39: would be really nice to have enum for those like 

```
enum BlockchainId {
eth = 1,
op = 10,
mantle = 5000
...
}
```

because right now it is hard to remember each chain id ;p

components/common/CharmEditor/components/nft/utils.ts
```diff
@@ -50,14 +52,22 @@ export function getNFTUrl({
 }): string | null {
   let link: null | string = null;
   switch (chain) {
-    case 42161:
-      link = `https://stratosnft.io/assets/${contract}/${token}`;
-      break;
     case 1:
```
msadura, line 55: once again (might not be related to this task so we might want separate card - with more chains it is harder to understand such code with just chain id as a  number

components/common/CharmEditor/components/nft/utils.ts
```diff
@@ -50,14 +52,22 @@ export function getNFTUrl({
...
     case 137:
+    case 43114:
       link = `https://opensea.io/assets/${openseaPathsByChain[chain]}/${contract}/${token}`;
       break;
+    // fantom chain
+    case 250:
+      link = `https://artion.io/explore/${contract}/${token}`;
+      break;
+    // Aribitrum One
+    case 42161:
```
msadura, line 67: do we need this? opensea supports arbitrum i.e `https://opensea.io/assets/arbitrum/0x5d482d501b369f5ba034dec5c5fb7a50d2d6ca20/84914531375613224443884725751445934403308641051756060792297532171538521196491`

lib/blockchain/getNFTs.ts
```diff
@@ -27,62 +37,81 @@ export type NFTData = {
...
-}) {
-  const addresses = wallets.map((w) => w.address);
-  const nfts = await alchemyApi.getNFTs(addresses, chainId);
-  const mappedNfts = nfts.map((nft) => {
-    const walletId = wallets.find((wallet) => wallet.address === nft.walletAddress)?.id ?? null;
-    return mapNftFromAlchemy(nft, walletId, chainId);
-  });
-  return mappedNfts.filter(isTruthy);
+export async function getNFTs({ wallets }: { wallets: UserWallet[] }) {
+  const [alchemyNFts, mantleNFTs] = await Promise.all([
```
msadura, line 41: I would call it `[alchemyNFts, ankrNFTs]` in case we will have more chains supported by ankr

scripts/testAnkr.ts
```diff
@@ -0,0 +1,11 @@
+import { getNFT } from 'lib/blockchain/provider/ankr';
```
msadura, line 1: probably to be removed, but up to you

components/common/CharmEditor/components/nft/NFTNodeView.tsx
mattcasey: i like this idea but its a big change for me, would want to try it someplace else

components/common/CharmEditor/components/nft/utils.ts
mattcasey: i was preserving the original logic, but it looks like https://stratosnft.io is winding down. so i'm going to update to opensea

PR #2534, opened 2023-07-28
alerts for failed deployment

yiwork:
### <samp>🤖 Generated by Copilot at 09740d5</samp>

Add a postdeploy hook script to send SNS notifications on deployment failures. The script reads environment variables from a file specified by `EBSTALK_ENV_FILE` in the `.env` file and uses them to publish a message to an SNS topic.

### WHY
Using platform postdeployment hook to send alert if deployment failed.


mattcasey (approved):

.platform/hooks/postdeploy/deployment_failure_sns.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash -e
```
mattcasey, line 1: i'm guessing this is a test? but neat!

PR #2539, opened 2023-07-31
Add dev webhook test button

Devorein:
### <samp>🤖 Generated by Copilot at 2bb14b3</samp>

This pull request adds the feature to test the webhook url for a space in the app settings. It modifies the `ApiSettings` component, the `charmClient` module, and the `test-webhook` API route to implement the test request, response, and feedback logic. It also uses the `webhookPublisher` module to generate and verify the webhook signatures.

### WHY
[Card](https://app.charmverse.io/charmverse/page-5385488274767563)


components/settings/api/ApiSettings.tsx
```diff
@@ -83,6 +84,31 @@ export function ApiSettings({ space }: { space: Space }) {
...
+      setIsTestingWebhook(true);
+      const { status } = await charmClient.testSpaceWebhook({
+        spaceId: space.id,
+        webhookUrl
+      });
+
+      if (status === 200) {
+        showMessage('Webhook successfully sent', 'success');
+      } else {
+        showMessage('Webhook failed to send', 'error');
```
msadura, line 102: maybe it will be helpful for devs if you add error status code?

components/settings/api/ApiSettings.tsx
```diff
@@ -123,17 +149,25 @@ export function ApiSettings({ space }: { space: Space }) {
...
-                size='small'
-                disabled={!isAdmin}
-                data-test='webhook-url-input'
-                fullWidth
-                error={!!errors.webhookUrl?.message}
-                helperText={errors.webhookUrl?.message}
-                placeholder='https://your-api.com/webhook'
-              />
+              <Stack flexDirection='row' gap={1}>
+                <TextField
```
msadura, line 153: please add url validation to this field - right now I can put random string and test it

components/settings/api/ApiSettings.tsx
```diff
@@ -123,17 +149,25 @@ export function ApiSettings({ space }: { space: Space }) {
...
+                  type='text'
+                  size='small'
+                  disabled={!isAdmin}
+                  data-test='webhook-url-input'
+                  fullWidth
+                  error={!!errors.webhookUrl?.message}
+                  helperText={errors.webhookUrl?.message}
+                  placeholder='https://your-api.com/webhook'
+                />
+                <Button
```
msadura, line 164: please use `import { Button } from 'components/common/Button';`

pages/api/spaces/[id]/test-webhook.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const { webhookUrl } = req.body as { webhookUrl: string };
+  const signingSecret = createSigningSecret();
+  const secret = Buffer.from(signingSecret, 'hex');
+  const webhookData = {
+    event: 'test',
+    spaceId,
+    createdAt: new Date().toISOString()
+  };
+  const signedJWT = await signJwt('webhook', webhookData, secret);
```
Comment (motechFR) [inline pages/api/spaces/[id]/test-webhook.ts:24]: How will the user verify the message from the webhook? I'd suggest providing some content on how they can go about this

I'd also suggest extracting the signing into its own lib method, and having a unit test with 2 cases
- Checking we can verify the signature
- Checking a wrong signature fails

components/settings/api/ApiSettings.tsx
motechFR: +1 on error code. Maybe show the whole response error body in some tab that shows?

Devorein: Initially, I wanted to show both the request payload and response but since the response error body only shows the status, I just went with snackbar

motechFR:
Some UX suggestions
- Reuse the space signing secret for test webhooks so users can work on verification flows

- Show the secret, but hidden with a button to toggle it.

- We might also want a way to regenerate the secret. If so, should contain a modal warning the user

pages/api/spaces/[id]/test-webhook.ts
Comment (mattcasey) [reply pages/api/spaces/[id]/test-webhook.ts:24]: @motechFR @Devorein this is all out of scope right now. The goal was just to give them a test button. No documentation.

PR #2546, opened 2023-08-01
Analytics/block quota

motechFR:
### <samp>🤖 Generated by Copilot at 031acaf</samp>

This pull request implements the block quota feature for spaces, which limits the number of blocks that a space can have based on its Stripe subscription. It adds a new metric to track the block quota in Mixpanel, updates the logic to set or update the block quota when creating or updating subscriptions, and adds a script to import the block quotas from Stripe to the database and Mixpanel. It also modifies the GitHub action to use a specific branch of the core repository that contains related changes.

### WHY
Adding the script and data for block quota

### DIFF
--- a/pages/api/v1/webhooks/stripe/index.ts
+++ b/pages/api/v1/webhooks/stripe/index.ts
@@ -100,7 +100,7 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse): 
           deletedAt: null
         };
 
-        const blockQuota = stripeSubscription.items.data[0]?.quantity as number;
+        const blockQuota = stripeSubscription.items.data[0]?.quantity;
 
         await prisma.$transaction([
           prisma.stripeSubscription.upsert({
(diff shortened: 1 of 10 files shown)


pages/api/v1/webhooks/stripe/index.ts
```diff
@@ -117,13 +119,11 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse):
               id: space.id
             },
             data: {
-              paidTier: paidTier === 'enterprise' ? 'enterprise' : 'community'
+              paidTier: paidTier === 'enterprise' ? 'enterprise' : 'community',
+              blockQuota: blockQuota && space.blockQuota !== blockQuota ? blockQuota : undefined
```
valentinludu, line 123: Shouldn't this be?
`blockQuota: blockQuota && space.blockQuota !== blockQuota ? blockQuota :  (space.blockQuota || 0)`

I think if the blockQuota is equal to space.blockQuota you want space.blockQuota instead of undefined.

pages/api/v1/webhooks/stripe/index.ts
```diff
@@ -100,6 +100,8 @@ export async function stripePayment(req: NextApiRequest, res: NextApiResponse):
           deletedAt: null
         };
 
+        const blockQuota = stripeSubscription.items.data[0]?.quantity as number;
```
valentinludu, line 103: You can add `|| 0` instead of asserting.

PR #2547, opened 2023-08-01
Resize image before upload to s3

Devorein:
### <samp>🤖 Generated by Copilot at 7a30441</samp>

This pull request adds a feature to resize and optimize images before uploading them to S3 using the `/api/image/resize` endpoint and the `useS3UploadInput` hook. It also updates some components and hooks that use the `useS3UploadInput` hook to pass an options object instead of a single callback function. It also adds and updates some packages and modules related to image processing and file handling.

### WHY
[Card](https://app.charmverse.io/charmverse/page-7735843749868132)


pages/api/image/resize.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const form = formidable({
+    uploadDir,
+    maxFileSize: DEFAULT_MAX_FILE_SIZE_MB * 1024 * 1024,
+    filter: (part) => {
+      return part.name === FORM_DATA_FILE_PART_NAME && (part.mimetype?.includes('image') || false);
+    }
+  });
+  const [, files] = await form.parse(req);
+  if (files && files[FORM_DATA_FILE_PART_NAME]) {
```
msadura, line 32: honestly I've got the feeling this solution introduces far more complexity than it should.

1) Do you need additional library to upload a file? If that is the case, looks like we have something called `form-data` already in project deps, did you check if that can be used?
2) why do you save teh file to the disk instead using buffer right away (sharp takes buffer anyway from what I understand)
3) and finally the whole flow looks weird to me - send file to our server, send it back and then upload it to s3. Would it work like this?
upload file to our server -> resize -> upload to s3 -> send back url to the user?

msadura (changes requested):

components/settings/space/components/LargeAvatar.tsx
```diff
@@ -94,7 +94,9 @@ export default function LargeAvatar(props: LargeAvatarProps) {
     updateAvatar?.(userAvatar);
   }
 
-  const { inputRef, openFilePicker, onFileChange } = useS3UploadInput(updateImageAvatar);
+  const { inputRef, openFilePicker, onFileChange } = useS3UploadInput({
+    onFileUpload: updateImageAvatar
```
msadura, line 98: would be nice to have avatars optimized as well

pages/api/image/resize.ts
Devorein: `form-data` unfortunately can't handle that use case. `According to its npm package page, it can be used to submit forms and file uploads to other web applications.` I saw no example to read incoming multi-part/form-data and parse it to get the files. As for the other two thank you for the suggestions and I updated the code accordingly 🚀

mattcasey (approved):

PR #2549, opened 2023-08-01
Fix app throwing an error on parse

motechFR:
### <samp>🤖 Generated by Copilot at 5f83f88</samp>

Refactored `checkIsContentEmpty` function in `lib/prosemirror/checkIsContentEmpty.ts` to handle null or undefined content and simplify logic. This improves the function's reliability and clarity for checking page or block content.

### WHY
Fix error in the API


mattcasey (approved):

mattcasey (changes requested):
actually, i think the issue may be upstream in addToDatabase

PR #2551, opened 2023-08-02
updates to gnosis safes to work with mantle

mattcasey:
### <samp>🤖 Generated by Copilot at f48dc15</samp>

This pull request enhances the integration of the Gnosis Safe app with the charmverse app, allowing users to import, display, and use their safes for multi-payment bounties. It fixes some bugs and updates some dependencies related to the Gnosis Safe transaction service. It also removes an unused file.


Devorein (approved):

mattcasey:
> Facing this issue on localhost image

Sorry about that! I was not using the 'checksum' address. We have a method `getAddress` which fixes the lower and uppercase letters so the address is a special checksum. The value we store in the db is lowercased to keep things simple

PR #2552, opened 2023-08-02
Feat/collabland connect

msadura:
### <samp>🤖 Generated by Copilot at 8210237</samp>

This pull request adds a new feature to the space settings, which is the ability to configure third-party integrations, such as collabland. It also improves the UI design and user experience of some components, such as the snapshot domain settings and the form labels. It also updates the types and functions related to the space API responses and queries, and adds unit tests for some of the new functions. The main files affected are `components/settings/space/SpaceSettings.tsx`, `components/settings/space/components/SpaceIntegrations.tsx`, `components/settings/space/components/ConnectCollabland.tsx`, `lib/public-api/createWorkspaceApi.ts`, `lib/public-api/interfaces.ts`, `lib/collabland/connectSpace.ts`, `lib/utilities/dataEncryption.ts`, `pages/api/spaces/[id]/collabland/code.ts`, and `pages/api/v1/collabland/connect.ts`.

### WHY
https://app.charmverse.io/charmverse/page-4384984192524961


mattcasey (approved):
did not test but code looks good

lib/utilities/dataEncryption.ts
```diff
@@ -0,0 +1,38 @@
+import crypto from 'crypto';
+
+import { authSecret } from 'config/constants';
```
mattcasey, line 3: i like that you just reused an existing secret

lib/utilities/dataEncryption.ts
```diff
@@ -0,0 +1,38 @@
...
+
+if (!authSecret) {
+  throw new Error('authSecret not defined');
```
mattcasey, line 6: files should not throw an error just on load

mattcasey (changes requested):
One change request: no file should blow up the stack just because it's imported. It might be imported but not actually needed, so instead you could throw an exception when a method gets called

PR #2558, opened 2023-08-04
Proposal properties UI

mattcasey:
### WHAT
Make all the properties on proposals look like the fields on focalboard (this also changes one field on bounties.. )

### WHY
image


components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx
```diff
@@ -22,7 +22,7 @@ export function MissingPagePermissions({ bountyPermissions, pagePermissions, tar
   // Compare each existing assignee against page permissions
   const assigneesMissingPermissions = visibleToSpace
     ? []
-    : ((bountyPermissions[target] ?? []).filter(
+    : ((bountyPermissions[target] ?? ([] as { id: string; group: string }[])).filter(
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx:25]: We can use the generic TargetPermissionGroup type here

components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx
```diff
@@ -22,7 +22,7 @@ export function MissingPagePermissions({ bountyPermissions, pagePermissions, tar
...
         (bountyPermissionAssignee) =>
           bountyPermissionAssignee.group !== 'public' &&
```
Comment (motechFR) [inline components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx:27]: This line could probably be deleted

components/common/BoardEditor/components/properties/TagSelect/TagSelect.tsx
```diff
@@ -150,31 +150,27 @@ export function SelectProperty({
...
-          }
+          showEmpty={displayType === 'details'}
         />
       </SelectPreviewContainer>
     );
   }
 
   return (
     <StyledSelect
+      canEditOptions={false} // TODO: allow editing options
```
motechFR, line 171: Notes:

ProposalCategory: false

ProposalTemplateSelect: false

Database:
- false if prop is proposal property
- false if page is readonly
- true if neither of above is false

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -97,11 +98,7 @@ function CardDetailProperty({
       }}
       className='octo-propertyrow'
     >
-      {readOnly && (
-        <div className='octo-propertyname octo-propertyname--readonly'>
-          <Button>{property.name}</Button>
-        </div>
-      )}
+      {readOnly && <PropertyLabel readOnly>{property.name}</PropertyLabel>}
```
motechFR, line 101: Yay for deleting focalboard CSS

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -104,7 +104,7 @@ function PropertyValueElement(props: Props) {
     propertyTemplate.type === 'proposalStatus'
   ) {
     propertyValueElement = (
-      <SelectProperty
+      <TagSelect
```
motechFR, line 107: Not a regression, but just noticed the proposal property name in databases is readonly in table view, but not the details view

lib/permissions/bounties/__tests__/setBountyPermissions.spec.ts
```diff
@@ -203,7 +203,7 @@ describe('setBountyPermissions', () => {
       bountyId: bounty.id
     });
 
-    expect(queryResult.reviewer.some((p) => p.group === 'space' && p.id === space.id)).toBe(true);
+    expect(queryResult.reviewer.some((p) => p.group === 'role' && p.id === space.id)).toBe(true);
```
motechFR, line 206: Makes sense, this is closer to actual use case

lib/permissions/bounties/interfaces.ts
```diff
@@ -15,16 +15,20 @@ export type BountyPermissionAssignment = {
...
-
 // Groups that can be assigned to various bounty actions
 export type BountyReviewer = Extract<AssignablePermissionGroupsWithPublic, 'role' | 'user'>;
 
 export type BountySubmitter = Extract<AssignablePermissionGroupsWithPublic, 'space' | 'role'>;
 
 export type BountyPermissionGroup = TargetPermissionGroup<BountyReviewer | BountySubmitter>;
 
+// The set of all permissions for an individual bounty
+export type BountyPermissions = {
```
motechFR, line 26: I noticed we have this interface in core under src/lib/testing/bounties.ts

That said, it's only used for a test and not exported. Wondering if we should have this there

In l:98 of this file in permissions API, we also use this type implicitly
src/lib/pagePermissions/isBountyPageEditableByApplicants.ts

motechFR (approved):

lib/permissions/bounties/interfaces.ts
mattcasey: the permissions are a little had to follow for bounties. i'm hoping we do some cleanup when it comes time for rewards... there seems to be too many permutations of the same type

motechFR: I agree. The model we have for proposals with explicit authors and reviewers as separate tables is a lot better.

That's also why I advocated not to spend too much time on permissions when we were doing permissions API.

We should rebuild this subsystem when we get round to rewards.

components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx
Comment (motechFR) [reply components/[pageId]/DocumentPage/components/BountyProperties/components/MissingPagePermissions.tsx:25]: We can pass an option to target permission group interface, for example in this case

TargetPermissionGroup<'user' | 'role'>

PR #2566, opened 2023-08-07
Subscription improvements and fixes

valentinludu:
### WHAT
copilot:summary

### WHY
Changes:
- show FAQ for multisig payment
- hide close modal button when subscription ended
- fix settings modal can't be closed after paying or switching to free because space and subscription were not refreshed
- fix pending payment button was not visible
- fix pending payment status in session storage was true for all spaces and disabled the pay button in other spaces


msadura (approved):
looks good

PR #2570, opened 2023-08-07
Editor content and sidebar sync for delete action

Devorein:
### <samp>🤖 Generated by Copilot at 6845662</samp>

This pull request implements real-time page deletion and archiving functionality using WebSockets and the document model. It modifies the `PageTreeItem` component, the WebSocket event handlers, and the database operations in `lib/websockets/documentEvents/documentEvents.ts` and `lib/websockets/spaceEvents.ts`. It also updates the types and messages in `lib/websockets/interfaces.ts` to support the new feature.

### DIFF
--- a/components/common/PageActions/FullPageActionsMenuButton.tsx
+++ b/components/common/PageActions/FullPageActionsMenuButton.tsx
@@ -12,7 +12,6 @@ import { DatabasePageActionList } from 'components/common/PageActions/components
 import type { PageActionMeta } from 'components/common/PageActions/components/DocumentPageActionList';
 import { DocumentPageActionList, documentTypes } from 'components/common/PageActions/components/DocumentPageActionList';
 import { ForumPostActionList } from 'components/common/PageActions/components/ForumPostActionList';
-import { useProposals } from 'components/proposals/hooks/useProposals';
 import { usePagePermissions } from 'hooks/usePagePermissions';
 import { usePostPermissions } from 'hooks/usePostPermissions';
 import type { PostWithVotes } from 'lib/forums/posts/interfaces';
--- a/components/common/PageLayout/components/TrashModal.tsx
+++ b/components/common/PageLayout/components/TrashModal.tsx
@@ -168,6 +168,7 @@ export default function TrashModal({ onClose, isOpen }: { onClose: () => void; i
     try {
       e.preventDefault();
       setIsMutating(true);
+      await restorePage(pageId);
       // Optimistically remove the restored page from modal
       setArchivedPages((_archivedPages) => {
         if (!_archivedPages) {
@@ -178,7 +179,6 @@ export default function TrashModal({ onClose, isOpen }: { onClose: () => void; i
         }
         return { ..._archivedPages };
       });
-      await restorePage(pageId);
     } catch (err: any) {
       showMessage(err.message ?? 'Failed to restore page', 'error');
     } finally {
--- a/hooks/usePages.tsx
+++ b/hooks/usePages.tsx
@@ -93,7 +93,7 @@ export function PagesProvider({ children }: { children: ReactNode }) {
   async function deletePage({ pageId, board }: { pageId: string; board?: Block }) {
     const page = pages[pageId];
     const totalNonArchivedPages = Object.values(pages).filter(
-      (p) => !p?.deletedAt && (p?.type === 'page' || p?.type === 'board')
+      (p) => !p?.deletedAt && (p?.type === 'page' || p?.type === 'board' || p?.type === 'card')
     ).length;
 
     const pageType = page?.type;
@@ -106,6 +106,19 @@ export function PagesProvider({ children }: { children: ReactNode }) {
           },
           type: 'page_deleted'
         });
+
+        if (board) {
+          await mutator.deleteBlock(
+            board,
+            'Delete board',
+            async () => {
+              // success
+            },
+            async () => {
+              // error
+            }
+          );
+        }
       } else {
         const { pageIds } = await charmClient.archivePage(page.id);
         let newPage: null | PageMeta = null;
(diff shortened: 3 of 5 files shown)


components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
```diff
@@ -399,21 +401,33 @@ function PageActionsMenu({ closeMenu, pageId, pagePath }: { closeMenu: () => voi
...
-      board,
-      pageId
-    });
-
-    if (!currentPage && newPage) {
-      // If we are in a page that doesn't exist, redirect user to the created page
-      router.push(`/${router.query.domain}/${newPage.id}`);
+
+    if (user) {
+      sendMessage({
```
mattcasey, line 412: I would keep the 'deletePage()' abstraction. we're gonna need to do this in other places - maybe inside of it, you'll want to exclude events internally for 'bounties' or 'proposals' since those don't get embedded as page content-  we can keep calling the webapp api for those

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
```diff
@@ -399,21 +401,33 @@ function PageActionsMenu({ closeMenu, pageId, pagePath }: { closeMenu: () => voi
...
+
+    // const board = boards.find((b) => b.id === page?.id);
+    // const newPage = await deletePage({
+    //   board,
+    //   pageId
+    // });
+
+    // if (!currentPage && newPage) {
+    //   // If we are in a page that doesn't exist, redirect user to the created page
+    //   router.push(`/${router.query.domain}/${newPage.id}`);
```
mattcasey, line 429: i would like to still test this case works. i think it is when you delete all the pages in your space

lib/websockets/documentEvents/documentEvents.ts
```diff
@@ -416,6 +420,26 @@ export class DocumentEventHandler {
...
         message.ds = message.ds.map(this.removeTooltipMarks);
 
+        // Go through the diffs and see if any of them are for deleting a page. If so, we need to archive the page and all of its children and then broadcast the change.
+        for (const ds of message.ds) {
+          if (ds.stepType === 'replace') {
+            const node = room.node.resolve(ds.from).nodeAfter?.toJSON() as PageContent;
+            if (node && node.type === 'page') {
+              const pageId = node.attrs?.id;
+
+              const modifiedChildPageIds = await modifyChildPages(pageId, session.user.id, 'archive');
```
mattcasey, line 430: Move the i/o outside of the loops. Also, I think we should wait until the document is updated before deleting anything

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
```diff
@@ -399,21 +401,33 @@ function PageActionsMenu({ closeMenu, pageId, pagePath }: { closeMenu: () => voi
...
-
-    if (!currentPage && newPage) {
-      // If we are in a page that doesn't exist, redirect user to the created page
-      router.push(`/${router.query.domain}/${newPage.id}`);
+
+    if (user) {
+      sendMessage({
+        payload: {
+          pageId,
+          userId: user.id
```
mattcasey, line 415: userId is already available in the socket service. we shouldn't need a userId from the frontend -it's not trustworthy anyway

lib/websockets/interfaces.ts
```diff
@@ -102,6 +102,14 @@ type SubscribeToWorkspace = {
   } & SocketAuthResponse;
 };
 
+type PageDeleted = {
+  type: 'page_deleted';
+  payload: {
+    pageId: string;
```
mattcasey, line 108: I think you can just use the `Resource` type here. Note we use 'id' and not 'pageId' for other events

lib/websockets/spaceEvents.ts
```diff
@@ -42,6 +45,73 @@ export class SpaceEventHandler {
...
+              });
+            }
+          } else if (documentRoom.participants.size !== 0) {
+            // Handle the case where the user is not present in the document but other users are present
+          } else {
+            // Handle the case when the document is not open by any user
+          }
+        }
+      } catch (error) {
+        log.error('Error deleting page', { error });
```
mattcasey, line 112: please include pageId/spaceId/userId too if avilable

lib/websockets/spaceEvents.ts
```diff
@@ -42,6 +45,73 @@ export class SpaceEventHandler {
...
+            id: message.payload.pageId
+          },
+          select: {
+            spaceId: true,
+            parentId: true
+          }
+        });
+
+        if (!pageWithSpaceId) {
+          this.sendError('Page not found');
```
mattcasey, line 61: this is not worth sending an error - just log a warning, maybe.

lib/websockets/spaceEvents.ts
```diff
@@ -42,6 +45,73 @@ export class SpaceEventHandler {
...
+          where: {
+            id: message.payload.pageId
+          },
+          select: {
+            spaceId: true,
+            parentId: true
+          }
+        });
+
+        if (!pageWithSpaceId) {
```
mattcasey, line 60: should just throw an error here

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
mattcasey: Correcting myself: I think it may only be available when a user 'subscribes'. But at that point ,i think we should just write the socketId to the instance of SpaceEventHandler so we can use it later.

Devorein: Nice, I was not sure whats the best approach to do that since we sent authToken to transfer the userId. Will do.

lib/websockets/interfaces.ts
Devorein: Yeah lets do that

components/common/PageLayout/components/PageNavigation/components/PageTreeItem.tsx
Devorein: Good point. Reverted back edits.

PR #2654, opened 2023-08-22
Charmverse proposal to lens post publication

Devorein:
### <samp>🤖 Generated by Copilot at 90e6cd3</samp>

This pull request adds a feature to automatically publish proposals to Lens, a decentralized social media platform, if the user opts in. It adds a new option to the account settings page, a new hook and component to handle the publication process, and a new utility function and interface to create and upload post publications. It also fixes some import paths and updates the user model and the package.json file. It uses the `lens-protocol` SDK and the `lensClient` object to interact with Lens.

### WHY
* Create a lens profile over here https://lens-do-it.vercel.app/
* Go to My Accounts in settings and then Publish
image
* Click on login to retrieve your lens profile connected with the wallet (there can be multiple, but only one is default, and that is used)
* Toggle on the Lens Publication
* Visit the proposal page and create a proposal. Move it to the feedback stage and automatically publish it to Lenster (https://testnet.lenster.xyz/). If you don't have an auto dispatcher on, it will ask you to sign the transaction.
* If you visit your profile it should look something like this
image

[Card](https://app.charmverse.io/charmverse/page-2825540469904342)


components/settings/LensProfileProvider.tsx
```diff
@@ -0,0 +1,174 @@
+import { log } from '@charmverse/core/log';
```
mattcasey, line 1: this file should be renamed to `useLensProfile.tsx` and put into a hooks folder

components/settings/account/components/LensPublication.tsx
```diff
@@ -0,0 +1,70 @@
...
+      });
+      updateUser({ ...user, autoLensPublish: newState });
+    }
+  }
+
+  return (
+    <Stack mt={2}>
+      <Legend
+        sx={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}
+        marginTop={(theme) => theme.spacing(4)}
```
mattcasey, line 31: you're already using sx prop, you should remove marginTop and use `mt: 4`

components/settings/account/components/LensPublication.tsx
```diff
@@ -0,0 +1,70 @@
...
+        id='multisig-section'
```
mattcasey, line 32: this id doesn't appear to be needed

components/settings/account/components/useNewLensPublication.tsx
```diff
@@ -0,0 +1,30 @@
+import useSWR from 'swr';
```
mattcasey, line 1: hooks should always be put in a 'hooks' folder

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+  amount: null,
+  referralFee: 0,
+  collectLimit: null,
+  timeLimit: false,
+  recipients: [],
+  followerOnlyCollect: false
+};
+
+const uploadToArweave = async (data: any): Promise<string> => {
+  const response = (await POST('https://metadata.lenster.xyz', data, {
```
mattcasey, line 21: I think you can remove 'as' and just use the generic response type: `await POST<string>()`

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+    headers: { 'Content-Type': 'application/json' },
+    // remove credentials to bypass CORS error
+    credentials: 'omit'
+  })) as string;
+  // Lenster response header content type is text/plain;charset=UTF-8, so we need to json parse it manually
+  const { id } = JSON.parse(response);
```
mattcasey, line 27: needs a try/catch

mattcasey: just saw your comment lol. carry on..

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+  return id;
+};
+
+const getUserLocale = () => {
```
mattcasey, line 31: this is a good method for lib/utilities/browser.ts

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+export async function createPostPublication({
+  lensProfile,
+  contentText,
+  proposalLink
+}: {
+  contentText: string;
+  proposalLink: string;
+  lensProfile: ProfileFragment;
+}) {
+  const collectModule = INITIAL_COLLECT_MODULE;
```
mattcasey, line 51: it seems unecessary to map this constant to a local variable

components/proposals/components/ProposalProperties/ProposalProperties.tsx
```diff
@@ -46,6 +47,7 @@ export type ProposalPropertiesInput = {
 };
 
 type ProposalPropertiesProps = {
+  proposalPage?: PageWithContent;
```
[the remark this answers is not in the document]
Devorein: This is necessary since we need to get access to the proposal page (json content) when creating the lens post. It is used in `components\[pageId]\DocumentPage\components\ProposalProperties.tsx:L84`

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+  const canUseRelay = lensProfile?.dispatcher?.canUseRelay;
+  const isSponsored = (lensProfile?.dispatcher as ProfileFragment['dispatcher'] & { sponsor: null | boolean })?.sponsor;
+
+  const isRevertCollectModule = collectModule.type === CollectModules.RevertCollectModule;
+  const useDataAvailability = !restricted && isRevertCollectModule;
+
+  const metadata: PublicationMetadataV2Input = {
+    version: '2.0.0',
+    metadata_id: uuid(),
+    content: `${contentText} \n\n CharmVerse Proposal Link: ${proposalLink}`,
```
mattcasey, line 64: I would change "CharmVerse Proposal Link" to just "CharmVerse Link" it would be cool if this could be an anchor link? it could just say "View on CharmVerse"

components/settings/LensProfileProvider.tsx
Devorein: was a bit confused about where to include it. Its been renamed to `useLensProfile` and placed inside `settings/account/hooks`

components/settings/account/components/LensPublication.tsx
Devorein: Oops, copy-pasted without rethinking. I will update the source as well.

components/settings/account/components/LensPublication.tsx
Devorein: Right. Removed.

components/settings/account/components/useNewLensPublication.tsx
Devorein: We don't make use of that hook anymore. Forgot to delete it. The code inside it was moved to `useLensProfile`

lib/lens/createPostPublication.ts
Devorein: Wow, thanks, updated. I remember trying that, and not working 🤦🏼‍♂️

lib/lens/createPostPublication.ts
Devorein: Transferred over there.

lib/lens/createPostPublication.ts
Devorein: point. I thought this will change later just like what was there in lenster repo but Iit stayed the same and I forgot to remove the unnecessary variable.

lib/lens/createPostPublication.ts
```diff
@@ -0,0 +1,117 @@
...
+    imageMimeType: null,
+    name: `Post by @${lensProfile.handle}`,
+    animation_url: null,
+    mainContentFocus: PublicationMainFocus.TextOnly,
+    attributes: [{ traitType: 'type', displayType: PublicationMetadataDisplayTypes.String, value: 'text_only' }],
+    media: [],
+    tags: [],
+    locale: getUserLocale(),
+    // lenster users Lenster
+    appId: 'CharmVerse'
```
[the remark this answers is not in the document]
Devorein: yes I checked with `CharmVerse` and it worked. I'm assuming its the client app's name.

lib/lens/createPostPublication.ts
Devorein: Changed and it is an anchor link 😮 It even shows the og image.
image

components/settings/LensProfileProvider.tsx
```diff
@@ -0,0 +1,174 @@
...
+    params: [{ chainId: switchedChainId }]
+  });
+}
+
+export function LensProfileProvider({ children }: { children: React.ReactNode }) {
+  const [lensProfile, setLensProfile] = useState<ProfileFragment | null>(null);
+  const { account, library, chainId } = useWeb3AuthSig();
+  const { user } = useUser();
+  const { space } = useCurrentSpace();
+  const { showMessage } = useSnackbar();
```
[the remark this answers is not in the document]
Devorein: You are right ! That was total over-engineering 🤦🏼‍♂️ Removed the provider, used regular hook and swr to fetch the lens profile data.

components/proposals/components/ProposalProperties/ProposalProperties.tsx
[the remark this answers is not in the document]
mattcasey: This component is a child of that component. So it's not necessary

PR #2671, opened 2023-08-24
simplify mocks for browser tests using charm editor

mattcasey:
### WHAT
copilot:summary

### WHY
The CharmEditor ends up importing a lot of unrelated stuff due to the InlineDatabase. Similarly, the Poll component uses snapshot and some dependency that breaks browser testing. These components can be mocked for the most part.


msadura (approved):
way simpler :D

PR #2690, opened 2023-08-28
Rename features page title

valentinludu:
### <samp>🤖 Generated by Copilot at b53ac5c</samp>

Added the ability to customize the titles of the bounties, forum, member directory, and proposals features in the space settings. Refactored the corresponding page and component files to use a `title` prop and a `useFeaturesAndMembers` hook to fetch and display the custom titles.

### WHY
Feedback from Friday demo.
If a user renames a feature, he should see the renamed title on that page, metatitle and breadcrumbs.


pages/[domain]/bounties/index.tsx
```diff
@@ -2,21 +2,25 @@ import BountiesPage from 'components/bounties/BountiesPage';
...
+import { useFeaturesAndMembers } from 'hooks/useFeaturesAndMemberProfiles';
 import { useHasMemberLevel } from 'hooks/useHasMemberLevel';
 import { setTitle } from 'hooks/usePageTitle';
 import { useSharedPage } from 'hooks/useSharedPage';
 
 export default function BountyPage() {
   const { bounties, loadingBounties } = useBounties();
   const { accessChecked } = useSharedPage();
+  const { features } = useFeaturesAndMembers();
+  const bountiesTitle = features.find((f) => f.id === 'bounties')?.title || 'Bounties';
```
Comment (mattcasey) [inline pages/[domain]/bounties/index.tsx:14]: i think we have to wait for the custom title to. come back, otherwise they'll see a flicker.

Comment (mattcasey) [reply pages/[domain]/bounties/index.tsx:14]: This also seems like a good case for SSR

mattcasey (changes requested):
We should wait til the API comes back, so that there's not a flicker on page load of different titles. 

Also, please add the custom titles to PageTitleWithBreadCrumbs.ts. Then I think we're good

pages/[domain]/bounties/index.tsx
Comment (valentinludu) [reply pages/[domain]/bounties/index.tsx]: There are some cases where we need it to be an array and some cases where we need an object.
I finally exported a mappedFeatures where we can use it as in your example (e.g. `features.bounties.title`)

Comment (valentinludu) [reply pages/[domain]/bounties/index.tsx]: The titles will not flicker because the PageLayout renders everything only after we have a space and a user.

mattcasey (approved):

PR #2694, opened 2023-08-28
Feat/proposal custom fields

msadura:
### <samp>🤖 Generated by Copilot at 0170ef3</samp>

This pull request adds the functionality to create, update, delete, and display custom properties of proposals using proposal blocks. It implements the proposal blocks endpoint in the `charmClient`, the proposal blocks library in the `lib/proposal/blocks` module, and the `useProposalBlocks` hook to access and update the blocks. It also refactors the `cardDetailProperties` component from the `focalboard` library and creates new components and hooks to adapt and render the proposal properties in the board editor and the proposal dialog. It adds test modules for the proposal blocks library and the `createBlock` and `updateBlock` functions. It updates the existing components, stories, and tests to include the fields property of the proposals.

### WHY
custom proposal properties


components/common/BoardEditor/components/cardProperties/CardDetailProperty.tsx
```diff
@@ -0,0 +1,121 @@
+import styled from '@emotion/styled';
```
msadura, line 1: this is just extracted from CardDetailProperties

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -37,115 +32,21 @@ type Props = {
...
+  const {
+    board,
+    card,
+    cards,
+    views,
+    activeView,
+    pageUpdatedAt,
+    pageUpdatedBy,
+    syncWithPageId,
+    mutator = defaultMutator
```
msadura, line 48: added option to pass mutator

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties.tsx
```diff
@@ -181,19 +82,12 @@ function CardDetailProperties(props: Props) {
     const srcIndex = cardPropertyIds.indexOf(sourceProperty.id);
     cardPropertyIds.splice(srcIndex, 1);
     cardPropertyIds.splice(destIndex, 0, sourceProperty.id);
-    await charmClient.patchBlock(
```
msadura, line 184: NOTE: we should not use charm editor directly because it will loose flexibility and always use mutator

.storybook/preview.tsx
```diff
@@ -139,4 +140,15 @@ export const withMuiTheme = (Story, context) => {
   );
 };
 
-export const decorators = [withMuiTheme];
+export const withIntl = (Story, context) => {
```
mattcasey, line 143: does this actually get used somewhere?

charmClient/apis/proposalsApi.ts
```diff
@@ -53,4 +58,20 @@ export class ProposalsApi {
   deleteProposalCategory(spaceId: string, categoryId: string) {
     return http.DELETE<{ ok: true }>(`/api/spaces/${spaceId}/proposal-categories/${categoryId}`);
   }
+
+  getProposalBlocks(spaceId: string) {
```
mattcasey, line 62: please start using hooks pattern going forward

seedData/proposalTemplates.ts
```diff
@@ -1648,7 +1649,8 @@ const templates: Omit<ExportedPage, StandardFields>[] = [
         spaceId: '30bfda0c-19a6-463d-bef7-daccbf433a04',
         title: 'Treasury',
         color: 'gray'
-      }
+      },
+      fields: {}
```
mattcasey, line 1653: i feel like 'null' would be a better default, as we not only have to check if fields exist but also that properties exist. And ideally, ALL proposals will be updated w/a migratio nand we wont need to check anymore. Just like for views, cards, we dont have to check that 'fields' is the right data model

mattcasey (approved):

.storybook/preview.tsx
msadura: focalboar's CardProperties use it, so stories related to them will crash without it. I can add it specifically to proposal stories, but it took me a while to find what actually caused the issue (the error was not saying that) so I decided to add it to all the stories to save us from the future troubles

PR #2717, opened 2023-09-01
Drag and drop pages from sidebar into editor

Devorein:
### <samp>🤖 Generated by Copilot at 0c81e5b</samp>

This pull request adds the ability to reorder pages by drag and drop in the `FidusEditor` component. It updates the websockets logic and types to handle the new `page_reordered` event with a `trigger` and a `pos` property. It also removes some unnecessary properties from the event payload and simplifies the test cases.

### WHY
[Card](https://app.charmverse.io/charmverse/page-7241947275861003)


lib/websockets/spaceEvents.ts
```diff
@@ -218,7 +219,35 @@ export class SpaceEventHandler {
         this.sendError(errorMessage);
       }
     } else if (message.type === 'page_reordered' && this.userId) {
-      const { pageId, currentParentId, newParentId, newIndex } = message.payload;
+      const { pageId: pagePathOrId, newParentId, newIndex, trigger, pos } = message.payload;
```
mattcasey, line 222: can you make sure this is a pageId? I think using 'pageIdOrPath' is an antipattern.

mattcasey (changes requested):
staging is broken and in local dev i cna't drag and drop in the sidebar at all

lib/websockets/spaceEvents.ts
Devorein: I tried to add a custom datatransfer value when dragging and dropping but wasn't able to get it to work. Tried one last time and figured out the issue 🥳 Removed pagePath completely and its a lot more reliable now.

PR #2723, opened 2023-09-04
Use latest version of prosemirror tables

mattcasey:
### <samp>🤖 Generated by Copilot at 4e24715</samp>

This pull request updates the `CharmEditor` component to use the original `prosemirror-tables` package instead of the `@skiff-org` fork, and improves the user experience of resizing table columns. It also removes an unused dependency, updates the storybook package, and fixes some code formatting and styling issues.

### WHY

Wanted to fix an issue with long table headers. Well the solution was just to let the text wrap, like Notion does. But I also wanted to finally upgrade our library and remove skiff, which was not adding anything extra for us anymore. I reused some code from Skiff's repo to add the "Add row" and "Add column" buttons. This will make it easier to add more toggles for the tables later


motechFR:
Found a bug; You need to resize any column at least once for add row/column to work

PR #2724, opened 2023-09-05
Hide suggestion blocks on editor viewing mode

Devorein:
### <samp>🤖 Generated by Copilot at 2d8997f</samp>

This pull request improves the charm editor by removing unused code, adding read-only mode styles, and fixing a syntax error. It affects the files `components/common/CharmEditor/plugins.ts`, `components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx`, `theme/@bangle.dev/styles.scss`, and `theme/styles.scss`.

### WHY
[Card](https://app.charmverse.io/charmverse/page-6634813701566527)


components/common/CharmEditor/plugins.ts
```diff
@@ -91,10 +90,7 @@ export function charmEditorPlugins({
     linkPlugins({ key: linksPluginKey }),
     pasteChecker.plugins({ onError }),
     new Plugin({
-      view: (_view) => {
-        if (readOnly) {
-          rejectAll(_view);
-        }
+      view: () => {
```
motechFR, line 93: Why was this change made?

motechFR (approved):

components/common/CharmEditor/plugins.ts
mattcasey: this was the original solution: I was mutating the doc and removing all suggestions in readonly mode, since we wanted to hide them. but then later on we added the abality for users to still edit the doc so they could save comments - and this means we're using an incorrect version of the doc

mattcasey (approved):

PR #2728, opened 2023-09-05
Fix showing comments in proposal dialog

motechFR:
### <samp>🤖 Generated by Copilot at 50007c2</samp>

This pull request enables comment threads for proposals viewed in a popup dialog. It adds a new prop `externalPageId` to the `ThreadsProvider` component to fetch and display the correct threads for a given proposal. It also improves some code style and performance aspects.

### WHY
Proposals shown in popup had comments coming from the global page id.

The solution was to localise a threads context next to proposals popup and give it an optional pageId.

I considered refactoring useThreads to not rely on a global context, and takes external pageId like other hooks (especially since useSWR will cache the data globally anyways) but felt like overkill.


mattcasey (changes requested):
useCurrentPage is intended to be used by all charm editor plugins to get the current page id (its not used anywhere else anymore). So this might fix threads but other things could be broken. 

Instead, can you try adding `DocumentPageProviders`  around the proposal dialog component so that it has its own id. That is how we manage it for database cards.

You might also need to verify that setCurrentPageId is also being called correctly. u can see how its called in PageDialog

mattcasey (approved):

PR #2740, opened 2023-09-06
Devops/optimise proposals list

motechFR:
### <samp>🤖 Generated by Copilot at 75c4e66</samp>

No summary available (Limit exceeded: required to process 105488 tokens, but only 50000 are allowed per call)

### WHY
Use get accessible proposal ID methods when computing notifications
Trim interfaces for notification tasks to the minimum
Add a peer method for getAccessibleProposalIds inside free permissions client


mattcasey (approved):
my e2e test is failing, all it does is try to create a proposal using a template

PR #2749, opened 2023-09-08
Fix proposals labels in the UI

motechFR:
### <samp>🤖 Generated by Copilot at 9bafa43</samp>

Refactored the UI for displaying the proposal status property in the board editor. Replaced the tag select component with a read-only chip component that shows the status text and color. Added a new `ProposalStatusChipTextOnly` component to `components/proposals/components/ProposalStatusBadge.tsx` to implement the chip UI.

### WHY
Show correct proposal labels and fix overflow for chips.

One potential issue, overflow handles differently for the Proposal Chip than it does for Tag Select.

We could:
1. Leave as is
2. Align the styling of ProposalChip
3. Go back to using TagSelect for proposal chip with mapped properties

Screenshot 2023-09-08 at 18 26 55


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -135,40 +139,50 @@ function PropertyValueElement(props: Props) {
...
+    // We should migrate over the proposals as datasource blocks to the same format as proposals table
+    return (
+      <ProposalStatusChipTextOnly
+        status={
+          (stringUtils.isUUID(propertyValue as string)
+            ? propertyTemplate.options.find((opt) => opt.id === propertyValue)?.value
+            : propertyValue) as ProposalStatus
+        }
+      />
+    );
```
motechFR, line 154: Rationale for using this, we only update proposal category from the proposal itself, and it's simpler to use than mapping existing property values

mattcasey (approved):

PR #2753, opened 2023-09-11
Removed dynamic mock data in member directory story 

Devorein:
### <samp>🤖 Generated by Copilot at 5f68ccc</samp>

Improved mock data for testing space members and member properties. Added `joinDate` and set `timezone` to null in `spaceMember.ts`. Used fixed dates in `memberProperty.ts`.

### DIFF
--- a/testing/mocks/spaceMember.ts
+++ b/testing/mocks/spaceMember.ts
@@ -3,6 +3,8 @@ import type { Member } from 'lib/members/interfaces';
 
 import { createMockUser } from './user';
 
+const joinDate = new Date('2023-07-19T00:00:00.000Z').toISOString();
+
 export function createMockSpaceMember(user = createMockUser()): Member {
   const username = user.username.toLocaleLowerCase().replace(' ', '_');
   return {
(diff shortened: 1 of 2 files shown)


testing/mocks/spaceMember.ts
```diff
@@ -11,11 +13,11 @@ export function createMockSpaceMember(user = createMockUser()): Member {
     avatar: user.avatar ?? undefined,
     deletedAt: user.deletedAt ?? undefined,
     isBot: user.isBot ?? false,
-    joinDate: new Date().toISOString(),
+    joinDate,
     onboarded: true,
     profile: {
       id: user.id,
-      timezone: 'America/New_York',
+      timezone: null,
```
mattcasey, line 20: why'd you remove timezone? this means that the tests will result in a different string depending on where the code is run

mattcasey (approved):
one question

testing/mocks/spaceMember.ts
Devorein: Unfortunately having a timezone results in it being displayed in the gallery view of the member directory with the exact time the stories were built, which means we will have to manually approve this story everytime.
image

PR #2755, opened 2023-09-11
Performance/optimise blocks

motechFR:
### <samp>🤖 Generated by Copilot at 0065b57</samp>

This pull request refactors the focalboard store and the charm client to simplify the loading of database blocks and remove unused features. It introduces a single async thunk `initialDatabaseLoad` for fetching all blocks for a given space, and a single endpoint `charmClient.getSubtree` for fetching blocks for a given page id or path. It also removes the comments and users features from the focalboard store and components, as they are now handled by separate services and components. Additionally, it adds the logic to load the views for a database page when it is expanded in the page navigation tree, and disables caching for the HTTP request to fetch blocks.

### WHY
Simplify block loading to only retrieve what we need at any point in time


charmClient/charmClient.ts
```diff
@@ -258,14 +258,30 @@ class CharmClient {
...
-      .GET<Block[]>('/api/blocks', { spaceId })
+      .GET<Block[]>(
+        '/api/blocks',
+        { spaceId },
+        {
+          headers: {
+            'Cache-Control': 'no-store, no-cache, must-revalidate, proxy-revalidate'
+          }
+        }
+      )
```
motechFR, line 269: This is temporary for the PR development so I always see the payload size. Should revert this change before merging

charmClient/charmClient.ts
```diff
@@ -256,16 +252,20 @@ class CharmClient {
     return http.POST<{ importedRolesCount: number }>('/api/guild-xyz/importRoles', payload);
   }
 
-  async getAllBlocks(spaceId: string): Promise<FBBlock[]> {
+  getBlock({ blockId }: { blockId: string }): Promise<FBBlock> {
+    return http.GET<Block>(`/api/blocks/${blockId}`).then(blockToFBBlock);
+  }
+
+  getSubtree({ pageIdOrPath, spaceId }: { pageIdOrPath?: string; spaceId?: string }): Promise<FBBlock[]> {
```
mattcasey, line 259: Please avoid using variables that mix types like this unless it's truly necessary. This makes the infra very hard to reason about. I don't see any place we are actually using a path for this particular method? the server-side code looks like it will blow up if it's not a uuid.

lib/votes/getVoteTasks.ts
```diff
@@ -1,4 +1,4 @@
-import type { BulkPagePermissionFlags } from '@charmverse/core/dist/cjs/permissions';
+import type { BulkPagePermissionFlags } from '@charmverse/core/permissions';
```
mattcasey, line 1: nice cleanup, i wonder if this might have been messing with our builds since i think we use both compiled versions sometimes

charmClient/charmClient.ts
motechFR: Thanks! I started out thinking might need this, looks like we don't will clean up before merging

lib/votes/getVoteTasks.ts
motechFR: The interesting thing that dist/cjs will always blow up a build if importing actual code. Type imports work just fine though (even though best not to have them)

Could be a nice little rule to add to our dependency-cruiser now that we have it working for non-circular already

mattcasey (approved):

PR #2760, opened 2023-09-11
add redirect for pages onto backend

mattcasey:
### <samp>🤖 Generated by Copilot at d0d7aa1</samp>

Refactored and simplified the page view tracking logic across the app by using a custom hook `useTrackPageView` that sends analytics data to the `charmClient` service. Removed unused and redundant code related to the old tracking methods and events. Improved the consistency and accuracy of the page view event types and properties. Added page view tracking to some components that were missing it. Reorganized some settings components into separate pages. Added logic to find the default page for a space domain based on the user's last page view.

### WHY

This is the main file I moved logic to the backend: https://github.com/charmverse/app.charmverse.io/pull/2760/files#diff-22ffda8d8af7a93df01ce6cec34538e350a42aba9b1b6c944b6699d41539e204R8

The other changes are mostly to make sure we track page views properly, because it depends on them working, and to get some of the e2e tests to pass


__e2e__/po/signup.po.ts
```diff
@@ -24,7 +24,7 @@ export class SignUpPage {
   }
 
   async waitForWorkspaceLoaded({ domain }: { domain: string }) {
-    await this.page.waitForURL(`**/${domain}`);
+    await this.page.waitForURL(`**/${domain}/**`);
```
motechFR, line 27: We should cleanup our E2E to avoid duplicates one day in the future

__e2e__/sidebarAddNewSpace.spec.ts
```diff
@@ -79,7 +81,7 @@ test.describe.serial('Add a new space from sidebar and load it', async () => {
 
     await nameInput.fill(uniqueDomainName2);
     await page.locator('data-test=create-workspace').click();
-    await page.waitForURL(`**/${uniqueDomainName2}`);
+    await page.waitForURL(`**/${uniqueDomainName2}/*`);
```
motechFR, line 84: Can we use waitForWorkspaceUrl here?

components/bounties/BountiesPage.tsx
```diff
@@ -40,23 +41,16 @@ const StyledButton = styled(Button)`
   }
 `;
 
-const views: { label: string; view: 'gallery' | 'board' }[] = [
-  { label: 'Open', view: 'gallery' },
-  { label: 'All', view: 'board' }
-];
+const views = [
+  { label: 'Open', id: 'gallery' },
```
motechFR, line 45: Another suggestion: "type" instead of ID to align with pages and blocks

components/settings/space/SpaceSettings.tsx
```diff
@@ -462,14 +459,8 @@ export function SpaceSettings({ space }: { space: Space }) {
...
-            const filteredSpaces = spaces.filter((s) => s.id !== space.id);
-            // redirect user to the next space if they have one
-            if (filteredSpaces.length > 0) {
-              await router.push(getDefaultWorkspaceUrl(filteredSpaces));
-            } else {
-              await router.push('/createSpace');
-            }
-            setSpaces(filteredSpaces);
+            // redirect to default workspace
+            window.location.href = window.location.origin;
```
motechFR, line 463: So much simpler 👍

lib/session/getDefaultPageForSpace.ts
```diff
@@ -0,0 +1,129 @@
+import type { PageMeta } from '@charmverse/core/pages';
+import { pageTree } from '@charmverse/core/pages/utilities';
+import type { Space } from '@charmverse/core/prisma';
+import { PageType } from '@charmverse/core/prisma';
+import { prisma } from '@charmverse/core/prisma-client';
+
+import { filterVisiblePages } from 'components/common/PageLayout/components/PageNavigation/PageNavigation';
```
motechFR, line 7: Can we avoid importing from components towards lib if possible?

testing/setupDatabase.ts
```diff
@@ -1143,20 +1143,22 @@ export async function generateForumComment({
 }
 
 export async function createPost(
```
motechFR, line 1145: We could even delete this code, we have testUtilsForums from core

motechFR:
Edge case: Opening a card in a modal from DB then navigating to it sends a second event

lib/session/getDefaultPageForSpace.ts
mattcasey: +1 this is a good cleanup

components/bounties/BountiesPage.tsx
mattcasey: hm i like 'type' thanks

__e2e__/sidebarAddNewSpace.spec.ts
mattcasey: ya, i wasnt even paying attention to which method was being used tbh

mattcasey:
> Edge case: Opening a card in a modal from DB then navigating to it sends a second event

So clicking "View as page" tracks another page view, you mean? I think there's pros and cons of tracking it this way, but one benefit is that it updates the pathname used so that when you come back to it, it remembers you last saw it as a full page.

> Second edge case: Clicking to another tab then back to app fires another app_loaded event

app_loaded was from a previous project. I don't think I changed this code, but might be good to investigate/record as a bug

PR #2762, opened 2023-09-12
Fixed issue with showing user avatar in page thread

Devorein:
### <samp>🤖 Generated by Copilot at b82f660</samp>

Refactored the comment data structure and rendering logic to simplify the code and avoid unnecessary user data. Removed the `CommentWithUser` type and used the `useMembers` hook to fetch the user information for each comment. Updated the comment API methods, types, and endpoints accordingly.

### WHY
[Post](https://discord.com/channels/894960387743698944/1150792279066419220)

### DIFF
--- a/components/common/CharmEditor/components/PageThread.tsx
+++ b/components/common/CharmEditor/components/PageThread.tsx
@@ -1,5 +1,6 @@
 import { useEditorViewContext } from '@bangle.dev/react';
 import type { PagePermissionFlags } from '@charmverse/core/permissions';
+import type { Comment } from '@charmverse/core/prisma-client';
 import styled from '@emotion/styled';
 import { Check } from '@mui/icons-material';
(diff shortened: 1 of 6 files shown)


components/common/CharmEditor/components/PageThread.tsx
```diff
@@ -284,7 +288,8 @@ const PageThread = forwardRef<HTMLDivElement, PageThreadProps>(
     const [isMutating, setIsMutating] = useState(false);
     const [editedCommentId, setEditedCommentId] = useState<null | string>(null);
     const menuState = usePopupState({ variant: 'popover', popupId: 'comment-action' });
-    const [actionComment, setActionComment] = useState<null | CommentWithUser>(null);
+    const [actionComment, setActionComment] = useState<null | Comment>(null);
+    const { members } = useMembers();
```
mattcasey, line 292: Since youre looking up by id, it would be better to import `membersRecord` next time instead

PR #2763, opened 2023-09-12
Sync page content with sidebar script

Devorein:
### <samp>🤖 Generated by Copilot at 34f696f</samp>

Improved error handling and logging for content sync script. The `scripts/updatePageContentForSync.ts` file now catches and logs any errors that occur during the synchronization process using the `writeToSameFolder` utility. The error logs are saved as JSON files in the same folder as the script.

### WHY
1. Insert the children pages in the parent page's content
2. Change the parent of the children pages


scripts/updatePageContentForSync.ts
```diff
@@ -11,60 +12,74 @@ export async function updatePageContentForSync() {
...
+    try {
+      const pageContent = (page.content ?? emptyDocument) as PageContent;
+      const nestedPageIds: string[] = [];
+      const pageContentNode = getNodeFromJson(pageContent);
+      pageContentNode.forEach((node) => {
+        if (node.type.name === 'page' && node.attrs) {
+          nestedPageIds.push(node.attrs.id);
+        }
+      });
+      const childPages = await prisma.page.findMany({
```
mattcasey, line 27: you already have all the pages in memory, no need to query these again i think. u could just include parentId in the query?

testing/prosemirror/builders.ts
```diff
@@ -70,7 +70,8 @@ export type NodeType =
   | 'tooltip-marker'
   | 'tweet'
   | 'video'
-  | 'underline';
+  | 'underline'
+  | 'linkedPage';
```
mattcasey, line 74: i try to keep these alphabetical so it's easy to find them

Devorein: Sure updated.

lib/prosemirror/applyStepsToNode.ts
```diff
@@ -1,5 +1,4 @@
-import type { Node } from '@bangle.dev/pm';
-import { Step } from '@bangle.dev/pm';
+import { Step, type Node } from '@bangle.dev/pm';
```
mattcasey, line 1: this doesn't look like our usual syntax

scripts/updatePageContentForSync.ts
mattcasey, line 1: this file belongs in scripts/migrations with a date attached

lib/pages/updatePageContentForSync.ts
mattcasey, line 1: We shouldn't put anything in`lib` that isn't actually used by our production code. This just belongs in the script folder as far as i can tell from this PR. But if we left it here it should at least go into lib/prosemirror/conversions.

lib/pages/updatePageContentForSync.ts
```diff
@@ -0,0 +1,191 @@
...
+  const { PAGE_SIZE } = config;
+  let cursor: string | undefined;
+  let completedPages = 0;
+  const totalPages = await prisma.page.count({});
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
+    const pages = await prisma.page.findMany({
+      where: {
+        id: cursor ? { gt: cursor } : undefined,
```
mattcasey, line 41: where did you see this pattern? we should be using `skip`.

lib/pages/updatePageContentForSync.ts
```diff
@@ -0,0 +1,191 @@
...
+    PAGE_SIZE: 100
+  }
+) {
+  const { PAGE_SIZE } = config;
+  let cursor: string | undefined;
+  let completedPages = 0;
+  const totalPages = await prisma.page.count({});
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
```
mattcasey, line 38: while (true) is not a safe way to write loops. i think it's ok because we are nto going to be using this in prod though

lib/pages/updatePageContentForSync.ts
```diff
@@ -0,0 +1,191 @@
...
+              }
+
+              return false;
+            }
+            default:
+              return true;
+          }
+        });
+
+        doc = applyStepsToNode(linkedPageConversionSteps, doc);
```
mattcasey, line 114: can we call applyStepsToNode once at the end?

lib/pages/updatePageContentForSync.ts
```diff
@@ -0,0 +1,191 @@
...
+        const childPagesNotInDocument = childPages.filter((childPage) => !nestedPageIds.has(childPage.id));
+        const nestedPageAppendStep: ProsemirrorJSONStep = {
+          from: doc.content.size,
+          stepType: 'replace',
+          slice: {
+            content: childPagesNotInDocument
+              .sort((page1, page2) => (page1.id > page2.id ? 1 : -1))
```
mattcasey, line 121: ids are random - there's no need to sort by id. Why are we not using the order of `childPageIds`?

lib/pages/__tests__/updatePageContentForSync.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+          id: childPage3.id,
+          path: childPage3.path,
+          type: childPage3.type
+        })
+      ).toJSON()
+    );
+
+    expect(updatedParentPage.version).toEqual(3);
+
+    expect(pageDiffs).toStrictEqual([
```
mattcasey, line 129: this is a very brittle test assertion, in general i would avoid using strictequal on large nested objects. this is actually testing a lot more than the code, for example it depends on the output of replaceWith

lib/pages/__tests__/updatePageContentForSync.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    const pageContent = _.doc(
+      _.paragraph('Paragraph 1'),
+      _.page({
+        id: childPage1Id,
+        type: 'page',
+        path: childPage1Path
+      }),
+      _.paragraph('Paragraph 2'),
+      _.page({
```
mattcasey, line 33: Can you add another page-to-be-linked instance, to make sure that the approach still works when we have more than one step

mattcasey:
I got a bunch of errors just running the script on a brand new space in dev. I was using tsx, though: 
```
error: Failed to update page 890cbfcb-096a-4ea6-b836-92ece9417c01 {
  error: RangeError: Invalid input for Node.fromJSON
      at Function.fromJSON (/Users/matt/charmverse/app.charmverse.io/node_modules/prosemirror-model/dist/index.cjs:1390:24)
      at Schema.nodeFromJSON (/Users/matt/charmverse/app.charmverse.io/node_modules/prosemirror-model/dist/index.cjs:2336:19)
      at getNodeFromJson (/Users/matt/charmverse/app.charmverse.io/lib/prosemirror/getNodeFromJson.ts:6:30)
      at updatePageContentForSync (/Users/matt/charmverse/app.charmverse.io/lib/pages/updatePageContentForSync.ts:81:19)
}```

File: next.config.js [hunk 10]
```diff
@@ -183,10 +178,10 @@ const config = {
           return {
             ..._entry,
             cron: './background/cron.ts',
-            websockets: './background/initWebsockets.ts'
+            websockets: './background/initWebsockets.ts',
+            updatePageContentForSync: './scripts/updatePageContentForSync.ts'
```
mattcasey, line 182: this script seems to work by using tsx, also please don't commit this to main if you do take this approach

scripts/updatePageContentForSync.ts
Devorein: Oops, you're right. Updated.

lib/pages/updatePageContentForSync.ts
Devorein: Updated thnx.

lib/pages/updatePageContentForSync.ts
Devorein: Yeah thats a good point, I didn't consider new pages being created while the script is running. This pattern is what we are using every where for our pagination (cursor based) https://www.prisma.io/docs/concepts/components/prisma-client/pagination#cursor-based-pagination
Its more performant than offset pagination but since performance is a minimal concern here I will change it to offset pagination.

lib/pages/updatePageContentForSync.ts
Devorein: Yeah agree, removed this after changing cursor pagination to offset

lib/pages/__tests__/updatePageContentForSync.spec.ts
Devorein: Yeah definitely

next.config.js
Devorein: Yeah I get the following error while running it using `ts-node`

import getBasePlacement from "../utils/getBasePlacement.js";
^^^^^^

SyntaxError: Cannot use import statement outside a module

I will remember to remove it from next.config before merging it

PR #2764, opened 2023-09-12
Fixed scroll to top issue when closing date calendar

Devorein:
### <samp>🤖 Generated by Copilot at 7c0076b</samp>

Added the `disableRestoreFocus` prop to various `Popover` and `Menu` components to improve accessibility and keyboard navigation. This prop prevents the focus from returning to the element that triggered the popover or menu when it closes, avoiding unexpected focus shifts.

### WHY
[Card](https://app.charmverse.io/charmverse/page-4852670458855628)


mattcasey (changes requested):
looks like we're just applying this everywhere, kinda makes sense. can u please just add it as a default prop on the theme instead? theme/index.ts

This is such a strange default, makes me feel like maybe we shouldn't be using popover but dialog or something else

PR #2767, opened 2023-09-13
Show loading screen during loading notifications

Devorein:
### <samp>🤖 Generated by Copilot at 7782051</samp>

Added loading feedback to notification components. Used a common `LoadingComponent` to show a loading indicator and label in the `NotificationModal` and `NotificationPreviewPopover` components. Added an `isLoading` state variable to the `useNotifications` hook and context to track the loading status of notifications data.

### WHY
[Card](https://app.charmverse.io/charmverse/page-4651833491179498)

### DIFF
--- a/components/common/PageLayout/components/Header/components/NotificationPreview/NotificationModal.tsx
+++ b/components/common/PageLayout/components/Header/components/NotificationPreview/NotificationModal.tsx
@@ -10,6 +10,7 @@ import TaskOutlinedIcon from '@mui/icons-material/TaskOutlined';
 import { Badge, Box, Dialog, DialogContent, Divider, IconButton, Tab, Tabs, Tooltip, Typography } from '@mui/material';
 import { Fragment } from 'react';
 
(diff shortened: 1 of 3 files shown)


components/common/PageLayout/components/Header/components/NotificationPreview/NotificationModal.tsx
```diff
@@ -208,45 +210,47 @@ export function NotificationModal() {
...
-                  <NotificationPreview
-                    large
-                    notification={notification}
-                    markAsRead={markAsRead}
-                    onClose={closeNotificationsModal}
-                  />
-                  <Divider />
-                </Fragment>
-              ))}
+              <LoadingComponent isLoading={isLoading} label='Fetching your notifications' size={24}>
```
msadura, line 213: This works but honestly, it would look much nicer  with proper skeleton loader rather than just a spinner. If you have not used skeleton from mui before: https://mui.com/material-ui/react-skeleton/

We have also a few examples in our codebase already, it is really simple to use and looks much better from user perspective

PR #2769, opened 2023-09-13
Page link node, schema & plugin

Devorein:
### <samp>🤖 Generated by Copilot at 1a230da</samp>

No summary available (Limit exceeded: required to process 100954 tokens, but only 50000 are allowed per call)

### WHY
Added a new node, schema and plugin for page-link which replaces the previous `page` node. Link to page uses `page-link`, while insert page or database uses `page` node


components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
```diff
@@ -16,7 +16,7 @@ export default function useNestedPage(currentPageId?: string) {
   const cardId = new URLSearchParams(window.location.search).get('cardId');
   const isInsideCard = cardId && cardId?.length !== 0;
   const addNestedPage = useCallback(
-    async (type?: Page['type']) => {
+    async (type?: Page['type'], isLinkedPage?: boolean) => {
```
mattcasey, line 19: it doesnt look like you use this at all?

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
```diff
@@ -73,19 +76,29 @@ function Component({ menuState }: { menuState: PluginState }) {
...
     const node = getNodeForRowPosition({ view, rowPosition: menuState.rowPos, rowNodeOffset: menuState.rowNodeOffset });
+    if (!node) {
+      return null;
+    }
+
+    const nodeTypeName = node.node.type.name;
     const tr = view.state.tr;
-    if (node?.node.type.name === 'page') {
+    const isLinkedPage = nodeTypeName === linkedPageNodeName;
+    if (nodeTypeName === 'page' || isLinkedPage) {
```
mattcasey, line 86: if this is a linked page, we should not be duplicating anything. why did u need to special-case linked nodes?

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
```diff
@@ -73,19 +76,29 @@ function Component({ menuState }: { menuState: PluginState }) {
 
   async function duplicateRow() {
     const node = getNodeForRowPosition({ view, rowPosition: menuState.rowPos, rowNodeOffset: menuState.rowNodeOffset });
+    if (!node) {
```
mattcasey, line 79: this introduces a bug where the popup would not be called. it's better to remove this and just let the function continue

components/common/CharmEditor/components/linkedPage/linkedPage.constants.ts
```diff
@@ -0,0 +1,4 @@
+export const linkedPageNodeName = 'page-link';
```
mattcasey, line 1: Can we use camel case instead? I'll admit we have every combination of snakecase, camelcase, etc in our component names. if we were following our general best practice and starting from scratch, we should be using snake case, but the majority seem to be camel case. Hyphenated vars is pretty much only for CSS classes and URL paths, and are a bit more work because you need brackets to access it: `schema.nodes['page-link']`.

mattcasey: Further... i like that u are mirroring 'nestedPage' with 'linkedPage' but i dislike that its swapped in the name for some reason to 'page-link'. maybe we just call the component "linkedPage" as well?

components/common/CharmEditor/components/linkedPage/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './linkedPage.constants';
```
mattcasey, line 1: no more barrel files in plugin components. this is what we're trying to avoid

components/common/CharmEditor/components/nestedPage/components/NestedPagesList.tsx
mattcasey, line 1: this file should be moved to the linkedPage folder

mattcasey (changes requested):
see comments/questions :)

Also: I think it would really help to have the shortcut icon version for linked pages. Users are gonna start asking us why moving links sometimes moves pages and sometimes not. I know it's extra work but i think we should just do it right now than have to deal with it later.

components/common/CharmEditor/components/nestedPage/hooks/useNestedPage.ts
Devorein: Nice catch. Updated.

components/common/CharmEditor/components/linkedPage/linkedPage.constants.ts
Devorein: Yeah lets go with that, I thought of going with `linkedPage` as well.

components/common/CharmEditor/components/nestedPage/components/NestedPagesList.tsx
Devorein: Updated

components/common/CharmEditor/components/linkedPage/index.ts
Devorein: Gotcha. Updated

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
Devorein: Nice catch. Reverted back the change. It was mostly to get rid of `node?` in places.

PR #2779, opened 2023-09-14
Show comments in sidebar on initial load

Devorein:
### <samp>🤖 Generated by Copilot at 069aa80</samp>

Fixed a bug in `CommentsSidebar` component that could cause comment loss. Improved the logic for checking empty document content.

### DIFF
--- a/components/[pageId]/DocumentPage/components/CommentsSidebar.tsx
+++ b/components/[pageId]/DocumentPage/components/CommentsSidebar.tsx
@@ -1,4 +1,5 @@
 import { useEditorViewContext } from '@bangle.dev/react';
+import { isEmptyDocument } from '@bangle.dev/utils';
 import type { PagePermissionFlags } from '@charmverse/core/permissions';
 import styled from '@emotion/styled';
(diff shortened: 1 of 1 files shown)


components/[pageId]/DocumentPage/components/CommentsSidebar.tsx
```diff
@@ -78,7 +79,10 @@ function CommentsSidebarComponent({ inline, permissions }: { inline?: boolean; p
   }
 
   const view = useEditorViewContext();
-  const extractedThreadIds = extractThreadIdsFromDoc(view.state.doc, specRegistry.schema);
+  // view.state.doc stays the same (empty content) even when the document content changes
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/components/CommentsSidebar.tsx:82]: This was the only way I managed to do it, view.state.doc stays empty even when the document content changes. We might also filter comments those are not mentioned in the document out when returning them from our api

Comment (mattcasey) [reply components/[pageId]/DocumentPage/components/CommentsSidebar.tsx:82]: it seems like the bug is actually doc not being updated. What did you do to try to fix that, or can u see what's going on?

PR #2781, opened 2023-09-14
Friendly urls

motechFR:
### <samp>🤖 Generated by Copilot at e2496b2</samp>

This pull request introduces several changes to the app.charmverse.io repository, mainly related to the generation and handling of friendly and unique page paths for pages created from Google Forms, posts, and card pages. The pull request updates the usePage hook, the pages API, the pages server functions, the forums utils functions, and the public API functions to use the new page path logic. The pull request also adds new tests, refactors some code, and modifies some GitHub actions and workflows to use the friendly-urls branch and tag.

### WHY
Add support for human-friendly URLs and keep retro-compat with old URLs


charmClient/apis/pagesApi.ts
```diff
@@ -29,7 +29,7 @@ export class PagesApi {
   }
 
   updatePage(pageOpts: Partial<Page>) {
-    return http.PUT<void>(`/api/pages/${pageOpts.id}`, pageOpts);
+    return http.PUT<PageMeta>(`/api/pages/${pageOpts.id}`, pageOpts);
```
mattcasey, line 32: We shouldn't need to return data in a PUT. It just makes the response take longer.

hooks/usePage.ts
```diff
@@ -26,6 +30,8 @@ type PageResult = {
 const noop = () => Promise.resolve(undefined);
 
 export function usePage({ spaceId, pageIdOrPath }: Props): PageResult {
+  const router = useRouter();
+  const { space } = useCurrentSpace();
```
mattcasey, line 34: spaceId is already passed in - i think it's extremely confusing to pull in a different space from some other context

hooks/usePage.ts
```diff
@@ -26,6 +30,8 @@ type PageResult = {
 const noop = () => Promise.resolve(undefined);
 
 export function usePage({ spaceId, pageIdOrPath }: Props): PageResult {
+  const router = useRouter();
```
mattcasey, line 33: pageId/path is already passed in. please avoid pulling in a different pageId from a new context

hooks/usePage.ts
```diff
@@ -35,9 +41,19 @@ export function usePage({ spaceId, pageIdOrPath }: Props): PageResult {
...
+      const updatedPage = await charmClient.pages.updatePage(updates);
+
+      // if (space && router.query.pageId === pageIdOrPath && updatedPage.path !== pageIdOrPath) {
+      //   router.push(`/${space.domain}/${pageIdOrPath}`, `/${space.domain}/${updatedPage.path}`, { shallow: true });
+      // }
+      // if (currentPageId && currentPageId === pageWithContent?.id && space) {
+      //
+      // }
+    },
+    [space, pageIdOrPath, router.query.pageId]
```
mattcasey, line 55: we are just calling updatePage with data passed in, i don't think any of these dependencies actually matter?

hooks/usePageFromPath.ts
```diff
@@ -11,6 +11,10 @@ export function usePageIdFromPath() {
 export function usePageFromPath(): PageMeta | null {
   const basePageId = usePageIdFromPath();
   const { pages } = usePages();
-  const basePage = Object.values(pages).find((page) => page?.id === basePageId || page?.path === basePageId);
+  const basePage = basePageId
```
mattcasey, line 14: We really only need to use additionalPaths on the initial app load, and I'd like to keep the page api response size down.

Instead of looking it up here, can we put a db query in getServerSideProps? You can then do a redirect to the correct path.

lib/pages/utils.ts
```diff
@@ -1,9 +1,29 @@
...
+}) {
+  const split = existingPagePath.split('-');
+  let suffix = split[split.length - 1]?.trim();
+
+  if (!suffix || !suffix.match(/^\d{8,}$/)) {
+    suffix = Math.random().toString().replace('0.', '');
+  }
+  const titleToPath = stringToValidPath({ input: title ?? '', wordSeparator: '-', autoReplaceEmpty: false });
+
+  return `page-${titleToPath}-${suffix}`;
```
mattcasey, line 24: remove 'page-' prefix

lib/pages/server/updatePage.ts
```diff
@@ -1,39 +1,56 @@
...
     data,
-    include: {
-      permissions: {
-        include: {
-          sourcePermission: true
-        }
-      }
-    }
-  });
+    select: pageMetaSelect()
```
mattcasey, line 54: i mentioned someplace else i'd like to not return updates from PUT. But I realized you probably want to make sure u have the right path or be able to update the URL without reloading?

I still would only return the path field, so it's obvious and we can avoid coupling PageMeta logic here. But that's not as big a deal

mattcasey (changes requested):
couple small requests/suggestions

mattcasey (approved):

PR #2783, opened 2023-09-15
Persist last visited board view

Devorein:
### <samp>🤖 Generated by Copilot at e861bd5</samp>

Refactored and simplified the code for managing focalboard views and local storage in various components. Removed unused and redundant code from the `viewsSlice` and the `SharedPage` component. Added a feature to remember the last viewed view for each board in the charm editor.

### WHY
[Card](https://app.charmverse.io/charmverse/page-9404902857241855)

### DIFF
--- a/components/[pageId]/DatabasePage/DatabasePage.tsx
+++ b/components/[pageId]/DatabasePage/DatabasePage.tsx
@@ -14,10 +14,7 @@ import {
 } from 'components/common/BoardEditor/focalboard/src/store/boards';
 import { initialDatabaseLoad } from 'components/common/BoardEditor/focalboard/src/store/databaseBlocksLoad';
 import { useAppDispatch, useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
-import {
-  getCurrentBoardViews,
-  setCurrent as setCurrentView
-} from 'components/common/BoardEditor/focalboard/src/store/views';
+import { getCurrentBoardViews } from 'components/common/BoardEditor/focalboard/src/store/views';
 import { Utils } from 'components/common/BoardEditor/focalboard/src/utils';
 import FocalBoardPortal from 'components/common/BoardEditor/FocalBoardPortal';
 import { useFocalboardViews } from 'hooks/useFocalboardViews';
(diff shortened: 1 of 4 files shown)


components/[pageId]/DatabasePage/DatabasePage.tsx
```diff
@@ -39,16 +36,26 @@ interface Props {
...
+  const urlViewId = router.query.viewId as string | undefined;
+  const [currentViewId, setCurrentViewId] = useState<string | undefined>(urlViewId);
   const boardViews = useAppSelector(getCurrentBoardViews);
+  const { setFocalboardViewsRecord, focalboardViewsRecord } = useFocalboardViews();
   // grab the first board view if current view is not specified
   const { showMessage } = useSnackbar();
-  const activeView =
-    typeof currentViewId === 'string' ? boardViews.find((view) => view.id === currentViewId) : boardViews[0];
+
+  const boardId = page.boardId;
```
Comment (motechFR) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:46]: Suggestion; Make this board.id

motechFR (approved):
Tested, works great

motechFR (changes requested):
Bug: I cannot add a new view on a linked database. The "+" button opens the sidebar for current view instead

components/[pageId]/DatabasePage/DatabasePage.tsx
Comment (mattcasey) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:46]: please use board.id instead, to keep the number of places down where we check what board you're on

components/[pageId]/DatabasePage/DatabasePage.tsx
```diff
@@ -39,16 +36,26 @@ interface Props {
...
+  const firstBoardView = boardViews.find((view) => view.parentId === boardId);
```
Comment (mattcasey) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:47]: why do you need to check parentId = boardId? when would we have boardViews that are from other boards?

mattcasey (changes requested):
There's logic in the sidebar to handle default views already, none of that was changed and we added a bunch of new logic instead. In the Loom video it looks like viewId is never set, when she clicks the second tab. I think that is the real bug that needs to be fixed

mattcasey:
we decided not to support this

PR #2793, opened 2023-09-16
Added e2e test for charmeditor drag and drop

Devorein:
### <samp>🤖 Generated by Copilot at 0737997</samp>

Added a new e2e test for the CharmEditor component using playwright. The test checks that the drag and drop feature works correctly and updates the page content in the database. The test is part of a PR to improve the CharmEditor functionality.


__e2e__/charmEditor.spec.ts
```diff
@@ -0,0 +1,123 @@
+import { prisma } from '@charmverse/core/prisma-client';
```
msadura, line 1: test looks nice, but I would place it under `__e2e__/charmEditor/dragAndDrop.spec.ts` or something like that - this is just small pioece of charmEditor so I would not put it into general `charmEditor.spec.ts`

msadura (approved):

__e2e__/charmEditor.spec.ts
Devorein: Good point. I will move it inside `charmEditor` folder. Thank you for the suggestion.

PR #2798, opened 2023-09-18
Show page icon menu on small screens

Devorein:
### <samp>🤖 Generated by Copilot at b2d0b4d</samp>

This pull request removes unused and redundant code from the `subMenuOption` component and makes the `dialog` and `menu` components responsive on different screen sizes by removing media queries and hidden elements. These changes improve the user interface and functionality of the app.


Devorein:
This is a temporary fix. Ideally, I think we should get rid of all the Menu components coming from focalboard, and use the existing one.

mattcasey (approved):

PR #2809, opened 2023-09-21
replace ether lib with viem in most places

mattcasey:
### WHAT
copilot:summary


msadura (approved):
woah this looks solid and I really like how you reorganized things. 
PS. were you able to get rid of ethers from initial bundle? Looks like now it should be needed only with gnosis payments?

mattcasey:
> PS. were you able to get rid of ethers from initial bundle? Looks like now it should be needed only with gnosis payments?

The _app.tsx file size has not changed, I think because some hooks may still use ethers. I just took on the utilities, there are still some places to fix but they need proper testing so skipped them for now.

PR #2815, opened 2023-09-22
Update next.js to 14

motechFR:
### <samp>🤖 Generated by Copilot at df29a7c</samp>

This pull request adds websockets support to the app, refactors the deploy staging workflow, and fixes some typos. It modifies the `package.json`, `.env.test.local.example`, and `.github/workflows` files.


mattcasey:
FYI I am expecting that we'll just use concurrently when running npm start to run both next.js and sockets. let me know if you have another idea

mattcasey:
I tried your suggestionm, using next's default app... and i can't find anything special for pages folder: https://nextjs.org/docs/pages/api-reference/create-next-app

next.config.js
```diff
@@ -74,6 +74,12 @@ const config = {
         destination: '/',
         permanent: true
       },
+      {
+        // Added for backwards compatibility
+        source: '/invite/page',
+        destination: '/invite-request-code',
```
mattcasey, line 80: this should be /invite/request-code right?

mattcasey (changes requested):
- pages are not loading from the socket server because Next.js broke support with socket.io. So instead, I just don't want to run them together. As mentioned before please use concurrently to run them both with npm start.
- /invite/request-code is kind of a weird name for the path but i can't think of much better. BUT it turns out that while we can't use a file named "page.tsx", we can have a "page" folder with an index.tsx inside. So can we just do that instead, and keep the current URLs?

mattcasey (approved):

PR #2816, opened 2023-09-22
Fix/missing comments

motechFR:
### <samp>🤖 Generated by Copilot at 3d14799</samp>

This pull request adds a feature to display and manage comments on database cards in the app. It modifies the `DocumentPage` component, the board editor, the charm client, and the API to handle the logic and UI of comments. It also adds e2e tests for the feature using `viewCardProps.spec.ts` and `document.po.ts`. It changes the domain of the space for testing in `makeAPIKey.ts`.

### WHY
Restore comments in focalboard

### DIFF
--- a/__e2e__/po/document.po.ts
+++ b/__e2e__/po/document.po.ts
@@ -22,8 +22,7 @@ export class DocumentPage extends GlobalPage {
     public documentTitle = page.locator(`data-test=editor-page-title`),
     public openAsPageButton = page.locator('data-test=open-as-page'),
     public joinSpaceButton = page.locator('data-test=public-bounty-space-action'),
-    public cardDetailProperties = page.locator('data-test=card-detail-properties'),
-    public cardComments = page.locator('data-test=card-page-comments')
+    public cardDetailProperties = page.locator('data-test=card-detail-properties')
   ) {
     super(page);
   }
@@ -53,7 +52,7 @@ export class DocumentPage extends GlobalPage {
   }
 
   getCardCommentContent(commentId: string) {
-    return this.cardComments.locator(`data-test=card-comment-content-${commentId} >> div[contenteditable]`).first();
+    return this.page.locator(`data-test=card-comment-content-${commentId} >> div[contenteditable]`).first();
   }
 
   async isPageEditable() {
--- a/__e2e__/po/document.po.ts
+++ b/__e2e__/po/document.po.ts
@@ -22,7 +22,8 @@ export class DocumentPage extends GlobalPage {
     public documentTitle = page.locator(`data-test=editor-page-title`),
     public openAsPageButton = page.locator('data-test=open-as-page'),
     public joinSpaceButton = page.locator('data-test=public-bounty-space-action'),
-    public cardDetailProperties = page.locator('data-test=card-detail-properties')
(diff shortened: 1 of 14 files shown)


__e2e__/po/document.po.ts
```diff
@@ -22,7 +22,8 @@ export class DocumentPage extends GlobalPage {
     public documentTitle = page.locator(`data-test=editor-page-title`),
     public openAsPageButton = page.locator('data-test=open-as-page'),
     public joinSpaceButton = page.locator('data-test=public-bounty-space-action'),
-    public cardDetailProperties = page.locator('data-test=card-detail-properties')
+    public cardDetailProperties = page.locator('data-test=card-detail-properties'),
+    public cardComments = page.locator('data-test=card-page-comments')
```
mattcasey, line 26: in general, i would not add a data- attribute unless it's an element we are testing. here, it's just a parent selector which really shouldn't matter if it exists or not

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -3,15 +3,20 @@ import styled from '@emotion/styled';
...
 import { useRouter } from 'next/router';
 import { memo, useEffect, useRef, useState } from 'react';
 import { useElementSize } from 'usehooks-ts';
 
 import { PageComments } from 'components/[pageId]/Comments/PageComments';
 import AddBountyButton from 'components/common/BoardEditor/focalboard/src/components/cardDetail/AddBountyButton';
 import CardDetailProperties from 'components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetailProperties';
-import { blockLoad, databaseViewsLoad } from 'components/common/BoardEditor/focalboard/src/store/databaseBlocksLoad';
+import CommentsList from 'components/common/BoardEditor/focalboard/src/components/cardDetail/commentsList';
+import { getCardComments, hasLoadedCardComments } from 'components/common/BoardEditor/focalboard/src/store/comments';
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:14]: all loading props should just be 'isLoading' to keep it standard., or isLoadingCardComments maybe. I am trying to undo the combination we have of 'isLoaded' vs 'isLoading' We should use 'isLoading'

components/common/BoardEditor/focalboard/src/store/comments.ts
```diff
@@ -0,0 +1,77 @@
...
+          .filter((c) => c.parentId === cardId)
+          .sort((a, b) => a.createdAt - b.createdAt)
+      : [];
+  };
+}
+
+export function hasLoadedCardComments(cardId?: string): (state: RootState) => boolean {
+  return (state: RootState): boolean => {
+    if (!cardId) {
+      return false;
```
mattcasey, line 60: this is an example of why we should try to be consistent with only "isLoading". If I am setting "isLoading" to the opposite of "hasLoaded", which we do sometimes, then "isLoading" will be true even though there is no cardId. But isLoading should be false unless a request is actually made.

mattcasey (approved):
Left some comments about isLoading... on a higher level, why are the comments not loaded when we grab the subtree of a card page? I wonder if the rootId or parentId is not being properly set, or if we need to do some refactoring so that DocumentPage has acess to the blocks

PR #2820, opened 2023-09-25
remove data from DeepDAO

mattcasey:
### <samp>🤖 Generated by Copilot at d567f1c</samp>

Removed DeepDao integration and refactored the profile module to use generic types and components for community events. Added a new function to fetch user communities from CharmVerse. Renamed some API endpoints and files to use consistent terminology. Updated the test space domain name.

### WHY
Data points are not what we are looking for.


Devorein (approved):

Devorein:
image
The organizations tab is not present anymore after the refactoring. When I click on my own card in member directory it takes me to edit profile modal, but there is no section where I can toggle visibility of my spaces.

PR #2824, opened 2023-09-26
Fix external links

msadura:
### <samp>🤖 Generated by Copilot at a6ff4c8</samp>

Improved and tested the `Link` component and added a `isExternalUrl` utility function. The `Link` component can now handle external links more reliably by using the `isExternalUrl` function to check the `href` prop. The `isExternalUrl` function parses and validates urls or paths and returns a boolean value. The `Link` component and the `isExternalUrl` function are both covered by unit tests in `Link.spec.tsx` and `isExternalUrl.spec.ts`, respectively.

### WHY
Fix external lens links (and all external link using our atomic Link component without need to add `external` prop)


motechFR (approved):
Great idea!

PR #2825, opened 2023-09-26
Fix additionalPaths for duplicated page

msadura:
### <samp>🤖 Generated by Copilot at 5f6beb2</samp>

Renamed and updated a module for duplicating pages within a workspace. Added an option to reset the additional paths of the duplicated pages based on their titles.

### WHY
Fix additionalPaths when page is being duplicated

### DIFF
--- a/lib/templates/importWorkspacePages.ts
+++ b/lib/templates/importWorkspacePages.ts
@@ -11,7 +11,7 @@ import { v4, validate } from 'uuid';
 import type { BountyWithDetails } from 'lib/bounties';
 import { isBoardPageType } from 'lib/pages/isBoardPageType';
 import { createPage } from 'lib/pages/server/createPage';
-import { getPagePath } from 'lib/pages/utils';
+import { generatePagePathFromPathAndTitle, getPagePath } from 'lib/pages/utils';
 import type { PageContent, TextContent, TextMark } from 'lib/prosemirror/interfaces';
 import { InvalidInputError } from 'lib/utilities/errors';
 import { typedKeys } from 'lib/utilities/objects';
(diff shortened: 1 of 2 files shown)


lib/templates/importWorkspacePages.ts
```diff
@@ -293,6 +295,15 @@ export async function generateImportWorkspacePages({
       }
     };
 
+    if (resetAdditionalPaths && newPageContent.data.additionalPaths) {
+      const additionalPath = generatePagePathFromPathAndTitle({
```
motechFR, line 299: Complement to above; How about a "resetPaths" option? This way we preserve paths when copying from templates, but always generate a new page path when calling from duplicatePage

In all cases, we never port over additionalPaths

lib/templates/importWorkspacePages.ts
```diff
@@ -293,6 +295,15 @@ export async function generateImportWorkspacePages({
       }
     };
 
+    if (newPageContent.data.additionalPaths) {
+      const additionalPath = generatePagePathFromPathAndTitle({
+        existingPagePath: newPageContent.data.path,
+        title: newPageContent.data.title as string
+      });
+
+      newPageContent.data.additionalPaths = resetPaths ? [additionalPath] : [];
```
motechFR, line 304: Here's what I meant

```typescript
if (resetPaths) {
  const newPath = generatePagePathFromPathAndTitle({
      existingPagePath: newPageContent.data.path,
      title: newPageContent.data.title as string
  })
 newPageContent.data.path = newPath;
}

// Always reset additional paths
newPageContent.data.additionalPaths = [];
```

motechFR, line 304: Here's what I meant

```typescript
if (resetPaths) {
  const newPath = generatePagePathFromPathAndTitle({
      existingPagePath: newPageContent.data.path,
      title: newPageContent.data.title as string
  })
 newPageContent.data.path = newPath;
}

// Always reset additional paths
newPageContent.data.additionalPaths = [];
```

PR #2837, opened 2023-09-27
Feat/discussion nofitications matt

mattcasey:
### <samp>🤖 Generated by Copilot at 85dcdae</samp>

No summary available (Limit exceeded: required to process 134991 tokens, but only 50000 are allowed per call)


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -199,6 +201,28 @@ function PropertyValueElement(props: Props) {
         }
         onChange={(newValue) => {
           mutator.changePropertyValue(card, propertyTemplate.id, newValue);
+          const previousValue = propertyValue
```
mattcasey, line 204: we might need this to be enabled via prop, propertyValueElement may be used by bounties, proposals, etc. which are not actually cards

Devorein (approved):

PR #2840, opened 2023-09-28
Remove space domain from pathname in page_view event

msadura:
### <samp>🤖 Generated by Copilot at 863c18f</samp>

Added space domain tracking to various page and dialog components. This involved updating the `trackPageView` function and the `ViewPageEvent` type to accept the `spaceDomain` as an optional parameter. The `useCurrentSpace` hook was used to get the current space domain from the browser path.

### WHY
Do not send space domain in pathname to avoid issues with redirecting after user changes domain name

### DIFF
--- a/charmClient/hooks/track.ts
+++ b/charmClient/hooks/track.ts
@@ -9,10 +9,13 @@ import { TrackApi } from '../apis/trackApi';
 const track = new TrackApi();
 
 export function trackPageView(page: Omit<PageEventMap['page_view'], 'userId'>) {
+  const fullPath = getBrowserPath();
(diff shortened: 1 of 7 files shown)


mattcasey (approved):

mattcasey (changes requested):
I didn't notice any changes to address the redirect part, where we need to append the space domain.  But maybe the code handles that already? I just marked as 'request changes' to make sure it's been tested.. :)

msadura:
> I didn't notice any changes to address the redirect part, where we need to append the space domain. But maybe the code handles that already? I just marked as 'request changes' to make sure it's been tested.. :)

yeah code handles that already - if domain is not present in path, it will be added

PR #2863, opened 2023-10-03
fix markdown parse and export of new lists

mattcasey:
### <samp>🤖 Generated by Copilot at b387ac7</samp>

No summary available (Limit exceeded: required to process 124447 tokens, but only 50000 are allowed per call)


motechFR (approved):
tested, works great

PR #2871, opened 2023-10-03
fix lookup for public pages on custom domains

mattcasey:
### <samp>🤖 Generated by Copilot at e8b5182</samp>

Enhanced the public pages API to handle custom domains better and query pages faster. Refactored the `getPublicPage` function in `pages/api/public/pages/[...pageId]/index.ts`.

### WHY
If you are logged in and visit a public page on custom domain, RouteGuard redirects you to /join. I thought the bug wsa in route guard, but then I realized that public/pages/[...pageId] was returning 404 (space not found), so there was no public page loaded

### DIFF
--- a/pages/api/public/pages/[...pageId]/index.ts
+++ b/pages/api/public/pages/[...pageId]/index.ts
@@ -160,11 +160,7 @@ async function getPublicPage(req: NextApiRequest, res: NextApiResponse<PublicPag
       : {
           OR: [
             {
-              // TODO: ask Marek why we need to support case-insensitivity for custom domains
-              customDomain: {
-                equals: spaceDomain,
-                mode: 'insensitive'
-              }
+              customDomain: spaceDomain
             },
             { domain: spaceDomain }
           ]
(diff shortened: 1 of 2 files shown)


pages/api/public/pages/[...pageId]/index.ts
```diff
@@ -154,8 +154,21 @@ async function getPublicPage(req: NextApiRequest, res: NextApiResponse<PublicPag
...
 
-  const space = await prisma.space.findUnique({
-    where: page ? { id: page.spaceId } : { domain: spaceDomain }
+  const space = await prisma.space.findFirst({
+    where: page
+      ? { id: page.spaceId }
+      : {
+          OR: [
+            {
+              // TODO: ask Marek why we need to support case-insensitivity for custom domains
```
Comment (msadura) [inline pages/api/public/pages/[...pageId]/index.ts:163]: in the past there was a bug you could create case-sensitive domain. I fixed that some time ago and migrated all the space domains so this should not be needed anymore

msadura (approved):

pages/api/public/pages/[...pageId]/index.ts
Comment (mattcasey) [reply pages/api/public/pages/[...pageId]/index.ts]: thanks, removed my comment!

PR #2872, opened 2023-10-04
Patch Metamask connector

motechFR:
### <samp>🤖 Generated by Copilot at c018ca4</samp>

Added features and fixed bug for web3 account connection. Used `injectedConnector` and `useConnect` to handle wallet connection, and added `useEffect` to reconnect on account change or unlock.

### WHY
Patch Metamask being disconnected.

We need to hook wagmi back into metamask if the wallet was locked on load.

### DIFF
--- a/hooks/useWeb3Account.tsx
+++ b/hooks/useWeb3Account.tsx
@@ -2,6 +2,7 @@ import { log } from '@charmverse/core/log';
 import type { UserWallet } from '@charmverse/core/prisma';
 import type { Web3Provider } from '@ethersproject/providers';
 import { verifyMessage } from '@ethersproject/wallet';
+import { injectedConnector } from 'connectors/config';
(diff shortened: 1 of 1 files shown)


hooks/useWeb3Account.tsx
```diff
@@ -263,6 +266,30 @@ export function Web3AccountProvider({ children }: { children: ReactNode }) {
...
+        }
+      };
+
+      window.ethereum.on('accountsChanged', handleAccountsChanged);
+
+      return () => {
+        window.ethereum.removeListener('accountsChanged', handleAccountsChanged);
+      };
+    }
+  }, [user]);
```
mattcasey, line 291: should just watch user.wallets, maybe even user.wallets.length?

mattcasey (approved):
seems fine.

PR #2873, opened 2023-10-04
Notion UX for notifications

Devorein:
### <samp>🤖 Generated by Copilot at 04370f1</samp>

This pull request adds a new notifications feature to the app that allows users to receive and manage notifications related to various pages and actions in the app. It implements the notifications logic in the `CharmClient`, the webhook publisher, and the backend API. It also creates new types and interfaces for different kinds of notifications and updates the existing ones. It adds a new popover component for notifications to the sidebar and a new custom component to the multi-tabs component. It refactors some of the existing code for clarity and consistency.


mattcasey:
Users should see either Updates or the Alarm in the app bar, not both... but since we are about to have new notifications for everyone, you could just remove the Alarm?
image

mattcasey:
Individual messages need to follow the layout from Notion:
image
- smaller user avatar
- 3 rows: action, page, then space

mattcasey:
Trying to archive gives error: `Error: Unknown arg `archived` in data.archived for type UserNotificationMetadataUpdateManyMutationInput`

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
```diff
@@ -0,0 +1,556 @@
...
+              }}
+            >
+              <IconButton
+                disabled={unArchivedNotifications.length === 0}
+                color={inboxState === 'unread' ? 'primary' : 'secondary'}
+                onClick={() => {
+                  setInboxState(inboxState === 'unarchived' ? 'unread' : 'unarchived');
+                }}
+              >
+                <PlaylistAddCheckCircleOutlinedIcon fontSize='small' />
```
mattcasey, line 220: I thought this meant 'mark as read' since ith as a checkmark. I think "FilterList" is more accurate and closer to notion here. Even closer would be "IoFilterCircleOutline" from react-icons/io5

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
```diff
@@ -0,0 +1,556 @@
...
+        : unreadNotifications
+      : archivedNotifications;
+
+  return (
+    <MultiTabs
+      endAdornmentComponent={
+        activeTab === 'Inbox' && (
+          <Tooltip
+            title={
+              unArchivedNotifications.length === 0
```
mattcasey, line 201: The icon is always active, so tooltip should ONLY be: "Show all" or "Show unread". no reason to hide the tooltip ever

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
```diff
@@ -0,0 +1,556 @@
...
+                    onClick={() => {
+                      if (!markAllReadButtonDisabled) {
+                        markNotifications({
+                          state: 'read',
+                          ids: unreadNotifications.map((notification) => notification.taskId)
+                        });
+                      }
+                    }}
+                  >
+                    <Button
```
mattcasey, line 251: Not sure why buttons are wrapped in a div and tooltip, please use the button props, it should handle this for you:
`<Button disabled={...} disabledTooltip='All notifications have been marked as read' onClick={} />`

lib/webhookPublisher/publisher.ts
```diff
@@ -3,6 +3,7 @@ import type { Space } from '@charmverse/core/prisma';
 import { prisma } from '@charmverse/core/prisma-client';
 
 import { addMessageToSQS } from 'lib/aws/SQS';
+import { createNotificationsFromEvent } from 'lib/notifications/createNotificationsFromEvent';
```
mattcasey, line 6: import is not used

pages/api/notifications/mark.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    case 'archived': {
+      await prisma.userNotificationMetadata.updateMany({
+        where: {
+          id: {
+            in: ids
+          }
+        },
+        data: {
+          archived: true,
```
mattcasey, line 44: 'archivedAt' would have been better, since it would be closer to our seenAt/deletedAt pattern

pages/api/notifications/mark.ts
```diff
@@ -0,0 +1,89 @@
...
+      });
+      break;
+    }
+
+    default: {
+      break;
+    }
+  }
+
+  return res.status(200).json({ ok: true });
```
mattcasey, line 86: just use .end() instead of sending a JSON object

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
Devorein: Great suggestion. I totally forgot about `react-icons`. Thank you and updated.

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
Devorein: Actually when you don't have any unarchived notifications its disabled and thus the tooltip is empty

image

components/common/PageLayout/components/Sidebar/components/NotificationsPopover.tsx
Devorein: I completely forgot about `disabledTooltip` prop. Added a div so that even when the button is disabled the tooltip will be shown. Also kept the div since that was needed to ensure both the buttons are equal width (since we wrap the buttons in a span with `disabledTooltip` prop)

lib/webhookPublisher/publisher.ts
Devorein: Removed. Thnx.

PR #2891, opened 2023-10-06
Ux proposal connect

motechFR:
### <samp>🤖 Generated by Copilot at 8427e3e</samp>

This pull request adds and integrates a new component for selecting a wallet to connect to the web3 provider in various scenarios where the user needs to interact with snapshot proposals. It also enhances the logic and UI for displaying the voting disabled status for snapshot proposals. It affects the files `WalletSelectorModal.tsx`, `SnapshotVoteDetails.tsx`, `PublishingForm.tsx`, and `useSnapshotVoting.ts`.

### WHY
Add a button to connect your wallet inside proposals and publish to snapshot

### DIFF
--- a/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
+++ b/components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
@@ -146,12 +146,3 @@ export function WalletSelectorModal() {
     </Modal>
   );
 }
-
-export function OpenWalletSelectorModal({ color }: { color?: string }) {
-  const { connectWallet } = useWeb3ConnectionManager();
-  return (
-    <Button variant='outlined' onClick={connectWallet} color={color}>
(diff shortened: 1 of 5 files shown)


components/_app/Web3ConnectionManager/components/WalletSelectorModal/WalletSelectorModal.tsx
```diff
@@ -145,3 +146,12 @@ export function WalletSelectorModal() {
     </Modal>
   );
 }
+
+export function OpenWalletSelectorModal({ color }: { color?: string }) {
```
msadura, line 150: 1) Please avoid exporting multiple components from the same file
2) I think better name for it would be`OpenWalletSelectorButton` as component itself is just a button

motechFR: Agreed, cleaner 👍

PR #2902, opened 2023-10-09
fix redirect when path includes space domain

mattcasey:
### <samp>🤖 Generated by Copilot at 2f17878</samp>

This pull request adds and tests a new utility function `getSubdomainPath` to handle subdomain-based routing in the app. It also fixes a bug and improves the flexibility of the same function in `lib/session/getDefaultPageForSpace.ts`. The other changes are minor formatting adjustments.

### DIFF
--- a/lib/utilities/browser.ts
+++ b/lib/utilities/browser.ts
@@ -252,7 +252,7 @@ export function fullyDecodeURI(uri: string) {
 // strip out custom or domain depending on the host
 export function getSubdomainPath(
   path: string,
-  config?: { domain: string; customDomain: string | null },
+  config?: { domain: string; customDomain?: string | null },
   host?: string
 ) {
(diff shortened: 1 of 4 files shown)


lib/utilities/browser.ts
```diff
@@ -273,9 +273,8 @@ export function getSubdomainPath(
   if (subdomain) {
     return path.replace(new RegExp(`^\\/${subdomain}`), '');
   }
-
   // if we are not using a custom domain or subdomain, make sure that the space domain exists in the URL
-  if (config && !customDomain && !path.startsWith(`/${config?.domain}`)) {
+  if (config && !customDomain && !path.startsWith(`/${config?.domain}/`)) {
```
mattcasey, line 277: fix is here..  a space with 'charmverse' as the domain and a page with a path like '/charmverse-1232132' would look like the domain was already in the path

PR #2906, opened 2023-10-09
Billing/extend block counts

motechFR:
### <samp>🤖 Generated by Copilot at b46261e</samp>

Refactored and improved the block counting functionality for spaces. Added new count functions for different block types, such as comments, forum, database, and member properties. Moved the count functions to a separate folder and updated the import paths. Added test files for the new count functions and a utility function for paginated prisma tasks. Deleted the old `countSpaceBlocks.ts` file.

### WHY
Clean up how we count blocks by splitting into specific methods.


lib/spaces/countSpaceBlocks/countSpacePages.ts
```diff
@@ -0,0 +1,76 @@
...
+    details: {
+      documents: 0,
+      rewards: 0,
+      proposals: 0,
+      databases: 0,
+      cards: 0
+    }
+  };
+
+  pageCounts.details.documents = await prisma.page.count({
```
mattcasey, line 28: these should all be run in parallel still.

lib/utilities/paginatedPrismaTask.ts
```diff
@@ -0,0 +1,92 @@
...
+type BatchProcessor<R> = (records: any[]) => R | Promise<R>;
+
+/**
+ * @param batchSize - Number of records to fetch at a time
+ */
+type PaginatedTask<M extends PrismaModel, R, V> = {
+  model: M;
+  queryOptions: PrismaArgs<M>;
+  batchSize?: number;
+  callback: BatchProcessor<R>;
```
mattcasey, line 30: 'callback' is confusing here because a reducer is a type of callback. When does the callback get run? would a better name be "onComplete? or "onSuccess"? how are errors propagated?

lib/utilities/paginatedPrismaTask.ts
```diff
@@ -0,0 +1,92 @@
...
+    throw new InvalidInputError(`Query options for ${model} must select at least id or include additional records`);
+  }
+
+  const prisma = new PrismaClient();
+
+  let cursor: string | undefined;
+
+  const allProcessedResults: R[] = [];
+
+  while (true) {
```
mattcasey, line 65: can this be rewritten without while true? that is very dangerous

lib/spaces/countSpaceBlocks/__tests__/countCommentBlocks.spec.ts
```diff
@@ -0,0 +1,227 @@
...
+          title: ' ',
+          schema: 1,
+          type: 'comment',
+          updatedBy: user.id,
+          id: uuid()
+        }
+      ]
+    });
+
+    const counts = await countCommentBlocks({ spaceId: space.id, batchSize: 2 });
```
mattcasey, line 213: why batch size of 2? that seems pretty expensive overhead in trips to the database

lib/utilities/paginatedPrismaTask.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    allProcessedResults.push(processedResults);
+
+    cursor = results[results.length - 1].id;
+
+    if (results.length < batchSize) break;
+  }
+
+  if (reducer) {
+    return reducer(allProcessedResults);
```
mattcasey, line 89: i dont think reducer belongs in this method, it's not doing any async stuff. this method feels overly complicated at the moment

mattcasey (changes requested):
i like the idea to organize but paginating requests was not part of the original scope, and i dont think this is a current problem that needed solving at the moment. i'd rather you have gotten proposal properties done than spend time on trying to optimize queries next time. the utility is a little too opinionated at the moment, and the batch sizes are too low imo

lib/utilities/paginatedPrismaTask.ts
motechFR: Fixed everything based on our chat today. We now have a mapper per record, an an optional onSuccess

lib/spaces/countSpaceBlocks/__tests__/countCommentBlocks.spec.ts
motechFR: Fixed to 10, thanks for spotting

lib/utilities/paginatedPrismaTask.ts
motechFR: Fixed! Went with recursion instead

lib/spaces/countSpaceBlocks/countSpacePages.ts
motechFR: Fair enough, let's see how it goes and we can always re-parallelise

mattcasey (approved):

PR #2907, opened 2023-10-09
adding concurrency grouping to organize the order of deployment. 

yiwork:
This will not cancel any in-progress workflow run, will just put the workflow run in queue, and will cancel any previously already enqueued and pending workflow run

### <samp>🤖 Generated by Copilot at 1f3ebb5</samp>

Add concurrency groups to various GitHub workflows to prevent redundant or conflicting runs. This improves the efficiency and reliability of the CI/CD pipeline for the app.

### DIFF
--- a/.github/workflows/deploy_staging.yml
+++ b/.github/workflows/deploy_staging.yml
@@ -7,6 +7,9 @@ on:
 env:
   SELECTED_PERMISSION_API_TAG: latest
 
+concurrency:
+  group: staging-${{ github.ref }}
+
 jobs:
   deploy:
     if: |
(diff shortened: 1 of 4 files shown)


mattcasey (approved):

.github/workflows/deploy_staging.yml
```diff
@@ -7,6 +7,9 @@ on:
 env:
   SELECTED_PERMISSION_API_TAG: latest
 
+concurrency:
+  group: staging-${{ github.ref }}
```
mattcasey, line 11: how is this different from the concurrency rule that exists already?

PR #2912, opened 2023-10-10
add migration script to remove notifyNewProposals

mattcasey:
### <samp>🤖 Generated by Copilot at a056384</samp>

This pull request removes the deprecated `notifyNewProposals` field from the space model and related components, scripts, and tests. It also adds a migration script to update existing spaces with the new notification toggles for proposal events.

### DIFF
--- a/lib/proposal/getProposalAction.ts
+++ b/lib/proposal/getProposalAction.ts
@@ -12,28 +12,22 @@ export type ProposalTaskAction =
 export function getProposalAction({
   isAuthor,
   currentStatus,
-  isReviewer,
-  notifyNewProposals,
-  notifyNewVotes
+  isReviewer
 }: {
   currentStatus: ProposalStatus;
   isAuthor: boolean;
   isReviewer: boolean;
-  notifyNewProposals: boolean;
-  notifyNewVotes: boolean;
 }): ProposalTaskAction | null {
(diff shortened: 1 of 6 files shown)


lib/proposal/getProposalAction.ts
```diff
@@ -12,28 +12,22 @@ export type ProposalTaskAction =
 export function getProposalAction({
   isAuthor,
   currentStatus,
-  isReviewer,
-  notifyNewProposals,
```
Devorein, line 16: if we remove this it will always send new proposal and vote active notification for all members regardless of whether they have that toggled in space notification

scripts/migrations/2023_10_10_remove_notifyNewProposals.ts
```diff
@@ -0,0 +1,20 @@
...
+async function migrateSpaces() {
+  const r = await prisma.space.updateMany({
+    where: {
+      notifyNewProposals: {
+        not: null
+      }
+    },
+    data: {
+      notificationToggles: {
+        proposals__start_discussion: false,
```
Devorein, line 13: Shouldn't the query be `notifyNewProposals: null`? Since not null means they want to be notified for these proposal actions.

Devorein (changes requested):

PR #2939, opened 2023-10-16
Space specific notifications UI

Devorein:
### <samp>🤖 Generated by Copilot at b4d5f61</samp>

This pull request adds and improves the notifications feature for the app.charmverse.io website. It displays the number of notifications on the header menu icon, the sidebar submenu, and the space avatars. It also filters the notifications by the current space and uses hooks to fetch and update the notifications data. It affects the files `Header.tsx`, `NotificationsPopover.tsx`, `SidebarSubmenu.tsx`, `SpaceListItem.tsx`, and `Sidebar.tsx`.

### DIFF
--- a/components/common/Avatar.tsx
+++ b/components/common/Avatar.tsx
@@ -61,6 +61,11 @@ const sizeVariantStyleMap: Partial<Record<AvatarSize, Record<AvatarVariant, Reac
     rounded: { borderRadius: '4px' },
     circular: null,
     square: null
+  },
+  xSmall: {
+    rounded: { borderRadius: '4px' },
+    circular: null,
+    square: null
   }
 };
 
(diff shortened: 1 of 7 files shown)


mattcasey (changes requested):
Functionally seems to work!
The UI is not quite the same as Notion:
- the number next to current space should only include count from other spaces, not the one i am looking at. For example, I currently have 5 and 2 notifications in two spaces, but it always says "7".
- the number next to current space should always be visible, not just on hover
- please make the space icon the same size as other icons below it, just like Notion. This gives us more room for space title now that the notification icon takes up room
- when the collapse icon is visible, we should give it a background so that it just completely hides the close button below it, instead of this ugly overlap:
image

mattcasey (changes requested):
see other comments

PR #2940, opened 2023-10-16
Send poll notification for forum posts

Devorein:
### WHAT
copilot:summary


lib/votes/extractPollIds.ts
mattcasey, line 1: Lets keep all prosemiror-related stuff inside lib/prosemirror or CharmEditor.

mattcasey (changes requested):
look good, but please keep prosemirror code in one place

lib/votes/extractPollIds.ts
Devorein: Oops. My bad. Thank you for reminding. Updated.

PR #2941, opened 2023-10-16
Enable renaming title column in boards

Devorein:
### WHAT
copilot:summary

### WHY
Also fixed inserting a column left and right to the selected column

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
@@ -63,7 +63,9 @@ function TableHeader(props: Props): JSX.Element {
   const columnWidth = (_templateId: string): number => {
     return Math.max(Constants.minColumnWidth, (activeView.fields.columnWidths[_templateId] || 0) + props.offset);
   };
-  const disableRename = proposalPropertyTypesList.includes(type as any) || DEFAULT_BLOCK_IDS.includes(templateId);
+  const disableRename =
+    proposalPropertyTypesList.includes(type as any) ||
+    DEFAULT_BLOCK_IDS.filter((blockId) => blockId !== Constants.titleColumnId).includes(templateId);
 
   const [tempName, setTempName] = useState(props.name || '');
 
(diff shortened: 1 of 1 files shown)


components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
```diff
@@ -63,7 +63,9 @@ function TableHeader(props: Props): JSX.Element {
   const columnWidth = (_templateId: string): number => {
     return Math.max(Constants.minColumnWidth, (activeView.fields.columnWidths[_templateId] || 0) + props.offset);
   };
-  const disableRename = proposalPropertyTypesList.includes(type as any) || DEFAULT_BLOCK_IDS.includes(templateId);
+  const disableRename =
+    proposalPropertyTypesList.includes(type as any) ||
+    DEFAULT_BLOCK_IDS.filter((blockId) => blockId !== Constants.titleColumnId).includes(templateId);
```
mattcasey, line 68: .filter.includes() is hard to read and logically follow. i would just change it to: `DEFAULT_BLOCK_IDS.includes(templateId) || templateId === Constants.titleColumnId`

components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
```diff
@@ -185,15 +187,13 @@ function TableHeader(props: Props): JSX.Element {
...
-
-              // const index = board.fields.cardProperties.findIndex((o: IPropertyTemplate) => o.id === templateId)
-              mutator.insertPropertyTemplate(board, activeView, index);
-            }
+            const containsTitle = activeView.fields.visiblePropertyIds.includes(Constants.titleColumnId);
+            const index = (
+              containsTitle
+                ? activeView.fields.visiblePropertyIds
+                : [Constants.titleColumnId, ...activeView.fields.visiblePropertyIds]
+            ).findIndex((i) => i === templateId);
```
mattcasey, line 195: this code is confusing - I think it figures out if the title exists in the list, and if not create a NEW List so that index is '0'. 
Less steps and more readable:
```
let index = activeView.fields.visiblePropertyIds.findIndex((i) => i === templateId);
if (index === Constants.titleColumnId && index === -1) {
  index = 0; // handle index for title column
}
  ```
I also think a comment would be good here on why we use 0. Does it not cause a problem with the true property that is at index 0?

mattcasey:
Tested and it works. left some suggestions for code cleanup

mattcasey (approved):

components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
Devorein: Good catch. Updated.

components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
Devorein: No it doesn't since the logic inside `mutator.insertPropertyTemplate` also considers the title property at index 0 if it doesn't exist in the `visiblePropertyIds` array.

PR #2959, opened 2023-10-18
standardize the sorting of board views / fix default board

mattcasey:
### <samp>🤖 Generated by Copilot at af0c0b4</samp>

This pull request refactors and improves the view selectors and components in the BoardEditor and CharmEditor modules. It makes the selectors more specific and consistent with the board and page contexts, and adds missing dependencies to the useEffect hooks. It also fixes some minor formatting issues in `mutator.ts`.

### WHY
The order of views in the sidebar is just alphabetical, it ignores how a user has organized the tabs. Also, when we select a default view, we are not respecting the order of views.


motechFR (approved):
Nice cleanup 👍

PR #2970, opened 2023-10-19
Feat/save new reward

msadura:
### <samp>🤖 Generated by Copilot at 6e295c0</samp>

Added a new page dialog feature to the app, which allows creating new pages with a charm editor and custom properties. Refactored the rewards page and the reward properties component to use the new page dialog and a new reward hook. Moved and exported some types and components for reuse and consistency. Removed unused code and fixed some imports.

### WHY
Save new reward with a button instead of right away.
Introduced NewPageDialog which is not global but can be used to present any kind of new page (we can port proposal to use it)


components/rewards/components/NewRewardButton.tsx
```diff
@@ -1,37 +1,31 @@
...
   return (
-    <Button data-test='create-suggest-bounty' onClick={onClickCreate}>
-      Create
-    </Button>
+    <>
+      <Button data-test='create-suggest-bounty' onClick={onClickCreate}>
+        Create
+      </Button>
+
+      <NewPageDialog onClose={clearFormInputs} onSave={createReward} isSaving={isSavingReward}>
```
msadura, line 26: this is how you compose new page dialog

components/rewards/components/RewardProperties/RewardPropertiesForm.tsx
```diff
@@ -0,0 +1,337 @@
+import { Box, Divider, Stack, Tooltip } from '@mui/material';
```
msadura, line 1: this is jsut moved logic from RewardProperties

lib/rewards/interfaces.ts
```diff
@@ -28,3 +30,31 @@ export type RewardWithUsersAndPageMeta = Bounty & {
...
+    | 'customReward'
+    | 'fields'
+  >
+> & { reviewers?: RewardReviewer[]; allowedSubmitterRoles?: string[] | null };
+
+export type RewardPageProps = Partial<
+  Pick<Page, 'title' | 'content' | 'contentText' | 'sourceTemplateId' | 'headerImage' | 'icon'>
+>;
+
+export type RewardCreationData = UpdateableRewardFields & {
```
mattcasey, line 53: Why was RewardCreationData removed from the code that creates rewards? It seems pretty specific

lib/rewards/interfaces.ts
```diff
@@ -28,3 +30,31 @@ export type RewardWithUsersAndPageMeta = Bounty & {
 export type ApplicationWithTransactions = Application & {
   transactions: Transaction[];
 };
+
+export type UpdateableRewardFields = Partial<
```
mattcasey, line 34: this belongs next to the lib method that updates rewards

lib/proposal/createProposal.ts
```diff
@@ -109,15 +109,15 @@ export async function createProposal({
     createPage({
       data: {
         content: pageProps?.content ?? undefined,
-        createdBy: userId,
         contentText: pageProps?.contentText ?? '',
         headerImage: pageProps?.headerImage,
+        sourceTemplateId: pageProps?.sourceTemplateId,
```
mattcasey, line 114: these fields were actually sorted alphabetically before, please revert the order

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -165,6 +166,7 @@ function DocumentPage({ page, refreshPage, savePage, readOnly = false }: Documen
 
   const router = useRouter();
   const isSharedPage = router.pathname.startsWith('/share');
+  const isRewardsPage = router.pathname.endsWith('/rewards');
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:169]: i'm assuming we can remove this once rewards are deployed? can you add a TODO comment

components/[pageId]/DocumentPage/components/PageHeader.tsx
```diff
@@ -62,25 +62,25 @@ const EditorHeader = styled.div<{ pageType: 'page' | 'board' }>`
...
-type PageHeaderValues = Partial<Pick<Page, 'title' | 'icon' | 'headerImage' | 'updatedAt'>>;
+export type PageHeaderValues = Partial<Pick<Page, 'title' | 'icon' | 'headerImage' | 'updatedAt'>>;
 
 type PageHeaderProps = {
   headerImage: string | null;
   icon: string | null;
   readOnly: boolean;
   title: string;
-  setPage: (p: PageHeaderValues) => void;
+  setPage: ((p: PageHeaderValues) => void) | null;
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageHeader.tsx:72]: i think a noop method would be a cleaner default

Comment (mattcasey) [reply components/[pageId]/DocumentPage/components/PageHeader.tsx:72]: On reviewing the code it looks like there's no need for this to be optional

components/common/PageDialog/components/NewPageDocument.tsx
```diff
@@ -0,0 +1,75 @@
...
+              {/* temporary? disable editing of page title when in suggestion mode */}
+              <PageHeader
+                headerImage={newPageValues.headerImage || null}
+                icon={newPageValues.icon || null}
+                updatedAt={new Date().toString()}
+                title={newPageValues.title || ''}
+                readOnly={false}
+                setPage={updateNewPageValues}
+              />
+              <div className='focalboard-body font-family-default'>
```
mattcasey, line 66: feels like we should be using CardPropertiesWrapper here

components/common/PageDialog/NewPageDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <Dialog
+      onClose={() => {
+        if (contentUpdated) {
+          setShowConfirmDialog(true);
+        } else {
+          close();
+        }
+      }}
+      toolbar={<div />}
```
mattcasey, line 55: toolbar should be optional

components/rewards/components/NewRewardButton.tsx
```diff
@@ -1,37 +1,31 @@
...
+        <RewardPropertiesForm onChange={setFormInputs} values={formInputs} />
```
mattcasey, line 27: not sure i am a fan of this pattern, because it requires so much context for card properties and other things. it's nice that right now Documentpage gets both the data and the components you need. That is very easy to follow. The only complaint i have with it is there's lots of if/else. 

So a better pattern might be something like RewardsDocumentPage, but i would not compose page properties where the dialog is added

mattcasey (approved):
it works, so i approved. but i am not sure about this approach. i don't know how it will work for normal DocumentPage, and i think we will end up copy/pasting a lot of the same code. The reason DocumentPage does so much is that it is used in 4 different places, 2 of which need to work for proposals, bounties, cards and page types (SharedPage and EditorPage)

lib/rewards/interfaces.ts
msadura: ok I used interfaces for types reusable in multiple places. I will revert htis change

components/common/PageDialog/NewPageDialog.tsx
msadura: in theory it is optional, but breaks header layout. I can fix that in dialog anyway

PR #2973, opened 2023-10-20
Fix the layout of users and roles

mattcasey:
### <samp>🤖 Generated by Copilot at 99f58cf</samp>

Refactored a common component for selecting users and roles in the board editor. Extracted the logic for rendering the selected options into a separate component called `SelectedOptions`. Simplified the usage of the `UserAndRoleSelect` component by removing an unnecessary prop.

### DIFF
--- a/components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
+++ b/components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
@@ -56,14 +56,14 @@ export const StyledUserPropertyContainer = styled(Box, {
   overflow: ${(props) => (props.displayType === 'table' ? 'hidden' : 'initial')};
 `;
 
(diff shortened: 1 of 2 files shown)


components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
```diff
@@ -152,7 +153,7 @@ export function UserAndRoleSelect({
   variant = 'standard',
   value: inputValue,
   'data-test': dataTest,
-  wrapColumn = true,
+  wrapColumn,
```
mattcasey, line 156: I removed a default 'true' value that was added. Seems dangerous to use this, especially since we have it optional and default false everywhere else.

msadura (approved):

PR #2983, opened 2023-10-23
Drag and drop nested pages within each other in editor

Devorein:
### <samp>🤖 Generated by Copilot at f36fd2f</samp>

This pull request enhances the charm editor by adding drag and drop support for pages between editors and the sidebar. It also improves the code quality and consistency of the websockets and plugins modules. The changes affect the files `components/common/CharmEditor/plugins.ts`, `lib/websockets/spaceEvents.ts`, `components/common/CharmEditor/CharmEditor.tsx`, `lib/websockets/interfaces.ts`, and `theme/@bangle.dev/styles.scss`.


lib/websockets/spaceEvents.ts
```diff
@@ -549,15 +628,15 @@ async function getPageDetails({
...
+  parentDocumentNode.forEach((node, nodePos) => {
     if (node.type.name === 'page' && node.attrs.id === id) {
       position = nodePos;
       return false;
     }
   });
 
   return {
-    pageNode,
+    parentDocumentNode,
```
mattcasey, line 639: Not sure this variable name is any better. "parent" implies there is a child, but we are dealing with the node of the page/document. 'documentNode' might be ok?

lib/websockets/spaceEvents.ts
```diff
@@ -21,6 +21,8 @@ import type { DocumentEventHandler } from './documentEvents/documentEvents';
 import type { ProsemirrorJSONStep } from './documentEvents/interfaces';
 import type { AbstractWebsocketBroadcaster } from './interfaces';
 
+const nonContentPageTypes = ['board', 'board_template', 'inline_board', 'linked_board', 'inline_linked_board'];
```
mattcasey, line 24: Can we avoid adding another configuration object? These are easy to forget about and I think we have something similar. I think this is the opposite of `documentTypes` in the DocumentPageActionList? We just need to make it an array first, so it can be used at runtime

lib/websockets/spaceEvents.ts
```diff
@@ -221,79 +229,114 @@ export class SpaceEventHandler {
...
+        dragPos,
+        isLinkedPage,
+        pageId,
+        newParentId,
+        newIndex,
+        trigger,
+        dropPos
+      } = message.payload;
+
+      if (pageId === newParentId) {
```
mattcasey, line 243: maybe add a comment explaining why we return null?

lib/websockets/interfaces.ts
```diff
@@ -128,8 +128,11 @@ type PageReordered = {
     pageId: string;
     newParentId: string | null;
     newIndex: number;
-    trigger: 'sidebar-to-sidebar' | 'sidebar-to-editor';
-    pos?: number;
+    trigger: 'sidebar-to-sidebar' | 'sidebar-to-editor' | 'editor-to-editor';
```
mattcasey, line 131: these should just be different events. that way you don't need optional fields. it is hard to know what these fields mean - dragPos vs dropPos, for example.

components/common/CharmEditor/plugins.ts
```diff
@@ -108,6 +116,54 @@ export function charmEditorPlugins({
 
             const data = ev.dataTransfer.getData('sidebar-page');
             if (!data) {
+              const domElementAtCoords = document.elementFromPoint(ev.clientX, ev.clientY);
```
mattcasey, line 119: this belongs in its own plugin

lib/websockets/spaceEvents.ts
Devorein: This is actually the parent document node, as in the parent of the id of the page passed to the function, I got confused with pageNode as I didn't realize whether its `page(id).content` or `page(id).parent.content`. This make it a bit clearer as to which document it is.

lib/websockets/spaceEvents.ts
Devorein: Ooo, nice catch. Sorry I forgot about that one, I will remove this one and use the one inside `DocumentPageActionList`. Thnx.

lib/websockets/spaceEvents.ts
Devorein: Sure. Updated.

lib/websockets/interfaces.ts
Devorein: Good point, I myself got confused with the events, updated.

components/common/CharmEditor/plugins.ts
```diff
@@ -59,6 +58,15 @@ const inlineVotePluginKey = new PluginKey(inlineVote.pluginKeyName);
 
 export const suggestionsPluginKey = new PluginKey('suggestions');
 
+function removeSelectedNodeClass() {
+  setTimeout(() => {
+    const selectedDomNode = document.querySelector('.ProseMirror-selectednode');
+    if (selectedDomNode) {
+      selectedDomNode.classList.remove('ProseMirror-selectednode');
```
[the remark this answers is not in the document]
Devorein: Correct, I thought this was for the drop cursor, I've removed manually updating this class. We are now using a custom class which prosemirror doesn't maintain/handle internally.

components/common/CharmEditor/plugins.ts
Devorein: Move it inside `components/common/CharmEditor/components/nestedPage/nestedPage.plugins.ts`, as a separate plugin since it needs to be at the top and the nestedPagePlugin is located in the middle.

PR #2984, opened 2023-10-23
Make proposal tempalte properties readonly

msadura:
### <samp>🤖 Generated by Copilot at b403d47</samp>

This pull request adds a feature to make some custom properties of a proposal read-only in the UI. It affects the components that render and edit the proposal properties, such as `ProposalProperties`, `DocumentPage`, and `CardDetailProperties`. It uses a `readOnlyCustomProperties` prop to determine which properties should be read-only based on a template or user role.

### WHY
- fix creating proposal with custom props from template
- make filled props from template readonly


mattcasey (approved):
looks good.  one recommendation is to reuse `sourceTemplate` on line 166, instead of looking it up again

PR #2986, opened 2023-10-24
convert to nested jest configs

mattcasey:
### <samp>🤖 Generated by Copilot at 24104a4</samp>

Refactored and simplified the jest configuration files and test commands for the app. Used the `next/jest` package and created separate jest configuration files for different test directories. Updated the GitHub workflow, the VSCode settings, and the `package.json` file to reflect the new jest configuration and test commands.

### WHY
In VS Code, the vscode-jest-runner extension (https://github.com/firsttris/vscode-jest-runner) will now dynamically figure out the jest config to use. Prior, I was not able to run browser tests using jest.config.all in VS code. I would have to go to the settings.json and change it to jest.browser.ts.

Looking for feedback on how this might affect others' work. I primarily just run tests from inside of VS Code, so the npm commands don't matter too much to me. An alternative solution might be to use jest 'projects' and a different VS Code plugin


msadura (approved):
Loving this!

PR #2987, opened 2023-10-24
maintain mux permissions for custom domains

mattcasey:
### <samp>🤖 Generated by Copilot at aef21b8</samp>

Refactored the code for generating and using the mux token and playback restriction id. Added a function to sync the playback restriction with the custom domains of the spaces. Created a script to run the function manually.

### WHY
We whitelist the domains you can play a video on, which is currently only *.charmverse.io and *.charmverse.co. This PR overwrites/updates the whitelist whenever a custom space domain is set (I think!)

I ran the script locally and it fixed videos on work.charmverse.fyi


msadura (approved):
looks good!

PR #2988, opened 2023-10-24
swap toggle with checkbox ui

mattcasey:
### <samp>🤖 Generated by Copilot at e1c921e</samp>

This pull request improves the representation and functionality of the checkbox property type in the board editor. It updates the icon, the spacing, and the label of the checkbox properties in different views, using the Material UI components. It also changes the term "Toggle" to "Checkbox" in the type display name.

### WHY
Request from Alex. Not sure why focalboard was using toggles, but Notion uses checkbox too 

image

image


msadura (approved):
Looks good, one potential change would be to rename Switch to Checkbox - usually switch refers to toggle in web, and checkbox is a checkbox.

msadura:
+ seems like there is typescript error

PR #2995, opened 2023-10-25
Sync and assign collabland roles after connecting space

Devorein:
…rd serer### WHAT

### <samp>🤖 Generated by Copilot at ef1640b</samp>

This pull request refactors the collabland integration to use AWS SQS for sending and receiving webhook messages, and adds a feature to sync collabland roles for space members from discord. It also updates the log message format, the tests, and the types to handle multiple space ids per message, and renames some files and functions to improve consistency and readability.


lib/collabland/__tests__/connectSpace.spec.ts
```diff
@@ -49,4 +62,166 @@ describe('connectSpace', () => {
...
   });
+
+  it('should sync and assign collabland roles for all members of space', async () => {
+    const discordServerId = v4();
+
+    const { space, user: adminUser } = await testUtilsUser.generateUserAndSpace({
+      isAdmin: true
+    });
+
+    await prisma.space.update({
```
msadura, line 73: this part should happen during connect, it should run against space without discordServerid

lib/collabland/__tests__/connectSpace.spec.ts
```diff
@@ -49,4 +62,166 @@ describe('connectSpace', () => {
...
+        discordUser: {
+          create: {
+            account: {},
+            discordId: discordUserIds[0]
+          }
+        }
+      }
+    });
+
+    await prisma.user.update({
```
msadura, line 112: there is `createDiscordUser` helper

lib/collabland/connectSpace.ts
```diff
@@ -37,7 +41,51 @@ export async function connectSpace({ state, discordServerId }: { state: string;
...
+            select: {
+              discordId: true
+            }
+          }
+        }
+      }
+    }
+  });
+
+  for (const spaceMember of spaceMembers) {
```
msadura, line 75: I am wondering if there would be a way to do this in bulk, or at least in async way. Right now it can block whole function for a while if there are many discord users in space

msadura: + since htis is logic responsible for syncing al lthe users we might need to reuse it and it deservees separate function

lib/aws/webhookSqs.ts
```diff
@@ -73,7 +73,7 @@ export async function processMessages({ processorFn }: ProcessMssagesInput) {
 
       if (result.success) {
         log.info(`Message process successful: ${result.message}`, {
-          spaceId: result.spaceId,
+          spaceIds: result.spaceIds,
```
mattcasey, line 76: please use spaceId as its a standard field to filter on. its better to just use the first one, or log each space individually (i would just go with the first one)

lib/collabland/connectSpace.ts
```diff
@@ -37,7 +41,51 @@ export async function connectSpace({ state, discordServerId }: { state: string;
     throw new InvalidInputError('Cannot find space to connect');
   }
 
+  await upsertSpaceRolesFromDiscord({
```
mattcasey, line 44: i'd probably do this after you update the space, but not a big deal

mattcasey (commented):
please use spaceId in log arguments

lib/collabland/__tests__/connectSpace.spec.ts
Devorein: Nice catch. Updated.

lib/collabland/__tests__/connectSpace.spec.ts
Devorein: Thank you. Removed previous code and used `createDiscordUser` helper

lib/aws/webhookSqs.ts
Devorein: Good point. Updating.

lib/collabland/connectSpace.ts
Devorein: Good point. Updated.

mattcasey (approved):

PR #2998, opened 2023-10-25
Blacklist space user

Devorein:
### <samp>🤖 Generated by Copilot at 1ab178a</samp>

This pull request adds the functionality to ban members from spaces and prevent them from accessing space content via invite links, token gates, or direct navigation. It also adds a modal component to inform banned users of their status and provide some options. It modifies several components, hooks, lib functions, and API endpoints to implement the logic and UI for this feature. It also updates the `@charmverse/core` dependency to use the latest version.


charmClient/charmClient.ts
```diff
@@ -140,10 +140,11 @@ class CharmClient {
     return http.GET<LoggedInUser>('/api/profile');
   }
 
-  createUser({ address, walletSignature }: Web3LoginRequest) {
+  createUser({ address, walletSignature, spaceId }: Web3LoginRequest) {
     return http.POST<LoggedInUser>('/api/profile', {
       address,
-      walletSignature
+      walletSignature,
+      spaceId
```
msadura, line 147: honestly I am not sure about this. "Profile" is an entity not directly dependent on a space, so I would not add checking if user is banned in this funciton and just verify if user can be created in a separate step

components/common/Modal/SpaceBanModal.tsx
```diff
@@ -0,0 +1,72 @@
...
+            </Button>
+          )}
+          {user && !spaces?.length && (
+            <Button sx={{ width: '100%' }} href='/createSpace' color='primary'>
+              Create a space
+            </Button>
+          )}
+        </Box>
+      </Modal>
+      {!isBanned && !isLoading && children}
```
mattcasey, line 69: I would let the parent component determine what to do with the children. It's easier to follow when you're reading the code, and seems like a more common pattern in our code.

charmClient/charmClient.ts
mattcasey: I agree with this, and it looks like we actually call checkUserSpaceBanStatus() when the 'charmClient.acceptInvite()' method is called, so there's no reason to change this method?

components/settings/account/components/NewIdentityModal.tsx
```diff
@@ -91,6 +93,16 @@ export function NewIdentityModal({ isOpen, onClose }: Props) {
       sendingMagicLink.current = true;
       // console.log('Handling magic link request');
       try {
+        if (space) {
+          const isUserBannedFromSpace = await charmClient.members.checkSpaceBanStatus({
```
mattcasey, line 97: we need to check all spaces that a user is in when they try to add an email address

components/settings/account/components/NewIdentityModal.tsx
```diff
@@ -91,6 +93,16 @@ export function NewIdentityModal({ isOpen, onClose }: Props) {
...
+            spaceId: space.id,
+            email
+          });
+
+          if (isUserBannedFromSpace.isBanned) {
```
mattcasey, line 102: it's nice to be proactive, but in the name of simplicity i would have just put this logic inside loginWithMagicLink() and let it fail when they try to use the link

components/common/Modal/SpaceBanModal.tsx
```diff
@@ -0,0 +1,72 @@
+import { Box } from '@mui/material';
```
mattcasey, line 1: This component belongs to SpaceAccessGate.tsx or a components folder next to it, because that's the only place it's used.

Please try to keep specific modals out of the common/Modal folder. We should only put things here that are used in multiple places :)

pages/invite/[inviteCode].tsx
```diff
@@ -1,38 +1,63 @@
...
   const inviteCode = context.query.inviteCode as string;
   const inviteLink = await getInviteLink(inviteCode);
+  const user = context.req.session?.user;
 
   if (!inviteLink) {
     return {
       props: {
-        error: 'Invitation not found'
+        error: 'invalid',
+        invite: null
```
Comment (mattcasey) [inline pages/invite/[inviteCode].tsx:23]: just make invite optional?

charmClient/charmClient.ts
Devorein: Yeah that makes sense. I have removed it.

components/common/Modal/SpaceBanModal.tsx
Devorein: Its also used in the `SpaceInviteLink` component, but I get your point. I will move it over there.

components/common/Modal/SpaceBanModal.tsx
Devorein: Right, for now we will let the user "fail" themselves and then we show the modal as a side-effect.

pages/invite/[inviteCode].tsx
Comment (Devorein) [reply pages/invite/[inviteCode].tsx:23]: Good idea. Updated.

components/settings/account/components/NewIdentityModal.tsx
Devorein: Good point. Updated.

components/settings/account/components/NewIdentityModal.tsx
Devorein: Yep we are now taking a more reactive approach. Its definitely a lot simpler now.

PR #3000, opened 2023-10-25
Japaenese events

mattcasey:
### <samp>🤖 Generated by Copilot at 5e4278c</samp>

The pull request enhances the title input component for foreign language input by using a custom hook to detect IME composition. The hook `useIMEComposition` is defined in a new file `hooks/useIMEComposition.tsx` and returns a boolean flag that indicates the IME status. The component uses the hook and refactors the keyboard event handlers to simplify the logic and improve the user experience.

### WHY
The title input is actually pretty complicated. Notion uses an h1 tag with contenteditable, FWIW. But it must:
- support wrapping lines for long titles (which regular text input does not, even with break-word CSS rule)
- allow us to jump to the document body on "Enter" being pressed, unless user is using Enter to submit a Japenese character. More info: https://cotoacademy.com/type-japanese-windows-10/ 

The text input is most ideal because then we could listen to a form submit event when user hits Return, but it does not support text wrapping. There's also no simple API to tell if a user is using a foreign language keyboard input. I finally found someone who managed to make it work by listening to several DOM events: https://www.stum.de/2016/06/24/handling-ime-events-in-javascript/. Not rocket science, indeed!

### DIFF
--- a/hooks/useIMEComposition.tsx
+++ b/hooks/useIMEComposition.tsx
@@ -1,0 +1,57 @@
+import { useState, useEffect } from 'react';
+import type { RefObject } from 'react';
+
+export type InputType = HTMLInputElement | HTMLTextAreaElement;
+
+// A hook to determine if someone is using foreign language input to type via IME events
+// reference: https://www.stum.de/2016/06/24/handling-ime-events-in-javascript/
+export function useIMEComposition(ref: RefObject<InputType>) {
+  const [isComposing, setIsComposing] = useState(false); // IME Composing going on
+  const [hasCompositionJustEnded, sethasCompositionJustEnded] = useState(false); // Used to swallow keyup event related to compositionend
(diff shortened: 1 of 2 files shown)


hooks/useIMEComposition.tsx
```diff
@@ -0,0 +1,57 @@
+import { useState, useEffect } from 'react';
+import type { RefObject } from 'react';
+
+export type InputType = HTMLInputElement | HTMLTextAreaElement;
+
+// A hook to determine if someone is using foreign language input to type via IME events
+// reference: https://www.stum.de/2016/06/24/handling-ime-events-in-javascript/
+export function useIMEComposition(ref: RefObject<InputType>) {
```
msadura, line 8: woah

msadura (approved):

PR #3005, opened 2023-10-26
Fix computing default space page to display

msadura:
### <samp>🤖 Generated by Copilot at 897550b</samp>

Added a condition to `getDefaultPageForSpaceRaw` to ensure the default page for a space is a top-level page. This supports the new subpage feature in `app.charmverse.io`.

### WHY
Quick workaround to computing default page to redirect to in new space - the issue here was that it sorted all pages, including child pages and their indexes. Actually page navigation (sidebar) uses a bit more complex logic so I believe the easiest way will be to just do not fetch child pages


mattcasey (approved):
i think this works, but please add a test

mattcasey (approved):

PR #3020, opened 2023-10-30
Drag and drop nested pages within each other in editor

Devorein:
### <samp>🤖 Generated by Copilot at 19c45c6</samp>

This pull request introduces a custom editor state and view class, a custom hook, and several custom plugins for the `CharmEditor` component, to enable nested page nodes and drag and drop functionality. It also updates the websocket interfaces and handlers to support page reordering in the sidebar and the editor. It also adds some style and keyboard event handling improvements.

### DIFF
--- a/components/_app/GlobalComponents.tsx
+++ b/components/_app/GlobalComponents.tsx
@@ -2,7 +2,6 @@ import { useRouter } from 'next/router';
 import { useEffect } from 'react';
 import { mutate } from 'swr';
 
-import { useAppDispatch } from 'components/common/BoardEditor/focalboard/src/store/hooks';
 import HexagonalAvatarMask from 'components/common/HexagonalAvatarMask';
 import Snackbar from 'components/common/Snackbar';
 import { MemberProfileDialogGlobal } from 'components/members/components/MemberProfileDialogGlobal';
@@ -10,7 +9,6 @@ import { ApplicationDialog } from 'components/rewards/components/RewardApplicati
 import { useImportDiscordRoles } from 'components/settings/roles/hooks/useImportDiscordRoles';
 import { useAppLoadedEvent } from 'hooks/useAppLoadedEvent';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
-import useKeydownPress from 'hooks/useKeydownPress';
 import { useMarkNotificationFromUrl } from 'hooks/useMarkNotificationFromUrl';
 import { getPagesListCacheKey, usePages } from 'hooks/usePages';
 import { useSettingsDialog } from 'hooks/useSettingsDialog';
--- a/components/_app/GlobalComponents.tsx
+++ b/components/_app/GlobalComponents.tsx
@@ -10,6 +10,7 @@ import { ApplicationDialog } from 'components/rewards/components/RewardApplicati
 import { useImportDiscordRoles } from 'components/settings/roles/hooks/useImportDiscordRoles';
 import { useAppLoadedEvent } from 'hooks/useAppLoadedEvent';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import useKeydownPress from 'hooks/useKeydownPress';
 import { useMarkNotificationFromUrl } from 'hooks/useMarkNotificationFromUrl';
 import { getPagesListCacheKey, usePages } from 'hooks/usePages';
 import { useSettingsDialog } from 'hooks/useSettingsDialog';
--- a/components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
+++ b/components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
@@ -2,8 +2,7 @@ import type { NodeViewProps } from '@bangle.dev/core';
 import { useEditorViewContext } from '@bangle.dev/react';
 import type { PageMeta } from '@charmverse/core/pages';
 import styled from '@emotion/styled';
-import { Box, Typography } from '@mui/material';
-import { useState } from 'react';
+import { Typography } from '@mui/material';
 
 import Link from 'components/common/Link';
 import { NoAccessPageIcon, PageIcon } from 'components/common/PageLayout/components/PageIcon';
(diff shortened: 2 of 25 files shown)


components/_app/GlobalComponents.tsx
```diff
@@ -10,6 +10,7 @@ import { ApplicationDialog } from 'components/rewards/components/RewardApplicati
 import { useImportDiscordRoles } from 'components/settings/roles/hooks/useImportDiscordRoles';
 import { useAppLoadedEvent } from 'hooks/useAppLoadedEvent';
 import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import useKeydownPress from 'hooks/useKeydownPress';
```
mattcasey, line 13: import no longer needed?

components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
```diff
@@ -90,6 +101,17 @@ export default function NestedPage({
         enableDragAndDrop(view, nodePos);
       }}
       data-type={node.attrs.type}
+      // Only works for firefox
+      onDragExitCapture={() => {
+        resetHoverPluginState(view);
+      }}
+      // Should only works for chrome, firefox also handles it but for ff we have onDragExitCapture
```
mattcasey, line 108: do you know what will happen in IE?

mattcasey: i assume "Firefox" means "all browsers except Chrome" in this case :)

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
mattcasey, line 1: please add a comment next to code that explains what was changed

lib/websockets/documentEvents/documentEvents.ts
```diff
@@ -454,7 +455,7 @@ export class DocumentEventHandler {
               const node = room.node.resolve(ds.from).nodeAfter?.toJSON() as PageContent;
               if (node && node.attrs && node.type === 'page') {
                 const { id: pageId, type: pageType = '', path: pagePath } = node.attrs;
-                if (pageId && pageType === null && pagePath === null) {
+                if (pageId && pageType !== 'forum_category' && STATIC_PAGES.find((c) => c.path !== pagePath)) {
```
mattcasey, line 458: It's not clear at first why we are filtering here by these values, or what is special about 'forum_category'.

Suggestion: use `isUUID()` from our string utils to determine that it is a valid pageId, and then you don't need a blacklist of values. OR create a top-level array called something like "linkedPageStaticTypes" that includes all possible values.

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
```diff
@@ -0,0 +1,223 @@
...
+    const view = this.editorView;
+    const draggedNode = view.state.doc.nodeAt(view.state.selection.$anchor.pos);
+    const hoveredPageDomNode =
+      domElementAtCoords?.getAttribute('data-page-type') === 'page'
+        ? domElementAtCoords
+        : domElementAtCoords?.parentElement?.getAttribute('data-page-type') === 'page'
+        ? domElementAtCoords.parentElement
+        : null;
+    const hoveredPageDomNodeId = hoveredPageDomNode?.getAttribute('data-id')?.split('page-')[1];
+    const currentHoveredPageDomNode = document.querySelector(`.${HOVERED_PAGE_NODE_CLASS}`);
```
mattcasey, line 149: does this element only appear once ever? can we use id instead of class to speed up the query?

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
```diff
@@ -0,0 +1,223 @@
...
+    // Remove the class from the previous hovered page node
+    if (currentHoveredPageDomNode?.getAttribute('data-id')?.split('page-')[1] !== hoveredPageDomNodeId) {
```
mattcasey, line 151: Suggestion: `domNode.dataset.id` is a little cleaner way to access and update data attributes. (you can even do `domNode.dataset.id = 'foo'` )

mattcasey: also, it could just be data-page-id="". I am not sure there's any reason to add a page- prefix into the string, which just adds more work?

mattcasey (approved):

components/_app/GlobalComponents.tsx
Devorein: Removed. Thnx for noticing.

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
Devorein: There are no new element created here, its just a class with a specific style. And its added to the NestedPage's Link component which is an anchor tag. I see no existing id being used for that element so we can add an id to it.

components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
Devorein: I reversed the logic, even though opera, safari and edge all have chrome in their useragent its safer to see if its firefox and not execute it.

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
Devorein: Yeah I'm not sure either why we added a prefix here, removed page and just opted for the actual id in `dataset.id`

components/common/CharmEditor/components/prosemirror/prosemirror-dropcursor/dropcursor.ts
Devorein: Added comments.

lib/websockets/documentEvents/documentEvents.ts
Devorein: Updated to use isUUID and moved the logic inside another fn

PR #3021, opened 2023-10-30
Add page comments for card blocks

Devorein:
### WHAT
copilot:summary

### WHY
image


__e2e__/po/document.po.ts
```diff
@@ -54,7 +54,7 @@ export class DocumentPage extends GlobalPage {
   }
 
   getCardCommentContent(commentId: string) {
-    return this.page.locator(`data-test=card-comment-content-${commentId} >> div[contenteditable]`).first();
+    return this.page.locator(`data-test=post-comment-${commentId} >> div[contenteditable]`).first();
```
mattcasey, line 57: if we have to rename "card" to "post" here, i would rename the data attribute to something else, or maybe just remove "post-" from the data attribute

mattcasey (approved):
i like how simple this was!

__e2e__/po/document.po.ts
Devorein: Updated. I wanted to do that in a separate PR, but didn't notice anything breaking.

PR #3033, opened 2023-11-02
Fixed post comment vote count flicker

Devorein:
### <samp>🤖 Generated by Copilot at b181781</samp>

Refactored the comment feature to use a common `Comment` component and the `charmClient` API. Deleted the redundant `PostComment` component. Improved the reusability and consistency of the comment functionality across the app.


components/forum/components/PostPage/PostPage.tsx
```diff
@@ -226,6 +227,70 @@ export function PostPage({
...
   const canEdit = !!permissions?.edit_post;
 
+  async function updateComment({ id, content, contentText }: UpdateCommentPayload) {
+    const comment = postComments.find((_comment) => _comment.id === id);
+    if (comment) {
+      const updatedComment = await charmClient.forum.updatePostComment({
+        commentId: id,
+        content,
+        contentText,
+        postId: comment.postId
```
mattcasey, line 237: Doesn't seem like we should need to send postId, since the comment is is universally unique. This would simplify the logic here

mattcasey (approved):

mattcasey:
Thanks for reusing the comment component!

PR #3041, opened 2023-11-05
Use esbuild for websockets and cron instead of next.js

mattcasey:
### <samp>🤖 Generated by Copilot at 001088a</samp>

This pull request improves the performance and size of the app by using esbuild to bundle and minify TypeScript files for websockets and cron jobs. It also simplifies the module structure and reduces code duplication in the `mention` components of the CharmEditor. It fixes some import issues and applies patches to make some libraries compatible with esbuild.


package.json
```diff
@@ -249,23 +248,23 @@
...
     "@next/eslint-plugin-next": "^13.5.6",
     "@playwright/test": "^1.39.0",
-    "@storybook/addon-essentials": "^7.5.1",
-    "@storybook/addon-interactions": "^7.5.1",
-    "@storybook/addon-links": "^7.5.1",
-    "@storybook/blocks": "^7.5.1",
-    "@storybook/manager-api": "^7.5.1",
-    "@storybook/nextjs": "^7.5.2",
-    "@storybook/react": "^7.5.1",
+    "@storybook/addon-essentials": "^7.5.3",
```
mattcasey, line 259: upgraded storybook because it was depending on an older version of esbuild

next.config.js
```diff
@@ -109,47 +109,6 @@ const config = {
     // Fix for: "Module not found: Can't resolve 'canvas'"
     _config.resolve.alias.canvas = false;
```
motechFR, line 111: Thanks for cleaning up that leftover from the core integration

motechFR (approved):

PR #3044, opened 2023-11-06
Fix dragAndDrop e2e test

Devorein:
### <samp>🤖 Generated by Copilot at 0d10f4f</samp>

Increased the `waitForTimeout` value in the `dragAndDrop.spec.ts` file to reduce flakiness in the charm editor end-to-end tests. This was part of a pull request to improve the test quality and reliability.

### DIFF
--- a/__e2e__/charmEditor/dragAndDrop.spec.ts
+++ b/__e2e__/charmEditor/dragAndDrop.spec.ts
@@ -82,7 +82,7 @@ test('Drag and drop one paragraph over another in the CharmEditor', async ({ doc
     }
   });
 
-  await documentPage.page.waitForTimeout(150);
+  await documentPage.page.waitForTimeout(500);
 
   const page = await prisma.page.findUniqueOrThrow({
     where: {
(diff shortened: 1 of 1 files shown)


__e2e__/charmEditor/dragAndDrop.spec.ts
```diff
@@ -82,7 +82,7 @@ test('Drag and drop one paragraph over another in the CharmEditor', async ({ doc
     }
   });
 
-  await documentPage.page.waitForTimeout(150);
+  await documentPage.page.waitForTimeout(500);
```
mattcasey, line 85: Do you know what we're waiting for? I am guessing we need the socket server to process the event or something. Instead of a hard-coded timeout, playwright gives a lot of utilities to wait for something in the page, or can you try using https://playwright.dev/docs/api/class-page#page-wait-for-function if, for instance we need to check the db is updated?

mattcasey (changes requested):
left request in comments

__e2e__/charmEditor/dragAndDrop.spec.ts
Devorein: Nice one. I looked into it a bit and it is precisely what should've been used in the first place. Got the test to pass locally and hopefully in CI too.

mattcasey (approved):

PR #3046, opened 2023-11-06
dont save proposal templates immediately

mattcasey:
Make it so we don't save templates immediately, following the same UX when you choose to create a new regular proposal, or one from existing template.
 
This was originally a larger cleanup PR, but I merged most of the code separately: https://github.com/charmverse/app.charmverse.io/pull/3047.


mattcasey:
i merged this on along with the invite code fix

msadura (approved):
I really like how it simplifies code and actually this auto-saving behavior of templates was  a bit weird so totally like those improvements!

PR #3051, opened 2023-11-07
Disable opening poll modal when opening post cards

Devorein:
### <samp>🤖 Generated by Copilot at 999a29d</samp>

Refactored and simplified the code for the poll and voting features in the `CharmEditor` and `DocumentPage` components. Removed unused variables and types from the `useVotes` hook and the `PollComponent`.

### WHY
[Card](https://app.charmverse.io/charmverse/opening-a-forum-post-with-a-poll-shows-poll-creation-dialog-for-a-split-second-49881997491780106)

### DIFF
--- a/components/common/CharmEditor/components/poll/PollComponent.tsx
+++ b/components/common/CharmEditor/components/poll/PollComponent.tsx
@@ -62,9 +62,7 @@ export function PollNodeView({
           snapshotProposalId={snapshotProposalId || null}
         />
         <div
-          onClick={(e) => {
-            // e.stopPropagation();
-            // e.preventDefault();
+          onClick={() => {
             setShowModal(true);
           }}
         >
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/components/poll/PollComponent.tsx
```diff
@@ -79,18 +77,30 @@ export function PollNodeView({
...
-      castVote={castVote}
-      deleteVote={deleteVote}
-      detailed={false}
-      // This makes sure that if something goes wrong in loading state, we won't stop users who should be able to vote from voting
-      disableVote={(pagePermissions && !pagePermissions.comment) || (postPermissions && !postPermissions.add_comment)}
-      vote={votes[pollId]}
-      updateDeadline={updateDeadline}
-    />
-  );
+  return null;
```
mattcasey, line 105: in what condition do we return null? and why is it not returning an empty div like other case

Devorein: If the poll id in the node attribute doesn't refer to any vote, which is highly unlikely to occur. But I will leave a comment and just return `<div />`

PR #3055, opened 2023-11-08
Allow drag and drop images into editor

Devorein:
### <samp>🤖 Generated by Copilot at f08f371</samp>

Improved the image insertion and drag and drop features in the `CharmEditor` component. Copied and modified the `safeInsert` function from the `prosemirror-utils` library to handle the insertion logic more robustly. Added a custom drag and drop plugin that uses the modified `safeInsert` function and works better than the default one.

### WHY
[Card](https://app.charmverse.io/charmverse/support-drag-and-drop-images-from-your-desktop-2563324601968251)


components/common/CharmEditor/components/prosemirror/prosemirror-utils/transforms.ts
```diff
@@ -0,0 +1,74 @@
+// Reference: https://github.com/atlassian/prosemirror-utils/blob/fbdf888a4c24bae5f124d166bd21692d7b5285be/src/transforms.ts#L168
```
mattcasey, line 1: why are we not just importing this from prosemirror-utils?

mattcasey: Looks like this is just a method you're re-implementing, so there's no reason to treat it like a "special citizen." Please put this into a standard folder like `lib/prosemiror/safeInsert.ts`.  Also please include a comment explaining why we copied and what was added, as well as a test on it. We should try to copy any existing tests as well,  or find another workaround that doesn't require replacing the whole method.

components/common/CharmEditor/plugins.ts
```diff
@@ -120,9 +120,7 @@ export function charmEditorPlugins({
       key: linkedPagePluginKey
     }),
     nestedPagePlugins(),
-    imagePlugins({
-      handleDragAndDrop: false
```
mattcasey, line 124: wow i can't believe we just had it disabled..

mattcasey: FYI I did some digging and it was me that added this when i fixed some issues with column UX, i disabled dragging on a few other components as well, so maybe i thought this was something that it wasn't. I will do some testing

mattcasey (changes requested):
see notes about the safeInsert method. otherwise looks good

Devorein:
I ended up making use of patch package to make the single-line change. 

```
if (content instanceof prosemirrorModel.Fragment) {
     return $pos.parent.canReplace(index, index, content);
-  } else if (content instanceof prosemirrorModel.Node) {
-    return $pos.parent.canReplaceWith(index, index, content.type);
   }
-  return false;
+  return $pos.parent.canReplaceWith(index, index, content.type);
```

The condition always returned false even if it was a Node. It could be due to two different classes of `Node` being compared with each other. There is no chance of the third condition as content is either `Fragment` or `Node`.

PR #3067, opened 2023-11-08
Simplify new rewards state 

mattcasey:
### <samp>🤖 Generated by Copilot at 8f20571</samp>

This pull request refactors and simplifies the components and hooks related to creating and displaying new pages in the app, such as proposals and rewards. It also improves the layout and styling of the database pages by using common components. It affects the files `NewPageDocument.tsx`, `useNewPage.tsx`, `NewPageDialog.tsx`, `NewProposalButton.tsx`, `ProposalsPage.tsx`, `NewRewardButton.tsx`, `useNewRewardPage.tsx`, `RewardsPage.tsx`, `centerPanel.scss`, `DatabasePageContent.tsx`, and `RewardsPageWithProviders.tsx`.

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.scss
@@ -18,10 +18,12 @@
   overflow: auto;
 
   > * {
-    padding: 0 24px;
+    padding-left: 24px;
+    padding-right: 24px;
 
     @media screen and (min-width: 768px) {
-      padding: 0 80px;
+      padding-left: 80px;
+      padding-right: 80px;
     }
 
     &.PageBanner {
--- a/components/common/PageLayout/components/DatabasePageContent.tsx
+++ b/components/common/PageLayout/components/DatabasePageContent.tsx
@@ -15,7 +15,7 @@ export function DatabaseStickyHeader({ children }: { children: ReactNode }) {
 
 export function DatabaseTitle({ children }: { children: ReactNode }) {
   return (
-    <Box pt={6} className='ViewTitle'>
+    <Box pt={8} className='ViewTitle'>
       <Box mb={2} data-test='board-title'>
         {children}
       </Box>
--- a/components/common/PageLayout/components/DatabasePageContent.tsx
+++ b/components/common/PageLayout/components/DatabasePageContent.tsx
@@ -1,0 +1,24 @@
+import { Box } from '@mui/material';
+import type { ReactNode } from 'react';
+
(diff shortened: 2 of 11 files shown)


components/common/BoardEditor/focalboard/src/components/centerPanel.scss
```diff
@@ -18,10 +18,12 @@
   overflow: auto;
 
   > * {
-    padding: 0 24px;
+    padding-left: 24px;
+    padding-right: 24px;
```
mattcasey, line 22: this was to address a weird padding issue that only happens in local development. in production the cSS gets loaded in a different order somehow

components/rewards/components/NewRewardButton.tsx
```diff
@@ -1,17 +1,15 @@
...
-import { useNewReward } from 'components/rewards/hooks/useNewReward';
-import { useCurrentSpace } from 'hooks/useCurrentSpace';
-import { usePages } from 'hooks/usePages';
-import { useUser } from 'hooks/useUser';
+import { useNewRewardPage } from 'components/rewards/hooks/useNewRewardPage';
 
 export function NewRewardButton() {
-  const { openNewPage } = useNewPage();
-  const { clearFormInputs, formInputs, setFormInputs, createReward, isSavingReward } = useNewReward();
+  const { isDirty, clearNewPage, openNewPage, newPageValues, updateNewPageValues } = useNewPage();
```
mattcasey, line 9: in this new version, you can see what values/methods are being used because they are being passed in. Before, the context for useNewPage meant that anyone could be calling or doing anything with that state. But the reality is that it's only used by 2-3 layers of components and hooks.

I think a good case for adding context is when there's a large gap between the level of state and the components that use it, like whether the app is dark/light mode, or when we want to avoid re-renders of intermediate components

components/common/PageLayout/components/DatabasePageContent.tsx
```diff
@@ -0,0 +1,24 @@
+import { Box } from '@mui/material';
+import type { ReactNode } from 'react';
+
+export function DatabaseContainer({ children }: { children: ReactNode }) {
```
mattcasey, line 4: this just unifies some of the components for Rewards/Proposals

components/rewards/components/NewRewardButton.tsx
msadura: I though having context will be more handy, but indeed your updated version seems more straightforward and I really like it

msadura (approved):

PR #3074, opened 2023-11-09
Support permissions for inline comments on proposals

Devorein:
### <samp>🤖 Generated by Copilot at 5bd51c0</samp>

Added support for thread access groups in the inline comments feature. This allows users to choose who can see their comments on different page types. Updated the API, the UI, and the tests to handle the new `accessGroup` property in the `ThreadCreatePayload` interface. Refactored some files and dependencies related to the comments feature.


lib/threads/interfaces.ts
```diff
@@ -14,14 +14,20 @@ export interface ThreadStatusUpdate {
...
 /**
  * @context Prosemirror content for knowing where to display this thread inside the CharmEditor
  */
-export interface ThreadCreate {
+export interface ThreadCreatePayload {
   comment: string | PageContent;
   pageId: string;
   userId: string;
   context: string;
+  accessGroup: ThreadAccessGroup;
```
motechFR, line 30: Should we make this an array from day 1 just for simplicity? Usually in all our permissions, we end up with Array<TargetPermissionGroup> anyways

lib/threads/getPageThreads.ts
```diff
@@ -0,0 +1,111 @@
...
+    where: {
+      spaceId,
+      userId
+    },
+    select: {
+      id: true
+    }
+  });
+
+  for (const thread of threads) {
```
motechFR, line 89: A suggestion, you could offload this work to the database

https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#json-object-arrays

lib/threads/__tests__/getPageThreads.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+  it(`Should get page threads based on proposal authors access`, async () => {
+    const { user: adminUser, space } = await generateUserAndSpace({
+      isAdmin: true
+    });
+
+    const proposalAuthor = await generateUser();
+    await addUserToSpace({
+      spaceId: space.id,
+      userId: proposalAuthor.id
+    });
```
motechFR, line 162: There is a shorter utility for doing this in one line of code

```typescript
import { testUtilsUser } from '@charmverse/core/test'

// You can also pass an isAdmin: true if needed
const member = await testUtilsUser.generateSpaceUser({spaceId: space.id})

```

components/common/CharmEditor/components/inlineComment/components/InlineCommentSubMenu.tsx
```diff
@@ -50,6 +61,11 @@ export function InlineCommentSubMenu({ pluginKey, pageId }: { pluginKey: PluginK
       }
     ]
   });
+  const [threadAccessGroup, setThreadAccessGroup] = useState<'space' | 'reviewers' | 'authors'>('space');
+
+  const handleThreadAccessGroupChange: SelectProps['onChange'] = (event) => {
+    setThreadAccessGroup(event.target.value as any);
+  };
```
motechFR, line 68: I don't know where this would occur exactly, but could we get a component test to make sure the inline comment marker shows or doesn't show?

motechFR:
@Devorein PR looking good so far. My main comment, is we should go for an array of groups with permissions from day 1, just to make this more flexible.

Re-posting some prisma docs for working with JSON arrays

https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#json-object-arrays

components/common/CharmEditor/components/inlineComment/components/InlineCommentSubMenu.tsx
Devorein: Thats actually a very good point, I will create an e2e test to make sure we have this working correctly.

lib/threads/__tests__/getPageThreads.spec.ts
Devorein: Thank you. I will use it from now.

lib/threads/getPageThreads.ts
Devorein: Thats actually a very performant suggestion, updated the query to do everything in a single query, but we needed to get the roles before hand.

lib/threads/createThread.ts
```diff
@@ -46,6 +53,10 @@ export async function createThread({ comment, pageId, userId, context }: ThreadC
           id: userId
         }
       },
+      accessGroup: {
+        id: accessGroup.id,
+        group: accessGroup.group
+      },
```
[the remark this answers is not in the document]
Devorein: Yep, we converted it into an array of access groups.

lib/threads/interfaces.ts
Devorein: Yep, we updated it to use an array.

mattcasey (approved):

PR #3075, opened 2023-11-09
Fix loop along with typecasting error

motechFR:
### <samp>🤖 Generated by Copilot at 4d0d55c</samp>

Fixed a TypeScript error when loading Stripe and simplified the return value of `getLoopProducts` function. These changes improve the reliability and readability of the code related to subscription management.

### DIFF
--- a/components/settings/subscription/loadStripe.ts
+++ b/components/settings/subscription/loadStripe.ts
@@ -2,9 +2,7 @@ import env from '@beam-australia/react-env';
 import type { Stripe } from '@stripe/stripe-js';
 import { loadStripe as _loadStripe } from '@stripe/stripe-js';
 
-// This typecasting prevents stripe throwing an error
-// https://stackoverflow.com/a/63784104
-const stripePublicKey = env('STRIPE_PUBLIC_KEY') as string;
+const stripePublicKey = env('STRIPE_PUBLIC_KEY');
 
 let stripePromise: Promise<Stripe | null>;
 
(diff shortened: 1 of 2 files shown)


lib/loop/loop.ts
```diff
@@ -27,7 +27,7 @@ const DEFAULT_HEADERS = {
 };
 
 export async function getLoopProducts(productId?: string) {
-  return GET<LoopItem[]>(`${loopApiUrl}/api/v1/items${productId ? `?id=${productId}` : ''}`, null, {
+  return GET<{ items: LoopItem[] }>(`${loopApiUrl}/api/v1/items${productId ? `?id=${productId}` : ''}`, null, {
```
motechFR, line 30: Seems API signature changed

components/settings/subscription/loadStripe.ts
```diff
@@ -2,7 +2,9 @@ import env from '@beam-australia/react-env';
 import type { Stripe } from '@stripe/stripe-js';
 import { loadStripe as _loadStripe } from '@stripe/stripe-js';
 
-const stripePublicKey = env('STRIPE_PUBLIC_KEY');
+// This typecasting prevents stripe throwing an error
+// https://stackoverflow.com/a/63784104
+const stripePublicKey = env('STRIPE_PUBLIC_KEY') as string;
```
mattcasey, line 7: I do not get a type error? if stripe doesn't want an undefined value, we should not call it

mattcasey (approved):
wow that's a shitty api break

components/settings/subscription/loadStripe.ts
motechFR: Turns out, I just needed to restart the app since stripe env var wasn't passed in on first start

PR #3079, opened 2023-11-10
Export zipped data

motechFR:
### WHAT
copilot:summary

### WHY
Enable users to export a database to a zipped file

### DIFF
--- a/components/common/PageActions/components/DatabasePageActionList.tsx
+++ b/components/common/PageActions/components/DatabasePageActionList.tsx
@@ -40,6 +40,7 @@ import { useUser } from 'hooks/useUser';
 import type { Board } from 'lib/focalboard/board';
 import type { BoardView } from 'lib/focalboard/boardView';
 import type { CardPage } from 'lib/focalboard/card';
+import { generateMarkdown } from 'lib/prosemirror/plugins/markdown/generateMarkdown';
 
 import { isValidCsvResult, addNewCards } from '../utils/databasePageOptions';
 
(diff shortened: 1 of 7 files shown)


mattcasey (changes requested):
it's weird that the CSV is generated on the frontend, but the markdown is generated on the backend. I think we should just do it all on the backend

components/common/PageActions/components/DatabasePageActionList.tsx
```diff
@@ -260,6 +324,15 @@ export function DatabasePageActionList({ pagePermissions, onComplete, page }: Pr
...
         </div>
       </Tooltip>
+      <ListItemButton onClick={() => exportZippedDatabase(board, view)}>
+        <FormatListBulletedIcon
+          fontSize='small'
+          sx={{
+            mr: 1
+          }}
+        />
+        <ListItemText primary='Export Pages & Data' />
```
mattcasey, line 334: this replaces "Export to CSV"

components/common/PageActions/components/DatabasePageActionList.tsx
```diff
@@ -260,6 +324,15 @@ export function DatabasePageActionList({ pagePermissions, onComplete, page }: Pr
           </ListItemButton>
         </div>
       </Tooltip>
+      <ListItemButton onClick={() => exportZippedDatabase(board, view)}>
+        <FormatListBulletedIcon
```
mattcasey, line 328: can we use the UploadFile icon? https://mui.com/material-ui/material-icons/?query=download&selected=UploadFile. It actually matches what Notion uses:
image

mattcasey (approved):

PR #3088, opened 2023-11-11
Improved ux for uploading images via dnd

Devorein:
### <samp>🤖 Generated by Copilot at 9912f71</samp>

This pull request adds and improves the image insertion feature for the charm editor. It allows users to drop image files into the editor and insert them as base64 data URLs, which are then uploaded to S3 as file objects. It also simplifies and enhances the image plugin and the resizable image component. It affects the `DocumentPage`, `PostPage`, `CharmEditor`, `InlineCharmEditor`, `ReactEditor`, `image`, and `ResizableImage` modules, and adds three new modules: `convertFileToBase64`, `getFileBinary`, and `getFileData`.

### WHY
1. Show a loading ux when uploading image via dnd
2. Drop image anywhere in the editor and add it to the document


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -208,7 +209,41 @@ function DocumentPage({ page, refreshPage, savePage, readOnly = false, close }:
...
       <div ref={printRef} className={`document-print-container ${fontClassName}`}>
         <ScrollContainer id='document-scroll-container' showPageActionSidebar={showPageActionSidebar}>
-          <Box display='flex' flexDirection='column' ref={containerRef}>
+          <div
+            style={{
+              display: 'flex',
+              flexDirection: 'column'
+            }}
+            ref={containerRef}
+            onDrop={(event) => {
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:218]: please keep this logic out of DocumetPage. I think it belongs in the image component folder

lib/file/getFileData.ts
```diff
@@ -0,0 +1,59 @@
+export function getFileData(data: DataTransfer, accept: string, multiple: boolean) {
```
mattcasey, line 1: all this logic should remain in the image component folder

lib/file/convertFileToBase64.ts
```diff
@@ -0,0 +1,8 @@
+export function convertFileToBase64(file: File) {
```
mattcasey, line 1: all this code belongs in image component folder

mattcasey (changes requested):
Works really well! left comments

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:218]: Updated. Now we can add this easily where ever we want to support this feature

lib/file/getFileData.ts
Devorein: Updated

lib/file/convertFileToBase64.ts
Devorein: Updated

PR #3089, opened 2023-11-12
Disable sending multiple notifications for same event

Devorein:
### <samp>🤖 Generated by Copilot at bb45d4a</samp>

No summary available (Limit exceeded: required to process 126795 tokens, but only 50000 are allowed per call)

### DIFF
--- a/lib/notifications/polls/createPollNotifications.ts
+++ b/lib/notifications/polls/createPollNotifications.ts
@@ -23,17 +23,9 @@ export async function createPollNotifications(webhookData: {
         },
         select: {
           spaceId: true,
-          page: {
-            select: { id: true }
-          },
-          post: {
-            select: { id: true }
-          },
-          author: {
-            select: {
-              id: true
-            }
-          }
+          pageId: true,
+          postId: true,
+          createdBy: true
         }
       });
 
@@ -48,23 +40,23 @@ export async function createPollNotifications(webhookData: {
         }
       });
 
-      const spaceUserIds = spaceRoles.map(({ userId }) => userId).filter((userId) => userId !== vote.author.id);
+      const spaceUserIds = spaceRoles.map(({ userId }) => userId).filter((userId) => userId !== vote.createdBy);
 
       const permissionClient = await getPermissionsClient({
         resourceId: spaceId,
         resourceIdType: 'space'
       });
 
-      if (vote.page) {
+      if (vote.pageId) {
         for (const spaceUserId of spaceUserIds) {
           const pagePermission = await permissionClient.client.pages.computePagePermissions({
-            resourceId: vote.page.id,
+            resourceId: vote.pageId,
             userId: spaceUserId
           });
           if (pagePermission.comment) {
             const { id } = await savePollNotification({
               createdAt: webhookData.createdAt,
-              createdBy: vote.author.id,
+              createdBy: vote.createdBy,
               spaceId,
               type: 'new_vote',
               userId: spaceUserId,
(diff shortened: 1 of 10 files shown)


lib/notifications/proposals/createProposalNotifications.ts
```diff
@@ -117,8 +117,9 @@ export async function createProposalNotifications(webhookData: {
         const accessibleProposalCategoryIds = accessibleProposalCategories.map(({ id }) => id);
 
         const isAuthor = proposalAuthorIds.includes(spaceRole.userId);
-        const isReviewer = proposalReviewerIds.includes(spaceRole.userId);
-        const isReviewerRole = proposalReviewerRoleIds.some((roleId) => roleIds.includes(roleId));
+        const isReviewer =
+          proposalReviewerIds.includes(spaceRole.userId) ||
+          proposalReviewerRoleIds.some((roleId) => roleIds.includes(roleId));
```
mattcasey, line 122: the previous code was easier to read

lib/notifications/polls/createPollNotifications.ts
```diff
@@ -82,18 +75,12 @@ export async function createPollNotifications(webhookData: {
...
-                  postCategories: [vote.post.category],
-                  userId: spaceUserId
-                });
+          const postPermission = await permissionClient.client.forum.computePostPermissions({
+            resourceId: vote.post.id,
+            userId: spaceUserId
+          });
 
-          if (categories.length !== 0 && categories[0].permissions.comment_posts && vote.author.id !== spaceUserId) {
+          if (postPermission.add_comment) {
```
mattcasey, line 83: dont we still need to check the vote author is not the user?

lib/notifications/polls/createPollNotifications.ts
```diff
@@ -56,19 +50,18 @@ export async function createPollNotifications(webhookData: {
...
-              : await premiumPermissionsApiClient.pages.computePagePermissions({
-                  resourceId: vote.page.id,
-                  userId: spaceUserId
-                });
-          if (pagePermission.comment && vote.author.id !== spaceUserId) {
+          const pagePermission = await permissionClient.client.pages.computePagePermissions({
+            resourceId: vote.page.id,
+            userId: spaceUserId
+          });
+          if (pagePermission.comment) {
```
mattcasey, line 64: dont we need to check the vote author?

lib/notifications/polls/createPollNotifications.ts
```diff
@@ -22,28 +21,23 @@ export async function createPollNotifications(webhookData: {
           id: voteId,
           status: 'InProgress'
         },
-        include: {
+        select: {
+          spaceId: true,
           page: {
-            select: { id: true, path: true, title: true }
+            select: { id: true }
```
mattcasey, line 27: looks like you dont need to join on the tables, you can just select createdBy, pageId and postId like you did with space?

mattcasey (changes requested):
looks pretty good but see questions

lib/notifications/proposals/createProposalNotifications.ts
Devorein: `isReviewer` should ideally check if the user is a reviewer or not, not just whether he is a reviewer based on direct user id reference neglecting the role.

lib/notifications/polls/createPollNotifications.ts
Devorein: That's done on line 51, we remove the poll author from the `spaceUserIds`

lib/notifications/polls/createPollNotifications.ts
Devorein: That's done on line 51, we remove the poll author from the `spaceUserIds`

lib/notifications/polls/createPollNotifications.ts
Devorein: Nice one. updated code.

PR #3091, opened 2023-11-13
Fix adding video node inside page/post comments

Devorein:
### <samp>🤖 Generated by Copilot at b6ea6f6</samp>

This pull request adds support for embedding different types of pages in comments and replies, such as posts, courses, or lessons. It does this by passing a `pageType` prop to the relevant components, such as `PageComments`, `CommentForm`, and `CharmEditor`. It also fixes a bug with YouTube video playback in the `CharmEditor` component.

### WHY
1. Add video node in page/post comments
2. Fix the issue with YouTube videos showing "Refused to connect"


components/common/CharmEditor/components/video/VideoNodeView.tsx
```diff
@@ -87,9 +86,14 @@ export function VideoNodeView({
...
-                    if (embedType) {
+                    const youtubeEmbedLink = extractYoutubeEmbedLink(videoUrl);
+                    if (youtubeEmbedLink) {
+                      updateAttrs({
+                        src: videoUrl
+                      });
+                    } else if (embedType) {
                       const pos = getPos();
-                      if (!pos) return;
+                      if (pos === undefined || pos === null) return;
```
mattcasey, line 96: nice catch!

mattcasey (changes requested):
i can't tell but where is pageType being used? using 'type' === 'post' sounds incorrect, because the charm editor doc is not being used as a post

PR #3092, opened 2023-11-13
Add delete icon on table node to delete

Devorein:
### <samp>🤖 Generated by Copilot at 0f4940f</samp>

Improved table editing and rendering in `CharmEditor` component. Removed unused code from `colors.ts` and `tableView.ts`. Added CSS styles for table delete button in `styles.scss`.

### WHY
image


components/common/CharmEditor/components/fiduswriter/collab/colors.ts
```diff
@@ -16,18 +16,6 @@ import type { ModCollab } from './index';
...
-  '--bg-turquoise',
-  '--bg-orange',
-  '--bg-yellow',
-  '--bg-teal',
-  '--bg-blue',
-  '--bg-purple',
-  '--bg-red',
-  '--bg-pink',
-  '--bg-gray'
-];
```
mattcasey, line 29: thx for the cleanup. i think we ended up using the colors of your avatar instead of a static list

mattcasey (changes requested):
we should just not support tables inside of comments.

PR #3094, opened 2023-11-13
Disconnect space from collabland server

Devorein:
### <samp>🤖 Generated by Copilot at d8efe58</samp>

This pull request adds the feature to disconnect a space from Collabland, a Discord-NFT integration service. It modifies the `ConnectCollabland` component to render a `DisconnectCollabland` component that redirects the user to the Collabland dashboard. It also adds a new API route, a server-side function, and unit tests for handling the disconnection logic. It removes an unused import from the `publishEvent` file.


lib/collabland/__tests__/disconnectSpace.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+import { testUtilsUser } from '@charmverse/core/test';
+import { v4 } from 'uuid';
+
+import { createDiscordUser } from 'testing/utils/discord';
+import { addUserToSpace } from 'testing/utils/spaces';
+
+import { disconnectSpace } from '../disconnectSpace';
+
+describe('disconnectSpace', () => {
+  it('should throw error if no user with the provided discord id exist', async () => {
```
mattcasey, line 12: i think it would be better if the function did not return an error in this case. It forces the code to deal with this specific error, because it's not really a problem. The action is already fulfilled, so the code that's calling it shouldn't need to worry then. If it's useful, you could return an error object like "{ notFound: true }"

lib/collabland/__tests__/disconnectSpace.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+        source: 'collabland',
+        spaceId: space3.id
+      }
+    });
+
+    expect(discordConnectedSpaces.map((space) => space.id)).toStrictEqual([space3.id]);
+    expect(space1CollablandRolesCount).toBe(0);
+    expect(space2CollablandRolesCount).toBe(0);
+    // Not removed since user is not admin
+    expect(space3CollablandRolesCount).toBe(1);
```
mattcasey, line 111: this 3rd case should be its own test

lib/collabland/disconnectSpace.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  });
+
+  if (!user) {
+    throw new DataNotFoundError('Cannot find user to disconnect');
+  }
+
+  const adminSpaceIds = user.spaceRoles.filter((role) => role.isAdmin).map((role) => role.spaceId);
+
+  await prisma.space.updateMany({
```
mattcasey, line 38: these queries should be wrapped in `prisma.$transaction([])`

lib/collabland/disconnectSpace.ts
```diff
@@ -0,0 +1,61 @@
...
+      discordUser: {
+        discordId: discordUserId
+      }
+    },
+    select: {
+      spaceRoles: {
+        where: {
+          space: {
+            discordServerId
+          }
```
mattcasey, line 22: i think u could add 'isAdmin: true' here and remove the filtering below?

lib/webhookConsumer/processWebhookMessage.ts
```diff
@@ -78,6 +79,31 @@ const messageHandlers: Record<MessageType, (message: WebhookMessage) => Promise<
...
+      const { userId: discordUserId, guildId: discordServerId } = payload;
+      const result = await disconnectSpace({ discordUserId, discordServerId });
+
+      return {
+        spaceIds: result?.spaceIds ?? [],
+        success: true,
+        message: 'Uninstalled miniapp.'
+      };
+    } catch (e: any) {
+      return {
```
mattcasey, line 102: this needs to capture the case that we didn't find anything to disconect, because it's not an error, the message still should be removed from SQS

mattcasey (changes requested):
needs to not fail if a space was already disconnected

lib/collabland/disconnectSpace.ts
Devorein: Good point. Updated.

lib/collabland/disconnectSpace.ts
Devorein: Oops, careless mistake. Updated thnx.

lib/collabland/__tests__/disconnectSpace.spec.ts
Devorein: Good point. Updated to not throw an error and just return `{spaceIds: []}` indicating no spaces where updated, which would be the same thing if a non admin user triggered this event.

lib/collabland/__tests__/disconnectSpace.spec.ts
Devorein: Sure.

lib/webhookConsumer/processWebhookMessage.ts
Devorein: rather than throwing an error we are now returning an empty spaceIds.

PR #3098, opened 2023-11-13
dont wait 2 seconds before sending app_loaded event

mattcasey:
### <samp>🤖 Generated by Copilot at d9b9756</samp>

Improved app load tracking by using `leading` and `trailing` options of `debounce` function in `hooks/useAppLoadedEvent.ts`. This prevents duplicate or delayed tracking events for the same app load action.

### WHY
We are seeing app_loaded is not account for all active users when looking at page_views. Based on Datadog RUM, we have sessions that are less than 2 seconds, so this could explain it. The debounce options trigger the function immediately, and only 2 seconds after that.


mattcasey:
merging now so that we can see the effects tomorrow

PR #3102, opened 2023-11-14
Whitelabel pages

valentinludu:
### <samp>🤖 Generated by Copilot at 78cc1e9</samp>

This pull request adds support for custom domain spaces to the app. It allows different clients to have their own branding and message on the login page and the header. It uses `@mui/material` components, `useBaseCurrentDomain` hook, `WorkspaceAvatar` component, and `charmClient` library to implement this feature. It also updates the `Header`, `Footer`, `LoginPageContent`, and `_app` components to use the custom domain and space data.

### WHY
Change images and text on pages where users are not logged in.

[Task](https://app.charmverse.io/charmverse/custom-login-for-custom-domains-18798186353187374)


hooks/useBaseCurrentDomain.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const BaseCurrentDomainContext = createContext<Readonly<IBaseCurrentDomainContext>>({
+  isSpaceLoading: true,
+  spaceFromPath: undefined,
+  customDomain: undefined
+});
+
+export function BaseCurrentDomainProvider({ children }: { children: ReactNode }) {
+  const [customDomain, setCustomDomain] = useState<string | null>();
+  const { data: spaceFromPath, isLoading: isSpaceLoading } = useSearchByDomain(customDomain || '');
```
mattcasey, line 24: just make useSearchByDomain take an optional string, so you don't have to add a default empty string

hooks/useBaseCurrentDomain.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { user } = useUser();
+
+  useEffect(() => {
+    if (typeof window !== 'undefined' && !user) {
+      const customDomainFromHost = getCustomDomainFromHost(window?.location.host);
+
+      setCustomDomain(customDomainFromHost);
+    }
+  }, [typeof window !== 'undefined', !user]);
```
mattcasey, line 33: i don't think "typeof window !== 'undefined'" should be a dependency

hooks/useBaseCurrentDomain.tsx
```diff
@@ -0,0 +1,43 @@
...
+  customDomain?: string | null;
+};
+
+export const BaseCurrentDomainContext = createContext<Readonly<IBaseCurrentDomainContext>>({
+  isSpaceLoading: true,
+  spaceFromPath: undefined,
+  customDomain: undefined
+});
+
+export function BaseCurrentDomainProvider({ children }: { children: ReactNode }) {
```
mattcasey, line 22: I don't think this needs to be added to our global context. You can just do this logic inside components/login/LoginPage.tsx, and pass the results down to the components.

pages/_app.tsx
```diff
@@ -251,13 +252,15 @@ function DataProviders({ children }: { children: ReactNode }) {
...
                                 <PaymentMethodsProvider>
                                   <PagesProvider>
-                                    <MemberPropertiesProvider>
-                                      <LensProvider config={lensConfig}>
-                                        <UserProfileProvider>
-                                          <PageTitleProvider>{children}</PageTitleProvider>
-                                        </UserProfileProvider>
-                                      </LensProvider>
-                                    </MemberPropertiesProvider>
+                                    <BaseCurrentDomainProvider>
```
mattcasey, line 255: i believe this can be added to the base layout component (BaseLayout/getLayout.tsx) instead of _app.tsx, so that it only gets run on those pages?

hooks/useBaseCurrentDomain.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [customDomain, setCustomDomain] = useState<string | null>();
+  const { data: spaceFromPath, isLoading: isSpaceLoading } = useSearchByDomain(customDomain || '');
+  const { user } = useUser();
+
+  useEffect(() => {
+    if (typeof window !== 'undefined' && !user) {
```
mattcasey, line 28: why does it matter if user exists? the name of the hook and provider don't imply anything related to logged out users

hooks/useBaseCurrentDomain.tsx
```diff
@@ -0,0 +1,43 @@
...
+      const customDomainFromHost = getCustomDomainFromHost(window?.location.host);
+
+      setCustomDomain(customDomainFromHost);
+    }
+  }, [typeof window !== 'undefined', !user]);
+
+  const value = useMemo(() => {
+    return { isSpaceLoading, spaceFromPath, customDomain };
```
mattcasey, line 36: Suggestion: To avoid flickering after customDomain is set, I think we need to take into account whether the customDomain has been determined in the useEffect. One suggestion would be to use null as its initial state but set it as a string or empty string in the useEffect. And then `isSpaceLoading = isSpaceLoading || customDomain === null`. This way, we can properly hide the UI until we determine if a custom domain exists in the path

components/login/LoginPageContent.tsx
```diff
@@ -47,6 +50,18 @@ export function LoginPageContent({ hideLoginOptions, isLoggingIn, children }: Pr
     }
   }, [discordLoginError]);
 
+  const isCustomDomain = customDomain && spaceFromPath?.name && spaceFromPath.isCustomDomainVerified;
```
mattcasey, line 53: this logic could probably be put inside the hook

components/login/LoginPageContent.tsx
```diff
@@ -47,6 +50,18 @@ export function LoginPageContent({ hideLoginOptions, isLoggingIn, children }: Pr
...
+
+  const image = isCustomDomain ? (
+    <WorkspaceAvatar
+      image={spaceFromPath?.spaceImage || ''}
+      name={spaceFromPath?.name || ''}
+      sx={{ width: '100%', paddingTop: '50%', paddingBottom: '50%', fontSize: '10rem' }}
```
mattcasey, line 59: Suggestion: it would be more ideal if we extended the 'size' prop of WorkspaceAvatar as needed, and didn't need to add an sx prop. Adding so much flexibility makes it harder to understand how changes to the component may affect places it is being used. The padding rules can probably be applied to a container element

mattcasey (changes requested):
left some suggestions but primarily i think we should move the provider to the base layout, instead of _app.tsx

hooks/useBaseCurrentDomain.tsx
valentinludu: I added the user check because the provider was in _app and wanted to make sure I don't trigger the logic when the user is logged in. Now that I moved it inside BaseLayout I know for sure that it will never have a user.
That being said, I deleted the user check.

pages/_app.tsx
valentinludu: Moved!

hooks/useBaseCurrentDomain.tsx
valentinludu: Now that I moved it inside BaseLayout, I know the user will never exist.
I added that so I know not to trigger the logic when user is logged in

components/login/LoginPageContent.tsx
valentinludu: Moved!

components/login/LoginPageContent.tsx
valentinludu: This is a specific thing, where I need to make a square, but don't know how big, that is why I am using this trick with the padding.

hooks/useBaseCurrentDomain.tsx
valentinludu: I am doing something similar.

customDomain can be string | null | undefined;

Undefined means we didn't load it yet.
Null means we don't have a custom domain.
String means we have a custom domain.

mattcasey (approved):

components/common/Avatar.tsx
```diff
@@ -5,10 +5,21 @@ import React from 'react';
 
 import { stringToColor } from 'lib/utilities/strings';
 
-export type AvatarSize = 'xSmall' | 'small' | 'medium' | 'large' | 'xLarge' | '2xLarge';
+export type AvatarSize = 'xSmall' | 'small' | 'medium' | 'large' | 'xLarge' | '2xLarge' | '3xLarge';
 export type AvatarVariant = 'circular' | 'rounded' | 'square';
 
-const sizeStyleMap: Record<AvatarSize, React.CSSProperties> = {
+const sizeStyleMap: Record<AvatarSize, SxProps<Theme>> = {
```
mattcasey, line 11: added to MUI theme -- very nice

PR #3106, opened 2023-11-15
Switching content in page dialog +query based navigation

msadura:
### <samp>🤖 Generated by Copilot at f6fe148</samp>

This pull request refactors and enhances the reward application feature in the app. It uses URL query parameters and a global page dialog modal to display and manage the reward applications in different contexts. It removes some unused and redundant components and hooks, and improves the routing and query management logic. It affects the following files: `usePageDialog.tsx`, `PageDialog.tsx`, `PageLayout.tsx`, `RewardApplicationPage.tsx`, `NewWorkButton.tsx`, `RewardApplications.tsx`, `RewardsPage.tsx`, `RewardsPageWithProviders.tsx`, `useCharmRouter.ts`, `PageDialogGlobal.tsx`, and `_app.tsx`. It also deletes the following files: `RewardApplicationDialog.tsx`, `RewardSubmissionsTable.tsx`, and `useApplicationDialog.tsx`.

### WHY
- page dialog will now handle applications (and possibly other types of conent)
- you can switch content by changing query param, there is no flicker


components/common/PageDialog/PageDialog.tsx
```diff
@@ -44,6 +58,13 @@ function PageDialogBase(props: Props) {
 
   const readOnlyPage = readOnly || !pagePermissions?.edit_content;
 
+  const contentType: ContentViewType | null = useMemo(() => {
```
mattcasey, line 61: not sure useMemo is necessary here. It's just some if/else logic to set a string value?

components/common/PageDialog/PageDialog.tsx
```diff
@@ -116,33 +143,47 @@ function PageDialogBase(props: Props) {
...
+          </Box>
+        )
       }
       onClose={close}
     >
-      {page && (
+      {page && contentType === 'page' && (
         <DocumentPage page={page} savePage={savePage} refreshPage={refreshPage} readOnly={readOnlyPage} close={close} />
       )}
+      {contentType === 'application' && pageDialogContext && (
```
mattcasey, line 180: i think i would have called this "applicationContext". It's a little confusing that now, in addition to the pageId prop for PageDialog, there is a pageDialogContext.pageId property as well. but you wouldn't know this is only used for applications

components/common/PageDialog/hooks/usePageDialog.tsx
```diff
@@ -25,10 +30,28 @@ export const usePageDialog = () => useContext(ContextElement);
...
+    updateURLQuery
+  } = useCharmRouter();
+
+  useEffect(() => {
+    setProps((prevProps) => ({
+      ...prevProps,
+      // application props based on router query rather than explicit showing / hiding dialog
+      applicationId:
+        query.applicationId && isUUID(query.applicationId as string) ? (query.applicationId as string) : undefined,
+      isNewApplication: query.applicationId === 'new',
```
mattcasey, line 44: nice idea to use the URL for "new" state

components/common/PageLayout/PageLayout.tsx
```diff
@@ -226,6 +227,8 @@ function PageLayout({ children }: PageLayoutProps) {
                 <HeaderSpacer />
                 {children}
               </PageContainer>
+
+              <PageDialogGlobal />
```
mattcasey, line 231: I would like to leave PageDialogGlobal out of the PageLayout component. It's very large and complicated already, but not every page that uses it needs a dialog component (like EditorPage). It was already included in RewardPageWithProviders.

components/rewards/RewardsPage.tsx
```diff
@@ -110,8 +98,7 @@ export function RewardsPage({ title }: { title: string }) {
       if (id && (!rewardId || id === rewardId)) {
         openPage(id);
       } else if (id) {
-        setUrlWithoutRerender(router.pathname, { applicationId: id });
-        showApplication(id);
+        updateURLQuery({ applicationId: id });
```
mattcasey, line 101: ah thank u, glad it works

components/common/PageDialog/hooks/usePageDialog.tsx
```diff
@@ -25,10 +30,28 @@ export const usePageDialog = () => useContext(ContextElement);
 
 export function PageDialogProvider({ children }: { children: ReactNode }) {
   const [props, setProps] = useState<PageDialogContext>({});
+  const {
+    router: { query },
+    updateURLQuery
+  } = useCharmRouter();
+
+  useEffect(() => {
+    setProps((prevProps) => ({
```
mattcasey, line 39: We should call showPage() instead of adding logic inside of PageDialogProvider, if it's feasible. This component intentionally has no logic, so to avoid coupling the logic of each page or component that uses it. For example, if applications are only going to be visible on /rewards, then the logic should go in RewardsPage.tsx instead.

mattcasey (approved):
Works well, and the file sizes do not increase. Had some comments on organization and boundaries

components/common/PageLayout/PageLayout.tsx
msadura: ok I moved it to EditorPage (as we need dialog for applicaitons in full-page mode)

components/common/PageDialog/hooks/usePageDialog.tsx
msadura: I will move this logic to external hook to keep this one clean, but I want to stick with navigation through query params, as showPage + setUrlWithourRerender is really anti-pattern imo, because you basically cannot trust query from router as it is not always up to date

PR #3114, opened 2023-11-15
When pasting a link to a page in CV, convert it to a page mention

mattcasey:
### WHY

When working on custom domains, it includes my custom domain. I want it to just be a 'relative' link. I think this is better UX too since we pull up the correct page title.

### DIFF
--- a/components/common/CharmEditor/components/link/link.plugins.ts
+++ b/components/common/CharmEditor/components/link/link.plugins.ts
@@ -39,7 +39,6 @@ export function plugins({ key }: { key: PluginKey }) {
         },
         apply(tr, pluginState) {
           const meta = tr.getMeta(key);
-          // console.log('apply state', key, pluginState, meta, pages);
           if (meta === undefined) {
             return pluginState;
           }
--- a/components/common/CharmEditor/components/link/link.plugins.ts
+++ b/components/common/CharmEditor/components/link/link.plugins.ts
@@ -2,7 +2,6 @@
 import type { Mark, PluginKey } from '@bangle.dev/pm';
 import { Plugin } from '@bangle.dev/pm';
 
-import { linkPlugins } from '../@bangle.dev/base-components/link';
 import { createTooltipDOM, tooltipPlacement } from '../@bangle.dev/tooltip';
 import {
   hideSuggestionsTooltip,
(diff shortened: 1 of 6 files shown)


components/common/CharmEditor/components/link/link.ts
```diff
@@ -105,14 +142,16 @@ export function markPasteRule(regexp: RegExp, type: MarkType, getAttrs: (...para
   });
 }
 
-const LINK_REGEX = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{2,256}\.[a-zA-Z]{2,}\b([-a-zA-Z0-9@:%_'"+.~#?&//=]*)/g;
+const LINK_REGEX =
```
mattcasey, line 145: made an update so that http://localhost paths work

components/common/CharmEditor/components/link/link.plugins.ts
```diff
@@ -11,32 +10,36 @@ import {
...
     new Plugin<LinkPluginState>({
       key,
       state: {
         init() {
           return {
             show: false,
             href: null,
-            tooltipContentDOM: tooltipDOMSpec.contentDOM
+            tooltipContentDOM: tooltipDOMSpec.contentDOM,
+            pages: {}
```
mattcasey, line 37: perhaps this should be stored in another plugin... then, other plugins could use it. i realized at the very end we have a 'UserMetaDataPlugin' for managing the user's info, for instance.

components/common/CharmEditor/components/link/link.ts
```diff
@@ -81,7 +88,37 @@ export function pasteLink(regexp: RegExp) {
...
+        slice.content.descendants((node) => {
+          const linkMark = node.marks.find((mark) => mark.type.name === 'link');
+          if (linkMark) {
+            url = linkMark.attrs.href;
+          }
+        });
+        if (url) {
+          const pageMap = key.getState(view.state).pages;
+          if (pageMap) {
+            const pagePathMaybe = url.split('/').pop() || '';
```
Devorein, line 106: Shouldn't this check if the origin is the same? Although its highly unlikely that another link would end with the same the url chunk as a page path

components/common/CharmEditor/components/link/link.ts
```diff
@@ -81,7 +88,37 @@ export function pasteLink(regexp: RegExp) {
...
+            const foundPageId = pageMap[pagePathMaybe];
+            if (foundPageId) {
+              insertNode('mention', view.state, view.dispatch, view, { id: v4(), type: 'page', value: foundPageId });
```
Devorein, line 109: It might not always be `type: 'page'`, for example for rewards, posts and proposals.

Devorein (approved):

PR #3118, opened 2023-11-16
Feat/filteriing proposals rewards

msadura:
### WHY

Enabling + fixing advanced filters for rewards and  proposals

Screenshot 2023-11-16 at 16 43 36


components/common/BoardEditor/components/ViewFilterControl.tsx
```diff
@@ -0,0 +1,44 @@
+import { Popover } from '@mui/material';
```
msadura, line 1: this is just moved from viewHeader.tsx

PR #3122, opened 2023-11-17
"Open page in" option for database views

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/allow-setting-default-view-type-of-a-board-particularly-on-proposals-page-8846576127534682)
image


components/[pageId]/DatabasePage/DatabasePage.tsx
```diff
@@ -119,27 +120,32 @@ export function DatabasePage({ page, setPage, readOnly = false, pagePermissions
...
+  const showCard = async (cardId: string | null) => {
+    if (cardId === null) {
+      setShownCardId(null);
+      return;
+    }
+
+    const cardPage = cardId && pages ? pages[cardId] ?? (await refreshPage(cardId)) : null;
+
+    if (activeView.fields.openPageIn === 'center_peek' || cardPage?.type.includes('template')) {
+      updateURLQuery({ viewId: router.query.viewId as string, cardId });
```
Comment (mattcasey) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:132]: happy to see we're trying to use the normal query for updating state

components/[pageId]/DatabasePage/DatabasePage.tsx
```diff
@@ -119,27 +120,32 @@ export function DatabasePage({ page, setPage, readOnly = false, pagePermissions
...
-      });
-    },
-    [router.query]
-  );
+    } else if (activeView.fields.openPageIn === 'full_page' && cardPage) {
+      navigateToSpacePath(`/${cardPage.path}`);
+    }
+  };
+
+  const showView = (viewId: string) => {
```
Comment (mattcasey) [inline components/[pageId]/DatabasePage/DatabasePage.tsx:139]: why did you remove useCallback? I added this to try and avoid the database re-rendering too often. i think it was worth keeping in

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -653,7 +664,12 @@ function CenterPanel(props: Props) {
                     readOnly={props.readOnly}
                     onCardClicked={cardClicked}
                     selectedCardIds={state.selectedCardIds}
-                    addCard={(show) => addCard('', show)}
+                    addCard={(show) =>
```
mattcasey, line 667: we should memoize all the methods passed to this component

components/common/BoardEditor/focalboard/src/components/viewSidebar/viewLayoutOptions.tsx
```diff
@@ -22,12 +23,14 @@ interface LayoutOptionsProps {
   board?: Board;
   view: BoardView;
   intl: IntlShape;
+  hideLayoutSelectOptions?: boolean;
```
mattcasey, line 26: this should be "showLayoutOptions".

components/common/CharmEditor/components/inlineDatabase/components/InlineDatabase.tsx
```diff
@@ -91,9 +93,9 @@ export function InlineDatabase({ containerWidth, readOnly: readOnlyOverride, nod
   const views = useAppSelector((state) => selectSortedViews(state, pageId));
   const router = useRouter();
   const dispatch = useAppDispatch();
-
+  const { navigateToSpacePath } = useCharmRouter();
   const [currentViewId, setCurrentViewId] = useState<string | null>(views[0]?.id || null);
-
+  const { pages, refreshPage } = usePages();
```
mattcasey, line 98: lets just update the serverside query in /pages/[domain]/[pageId]. we are already querying the db whenever u visit, so we could check if the pageId is a UUID and query by page id

mattcasey (changes requested):
- add back useCallbacks
- instead of looking up page path, because usePages() has performance cost, i'd like to just do lookup by pageId in the server-side lookup in pages/domain/pageId.ts

components/[pageId]/DatabasePage/DatabasePage.tsx
Comment (Devorein) [reply components/[pageId]/DatabasePage/DatabasePage.tsx:139]: Updated. Initially, I was trying to debug a bug that was due to a stale state, and removed the `useCallbacks`, to check whether that's the issue or not. That came out negative, but I forgot to reinstate those back.

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
Devorein: Good point. Updated.

components/common/CharmEditor/components/inlineDatabase/components/InlineDatabase.tsx
Devorein: That makes sense. Updated.

PR #3126, opened 2023-11-18
Blacklist and block user from creating space

Devorein:
### WHY
[Card](https://app.charmverse.io/charmverse/blacklist-users-from-creating-spaces-in-charmverse-38505691901904315)


components/common/CreateSpaceForm/CreateSpaceForm.tsx
```diff
@@ -170,6 +170,14 @@ export function CreateSpaceForm({ className, defaultValues, onCancel, submitText
     }
   }, []);
 
+  useEffect(() => {
```
mattcasey, line 173: this does'nt need to be an effect, just do:
```
const isBanned = typeof saveError === 'string' && saveError.includes('blocked');

```

lib/members/checkUserSpaceBanStatus.ts
```diff
@@ -11,7 +11,7 @@ export async function checkUserSpaceBanStatus({
   discordId?: string;
   emails?: string[];
   userId?: string;
-  spaceIds: string[];
+  spaceIds?: string[];
```
mattcasey, line 14: looks like this should still be required. why would someone call checkUserSpaceBanStatus() without any space Id?

lib/users/blacklistUser.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    include: {
+      discordUser: true,
+      verifiedEmails: true,
+      wallets: true,
+      googleAccounts: true
+    }
+  });
+
+  if (!user) {
```
mattcasey, line 17: using findUniqueOrThrow() is better

lib/users/checkUserBlacklistStatus.ts
```diff
@@ -0,0 +1,53 @@
...
+        {
+          userId
+        },
+        {
+          walletAddresses: {
+            hasSome: userWalletAddresses
+          }
+        },
+        {
+          discordId: userDiscordId
```
mattcasey, line 41: i think this will always return a user with discordId === null? thats probably why the original method used 'in'

mattcasey (changes requested):
see comments

lib/members/checkUserSpaceBanStatus.ts
Devorein: Yikes, nice catch. This was for sure a mistake. Thank you.

lib/users/blacklistUser.ts
Devorein: Updated. Thanks.

components/common/CreateSpaceForm/CreateSpaceForm.tsx
Devorein: Good one, removed the state as well.

lib/users/checkUserBlacklistStatus.ts
Devorein: The original one was an array, but you're right. Updated with the fix.

PR #3127, opened 2023-11-18
Disable updating expired invite link

Devorein:
### WHY

image

### DIFF
--- a/components/settings/invites/components/InviteLinks/components/InviteLinksTable.tsx
+++ b/components/settings/invites/components/InviteLinks/components/InviteLinksTable.tsx
@@ -15,9 +15,11 @@ import Countdown from 'react-countdown';
 import ButtonChip from 'components/common/ButtonChip';
 import TableRow from 'components/common/Table/TableRow';
 import TokenGateRolesSelect from 'components/settings/invites/components/TokenGates/components/TokenGateRolesSelect';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
 import { useIsAdmin } from 'hooks/useIsAdmin';
 import { useSpaceInvitesList } from 'hooks/useSpaceInvitesList';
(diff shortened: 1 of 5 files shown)


components/settings/invites/components/InviteLinks/components/InviteLinksTable.tsx
```diff
@@ -65,73 +65,82 @@ export function InvitesTable() {
...
-                  placement='top'
-                  title={copied[invite.id] ? 'Copied!' : 'Click to copy link'}
-                  disableInteractive
-                >
-                  <Box component='span'>
-                    <CopyToClipboard text={getInviteLink(invite.code)} onCopy={() => onCopy(invite.id)}>
-                      <Chip
-                        sx={{ width: 90 }}
+          {privateInvites?.map((invite) => {
+            const hasExpired =
```
mattcasey, line 69: looks like you can call isOutOfDate() from lib/invites instead of re-writing this logic inline.

lib/invites/updateInviteLinkRoles.ts
```diff
@@ -29,6 +31,15 @@ export async function updateInviteLinkRoles({
     throw new DataNotFoundError(`Invite link with id ${inviteLinkId} not found`);
   }
 
+  const hasExpired =
```
mattcasey, line 34: for the future i dont see why this rule needs to be added to the backend. it doesnt really matter to anyone if someone updates a link that is expired.

mattcasey (changes requested):
M (must): - re-use the logic from our lib to determine expiration. This is a good practice we should always follow when it comes to doing math or translating dates
S (should): - remove restrictions on the backend that dont really matter, as its duplicate work to maintain

lib/invites/updateInviteLinkRoles.ts
Devorein: Fair point. I will remove this part and the corresponding tests as well.

PR #3129, opened 2023-11-18
Fix zora chain nft error and title display

valentinludu:
Fix:
- zora was not on the LIT_CHAINS list
- zora tokenId was sent as a number instead of a string. The ZDK was throwing an error.
- nft titles

### DIFF
--- a/lib/blockchain/provider/zora/getNFT.ts
+++ b/lib/blockchain/provider/zora/getNFT.ts
@@ -26,7 +26,7 @@ export async function getNFT({
   const response = await provider.token({
     token: {
       address,
-      tokenId
+      tokenId: String(tokenId)
     },
     network: {
       network: ZDKNetwork.Zora,
--- a/lib/blockchain/provider/zora/getNFT.ts
+++ b/lib/blockchain/provider/zora/getNFT.ts
@@ -26,7 +26,7 @@ export async function getNFT({
   const response = await provider.token({
     token: {
       address,
-      tokenId: String(tokenId)
+      tokenId
     },
     network: {
       network: ZDKNetwork.Zora,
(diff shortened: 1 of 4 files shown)


Devorein (approved):

lib/blockchain/provider/zora/getNFT.ts
```diff
@@ -26,7 +26,7 @@ export async function getNFT({
   const response = await provider.token({
     token: {
       address,
-      tokenId
+      tokenId: String(tokenId)
```
mattcasey, line 29: why do we do this? looks like a hack

mattcasey: i guess we're doing it for ankr too 🤷

lib/tokenGates/utils.ts
```diff
@@ -31,3 +33,30 @@ export function getAccessType(condition: AccsDefaultParams): TokenGateAccessType
...
+  name: base.name,
+  symbol: 'ETH',
+  decimals: 18,
+  rpcUrls: base.rpcUrls.default.http.slice(),
+  blockExplorerUrls: [base.blockExplorers.default.url],
+  type: null,
+  vmType: 'EVM'
+};
+
+const zoraChain = {
```
mattcasey, line 49: please import from connectors/chains instead. we may replace a lot of it with wagmi eventually but at least for now we are consistent where all the details are coming from. they're not 100% wagmi right now

mattcasey (changes requested):
- always import chain config from connections/chains. this makes sure we're using the same config everywhere

lib/tokenGates/utils.ts
```diff
@@ -31,3 +33,30 @@ export function getAccessType(condition: AccsDefaultParams): TokenGateAccessType
 export function getAccessTypes(conditions: AccsDefaultParams[]): TokenGateAccessType[] {
   return conditions.map((c) => getAccessType(c));
 }
+
+const baseChain = {
```
mattcasey, line 37: this is a config file, it doesnt really belong in a "utils" file. please put it in a constants.ts or config.ts

lib/blockchain/provider/zora/getNFT.ts
valentinludu: Condition parameters can be number | string and I did not know because the type was 'any'. Now I am inforcing the String() where I search for the parameters, instead of here.

mattcasey (approved):
talked about it and its fine to import wagmi config directly

PR #3132, opened 2023-11-19
Fixed cryptoprice block not working in public pages

Devorein:
### WHY

Before:
image

Now:
image

### DIFF
--- a/lib/middleware/requireUser.ts
+++ b/lib/middleware/requireUser.ts
@@ -8,24 +8,3 @@ export function requireUser(req: NextApiRequest, res: NextApiResponse, next: Nex
     next();
   }
 }
-
-/**
- * Allow an endpoint to be consumed if it originates from a share page
- */
-export function requireUserOrSharePage(req: NextApiRequest, res: NextApiResponse, next: NextHandler) {
-  const referer = req.headers.referer as string;
-
-  const url = new URL(referer);
-  url.hash = '';
-  url.search = '';
-  const pathnameParts = referer ? url.pathname.split('/') : [];
-  const spaceDomain = pathnameParts[1];
-
-  if (spaceDomain === 'share') {
-    next();
-  } else if (!req.session.user) {
-    res.status(401).send({ error: 'Please log in' });
-  } else {
-    next();
-  }
-}
--- a/theme/@bangle.dev/styles.scss
+++ b/theme/@bangle.dev/styles.scss
@@ -270,7 +270,7 @@ ol li {
   pointer-events: none;
 }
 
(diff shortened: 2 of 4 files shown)


theme/@bangle.dev/styles.scss
```diff
@@ -270,7 +270,7 @@ ol li {
   pointer-events: none;
 }
 
-.ProseMirror-selectednode {
+.bangle-editor-core:not(.readonly) .ProseMirror-selectednode {
```
mattcasey, line 273: this seems hacky, are you 100% positive we never want to show a selected node in readonly? why would someone not be able to select a node? they can highlight text, for example

lib/middleware/requireUser.ts
```diff
@@ -8,24 +8,3 @@ export function requireUser(req: NextApiRequest, res: NextApiResponse, next: Nex
     next();
   }
 }
-
-/**
- * Allow an endpoint to be consumed if it originates from a share page
- */
-export function requireUserOrSharePage(req: NextApiRequest, res: NextApiResponse, next: NextHandler) {
```
mattcasey, line 15: when you remove logic in a PR please explain why

theme/@bangle.dev/styles.scss
mattcasey: you should also include a comment why this was added

mattcasey (approved):
In general on your PRs, please provide more context about why things are added/removed especially if it's kind of vague changes not related to the PR title

theme/@bangle.dev/styles.scss
Devorein: That's actually a good point, there is actually a plugin called `node-view` that handles adding and removing that class. Initially I wanted to fix a bug where the CryptoPrice block always stayed selected after the document loaded, this was a temporary fix, but I forgot to remove it after realizing it was not working. I will remove the class

lib/middleware/requireUser.ts
Devorein: My bad, I'm adding it right away.

lib/middleware/requireUser.ts
Devorein, line 1: Removed `requireUserOrSharePage` since we are no longer user `share` style for public pages and its only being used in the cryptoPrice endpoint.

lib/middleware/requireUser.ts
Devorein, line 15: Removed `requireUserOrSharePage` since we are no longer using `/share` style for public pages and it's only being used in the cryptoPrice endpoint.

PR #3136, opened 2023-11-19
Fixed text filter for title property

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/db-title-text-filter-does-not-work-as-expected-5834628543749347)


mattcasey (changes requested):
i'd like to avoid adding pages into the selectors - pretty sure this will have negative effects on performance. if we are going to use page context, it should probably be added to Redux as a new store and not passed in as a prop.

But we already add the page titles to cards in the CetnerPanel compoentn - isn't this enough? https://github.com/charmverse/app.charmverse.io/blob/main/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx#L148

If not, i'd like to take a step back and do it properly by adding pages to redux instead of adding another workaround. this would help performance of db's too i think

Devorein:
`[main/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx#L148](https://github.com/charmverse/app.charmverse.io/blob/main/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx?rgh-link-date=2023-11-20T19%3A06%3A46Z#L148)`

This adds the title property after the filter is done, so if the filter contains a title-based clause then the card won't even reach to that stage

components/common/BoardEditor/focalboard/src/svg/dashboard-onboarding.tsx
Devorein, line 1: This component was unused

components/common/BoardEditor/focalboard/src/svg/search-illustration.tsx
Devorein, line 1: This component was also unused

components/common/BoardEditor/focalboard/src/components/viewHeader/filterComponent.tsx
```diff
@@ -133,7 +133,13 @@ const FilterComponent = React.memo((props: Props) => {
         </Stack>
       )}
 
-      <Button variant='outlined' color='secondary' size='small' onClick={addFilterClicked} sx={{ mt: 1 }}>
+      <Button
```
Devorein, line 136: Removed redundant margin-top for filter
image

pages/_app.tsx
```diff
@@ -249,15 +247,17 @@ function DataProviders({ children }: { children: ReactNode }) {
...
+                                    <PagesProvider>
+                                      <MemberPropertiesProvider>
+                                        <LensProvider config={lensConfig}>
+                                          <UserProfileProvider>
+                                            <PageTitleProvider>{children}</PageTitleProvider>
+                                          </UserProfileProvider>
+                                        </LensProvider>
+                                      </MemberPropertiesProvider>
+                                    </PagesProvider>
+                                  </FocalBoardProvider>
```
Devorein, line 260: FocalboardProvider needs to wrap PagesProvider, otherwise `useAppDispatch` is not available inside `usePages`

components/common/BoardEditor/focalboard/src/store/cards.ts
```diff
@@ -70,23 +73,75 @@ const cardsSlice = createSlice({
...
+          state.cardPages[page.id] = { title: page.title };
+        }
+      });
+      for (const card of Object.values(state.cards)) {
+        const cardPage = action.payload[card.id];
+        if (cardPage) {
+          card.fields.properties[Constants.titleColumnId] = cardPage.title ?? '';
+        }
+      }
+    });
```
Devorein, line 111: This is a bit of a hack. Stored another pagesMap in the slice of cards, the reason being its not possible to know which will thunk will be fulfilled first, `pagesLoad` or `initialDatabaseLoad`. So state.cards might be empty when pages are loaded in redux state, which means no cards will have its title property overriden and vice versa. We can remove this if we don't plan to keep the card title updated with the page title on initial load, or there could be a much more efficient way.

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -15,6 +15,7 @@ import type { WrappedComponentProps } from 'react-intl';
 import { injectIntl } from 'react-intl';
 import type { ConnectedProps } from 'react-redux';
 import { connect } from 'react-redux';
+import { useSWRConfig } from 'swr';
```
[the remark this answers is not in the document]
Devorein: Updated. Thnx.

components/common/BoardEditor/focalboard/src/store/cards.ts
```diff
@@ -70,23 +73,75 @@ const cardsSlice = createSlice({
...
+
+    builder.addCase(initialDatabaseLoad.fulfilled, (state, action) => {
+      const cardPages = state.cardPages;
       for (const block of action.payload) {
         if (block.type === 'card' && block.fields.isTemplate) {
           state.templates[block.id] = block as Card;
         } else if (block.type === 'card' && !block.fields.isTemplate) {
           state.cards[block.id] = block as Card;
+          if (cardPages) {
```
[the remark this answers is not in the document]
Devorein: Removed this part of code after our discussion tmrw.

PR #3138, opened 2023-11-20
Show view header with 0 proposal/rewards

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/show-the-view-headers-when-there-are-0-proposals-or-rewrads-8372058249562722)
image

image


mattcasey (approved):
It works but i think its weird we don't show View and Filter. They are not visible whether i just created a space or deleted all proposals:

image

PR #3139, opened 2023-11-20
Show updated sidebar features title in editor

Devorein:
…d page

### WHY

[Card](https://app.charmverse.io/charmverse/if-change-name-of-forum-link-to-page-still-shows-forum-16800419950317025)
image


components/common/CharmEditor/CharmEditor.tsx
```diff
@@ -236,6 +239,13 @@ function CharmEditor({
   const { user } = useUser();
   const { threads } = useThreads();
 
+  const spaceFeatures = (currentSpace?.features as FeatureJson[]) ?? [];
+
+  const extendedFeatures = STATIC_PAGES.map(({ feature, ...restFeat }) => ({
```
mattcasey, line 244: this logis is already managed via the useFeaturesAndMembers() hook, and i think it should be added to NestedPage/LinkedPagesList components instead of CharmEditor. remember please keep this file compleixty down

mattcasey (changes requested):
see comment

components/common/CharmEditor/CharmEditor.tsx
Devorein: Yeah that's true, it was mostly for performance reasons so that the hook is not called for all the NestedPage/LinkedPage components, since there could be a lot of them in the editor. I will remove it as a prop.

PR #3141, opened 2023-11-20
Replace bounties to rewards in UI

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/update-the-word-bounties-to-rewards-in-settings-7544261640650638)


mattcasey (approved):
we need another story to use the 'renamed' version of Rewards, for example in breadcrumbs

msadura (approved):

PR #3142, opened 2023-11-20
Feat/local sort filter

msadura:
### WHY

Custom sort and filter for rewards and proposals

https://app.charmverse.io/charmverse/tech-task-list-49366552253645923?viewId=50d940bb-fdbd-40ba-9816-616d6138a663&cardId=2496d4ac-59fd-487b-ab53-1b67d1fd5b23


charmClient/charmClient.ts
```diff
@@ -332,6 +332,11 @@ class CharmClient {
       const fbBlockInput = Object.assign(currentFBBlock, updates, {
         fields: { ...(currentFBBlock.fields as object), ...updatedFields }
       });
+
+      // delete local fields before saving
+      delete fbBlockInput.fields.localSortOptions;
```
mattcasey, line 337: did u consider a way to do this that doesn't require mutating the block? this seems pretty brittle, and we might accidentally write them to the db

mattcasey (approved):
looks fine but i think its a bad idea to mutate the fields with properties that "should not be saved." I would just go ahead and expect they will get saved on accident somewhere in the future

PR #3146, opened 2023-11-20
add reward token information to rewards table

mattcasey:
### WHY
asked for by Alex/Chris

```
| [logo] [Amount]  | [Token on Chain] | [Custom]
3 columns
Then I can filter for ETH on Base and then sort by amount.
so 3 new columns
```
image


components/rewards/components/RewardsBoard/utils/boardData.ts
```diff
@@ -85,3 +79,82 @@ export function getDefaultTableView({ storedBoard }: { storedBoard: RewardProper
 
   return view;
 }
+
+const rewardStatusOptions: { id: BountyStatus; value: string; color: keyof (typeof Constants)['menuColors'] }[] = [
```
mattcasey, line 83: i move this into here because it wasn't used anywhere else.

lib/focalboard/board.ts
```diff
@@ -36,6 +36,8 @@ export type PropertyType =
   | 'createdBy'
   | 'updatedTime'
   | 'updatedBy'
+  | 'tokenAmount'
```
Devorein, line 39: is `tokenAmount` just a number?

Devorein (commented):
Some small bugs/UI suggestions:-

1. Custom reward value cell seems to be editable directly from table view, just like title
2. Token Chain column uses the text icon, based on its filter value it should be a multi-select
3. Sorting based on token amount is not working
4. Desc sorting in `Token chain` column is reversed
5. Renaming the newly introduced columns not working
6. Removing "Delete", "Duplicate", "Insert Left/Right" options from the column menu
image

lib/focalboard/board.ts
mattcasey: currently yes but the final design is to include the token in display, similar to contract

PR #3151, opened 2023-11-21
Cleanup custom boards blocks

msadura:


lib/proposal/blocks/createBlock.ts
```diff
@@ -23,28 +25,32 @@ export async function createBlock({
...
 
+  const id = type === 'board' ? DEFAULT_BOARD_BLOCK_ID : data.id || v4();
+
   // there should be only 1 block with properties for space
-  if (type === 'board') {
-    // there should be only 1 block with properties for space
-    const currentPropertiesBlock = await tx.proposalBlock.findFirst({ where: { spaceId, type } });
+  if (id.startsWith('__')) {
+    // there should be only 1 block each space with particular internal id
+    const existingBlock = await tx.proposalBlock.findFirst({ where: { spaceId, id } });
```
motechFR, line 33: we could use prisma.findUnique here with the composite field

lib/proposal/blocks/updateBlock.ts
```diff
@@ -20,6 +23,13 @@ export function updateBlock({
   }
 
   if (data.type === 'board' || data.type === 'view') {
+    let upsertFields = data.fields;
+
+    if (data.type === 'board') {
+      // add default bord fields, TODO: insert all default view ids (__defaultView, __defaultBoardView, __defaultCalendarView)
+      upsertFields = createBoard({ block: { fields: { ...data.fields, viewIds: [] } } }).fields;
```
motechFR, line 30: TODO: insert the viewIds before this is merged ;)

lib/proposal/blocks/updateBlock.ts
```diff
@@ -20,6 +23,13 @@ export function updateBlock({
   }
 
   if (data.type === 'board' || data.type === 'view') {
+    let upsertFields = data.fields;
```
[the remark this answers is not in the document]
msadura: nice ideal, actually focalboar mutator uses only update method which is responsible for upserting, so I will update it

PR #3154, opened 2023-11-22
Include additional blocks for space when counting total block quota

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/create-spaces-for-op-projects-with-additional-free-blocks-37654345295167957)


components/settings/subscription/hooks/useSpaceSubscription.ts
```diff
@@ -31,7 +34,8 @@ export function useSpaceSubscription() {
     [spaceSubscription?.status]
   );
 
-  const spaceBlockQuota = defaultFreeBlockQuota * 1000 + (spaceSubscription?.blockQuota || 0) * 1000;
+  const spaceBlockQuota =
+    defaultFreeBlockQuota * 1000 + (spaceSubscription?.blockQuota || 0) * 1000 + additionalBlockQuota;
```
valentinludu, line 38: if defaultBlockQuota and spaceSubscription.blockQuota are in the form of 30(K), then I suggest to make additionalBlockQuota the same and multiply it here by 1000 like the rest of them.

valentinludu (approved):
Looks good. Added some small stuff.

components/settings/subscription/hooks/useSpaceSubscription.ts
Devorein: Good point, will update to reflect `spaceSubscription.blockQuota` using thousands

components/settings/subscription/hooks/useAdditionalBlockQuota.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function useAdditionalBlockQuota() {
+  const { space: currentSpace } = useCurrentSpace();
+  const { user } = useUser(); // a user session is required to get a result
+
+  const {
+    data: additionalBlockQuota = 0,
+    isLoading,
+    mutate: refetchAdditionalBlockCount
+  } = useSWR(currentSpace && user ? `/spaces/${currentSpace.id}/additional-block-quota` : null, () =>
```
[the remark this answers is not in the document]
Devorein: Matt suggested totally getting rid of this endpoint and reusing the block-count endpoint to retrieve the additional quota. So we reused `useBlockCount` hook.

PR #3159, opened 2023-11-22
Fix and update reward templates

mattcasey:
### WHAT

- make all fields optional for reward templates
- fix: reward templates are appearing in the main board

### WHY

[
](https://discord.com/channels/894960387743698944/1176944385678643301/1176944385678643301)


components/rewards/hooks/useNewReward.tsx
```diff
@@ -57,13 +55,6 @@ export function useNewReward() {
             throw err;
           })
           .then((reward) => {
-            mutateRewards(
```
mattcasey, line 60: this was the fix for reward templates - the hook is assuming that rewards are being updated, but it's actually the reward templates. both of the contexts already update themsevles, so there's no need for mutate

PR #3164, opened 2023-11-23
Feat/web3 credentials

motechFR:
### WHAT

This enables attestations stored on Ceramic

### WHY

Enabling Verifiable Credentials in-app

### DIFF
--- a/charmClient/apis/credentialsApi.ts
+++ b/charmClient/apis/credentialsApi.ts
@@ -1,0 +1,9 @@
+import * as http from 'adapters/http';
+import type { CharmVerseCredentialInput } from 'lib/credentials/attest';
+import type { CredentialData } from 'lib/credentials/schemas';
+
+export class CredentialsApi {
+  attest(data: CharmVerseCredentialInput): Promise<any> {
+    return http.POST(`/api/credentials/attest`, data);
+  }
+}
--- a/charmClient/charmClient.ts
+++ b/charmClient/charmClient.ts
@@ -38,6 +38,7 @@ import type { TelegramAccount } from 'pages/api/telegram/connect';
 import { BlockchainApi } from './apis/blockchainApi';
 import { BountiesApi } from './apis/bountiesApi';
 import { CommentsApi } from './apis/commentsApi';
+import { CredentialsApi } from './apis/credentialsApi';
 import { DiscordApi } from './apis/discordApi';
 import { FileApi } from './apis/fileApi';
 import { ForumApi } from './apis/forumApi';
@@ -119,6 +120,8 @@ class CharmClient {
   gnosisSafe = new GnosisSafeApi();
 
   rewards = new RewardsApi();
+
+  credentials = new CredentialsApi();
 
   async socket() {
     return http.GET<SocketAuthResponse>('/api/socket');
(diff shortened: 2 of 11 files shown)


lib/credentials/config/credentials.gql
```diff
@@ -0,0 +1,26 @@
...
+  issuer: String! @string(maxLength: 42, minLength: 42)
+  recipient: String! @string(maxLength: 42)
+  content: String! @string(maxLength: 100000)
+  sig: String! @string(maxLength: 100000)
+  type: CredentialType!
+  verificationUrl: String! @string(maxLength: 100000)
+  chainId: Int!
+  schemaId: String! @string(maxLength: 1000)
+  timestamp: DateTime!
+}
```
motechFR, line 26: @mattcasey Here is the GQL schema representing what's actually stored on Ceramic.

Once we lock in the model, this is where we'll update

lib/credentials/config/deploy-composites.ts
```diff
@@ -0,0 +1,49 @@
...
+  return `${folder}/generated/${filename}`;
+}
+
+export const compositeDefinitionFile = getCompositesFolder('composite-definition.json');
+
+/**
+ * @param {Ora} spinner - to provide progress status.
+ * @return {Promise<void>} - return void when composite finishes deploying.
+ */
+export async function writeComposite({ spinner }: { spinner: OraSpinner } = { spinner: ora() }) {
```
motechFR, line 29: This is the script that actually creates the json definition consumed by graphql server for read / write operations.

Currently affected by a esm / cjs issue. Overridable by using type: module in package.json when running script

lib/credentials/config/queriesAndMutations.ts
```diff
@@ -0,0 +1,100 @@
...
+  issuer: string;
+  recipient: string;
+  content: string;
+  sig: string;
+  type: CredentialType;
+  verificationUrl: string;
+  chainId: number;
+  schemaId: string;
+  timestamp: Date;
+};
```
motechFR, line 28: Interface directly matching graphql definition. We also have SignedCredential inside attestations methods which was created at start of project.

Might wanna dedupe PublishedSignedCredential and SignedCredential

lib/credentials/config/queriesAndMutations.ts
```diff
@@ -0,0 +1,100 @@
...
+        }
+      }
+    }
+  });
+  return response.data.createSignedCredentialThree.document;
+}
+
+const GET_CREDENTIALS = gql`
+  query GetCredentials($filter: SignedCredentialThreeFiltersInput!) {
+    signedCredentialThreeIndex(filters: $filter, first: 1000) {
```
motechFR, line 70: TODO Later, implement pagination. This isn't necessary though till we start hitting decent limits of credentials issued

lib/credentials/config/credentials.gql
```diff
@@ -0,0 +1,26 @@
+# GraphQL Type Definitions
+
+# Enum for CredentialType
+enum CredentialType {
+  proposal
+}
+
+# scalar JSON
+type SignedCredentialThree
```
motechFR, line 9: Each iteration of the model, Im changing the name as per received suggestion. Once we reach final version, will create a better name

PR #3165, opened 2023-11-24
Feat/rewards additional views

msadura:
### WHY

https://app.charmverse.io/charmverse/enable-calendar-and-board-view-15912610643980463

Screenshot 2023-11-24 at 14 29 13
Screenshot 2023-11-24 at 14 29 06
Screenshot 2023-11-24 at 14 29 01


pages/api/rewards/index.ts
```diff
@@ -139,6 +140,12 @@ async function createRewardController(req: NextApiRequest, res: NextApiResponse<
   logWorkspaceFirstBountyEvents(createdReward);
   logUserFirstBountyEvents(createdReward);
 
+  // Upsert reward board blocks when 1st bounty is created
+  const numberOfBounties = await prisma.bounty.count({ where: { spaceId } });
```
Devorein, line 144: Do we need to do this if we run a separate migration?

Devorein (approved):
Found only one issue. No views except for the table are available on public rewards. Otherwise everything looks and works great.
image

pages/api/rewards/index.ts
msadura: run migration for existing spaces with rewards. In the future it will create blocks only if there is only at least 1 reward

msadura: actually, I am not sure if it is worth it, maybe we should just create those blocks on space creation... lets discuss on dev meeting :)

PR #3166, opened 2023-11-24
Fixed inline database calendar view drag and drop

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/inline-db-calendar-view-can-t-move-events-to-new-dates-via-drag-and-drop-045683519793294325)


mattcasey:
This code is necessary for the sidebar to respond to events in teh editor when suggestions are selected. Note that when u click a suggestion, the suggestion you highlighted in the sidebar shifts a bit to the left. (Comments do not do this but they should)


The bug seems to be related to events inside the calendar. I don't see why we need to mess with the CharmEditor? Did you check to see if we can prevent events bubbling to the editor? Check the global ReactDNDProvider which has special logic to ignore elements

mattcasey (changes requested):
see comment. i made a video of the effect that happens when u select a suggestion. it doesn't work anymore with this update: https://www.loom.com/share/fc7ec4a7c5ac4b03a7adc3329d167061

PR #3168, opened 2023-11-25
Consistent user display avatar and username

Devorein:
### WHY

Before:
image

After:
image

### DIFF
--- a/components/common/UserDisplay.tsx
+++ b/components/common/UserDisplay.tsx
@@ -80,21 +80,16 @@ export const AnonUserDisplay = memo(AnonUserDisplayComponent);
  * @linkToProfile Whether we show a link to user's public profile. Defaults to false.
  */
 interface UserDisplayProps extends StyleProps {
-  user?: {
-    avatar?: string | null;
-    username: string;
-    path?: string | null;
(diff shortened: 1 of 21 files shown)


components/common/UserDisplay.tsx
```diff
@@ -90,6 +91,8 @@ interface UserDisplayProps extends StyleProps {
 
 function UserDisplay({ showMiniProfile = false, user, ...props }: UserDisplayProps) {
   const { showUserId } = useMemberDialog();
+  const { membersRecord } = useMembers();
```
Devorein, line 94: There are times when we pass the member directly in `UserDisplay`. We can pass the member directly, but that would be error prone and we will need to update quite a few places.

mattcasey (changes requested):
i think the main issue here is we pass in the logged in user in a lot of places instead of the member record. I would like a proper fix, though, because this makes the component "lie" from the standpoint of the call sites which means we're introducing tech debt. We might actually want to use the avatar/username that's on the passed-in user, I am not sure. But if we don't, then this is easily fixed by changing the prop to userId, and then Typescript will tell u what needs to be changed.

PR #3169, opened 2023-11-25
Use checkbox for boolean filter

Devorein:
### WHY

Notion
image

Now
image


mattcasey (approved):
this is better, but the checkbox/toggle IMO is pretty weird to begin with. Notion just uses a dropdown of "Checked/Unchecked"
image

It looks in your screenshot they display the checkbox to save space, if you're using the "Advanced filter"  but its still a dropdown to change it

image

Devorein:
Yes, I did notice that but was not sure whether we would go with a dropdown menu since there are only two options. It would require two clicks if we follow what Notion did. I will ask Chris about this and get his opinion.

PR #3170, opened 2023-11-25
Handle exporting empty database pages without throwing error

Devorein:
### WHY

1. Fix the exported zip name. Previously it was `.zip.zip`
2. Pass an empty prosemirror document to `BangleEditorState` if there is no content, rather than passing an empty string, as otherwise it would try use dom methods (which are not present in node env) and throw error.

### DIFF
--- a/components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
+++ b/components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
@@ -55,7 +55,8 @@ const createDocument = ({
     return schema.nodeFromJSON(content);
   }
 
-  if (typeof content === 'string') {
+  // Make sure document is present, otherwise it will throw an error
+  if (typeof content === 'string' && typeof window !== 'undefined') {
     const element = document.createElement('div');
     element.innerHTML = content.trim();
     return DOMParser.fromSchema(schema).parse(element, parseOptions);
--- a/components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
+++ b/components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
@@ -55,8 +55,7 @@ const createDocument = ({
     return schema.nodeFromJSON(content);
   }
 
-  // Make sure document is present, otherwise it will throw an error
-  if (typeof content === 'string' && typeof window !== 'undefined') {
+  if (typeof content === 'string') {
     const element = document.createElement('div');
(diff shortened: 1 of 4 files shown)


components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
```diff
@@ -55,7 +55,8 @@ const createDocument = ({
     return schema.nodeFromJSON(content);
   }
 
-  if (typeof content === 'string') {
+  // Make sure document is present, otherwise it will throw an error
+  if (typeof content === 'string' && typeof window !== 'undefined') {
```
mattcasey, line 59: in what situation are we using the bangle editor state outside of a browser?

lib/prosemirror/plugins/markdown/generateMarkdown.ts
```diff
@@ -23,7 +24,7 @@ export async function generateMarkdown({
 
   const state = new BangleEditorState({
     specRegistry,
-    initialValue: content ? Node.fromJSON(specRegistry.schema, content) : '',
+    initialValue: content ? Node.fromJSON(specRegistry.schema, content) : emptyDocument,
```
mattcasey, line 27: bangle editor state seems to already include emptyDocument. i want to avoid needing this in as many places as possible, if it's needed at all...

mattcasey (changes requested):
left some questions

components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
motechFR: We are using the editor state here in 

lib/prosemirror/plugins/markdown/generateMarkdown.ts

So we can generateMarkdown on the server side. Agree it would be nice not to import from a /components folder to the server

lib/prosemirror/plugins/markdown/generateMarkdown.ts
motechFR: Reverted this to undefined, doesn't seem to generate error anymore

components/common/CharmEditor/components/@bangle.dev/core/bangle-editor-state.ts
Devorein: When we export a full database, we make an api call to our backend to provide the csv + .md files in a zip format.

mattcasey (approved):

PR #3171, opened 2023-11-25
Fixed snackbar close ux

Devorein:
### WHY

Previously when the snackbar was about to close, we showed the initial state (info and empty message), which created a bad UX. Check out this loom recording, first portion is with the update, 2nd portion is what we have currently.

https://www.loom.com/share/d0ef1bb91ae641ee90c832bde7ef6870?sid=13b3008c-fdce-4761-bd9e-8997d3bfff41

### DIFF
--- a/hooks/useSnackbar.tsx
+++ b/hooks/useSnackbar.tsx
@@ -64,12 +64,9 @@ export function SnackbarProvider({ children }: { children: ReactNode }) {
       setAutoHideDuration,
       isOpen,
       handleClose,
-      showMessage: (msg: ReactNode, newSeverity?: AlertColor, anchorOrigin?: SnackbarOrigin) => {
+      showMessage: (msg: ReactNode, newSeverity?: AlertColor) => {
         newSeverity = newSeverity ?? 'info';
         setIsOpen(true);
-        if (anchorOrigin) {
-          setOrigin(anchorOrigin);
-        }
         setSeverity(newSeverity);
         setMessage(msg);
       },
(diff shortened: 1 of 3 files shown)


hooks/useSnackbar.tsx
```diff
@@ -64,12 +64,9 @@ export function SnackbarProvider({ children }: { children: ReactNode }) {
       setAutoHideDuration,
       isOpen,
       handleClose,
-      showMessage: (msg: ReactNode, newSeverity?: AlertColor, anchorOrigin?: SnackbarOrigin) => {
```
Devorein, line 67: `anchorOrigin` was not being used anywhere, hence why it was removed.

mattcasey (approved):

mattcasey:
looks good.. i wonder if a more permanent fix would be to make sure that the open and severity props are in sync

hooks/useSnackbar.tsx
```diff
@@ -37,18 +35,14 @@ export const SnackbarContext = createContext<Readonly<IContext>>({
...
   const [message, setMessage] = useState<null | ReactNode>(null);
   const [autoHideDuration, setAutoHideDuration] = useState<null | number>(5000);
 
   const resetState = () => {
     setIsOpen(false);
     setActions([]);
-    setOrigin({ vertical: 'bottom', horizontal: 'left' });
-    setMessage(null);
     setAutoHideDuration(5000);
-    setSeverity('info');
```
Devorein, line 51: This was the fix. The message and severity automatically get set when `showMessage` is called. no need to reset them when we are closing the snackbar.

PR #3173, opened 2023-11-26
Fixed inline database calendar view drag and drop

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/inline-db-calendar-view-can-t-move-events-to-new-dates-via-drag-and-drop-045683519793294325)

### DIFF
--- a/components/common/CharmEditor/components/suggestions/suggestions.plugins.tsx
+++ b/components/common/CharmEditor/components/suggestions/suggestions.plugins.tsx
@@ -28,7 +28,7 @@ export function plugins({ onSelectionSet, key }: { onSelectionSet?: (state: Edit
         },
         apply(tr, prev, oldState, state) {
           // react to when something is clicked, or when a selection is set by the sidebar component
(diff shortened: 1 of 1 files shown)


components/common/CharmEditor/components/suggestions/suggestions.plugins.tsx
```diff
@@ -28,7 +28,7 @@ export function plugins({ onSelectionSet, key }: { onSelectionSet?: (state: Edit
         },
         apply(tr, prev, oldState, state) {
           // react to when something is clicked, or when a selection is set by the sidebar component
-          if (tr.selectionSet || oldState.selection.eq(state.selection)) {
+          if (tr.selectionSet || !oldState.selection.eq(state.selection)) {
```
mattcasey, line 31: if the opposite works, do we even need this check?

mattcasey (changes requested):
clicking the cards in the sidebar no longer "highlights" them

PR #3175, opened 2023-11-26
Rename more bounties to rewards in copy and link

Devorein:
### WHY

1. A lot of copies still referred to "bounties"
2. A lot of links still had "bounties"


lib/webhookPublisher/entities.ts
Devorein, line 1: Using select to make the queries more performant, since its invoked a lot in other places and within the module itself.

msadura:
why tempalte files got removed?

motechFR:
@Devorein I have pulled these changes into #3180

PR #3181, opened 2023-11-27
Refresh check to ensure banner disappears

motechFR:
### WHAT

The banner was not disappearing

### WHY

This ensures the permissions banner disappears properly when invoked


mattcasey (approved):

scripts/migrateSpaceResources.ts
```diff
@@ -0,0 +1,101 @@
+import { stringUtils } from '@charmverse/core/dist/cjs/utilities';
```
mattcasey, line 1: imports from cjs

PR #3184, opened 2023-11-27
bump dev tools with security issue

mattcasey:
### WHY
Fixes a subdependency https://github.com/charmverse/app.charmverse.io/security/dependabot/80


valentinludu (commented):
Looks like in V6 you need to
`import '@testing-library/jest-dom';`
instead of
`import '@testing-library/jest-dom/extend-expect';`

mattcasey:
> Looks like in V6 you need to `import '@testing-library/jest-dom';` instead of `import '@testing-library/jest-dom/extend-expect';`

thanks for the tip!

PR #3185, opened 2023-11-27
rewrite s3 domain of user images

mattcasey:


motechFR:
Potential other locations

components/settings/space/components/LargeAvatar.tsx
components/common/CharmEditor/components/file/File.tsx

I used the useS3UploadInput to get a sense of where we are using S3

motechFR (approved):

PR #3187, opened 2023-11-27
Fix page content check

motechFR:
### WHAT

Ensure the document content check only runs if content exists

### WHY

Generated page content was causing error to throw


lib/security/checkPageContent.ts
```diff
@@ -11,7 +11,7 @@ export function checkPageContent(
   pageContent: any | null,
   meta: { spaceId?: string; blockId?: string; postId?: string; pageId?: string } = {}
 ): void {
-  if (pageContent) {
+  if (pageContent && pageContent.content) {
```
mattcasey, line 14: FYI the test data is invalid: `content: {}` is invalid. thats why we usually dont have to check for .content.

PR #3189, opened 2023-11-28
fix events not being able to be dragged on inline calendars

mattcasey:
### WHY
I am not sure why this works! I was just looking at what React components are rendering during drag and drop on an inline db and noticed CenterPanel was rendering due to 'showCard' prop, so I added a useCallback around it. Now I can drag the event fine. I can't really explain it, but we should be memoizing it anyway

### DIFF
--- a/hooks/useCharmRouter.ts
+++ b/hooks/useCharmRouter.ts
@@ -1,41 +1,45 @@
 import type { ParsedUrlQueryInput } from 'node:querystring';
 
 import { useRouter } from 'next/router';
+import { useMemo } from 'react';
 
 // utils for interacting with useRouter
 
 export function useCharmRouter() {
   const router = useRouter();
 
-  return {
-    router,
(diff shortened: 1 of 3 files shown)


hooks/useCharmRouter.ts
```diff
@@ -1,41 +1,45 @@
...
-        shallow: true
-      });
-    },
-    // update the URL, trigger a digest but do not re-run SSR
-    updateURLQuery(query: ParsedUrlQueryInput) {
-      // filter empty query params
-      const updatedQuery = Object.fromEntries(
-        Object.entries({ ...router.query, ...query }).filter(([_, v]) => v != null)
-      );
+  return useMemo(
```
mattcasey, line 11: pretty sure this is safe, just wanted to memoize all the functions that we expose

Devorein (approved):

Devorein (approved):

PR #3193, opened 2023-11-28
Required member properties during onboarding

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/onboarding-flow-for-required-member-properties-9749073104970365?domain=charmverse&pageId=page-9749073104970365)

### DIFF
--- a/components/_app/components/UserOnboardingDialog.tsx
+++ b/components/_app/components/UserOnboardingDialog.tsx
@@ -102,7 +102,7 @@ function UserOnboardingDialog({
   const [userDetails, setUserDetails] = useState<EditableFields>({});
   const [memberDetails, setMemberDetails] = useState<UpdateMemberPropertyValuePayload[]>([]);
   const { mutateMembers } = useMembers();
-  const [isFormClean, setIsFormClean] = useState(false);
+  const [isFormClean, setIsFormClean] = useState(true);
   const isInputValid =
     requiredProperties.length === 0 ||
     (isValid && (!isTimezoneRequired || !!userDetails.timezone) && (!isBioRequired || !!userDetails.description));


components/_app/components/OnboardingEmailForm.tsx
```diff
@@ -96,14 +100,22 @@ export function OnboardingEmailForm({ onClick, spaceId }: { onClick: VoidFunctio
...
+          />
         </FormGroup>
         <Stack flexDirection='row' gap={1} justifyContent='flex-end'>
-          <Button
-            loading={isMutating}
-            data-test='member-email-next'
-            type='submit'
-            disabled={Object.keys(errors).length !== 0}
-          >
+          <Button loading={isMutating} data-test='member-email-next' type='submit' disabled={!isValid}>
```
mattcasey, line 118: nice cleanup, dunno why i didn't use isValid before

components/_app/components/UserOnboardingDialog.tsx
```diff
@@ -31,22 +35,42 @@ type Step = 'email_step' | 'profile_step';
 export function UserOnboardingDialogGlobal() {
   const { space } = useCurrentSpace();
   const { user } = useUser();
-  const { showOnboardingFlow, completeOnboarding } = useOnboarding({ user, spaceId: space?.id });
 
   // Wait for user to load before deciding what to show
-  if (!user) {
+  if (!user || !space) {
```
mattcasey, line 40: not sure this check was really needed. useOnboarding works with an undefined spaceId too, we only need it when you close the popup. but it probably doesn't matter too much either

components/_app/components/UserOnboardingDialog.tsx
```diff
@@ -31,22 +35,42 @@ type Step = 'email_step' | 'profile_step';
...
-  // Show member profile for onboarding
-  if (showOnboardingFlow) {
+
+  return <LoggedInUserOnboardingDialog user={user} spaceId={space.id} />;
+}
+
+function LoggedInUserOnboardingDialog({ user, spaceId }: { spaceId: string; user: LoggedInUser }) {
+  const { showOnboardingFlow, completeOnboarding } = useOnboarding({ user, spaceId });
+
+  useEffect(() => {
```
mattcasey, line 50: thanks for wrapping this in useEffect

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+import useSWRImmutable from 'swr/immutable';
+import * as yup from 'yup';
+
+import charmClient from 'charmClient';
+import { useCurrentSpace } from 'hooks/useCurrentSpace';
+import type { MemberPropertyValueType } from 'lib/members/interfaces';
+
+import { useMemberPropertyValues } from './useMemberPropertyValues';
+
+export function useRequiredMemberProperties({ userId }: { userId: string }) {
```
mattcasey, line 13: you have us pass in "userId" but then you retrieve spaceId from useCurrentSpace(). Please pick one or the other approach for getting the details... if it were me i would just call useUser() here, and not have to pass userId in at all

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+        .flat(),
+    [memberPropertyValues, currentSpace?.id]
+  );
+
+  const requiredProperties = useMemo(
+    () => memberProperties.filter((p) => p.required && !['role', 'join_date'].includes(p.type)) ?? [],
+    [memberProperties]
+  );
+
+  const requiredPropertiesWithoutValue = useMemo(() => {
```
mattcasey, line 32: this value could just be a boolean - everywhere we use it we just check for .length

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+import type { MemberPropertyValueType } from 'lib/members/interfaces';
+
+import { useMemberPropertyValues } from './useMemberPropertyValues';
+
+export function useRequiredMemberProperties({ userId }: { userId: string }) {
+  const { memberPropertyValues = [], isLoading } = useMemberPropertyValues(userId);
+  const { space: currentSpace } = useCurrentSpace();
+  const { data: userDetails } = useSWRImmutable(`/current-user-details`, () => charmClient.getUserDetails());
+
+  const memberProperties = useMemo(
```
mattcasey, line 18: I don't think these useMemo's do anything and you should probably remove them. They can really complicate state - for example, in useThreads, the useMemo was causing values to be out of sync with ones that came from elsewhere, where because the value from useMemo is only calculated after the initial render.

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+    () =>
+      memberPropertyValues
+        .filter((mpv) => mpv.spaceId === currentSpace?.id)
+        .map((mpv) => mpv.properties)
+        .flat(),
+    [memberPropertyValues, currentSpace?.id]
```
mattcasey, line 24: On line 14, memberPropertyValues is always being recreated as a new empty array if there's no value. You should change that so that it does not constantly retrigger useMemo here

components/common/BoardEditor/focalboard/src/components/dialog.tsx
```diff
@@ -36,15 +36,10 @@ const FooterActionsContainer = styled(Box)`
   max-width: 100%;
   margin: 0 auto;
   position: relative;
-  padding: 0 40px 0 30px;
 
   ${({ theme }) => theme.breakpoints.up('lg')} {
     width: 860px;
   }
-
-  ${({ theme }) => theme.breakpoints.up('md')} {
```
mattcasey, line 45: Please leave this here, other dialogs depend on the padding matching the normal page width. It looks like member onboarding uses a special "fluidSize" which is why the button no longer lines up. So I think Dialog could adjust the padding here based on  the fluidSize prop.

mattcasey (changes requested):
- Please do not merge this with changes to name, until Alex/Chris I meet on Friday. If you can just undo those changes, and leave name as a regular property for now it would be better
- If we do go with a universal name it will probably be user.username. Why did you want to add it to UserDetails?

components/_app/components/UserOnboardingDialog.tsx
```diff
@@ -145,7 +192,22 @@ function UserOnboardingDialog({
       fluidSize={currentStep === 'email_step'}
       title={title}
       onClose={currentStep !== 'email_step' ? handleClose : undefined}
-      hideCloseButton={currentStep === 'email_step'}
+      hideCloseButton={currentStep === 'email_step' || requiredProperties.length !== 0}
+      footerActions={
+        currentStep === 'profile_step' ? (
+          <Box mr={4.5}>
```
mattcasey, line 198: i mentioned elsewhere but the dialog component should just adjust its footer spacing when fluidSize is true

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+          !['bio', 'timezone'].includes(rp.type)
+      )
+      .map((p) => p.memberPropertyId);
+
+    const isTimezoneRequired = requiredProperties.find((p) => p.type === 'timezone')?.required;
+    const isBioRequired = requiredProperties.find((p) => p.type === 'bio')?.required;
+    const isNameRequired = requiredProperties.find((p) => p.type === 'name')?.required;
+
+    if (isTimezoneRequired && !userDetails?.timezone && !isLoading) {
+      propertiesWithoutValue = [...propertiesWithoutValue, 'timezone'];
```
mattcasey, line 46: why not just propertiesWithoutValue.push()? it's easier to read

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+          !memberProperties.find((mp) => mp.memberPropertyId === rp.memberPropertyId)?.value &&
+          !['bio', 'timezone'].includes(rp.type)
+      )
+      .map((p) => p.memberPropertyId);
+
+    const isTimezoneRequired = requiredProperties.find((p) => p.type === 'timezone')?.required;
+    const isBioRequired = requiredProperties.find((p) => p.type === 'bio')?.required;
+    const isNameRequired = requiredProperties.find((p) => p.type === 'name')?.required;
+
+    if (isTimezoneRequired && !userDetails?.timezone && !isLoading) {
```
mattcasey, line 45: seems like u could jsut check isLoading first and return early

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export function useRequiredMemberPropertiesForm({ userId }: { userId: string }) {
+  const { memberProperties, userDetails, requiredProperties, requiredPropertiesWithoutValue } =
+    useRequiredMemberProperties({ userId });
+
+  const editableRequiredProperties = requiredProperties.filter(
+    (p) =>
+      ![
+        // Handled by oauth
```
mattcasey, line 75: these aren't actually handled by oauth yet?

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+  const requiredPropertiesWithoutValue = useMemo(() => {
+    let propertiesWithoutValue = requiredProperties
+      .filter(
+        (rp) =>
+          !memberProperties.find((mp) => mp.memberPropertyId === rp.memberPropertyId)?.value &&
+          !['bio', 'timezone'].includes(rp.type)
+      )
+      .map((p) => p.memberPropertyId);
+
+    const isTimezoneRequired = requiredProperties.find((p) => p.type === 'timezone')?.required;
```
mattcasey, line 41: should use .some() instead of .find()

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+  const memberProperties = useMemo(
+    () =>
+      memberPropertyValues
+        .filter((mpv) => mpv.spaceId === currentSpace?.id)
+        .map((mpv) => mpv.properties)
+        .flat(),
+    [memberPropertyValues, currentSpace?.id]
+  );
+
+  const requiredProperties = useMemo(
```
mattcasey, line 27: requiredProperties might work better as a map, most of the time you're looking for certain types

components/members/hooks/useRequiredMemberProperties.ts
mattcasey: another simplification could be to just wrap the whole response object with useMemo(). I did this in use useCharmRouter

components/settings/profile/components/UserDetailsForm.tsx
```diff
@@ -133,14 +165,37 @@ export function UserDetailsFormWithSave({
...
+  const isNameRequired = requiredProperties.find((p) => p.type === 'name');
+
+  const [userDetails, setForm] = useState<EditableFields>({
+    description: '',
+    timezone: '',
+    name: ''
+  });
+
+  useEffect(() => {
+    setForm({
```
mattcasey, line 185: Since this watches and resets the value whenever defaultUserDetails changes, is the 'userDetails' state even necessary? can you just use `defaultUserDetails` directly

components/_app/components/UserOnboardingDialog.tsx
Devorein: Yeah and we had a condition downward to return null if no space is present.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: The reason being `useMemberPropertyValues` requires an `userId` to be passed, using `useUser` will cause it to also take in `undefined` as a value.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: Ow I didn't realize that. Thnx for pointing it out. Updated.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: If we use the memberPropertyId as the key, the we would need to find the id of the member property that we want to check the required status first, which would require another loop, if we use the type there could be duplicates, having a key as the type and value as an array of properties would also be redundant if we want to check a specific property with specific id. We are checking the required status of mostly the user level properties i.e bio and timezone, so I can just return them from this hook.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: Agree. Updated.

components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -0,0 +1,133 @@
...
+    () => memberProperties.filter((p) => p.required && !['role', 'join_date'].includes(p.type)) ?? [],
+    [memberProperties]
+  );
+
+  const requiredPropertiesWithoutValue = useMemo(() => {
+    let propertiesWithoutValue = requiredProperties
+      .filter(
+        (rp) =>
+          !memberProperties.find((mp) => mp.memberPropertyId === rp.memberPropertyId)?.value &&
+          !['bio', 'timezone'].includes(rp.type)
```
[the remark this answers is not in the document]
Devorein: timezone and bio are user-level properties and they are handled in separate components and states where they are used. Consolidating the states is also quite difficult as they need to be sent to two different endpoints. While role and joined_at are not really editable by the member. I should've added a bit of comment on top of those to make it clearer.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: That was causing issue for some reason. I spent almost half an hour trying to figure out why thats the case. ended up using this less formal method.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: Not yet but they will be in the near future, so I'm leaving these out from being required for now.

components/members/hooks/useRequiredMemberProperties.ts
Devorein: Updated to make use of a single `useEffect`, in terms of the other deps, removed those as well and its working fine.

components/settings/profile/components/UserDetailsForm.tsx
Devorein: `isInputValid` depends on the `userDetails` state, and that is kept upto date using the `setForm` fn. Otherwise `defaultUserDetails` would be a static value always.

components/common/BoardEditor/focalboard/src/components/dialog.tsx
Devorein: Ahh gotcha. I will fix it for `UserOnboardingDialog` then

PR #3195, opened 2023-11-28
Security/delete static assests

motechFR:
### WHAT

Enable deletion of all s3 objects linked to a user

### WHY

Cleanup spam content from malicious users

### DIFF
--- a/lib/aws/deleteUserS3Assets.ts
+++ b/lib/aws/deleteUserS3Assets.ts
@@ -7,13 +7,13 @@ import { awsS3Bucket } from 'config/constants';
 
 import { getS3ClientConfig } from './getS3ClientConfig';
 import { listS3BucketContents } from './listS3Assets';
-import { getUserUploadsPrefix } from './uploadToS3Server';
+import { getUserS3FilePrefix } from './uploadToS3Server';
 
 export async function deleteUserS3Assets({ userId }: { userId: string }): Promise<void> {
   if (!stringUtils.isUUID(userId)) {
     throw new InvalidInputError('User id must be valid');
   }
-  const folderPath = getUserUploadsPrefix({ userId });
+  const folderPath = getUserS3FilePrefix({ userId });
 
   // List all objects in the subfolder
   const listedObjects = await listS3BucketContents({ prefix: folderPath });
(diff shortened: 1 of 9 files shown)


lib/aws/deleteUserS3Assets.ts
```diff
@@ -0,0 +1,38 @@
...
+    Bucket: awsS3Bucket,
+    Delete: { Objects: listedObjects.Contents.map(({ Key }) => ({ Key })) }
+  };
+
+  // Delete the objects
+  const s3Client = new S3Client(getS3ClientConfig());
+  await s3Client.send(new DeleteObjectsCommand(deleteParams));
+
+  // Recursively delete if there are more objects to be deleted
+  if (listedObjects.IsTruncated) {
```
mattcasey, line 35: i think a log here would be helpful in case this takes a long time, to see that it is re-curring

lib/aws/uploadToS3Server.ts
```diff
@@ -58,6 +57,10 @@ export function getFilePath({ spaceId, url }: { spaceId: string; url: string })
   return `spaces/${spaceId}/${uuid()}/${generateFilename(url)}`;
 }
 
+export function getUserUploadsPrefix({ userId }: { userId: string }) {
+  return `user-content/${userId}`;
+}
+
 export function getUserS3FilePath({ userId, url }: { userId: string; url: string }) {
-  return `user-content/${userId}/${uuid()}/${generateFilename(url)}`;
+  return `${getUserUploadsPrefix({ userId })}/${uuid()}/${generateFilename(url)}`;
```
mattcasey, line 65: 'getUserS3FilePrefix' would be more consistent

scripts/blacklistSpace.ts
```diff
@@ -77,6 +78,7 @@ async function deleteSpace(domain: string) {
   const adminUserIds = result.spaceRoles.map((sr) => sr.userId);
   for (const adminUserId of adminUserIds) {
     await blacklistUser(adminUserId);
+    await deleteUserS3Assets({userId: adminUserId})
```
mattcasey, line 81: i think this should be called inside of blacklistUser

mattcasey (approved):
left some suggestions

motechFR:
Thanks for feedback @mattcasey, addressed the comments

PR #3198, opened 2023-11-29
Token gate ens

valentinludu:
Accept ens name in the wallet option.

FYI:
- The old LIT approach doesn't work anymore with the current lit-protocol version.


mattcasey (approved):
remember to merge this to main, or to the feat branch before it gets merged ! :)

PR #3200, opened 2023-11-29
Admin settings for custom proposal flows

mattcasey:


motechFR:
Would love to test this out properly, how do I access the settings? Couldn't see them in staging

components/settings/proposals/components/EvaluationPermissions.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <PropertyLabel readOnly>{capitalize(level)}</PropertyLabel>
+          <UserAndRoleSelect
+            readOnly={readOnly}
+            wrapColumn
+            variant='outlined'
+            value={valuesByLevel[level] || []}
+            systemRoles={extraEvaluationRoles}
+            onChange={async (options) => updatePermissionLevel(level, options)}
+          />
+        </Box>
```
motechFR, line 67: Suggestion, some hard-coded smart defaults.

For example, during an evaluation stage, the reviewers will need to evaluate, but for this they need to view the proposal

components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
```diff
@@ -124,6 +126,24 @@ function SelectedOptions({
                   }
                 />
               )}
+              {option.group === 'system_role' && (
```
motechFR, line 129: Suggest we add some form of divider between system and custom roles. It wasn't immediately obvious the difference between proposal and custom roles

Additional suggestion; Changing labels to "Proposal Author" "Proposal Reviewer" "Space Member"

Screenshot 2023-11-30 at 15 09 15

__integration-tests__/server/pages/api/proposals/index.public.spec.ts
```diff
@@ -60,6 +60,7 @@ describe('POST /api/proposals - Create a proposal', () => {
         {
```
motechFR, line 60: Future-proofing alternative suggestion for lower cost of maintenance;

If we need to add a new null value to the object, we could use

expect.arrayContaining([
  // Only look for minimum amount of keys
  expect.objectContaining({obj})
])

charmClient/hooks/helpers.ts
```diff
@@ -4,6 +4,8 @@ import useSWRMutation from 'swr/mutation';
 
 import * as http from 'adapters/http';
 
+export type MaybeString = string | null | undefined;
```
motechFR, line 7: This is cool, spent too much time typing out value?: string | null in hooks

__integration-tests__/server/pages/api/proposals/index.public.spec.ts
mattcasey: Thanks... i really dont want to have to add null values, was just trying to make the tests+typescript happy.

components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
mattcasey: thre's icons with tooltips now

PR #3204, opened 2023-11-30
Make reward application rows distinct with empty value

Devorein:
### WHY

image

image

### DIFF
--- a/components/common/BoardEditor/components/kanban/KanbanGroupColumn.tsx
+++ b/components/common/BoardEditor/components/kanban/KanbanGroupColumn.tsx
@@ -24,6 +24,7 @@ type Props = {
   showCard: (cardId: string | null) => void;
   isManualSort: boolean;
   disableAddingCards?: boolean;
+  hideLinkedBounty?: boolean;
 };
 
 export function KanbanGroupColumn({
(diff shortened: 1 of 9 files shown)


components/common/BoardEditor/focalboard/src/components/table/tableRow.tsx
```diff
@@ -258,6 +260,7 @@ function TableRow(props: Props) {
               displayType='table'
               columnRef={columnRef}
               wrapColumn={activeView.fields.columnWrappedIds?.includes(template.id)}
+              subRowsEmptyValueContent={subRowsEmptyValueContent}
```
motechFR, line 263: I would add a comment for future maintainers about the reason for this field

motechFR (approved):

components/rewards/RewardsPage.tsx
```diff
@@ -241,6 +241,7 @@ export function RewardsPage({ title }: { title: string }) {
                     onCardClicked={() => {}}
                     onDeleteCard={onDelete}
                     expandSubRowsOnLoad
+                    subRowsEmptyValueContent='--'
```
motechFR, line 244: Suggestion: Would a single dash look better?

components/common/BoardEditor/components/kanban/KanbanGroupColumn.tsx
```diff
@@ -58,6 +60,7 @@ export function KanbanGroupColumn({
           onDrop={onDropToCard}
           showCard={showCard}
           isManualSort={isManualSort}
+          hideLinkedBounty={hideLinkedBounty}
```
motechFR, line 63: What does this field do?

components/rewards/RewardsPage.tsx
Devorein: Chris specifically mentioned to use `--` in the product card
> Use "--" for all columns that are currently empty on applications

PR #3206, opened 2023-11-30
Renamed Assigned to Applicant in Rewards table

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/rename-assigned-column-to-applicant-09813645492192125)


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -30,7 +30,7 @@ import {
   REWARD_CHAIN,
   REWARD_TOKEN,
   REWARD_CUSTOM_VALUE,
-  ASSIGNEES_BLOCK_ID,
+  APPLICANTS_BLOCK_ID,
```
motechFR, line 33: One suggestion; could you rename this to REWARD_APPLICANTS_BLOCK_ID ?

This will simplify maintenance when we're looking at custom block names

motechFR (approved):

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Agree, updated.

PR #3209, opened 2023-11-30
Rewards link

mattcasey:
### WHY

- "Open as page" to applications inside dialog
- removed a case we don't use for RewardStatusBadge, which had an incorrect link to rewards


motechFR (approved):
Tested & works, e2e test for this might be handy

PR #3215, opened 2023-12-01
Bug/fix page deletion

motechFR:
### WHAT

Fix loopholes allowing user to delete or restore a page

### WHY

Unexpected situation of user with view permission able to delete / restore pages

### DIFF
--- a/components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
+++ b/components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
@@ -60,7 +60,7 @@ export default function PageDeleteBanner({ pageType, pageId }: { pageType: PageT
         await navigateToSpacePath(`/${path}`);
       }
     } catch (err) {
-      // Do nothing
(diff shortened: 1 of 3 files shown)


components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx
```diff
@@ -19,29 +21,48 @@ export default function PageDeleteBanner({ pageType, pageId }: { pageType: PageT
...
-      await navigateToSpacePath(`/${path}`);
+    try {
+      setIsMutating(true);
       await charmClient.deletePage(pageId);
+      const path = Object.values(pages).find((page) => page?.type !== 'card' && !page?.deletedAt)?.path;
+      if (path) {
+        await navigateToSpacePath(`/${path}`);
+      }
+    } catch (err) {
+      // Do nothing
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/PageDeleteBanner.tsx:63]: can u include in the comment why "do nothing"? why dont we even log a warning?

PR #3216, opened 2023-12-01
Prevent reordering flows from a template

motechFR:
### WHAT

If proposal comes from a template, don't allow non admin to reorder criteria

### WHY

Author was able to reorder criteria, which makes it more difficult for reviewers of multiple proposals using the same template


mattcasey (approved):
thanks for catching this

PR #3219, opened 2023-12-01
Fix card notifications

motechFR:
### WHAT

Fix loading of notifications and cleanup of card notifications

### WHY

Notifications list was failing because the linked page was deleted

### DIFF
--- a/lib/notifications/cards/getCardNotifications.ts
+++ b/lib/notifications/cards/getCardNotifications.ts
@@ -59,33 +59,39 @@ export async function getCardNotifications({ id, userId }: QueryCondition): Prom
     });
   });
 
-  return cardNotifications.map((notification) => {
-    const notificationMetadata = notification.notificationMetadata;
-    const page = notification.card.page as Page;
-    const cardNotification = {
-      id: notification.id,
-      createdAt: notificationMetadata.createdAt.toISOString(),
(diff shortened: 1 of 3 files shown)


mattcasey (approved):

lib/pages/modifyChildPages.ts
```diff
@@ -54,6 +54,13 @@ export async function modifyChildPages(parentId: string, userId: string, action:
             }
           ]
         }
+      }),
+      prisma.cardNotification.deleteMany({
```
mattcasey, line 58: deleting them is fine but the notifications should work even if the connection is lost. notifications are supposed to be immutable

lib/notifications/cards/getCardNotifications.ts
```diff
@@ -59,33 +59,39 @@ export async function getCardNotifications({ id, userId }: QueryCondition): Prom
...
-            }
-          : null,
-      archived: !!notificationMetadata.archivedAt,
-      read: !!notificationMetadata.seenAt,
-      group: 'card'
-    } as CardNotification;
+  return (
+    cardNotifications
+      // Don't show notifications for a card whose page was deleted
+      .filter((notification) => !!notification.card.page)
```
mattcasey, line 65: nice... this is why i hate type assertions. So many bugs where a type assertion was added

PR #3220, opened 2023-12-01
remember if user closes the sidebar and only show sidebar if its time to review

mattcasey:


components/[pageId]/DocumentPage/components/ProposalProperties.tsx
```diff
@@ -165,7 +165,7 @@ export function ProposalProperties({
 
   // open the rubric sidebar by default
   useEffect(() => {
-    if (enableSidebar) {
+    if (enableSidebar && canSeeEvaluation) {
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/ProposalProperties.tsx:168]: i think this was a regression

PR #3235, opened 2023-12-05
Proposal setup using new evaluation workflows

mattcasey:
Users can now create, edit, and walk thru the new process of proposals. The review and evaluation process, including votes, is managed via sticky footers and sidebars, instead of embedded  in the document. I still need to verify notifications, proposal as a source, and that the current flow is not broken (this one before I merge).

Notable changes:
- A new page and component`NewProposalPage.tsx` at `/{spaceDomain}/proposals/new`

Clean up:
- removed useProposalPermissions hook, instead we just return the permissions now when u request a specific proposal (we should return permissions when u request anything, probably, since we already have to request them on the backend. It saves a whole round trip to permissions api )
- moved proposal sidebar and property components to the components/proposals page, so that it's altogether. Could probably do the same for RewardProperties


testing/setupDatabase.ts
```diff
@@ -803,7 +802,7 @@ export async function createProposalWithUsers({
       ).id
     : proposalCategoryId;
 
-  const proposalPage = await createPageDb<PageWithProposal>({
+  const proposalPage = await createPageDb({
```
motechFR, line 805: General testing notes, we have some proposal test utility generators in core. Useful to extend those if we'll be consuming this for testing proposal flows

stories/ProposalsPage/Proposals.stories.tsx
```diff
@@ -259,23 +272,25 @@ ProposalInEvaluation.parameters = {
...
+            evaluationId: criteria.evaluationId
           }))
         ];
         const proposal = createMockProposal({
           authors: [{ proposalId: '', userId: members[0].id }],
           reviewers: [
-            { id: '1', proposalId: '', roleId: null, userId: userProfile.id },
-            { id: '2', proposalId: '', roleId: null, userId: members[0].id },
-            { id: '3', proposalId: '', roleId: null, userId: members[1].id }
+            { evaluationId: null, id: '1', proposalId: '', roleId: null, userId: userProfile.id },
```
motechFR, line 291: I suggest we use the TargetPermissionGroup syntax here ({ group: 'role' | 'user', id: string})

scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -0,0 +1,140 @@
...
+        select: {
+          snapshotProposalId: true,
+          votes: true
+        }
+      }
+    }
+  });
+  const proposalsToUpdate = proposals.filter((p) => p.evaluations.length === 0);
+  await Promise.all(
+    proposalsToUpdate.map(async (p) => {
```
motechFR, line 23: Suggest that we use a transaction per proposal here, since we're doing quite a few things at once

testing/setupDatabase.ts
mattcasey: I saw that... this change was to reduce the usage of PageWithProposal. I just wanted to add evaluations to the response, but it was being used in lots of places that didn't really even care about the response shape. I did notice that we have "generateProposal" in both app and core - and core gives you a _page_ for some reason :P

stories/ProposalsPage/Proposals.stories.tsx
mattcasey: These permissions match the database schema

mattcasey: This is the database model

scripts/migrations/2023_11_30_convertProposals.ts
mattcasey: I believe that's what I'm doing?

lib/proposal/getProposal.ts
mattcasey, line 1: we were only using this method in tests

components/proposals/components/ProposalDialog/hooks/useNewProposal.tsx
```diff
@@ -87,37 +90,32 @@ export function useNewProposal({ newProposal }: Props) {
...
+      mutate(`/api/spaces/${currentSpace.id}/proposals`);
+      setContentUpdated(false);
     }
   }
 
-  const disabledTooltip = useMemo(() => {
-    if (!formInputs.title) {
-      return 'Title is required';
-    }
+  let disabledTooltip: string | undefined;
```
msadura, line 98: would be nice to have this as external function for clearer logic, I actually faced the same in rewards

pages/api/proposals/[id]/index.ts
```diff
@@ -69,23 +70,18 @@ async function getProposalController(req: NextApiRequest, res: NextApiResponse<P
...
   if (!canSeeAnswers) {
     proposal.draftRubricAnswers = [];
     proposal.rubricAnswers = [];
   }
 
-  // filter out evaluation-specific reviewers
+  // Support old model: filter out evaluation-specific reviewers
   proposal.reviewers = proposal.reviewers.filter((reviewer) => !reviewer.evaluationId);
 
+  (proposal as unknown as ProposalWithUsersAndRubric).permissions = permissions;
```
Comment (motechFR) [inline pages/api/proposals/[id]/index.ts:83]: Suggestion: Let's create a webapp-only type to reflect proposal with users, rubric and permissions

Comment (mattcasey) [reply pages/api/proposals/[id]/index.ts:83]: this is a webapp only type already

Comment (mattcasey) [reply pages/api/proposals/[id]/index.ts]: Ah FTR I realized that this inherits ProposalWithUsers - but we should b able to just copy over what we need. I think what we actually need are two types: PopulatedProposal and PopulatedProposalLite (when we get proposals for the db). made a card here: https://work.charmverse.fyi/work.charmverse.fyi/tech-task-list-49366552253645923?viewId=50d940bb-fdbd-40ba-9816-616d6138a663&cardId=71cf0f5a-a1a0-42a4-876e-ecc000693fa0

.vscode/settings.json
```diff
@@ -18,10 +18,15 @@
...
   },
   "editor.detectIndentation": false,
   "editor.tabSize": 2,
   "jestrunner.jestCommand": "npx dotenv -e .env.test.local -- node 'node_modules/.bin/jest'",
   "typescript.tsdk": "node_modules/typescript/lib",
+  "[javascript][typescript][typescriptreact]": {
+    "editor.codeActionsOnSave": {
+      "source.fixAll.eslint": "explicit"
+    }
+  },
```
mattcasey, line 31: my vscode updated this on its own?

motechFR (approved):

PR #3236, opened 2023-12-05
Show parent board title chip for reward cards

Devorein:
### WHY

Conditions:
1. Must be inside the modal view
2. Must be in the rewards page (Since its obvious what the parent board is when opening a reward attached card in a regular database)
3. Must be of card type but has a reward attached

image

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -84,14 +84,23 @@ export interface DocumentPageProps {
   savePage: (p: Partial<Page>) => void;
   readOnly?: boolean;
   close?: VoidFunction;
+  insideModal?: boolean;
   enableSidebar?: boolean;
 }
 
-function DocumentPage({ page, refreshPage, savePage, readOnly = false, close, enableSidebar }: DocumentPageProps) {
+function DocumentPage({
+  insideModal = false,
+  page,
+  refreshPage,
+  savePage,
+  readOnly = false,
+  close,
+  enableSidebar
+}: DocumentPageProps) {
   const { cancelVote, castVote, deleteVote, updateDeadline, votes, isLoading } = useVotes({ pageId: page.id });
 
   const isLargeScreen = useLgScreen();
-  const { navigateToSpacePath } = useCharmRouter();
+  const { navigateToSpacePath, router } = useCharmRouter();
   const {
     activeView: sidebarView,
     persistedActiveView,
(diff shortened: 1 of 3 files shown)


components/[pageId]/DocumentPage/components/PageHeader.tsx
```diff
@@ -95,13 +101,23 @@ function PageHeader({
...
-        icon={icon}
-        readOnly={readOnly}
-        setPage={setPage}
-      />
+      {showParentChip && parentInfo ? (
+        <Chip
+          label={parentInfo.title}
+          size='small'
+          onClick={() => {
+            navigateToSpacePath(`/${parentInfo.id}`);
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/components/PageHeader.tsx:109]: Using parentId since it will be costly to get the path via `usePages`

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -158,8 +167,8 @@ function DocumentPage({ page, refreshPage, savePage, readOnly = false, close, en
   const pageTop = getPageTop(page);
 
   const { threads, isLoading: isLoadingThreads, currentPageId: threadsPageId } = useThreads();
-  const router = useRouter();
   const isSharedPage = router.pathname.startsWith('/share');
+  const isRewardsPage = router.pathname === '/[domain]/rewards';
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:171]: Check if we are in the rewards route, as the parent board title chip is only visible there.

Comment (mattcasey) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:171]: this would make a good comment in code

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -354,6 +363,17 @@ function DocumentPage({ page, refreshPage, savePage, readOnly = false, close, en
...
+                  parentInfo={
+                    board
+                      ? {
+                          title: board.title,
+                          id: board.id
+                        }
+                      : null
+                  }
+                  showParentChip={
+                    !!(page.type === 'card' && page.bountyId && page.parentId && board && insideModal && isRewardsPage)
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:375]: suggestion: move this logic and define `showParentChip` right below `isRewardsPage`

mattcasey (changes requested):
The chip should use an href to link the user, instead of an onClick handler.

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:171]: Added

components/[pageId]/DocumentPage/components/PageHeader.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/components/PageHeader.tsx:109]: Added

PR #3238, opened 2023-12-06
Feat/custom rewards views

msadura:
### WHAT
Addint custom views to rewads page

https://app.charmverse.io/charmverse/views-should-be-customizable-and-appear-in-the-sidebar-like-a-normal-database-3816155843453677

Screenshot 2023-12-06 at 14 38 17


lib/blockchain/provider/alchemy/client.ts
```diff
@@ -94,7 +94,7 @@ export async function getNFTs({
         ...(filterSpam
           ? {
               spamConfidenceLevel: 'HIGH',
-              excludeFilters: ['SPAM']
+              'excludeFilters[]': ['SPAM']
```
Devorein, line 97: Is this one intentional? Looks a bit weird.

Devorein (approved):

lib/blockchain/provider/alchemy/client.ts
msadura: yes it is from alchemy docs https://docs.alchemy.com/reference/getnftsforowner-v3
Previously it was added directly in GET method for all requests which was wrong

PR #3240, opened 2023-12-06
Required social member properties

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/required-user-identities-38186121786501137)


components/members/hooks/useRequiredMemberProperties.ts
```diff
@@ -108,20 +165,126 @@ export function useRequiredMemberPropertiesForm({ userId }: { userId: string })
...
+    isDirty,
+    setValue,
+    getValues,
+    isSubmitting,
+    onSubmit,
+    onFormChange
+  };
+}
+
+export function useRequiredUserDetailsForm({ userId }: { userId: string }) {
```
mattcasey, line 213: this should be in its own file

mattcasey (approved):

components/members/hooks/useRequiredMemberProperties.ts
Devorein: Yeah sure. But since I branched #3247 off of this I will move it on that branch to avoid merge conflicts.

PR #3241, opened 2023-12-06
Extend public api user search to allow userId

motechFR:
### WHAT

Requester can now pass a user id in user search endpoint

### WHY

Until now, the endpoint only accepted email or wallet

### DIFF
--- a/pages/api/v1/users/search.ts
+++ b/pages/api/v1/users/search.ts
@@ -11,7 +11,6 @@ const handler = superApiHandler();
 handler.get(searchUser);
 
 /**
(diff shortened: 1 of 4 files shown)


pages/api/v1/users/search.ts
```diff
@@ -67,13 +71,14 @@ type SearchUserResponseBody = UserProfile;
 async function searchUser(req: NextApiRequest, res: NextApiResponse<SearchUserResponseBody>) {
```
motechFR, line 71: Extending this API as the underlying lib method already supports userId lookup

mattcasey (approved):

PR #3245, opened 2023-12-07
Chore/remove rewards in progress

msadura:
### WHAT

Removing occurences of inProgres status from FE


mattcasey (approved):
nice cleanup thank u

PR #3247, opened 2023-12-07
Required + Primary identity for connected accounts

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/required-user-identities-38186121786501137)


charmClient/apis/spacesApi.ts
```diff
@@ -1,5 +1,5 @@
 import type { SpaceDefaultPublicPageToggle } from '@charmverse/core/permissions';
-import type { Space, Prisma } from '@charmverse/core/prisma';
+import type { Space, Prisma, PrimaryMemberIdentity } from '@charmverse/core/prisma';
```
mattcasey, line 2: unused import

components/members/components/ConnectedAccounts.tsx
```diff
@@ -0,0 +1,241 @@
+import { Stack, Typography } from '@mui/material';
```
mattcasey, line 1: this is a child of UserOnboardingDialog and belongs in an components/_app/components/UserOnboardingDialog folder.

components/_app/components/UserOnboardingDialog.tsx
```diff
@@ -60,12 +64,20 @@ function LoggedInUserOnboardingDialog({ user, space }: { space: Space; user: Log
...
         currentUser={user}
         completeOnboarding={completeOnboarding}
+        hasEmptyRequiredProperties={hasEmptyRequiredProperties}
       />
     );
   }
 
-  if (nonEmptyRequiredProperties) {
-    return <UserOnboardingDialog space={space} key={user.id} currentUser={user} />;
+  if (hasEmptyRequiredProperties || query.onboarding === 'true') {
```
mattcasey, line 72: `query.onboarding`: I am not clear if it's necessary to use the URL, since connected accounts use a pop up? And since it's only children of UserOnboardingDialog setting `onboarding: true`, I wonder could this just be replaced with a simple useState() that gets passed down?

components/_app/components/UserOnboardingDialog.tsx
```diff
@@ -180,6 +204,7 @@ function UserOnboardingDialog({
         <OnboardingEmailForm onClick={goNextStep} spaceId={space.id} />
       ) : currentStep === 'profile_step' ? (
         <>
+          {hasEmptyRequiredProperties ? <Alert severity='info'>Please fill out all required fields</Alert> : null}
```
mattcasey, line 207: this was a great addition, thanks

components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
```diff
@@ -237,11 +237,7 @@ function ShowMoreMenuItem({ onClick, children }: { onClick: VoidFunction; childr
 }
 
 function filterByUserCustomName(member: Member, searchText: string) {
-  return member.properties
-    .find((prop) => prop.type === 'name')
-    ?.value?.toString()
-    .toLowerCase()
-    .match(searchText);
+  return member.username?.toString().toLowerCase().match(searchText);
```
mattcasey, line 240: this filter now functions just like filterByUsername() and should be removed

components/login/components/WalletSign.tsx
```diff
@@ -63,6 +65,40 @@ export function WalletSign({
...
+  onError
+}: Props) {
+  const {
+    connectWallet,
+    generateWalletAuth,
+    isConnectingIdentity,
+    isWalletSelectorModalOpen,
+    showLoadingState,
+    verifiableWalletDetected
+  } = useWalletSign({
```
mattcasey, line 95: honestly this was not much of a cleanup, it only makes it harder to know how this state is set up. i don't think we should always use t his pattern

components/login/components/WalletSign.tsx
```diff
@@ -22,21 +22,23 @@ interface Props {
   onError?: (err: any) => void;
 }
 
-export function WalletSign({
-  signSuccess,
-  buttonStyle,
-  buttonSize,
-  ButtonComponent = PrimaryButton,
-  buttonOutlined,
+export function useWalletSign({
```
mattcasey, line 25: Are you sure you needed to export this as a hook? WalletSign takes a generic `ButtonComponent`. It would be 10x better to follow the pattern we already have, maybe add more props if you need it, than add another one.

components/settings/profile/components/IdentityIcon.tsx
```diff
@@ -5,28 +5,60 @@ import PersonIcon from '@mui/icons-material/Person';
 import TelegramIcon from '@mui/icons-material/Telegram';
 import { SiDiscord } from 'react-icons/si';
 
-export function IdentityIcon({ type }: { type: IdentityType | null }) {
+export function IdentityIcon({
+  type,
+  height = 40,
```
mattcasey, line 10: it would be better to have 'size' than height/width as props. That way we can control the varieties available and maintain consistency

components/settings/space/SpaceSettings.tsx
```diff
@@ -129,6 +135,12 @@ export function SpaceSettings({
 
   useTrackPageView({ type: 'settings/space' });
 
+  useEffect(() => {
```
mattcasey, line 138: this seems unecessary since space is never undefined, and I don't think you can change the space id without closing the dialog

mattcasey: please add a comment explaining why it is necessary

components/settings/space/SpaceSettings.tsx
```diff
@@ -266,6 +282,53 @@ export function SpaceSettings({
...
+                </MenuItem>
+                <MenuItem value='google'>
+                  <Stack flexDirection='row' alignItems='center' gap={1}>
+                    <FaGoogle style={{ width: 18, height: 18 }} />
+                    <Typography variant='body2'>Google</Typography>
+                  </Stack>
+                </MenuItem>
+                <MenuItem value='telegram'>
+                  <Stack flexDirection='row' alignItems='center' gap={1}>
+                    <FaTelegram style={{ width: 18, height: 18 }} />
```
mattcasey, line 319: wouldn't we use IdentityIcon here instead of these black icons? they look a little intense..

lib/members/getSpaceMembers.ts
```diff
@@ -75,14 +86,43 @@ export async function getSpaceMembers({
     }
   });
 
+  const memberUsernameRecord: Record<string, string> = {};
+
+  for (const spaceRole of spaceRoles) {
+    const memberUsername = await getMemberUsername({ spaceRoleId: spaceRole.id });
```
mattcasey, line 92: The only data I think you need to query here is the space preference for primary identity, because we grabbed the other information for each user already. I would re-write "getMemberUsername" to just be a utility that takes a populated user object and a identity type.

Please be mindful of extra queries especially when you already have the date that you need (from the JOINs above). This was also written to run in series instead of parallel.

lib/notifications/mailer/sendNotificationEmail.ts
```diff
@@ -179,48 +180,18 @@ async function sendEmail({
   user: Pick<User, 'id' | 'username' | 'id' | 'avatar' | 'email'>;
 }) {
   const notificationSpaceId = notification.spaceId;
-  const userId = notification.createdBy.id;
 
-  const memberProperty = await prisma.memberProperty.findFirst({
+  const spaceRoleId = await prisma.spaceRole.findFirstOrThrow({
```
mattcasey, line 184: spaceRoleId is a weird thing to have to query for. And it's only because getMemberUserName internally uses it to figure out the space and user.

lib/notifications/mailer/sendNotificationEmail.ts
```diff
@@ -179,48 +180,18 @@ async function sendEmail({
...
-      select: {
-        value: true
-      }
-    });
-
-    if (notificationAuthorNamePropertyValue?.value) {
-      notification.createdBy.username = notificationAuthorNamePropertyValue.value as string;
-    }
-
-    const notificationTargetNamePropertyValue = await prisma.memberPropertyValue.findFirst({
```
mattcasey, line 210: don't we still need to get the username for the target?

mattcasey (changes requested):
Seems to work when i test locally with Discord! left some comments

components/members/components/ConnectedAccounts.tsx
Devorein: Good point. Updated.

components/common/CharmEditor/components/mention/components/MentionSuggest.tsx
Devorein: Oops. Overlooked that. Thnx.

mattcasey (approved):
thanks for the cleanup!

PR #3250, opened 2023-12-08
add category to forum breadcrumbs

mattcasey:
Makes it easier to find other posts in the topic from full page view:

image


components/forum/components/PostPage/PostPage.tsx
```diff
@@ -337,12 +342,24 @@ export function PostPage({
         >
           <Container top={50}>
             <Box minHeight={300} data-test='post-charmeditor'>
+              {isPublishedPost && isInsideDialog && (
```
mattcasey, line 345: I also hid the category input unless you have edit access, since users can now see it from the breadcrumb or this chip that appears in the dialog

PR #3252, opened 2023-12-09
Support for expand option when fetching proposal comment user

Devorein:
### WHY

[Discussion](https://discord.com/channels/894960387743698944/930253269459148900/1182714193242247208)

### DIFF
--- a/pages/api/v1/proposals/[proposalId]/comments/index.ts
+++ b/pages/api/v1/proposals/[proposalId]/comments/index.ts
@@ -49,9 +49,11 @@ handler.post(
  *               type: string
  *               example: "## This is a comment."
  *         createdBy:
- *           type: string
- *           description: User ID of the user who created the comment
- *           example: "69a54a56-50d6-4f7b-b350-2d9c312f81f3"
+ *           oneOf:
+ *             - type: string
+ *               description: User ID of the user who created the comment
+ *               example: "69a54a56-50d6-4f7b-b350-2d9c312f81f3"
+ *             - $ref: '#/components/schemas/SearchUserResponseBody'
  *         createdAt:
  *           type: string
  *           description: ISO Timestamp of comment creation date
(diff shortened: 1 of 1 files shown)


lib/public-api/searchUserProfile.ts
```diff
@@ -4,6 +4,38 @@ import { prisma } from '@charmverse/core/prisma-client';
...
+    select: {
+      address: true
+    }
+  },
+  googleAccounts: {
+    select: {
+      email: true
+    }
+  }
+} satisfies Prisma.UserSelect;
```
mattcasey, line 27: interesting, last time i tried this, Jest or something blew up. but if it works, that's great

pages/api/v1/proposals/[proposalId]/comments/index.ts
```diff
@@ -200,6 +220,12 @@ async function getProposalComments(req: NextApiRequest, res: NextApiResponse<Pub
     }
   });
 
+  const expand = Array.isArray(req.query.expand)
+    ? req.query.expand
+    : typeof req.query.expand === 'string'
+    ? [req.query.expand]
+    : [];
```
Comment (mattcasey) [inline pages/api/v1/proposals/[proposalId]/comments/index.ts:227]: We should validate what we get from expand and we return a 400 if there's an unrecognized value

pages/api/v1/proposals/[proposalId]/comments/index.ts
```diff
@@ -84,6 +93,7 @@ async function mapReducePageComments({
 }: {
   comments: (Pick<PageComment, 'id' | 'parentId' | 'content' | 'contentText' | 'createdAt' | 'createdBy'> & {
     votes: Pick<PageCommentVote, 'upvoted'>[];
+    user?: UserProfile;
```
Comment (mattcasey) [inline pages/api/v1/proposals/[proposalId]/comments/index.ts:96]: Let's copy Stripe here and just replace the value with the populated object. So in this case, 'comment.createdBy' should be the user object. 

I don't know how Swagger handles types of different kinds, but it's ok to not document this type in the API Docs for now, if there's any issue.

mattcasey (changes requested):
left comments

pages/api/v1/proposals/[proposalId]/comments/index.ts
Comment (Devorein) [reply pages/api/v1/proposals/[proposalId]/comments/index.ts:227]: Great point. Updated.

pages/api/v1/proposals/[proposalId]/comments/index.ts
Comment (Devorein) [reply pages/api/v1/proposals/[proposalId]/comments/index.ts]: Good point. I wanted to do it initially, but it was causing quite a few type issues. Its been resolved now.

PR #3254, opened 2023-12-11
Database table rows selection

Devorein:


components/common/BoardEditor/focalboard/src/components/table/calculation/calculationRow.scss
```diff
@@ -26,10 +24,10 @@
 .Table {
   .octo-table-cell.Calculation:focus-within {
     border: unset;
-    background-color: rgb(var(--center-channel-bg-rgb));
+    background-color: rgba(var(--center-channel-color-rgb), 0.05);
```
Devorein, line 27: This might now be the correct css variable but hovering hover the calculation cells in the footer should highlight them (notion)

components/common/BoardEditor/focalboard/src/components/table/tableRow.scss
```diff
@@ -19,13 +19,4 @@
   &.selected {
     background-color: rgba(90, 200, 255, 0.2);
   }
-
-  // Unlike cards and gallery, table rows allow inline editing, so we won't disable this style on iOS (see kanbanCard.scss and galleryCard.scss)
-  &:hover {
```
Devorein, line 24: Removed hover style on table rows (notion)

components/common/BoardEditor/focalboard/src/components/viewHeader/viewHeader.scss
```diff
@@ -13,6 +13,12 @@
     margin-left: 0.3em;
   }
 
+  @media (max-width: 768px) {
```
Devorein, line 16: This shows the select property section about the view options in mobile view. Not sure how to do this in a better manner.
image

charmClient/charmClient.ts
```diff
@@ -427,7 +427,7 @@ class CharmClient {
     });
   }
 
-  createEvent({ payload, spaceId }: { spaceId: string; payload: CreateEventPayload }) {
+  createEvents({ payload, spaceId }: { spaceId: string; payload: CreateEventPayload[] }) {
```
Devorein, line 430: Pluralized since it can create multiple events in one go now.

components/common/BoardEditor/focalboard/src/components/calculations/calculation.tsx
```diff
@@ -60,7 +60,7 @@ function Calculation(props: Props): JSX.Element {
 
       <span className='calculationLabel'>{optionDisplayNameString(valueOption!, intl)}</span>
 
-      {value === Options.none.value && <ExpandLessIcon fontSize='small' />}
+      {value === Options.none.value && <ExpandMoreIcon fontSize='small' />}
```
Devorein, line 63: Notion uses expand more.

components/common/BoardEditor/focalboard/src/components/table/tableHeader.tsx
```diff
@@ -339,7 +339,7 @@ function TableHeader(props: Props): JSX.Element {
         width: columnWidth(props.template.id),
         opacity: isDragging ? 0.5 : 1,
         transition: `background-color 150ms ease-in-out`,
-        backgroundColor: isOver ? 'var(--charmeditor-active)' : 'initial'
+        backgroundColor: isOver ? 'var(--charmeditor-active)' : ''
```
Devorein, line 342: This was stopping the hover effect to take place inside the table header cells (notion)

components/common/BoardEditor/focalboard/src/components/table/tableRow.tsx
```diff
@@ -171,18 +207,47 @@ function TableRow(props: Props) {
       className={className}
       onClick={(e) => props.onClick?.(e, card)}
       ref={cardRef}
-      style={{ opacity: isDragging ? 0.5 : 1, backgroundColor: isNested ? 'var(--input-bg)' : 'transparent' }}
+      style={{
+        opacity: isDragging ? 0.5 : 1,
+        backgroundColor: isNested ? 'var(--input-bg)' : 'transparent',
+        ...(isChecked && {
+          background: 'rgba(35, 131, 226, 0.14)',
```
Devorein, line 214: Taken from notion, maybe we should use our select color variable?

pages/api/blocks/index.ts
```diff
@@ -294,8 +294,8 @@ async function updateBlocks(req: NextApiRequest, res: NextApiResponse<Block[]>)
 }
 
 async function deleteBlocks(req: NextApiRequest, res: NextApiResponse<Block[]>) {
-  // TODO: remove use of req.body after browsers update - 06/2023
-  const blockIds: string[] = (req.query.blockIds || req.body) ?? [];
+  const queryBlockIds = (req.query.blockIds ?? []) as string[];
+  const blockIds = typeof queryBlockIds === 'string' ? [queryBlockIds] : queryBlockIds;
```
mattcasey, line 298: thx for fixing

pages/api/spaces/[id]/event.ts
```diff
@@ -14,19 +14,21 @@ handler.use(requireUser).post(createEvent);
 async function createEvent(req: NextApiRequest, res: NextApiResponse) {
   const userId = req.session.user.id;
   const { id: spaceId } = req.query as { id: string };
-  const payload = req.body as CreateEventPayload;
+  const events = req.body as CreateEventPayload[];
```
Comment (mattcasey) [inline pages/api/spaces/[id]/event.ts:17]: you should check this is an array so that we don't break requests with the previous code

Comment (Devorein) [reply pages/api/spaces/[id]/event.ts:17]: Yes, it only affects notifications, and it's a lot more optimal. Otherwise, we will have to send M x N API requests where N is the number of persons selected and M being the number of cards updated

charmClient/charmClient.ts
```diff
@@ -289,7 +289,7 @@ class CharmClient {
   }
 
   async deleteBlocks(blockIds: string[], updater: BlockUpdater): Promise<void> {
-    const rootBlocks = await http.DELETE<Block[]>(`/api/blocks`, blockIds);
+    const rootBlocks = await http.DELETE<Block[]>(`/api/blocks`, { blockIds });
```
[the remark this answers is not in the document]
mattcasey: nice catch

components/common/BoardEditor/components/properties/TagSelect/TagSelect.tsx
```diff
@@ -133,6 +135,12 @@ export function TagSelect({
     return options.map((o) => mapPropertyOptionToSelectOption(o));
   }, [options]);
 
+  useEffect(() => {
```
msadura, line 138: if you want to keep state in parent then internal isOpened state for this compoennt does not make sense

components/common/BoardEditor/components/properties/UserSelect.tsx
```diff
@@ -21,6 +21,7 @@ type Props = {
   displayType?: PropertyValueDisplayType;
   wrapColumn?: boolean;
   'data-test'?: string;
+  open?: boolean;
```
msadura, line 24: in previous file there is `isOpen`, `isOpened` and here it is `open` - would be nice to stick to some convention

components/common/BoardEditor/components/properties/UserSelect.tsx
```diff
@@ -112,10 +113,15 @@ export function UserSelect({
...
   showEmptyPlaceholder,
   wrapColumn,
+  open,
   'data-test': dataTest,
   error
 }: Props): JSX.Element | null {
-  const [isOpen, setIsOpen] = useState(false);
+  const [isOpen, setIsOpen] = useState(open);
+
+  useEffect(() => {
```
msadura, line 122: same again - why making redundant states?

components/common/BoardEditor/focalboard/src/components/table/table.tsx
```diff
@@ -231,6 +252,13 @@ function Table(props: Props): JSX.Element {
     [board, groupByProperty]
   );
 
+  const selectionContextValue = useMemo(() => {
```
msadura, line 255: this const does not make sense here. You are wrapping things returned from a hook with additional layer of complexity, while you could do:
`const selectionContext = useAreaSelection({ readOnly, container: selectContainerRef });`

and you can use it the same way. If you need memoized value then `useAreaSelection` should do that

msadura (changes requested):
just small polishes and will be ready to go

components/common/BoardEditor/components/properties/TagSelect/TagSelect.tsx
Devorein: The open state is managed by the component itself, but isOpen is provided to initial value of the state. I will remove useEffect since its redundant.

components/common/BoardEditor/components/properties/UserSelect.tsx
Devorein: Converted everything to `opened` to indicate the initial state of the `isOpen` internal state

components/common/BoardEditor/components/properties/UserSelect.tsx
Devorein: This is for the initial value of the internal state, remove the `useEffect`

components/common/BoardEditor/focalboard/src/components/table/table.tsx
Devorein: Updated

msadura (approved):

PR #3257, opened 2023-12-12
Fixed storybook build issue

Devorein:
### WHY

Note: I added all the providers necessary in the GlobalContext component; that way we don't need to concern ourselves whether all the providers are being passed or not. Just wrap your component inside `GlobalContext` and you should be good to go. If we need to modify the initial state of a context provider, it should be passed via a prop.


msadura (approved):
lets get it to main and fix ci :)

motechFR (approved):

PR #3260, opened 2023-12-13
Unlock Protocol

valentinludu:
Unlock Protocol implementation

Main changes and features:

- Added Unlock to the Token Gate modal
- TokenGate model schema will have a type and conditions can be any json. + Added validation of the payload for POST requests. In the future we will probably have many types of token gates
- Moved mutations from charmClient to hooks/tokenGates for better state management from the swr library
- Use unlock-protocol library to manage lock details
- Payload validation with yup - for conditions

Tested:
✅ Regression testing for lit protocol token gates
✅ User without a key
✅ Member with expired key
✅ Member with a valid key
✅ Chains tested: Polygon & Goerli


connectors/chains.ts
```diff
@@ -1,5 +1,6 @@
 import { log } from '@charmverse/core/log';
 // ref: https://wagmi.sh/core/chains
+import networks from '@unlock-protocol/networks';
```
mattcasey, line 3: it doesn't look like we need this to define our chain config at all, it just ends up exposting 'unlockChains'. It's a grey area but just to reduce file size everywhere, it might be a good idea to put this config inside of lib/tokenGates/unlock instead

lib/tokenGates/unlock/getLockMetadata.ts
```diff
@@ -0,0 +1,11 @@
+import { LocksmithService } from '@unlock-protocol/unlock-js';
```
mattcasey, line 1: this is a really massive lib, actually! VS Code says 1.4mb. it would be nice if we didn't need it, since our dev env is already pretty bloated

lib/tokenGates/unlock/getLockMetadata.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import type { GetLockPayload } from './getLockDetails';
+
+export async function getLockMetadata(values: Pick<GetLockPayload, 'chainId' | 'contract'>) {
```
mattcasey, line 5: i wouldn't have reused the input type here from getLockDetails. For one, both files are importing from each other which is hard to follow. Second, there's no reason the interface to getLockMetadata needs to be the same as getLockDetails. In fact, it requires you to cast chainId to Number twice.

lib/tokenGates/validateTokenGateConditions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const UnlockProtocolSchema = yup.object().shape({
+  contract: yup
+    .string()
+    .required()
+    .test('isAddress', 'Invalid address', (value) => isValidChainAddress(value)),
+  chainId: yup.number().required()
+});
+
+const LitProtocolSchema = yup.object().shape({
```
mattcasey, line 15: nice, i like that you used yup instead of writing the rules yourself

mattcasey: would be interesting to see some tests on this

lib/tokenGates/verifyTokenGates.ts
```diff
@@ -49,28 +49,40 @@ export async function verifyTokenGates({ spaceId, userId, tokens }: Props): Prom
...
+                verified: true,
+                grantedRoles: matchingTokenGate.tokenGateToRoles.map((tgr) => tgr.role.id)
+              };
+            }
           }
+        } else {
+          return {
+            ...matchingTokenGate,
+            jwt: tk.signedToken,
+            verified: true,
```
mattcasey, line 83: is this a bug? don't we need to still verify the gate thru unlock?

pages/api/token-gates/evaluate.ts
```diff
@@ -1,21 +1,22 @@
...
-async function verifyWallet(req: NextApiRequest, res: NextApiResponse) {
-  const user = req.session.user;
+async function verifyWallet(req: NextApiRequest, res: NextApiResponse<TokenGateEvaluationResult>) {
+  const body = req.body as TokenGateEvaluationAttempt;
 
-  const result = await evaluateTokenGateEligibility({
-    ...req.body,
-    userId: user.id
-  });
+  const result = await evaluateTokenGateEligibility(body);
```
mattcasey, line 19: interesting, we were not even using userId it seems

charmClient/apis/tokenGates.ts
```diff
@@ -1,44 +1,8 @@
...
-  }
-
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  verifyTokenGate(
-    verification: Omit<TokenGateVerificationRequest, 'userId'>
-  ): Promise<{ error?: string; success?: boolean }> {
-    return http.POST('/api/token-gates/verify', verification);
-  }
-
-  evaluateTokenGateEligibility(
```
mattcasey, line 36: thanks for cleaning up

pages/api/token-gates/index.ts
```diff
@@ -98,23 +84,22 @@ async function getTokenGates(req: NextApiRequest, res: NextApiResponse<TokenGate
...
     }
-  });
-  // Add identifiable names to token gates
-  const updatedResult = await updateTokenGateDetails(result as TokenGateWithRoles[]);
+  })) as TokenGateWithRoles[];
+
+  const tokenGatesWithDetails = await updateTokenGatesDetails(result);
 
-  res.status(200).json(updatedResult);
+  res.status(200).json(tokenGatesWithDetails);
```
mattcasey, line 102: It looks like we don't even use the response in the UI. I think this could just be res.status(200).end()

charmClient/hooks/helpers.ts
```diff
@@ -34,6 +34,14 @@ export function usePUT<T, U = unknown>(path: string) {
   });
 }
 
+// To be used when you need to trigger a get request on demand
+export function useGETTrigger<T, U = unknown>(path: string) {
```
mattcasey, line 38: i needed this! thanks. One suggestion is to just call it useGETtrigger. I dont think trigger needs to be capitalized

lib/tokenGates/evaluateEligibility.ts
```diff
@@ -153,3 +120,70 @@ export async function evaluateTokenGateEligibility({
...
+
+export async function getTokenGateResults(tokenGate: TokenGateWithRoles, authSig: AuthSig) {
+  if (tokenGate.type === 'unlock') {
+    return getUnlockProtocolTokenGateResults(tokenGate, authSig);
+  } else {
+    return getLitTokenGateResults(tokenGate, authSig);
+  }
+}
+
+export async function getUnlockProtocolTokenGateResults(tokenGate: TokenGateWithRoles<'unlock'>, authSig: AuthSig) {
```
mattcasey, line 132: dont think this needs to be exported

lib/tokenGates/evaluateEligibility.ts
```diff
@@ -123,7 +90,7 @@ export async function evaluateTokenGateEligibility({
     })
   );
 
-  const successGates = tokenGateResults.filter((result) => result !== null) as TokenGateJwt[];
+  const successGates = tokenGateResults.filter(isTruthy);
```
mattcasey, line 93: nice

pages/api/token-gates/evaluate.ts
valentinludu: yes, because we had req.body as any, typescript didn't throw an error

charmClient/apis/tokenGates.ts
valentinludu: Sure. It's nice now that we don't have to create 3 states for data, loading and error in the frontend and we get everything out of the box from swr

charmClient/hooks/helpers.ts
valentinludu: yeah, it looks weird with TT

lib/tokenGates/verifyTokenGates.ts
valentinludu: I will think of a better solution here

lib/tokenGates/validateTokenGateConditions.ts
valentinludu: I know we discussed about validating our endpoints payload so this can be useful in the future.

pages/api/token-gates/index.ts
valentinludu: This is the GET method. I think you are referring to the POST method. Yes, I will remove the response.

lib/tokenGates/evaluateEligibility.ts
valentinludu: Done. I removed export from 3 func that were not exported.

lib/tokenGates/unlock/getLockMetadata.ts
valentinludu: Yep. I remove id. I am checking outside if it is a number or  not and then I am sending it to the getLockMetadata

lib/tokenGates/verifyTokenGates.ts
valentinludu: I added the verification of unlock token gates here

mattcasey (approved):

PR #3263, opened 2023-12-13
Logout user from all tabs when logging out in one tab

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/send-user-to-login-page-when-they-log-out-from-another-tab-7662265015609018)

### DIFF
--- a/pages/api/session/logout.ts
+++ b/pages/api/session/logout.ts
@@ -5,16 +5,25 @@ import nc from 'next-connect';
 import { onError, onNoMatch } from 'lib/middleware';
 import { removeOldCookieFromResponse } from 'lib/session/removeOldCookie';
 import { withSessionRoute } from 'lib/session/withSession';
+import { relay } from 'lib/websockets/relay';
 
 const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
 
 handler.post(logout);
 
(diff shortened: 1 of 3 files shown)


pages/api/session/logout.ts
```diff
@@ -5,17 +5,26 @@ import nc from 'next-connect';
...
 
 async function logout(req: NextApiRequest, res: NextApiResponse<{ ok: boolean }>) {
-  log.debug('User logged out', { userId: req.session.user?.id });
+  const userId = req.session.user?.id;
+  log.debug('User logged out', { userId });
 
   req.session.destroy();
   await removeOldCookieFromResponse(req, res, false);
 
+  relay.broadcastToAll({
```
valentinludu, line 21: Niiice.
I didn't know we have this broadcastToAll.

valentinludu (approved):

pages/api/session/logout.ts
msadura: hold on, are you sending single user logout event to ALL the users connected users? This definitely does not work this way.

Devorein: I'm passing the `userId` of the user that is logging out so that the frontend can check against the currently logged-in user. I can try a different approach using `this.sockets[userId].emit(message)` but not sure if that's the correct approach.

PR #3266, opened 2023-12-13
Open card property menu on clicking handler

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/delete-option-not-coming-up-from-block-handle-on-proposal-custom-property-6017166448466353)

### DIFF
--- a/components/common/BoardEditor/components/cardProperties/CardDetailProperty.tsx
+++ b/components/common/BoardEditor/components/cardProperties/CardDetailProperty.tsx
@@ -84,6 +84,7 @@ export function CardDetailProperty({
         <Box>
           <PropertyNameContainer
             className='octo-propertyname'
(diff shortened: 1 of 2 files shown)


mattcasey (approved):

components/common/BoardEditor/components/cardProperties/CardDetailProperty.tsx
```diff
@@ -84,14 +84,15 @@ export function CardDetailProperty({
         <Box>
           <PropertyNameContainer
             className='octo-propertyname'
+            {...bindTrigger(changePropertyPopupState)}
```
mattcasey, line 87: i was worried this would trigger on drag/drop but it doesn't. nice improvement!

Devorein: Yeah same.

PR #3267, opened 2023-12-13
Populate person property in a filter when creating a new card

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/should-create-new-cards-with-the-filtered-member-the-way-we-would-filtered-multiselect-6425044077588473)

### DIFF
--- a/lib/focalboard/cardFilter.ts
+++ b/lib/focalboard/cardFilter.ts
@@ -328,11 +328,6 @@ class CardFilter {
     filterClause: FilterClause,
     templates: readonly IPropertyTemplate[]
   ): { id: string; value?: string | string[] } {
-    const template = templates.find((o) => o.id === filterClause.propertyId);
-    if (!template) {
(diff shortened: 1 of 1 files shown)


mattcasey (approved):

lib/focalboard/cardFilter.ts
```diff
@@ -328,11 +328,6 @@ class CardFilter {
     filterClause: FilterClause,
     templates: readonly IPropertyTemplate[]
   ): { id: string; value?: string | string[] } {
-    const template = templates.find((o) => o.id === filterClause.propertyId);
```
mattcasey, line 331: looks like just a cleanup?

mattcasey (approved):

lib/focalboard/cardFilter.ts
Devorein: yep, the variable is re-declared again just below

PR #3269, opened 2023-12-13
expect summonUserId to be provided when member is added

mattcasey:
### WHY
Summon team agreed they will just send us the user's summon id instead.

Another change I made is that summonUserId is not required to use this endpoint

### DIFF
--- a/pages/api/v1/spaces/[spaceId]/members/index.ts
+++ b/pages/api/v1/spaces/[spaceId]/members/index.ts
@@ -1,11 +1,11 @@
 import { UnauthorisedActionError } from '@charmverse/core/errors';
+import { log } from '@charmverse/core/log';
 import type { NextApiRequest, NextApiResponse } from 'next';
 
 import { isTestEnv } from 'config/constants';
-import { NotFoundError, requireKeys, requireSuperApiKey } from 'lib/middleware';
-import { getSummonProfile } from 'lib/profile/getSummonProfile';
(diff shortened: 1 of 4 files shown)


pages/api/v1/spaces/[spaceId]/members/index.ts
```diff
@@ -29,7 +29,20 @@ handler.use(requireKeys([{ key: 'wallet', valueType: 'wallet' }], 'body')).post(
  *       content:
  *          application/json:
  *             schema:
- *                $ref: '#/components/schemas/CreateSpaceMemberRequestBody'
+ *                 CreateSpaceMemberRequestBody:
```
Comment (mattcasey) [inline pages/api/v1/spaces/[spaceId]/members/index.ts:32]: i moved this interface here because it's not used any place else. reduces clutter on the api docs too

pages/api/v1/spaces/[spaceId]/members/index.ts
```diff
@@ -39,6 +52,12 @@ handler.use(requireKeys([{ key: 'wallet', valueType: 'wallet' }], 'body')).post(
  *                $ref: '#/components/schemas/UserProfile'
  */
 
+type CreateSpaceMemberRequestBody = {
+  email?: string;
+  summonUserId?: string;
```
Comment (mattcasey) [inline pages/api/v1/spaces/[spaceId]/members/index.ts:57]: summonUserId: this is what i wish we had called it internally but "Summon" was not a thing at the time.  Eventually we might migrate the column, but i wanted the api to be accurate

valentinludu (approved):

PR #3272, opened 2023-12-14
Refetch tokengates after creation

valentinludu:
Fix for refetching token gates after we create them.

Looks like mutate() from swr doesn't work.


components/common/TokenGateModal/hooks/useTokenGateModalContext.tsx
```diff
@@ -138,7 +145,7 @@ export function TokenGateModalProvider({ children, onClose }: { children: ReactN
         id: tokenGateId
       });
 
-      mutate(`/api/token-gates`);
+      refreshTokenGates();
```
mattcasey, line 148: much better to do this anyway thanks

mattcasey (approved):

PR #3273, opened 2023-12-14
Proposals/permissions workflow adapt

motechFR:
### WHAT

Opt in cvt spaces to new proposal permissions model

### WHY

Facilitate workflow development with new feature


pages/api/spaces/[id]/proposals/index.ts
```diff
@@ -1,43 +1,40 @@
...
+    useProposalEvaluationPermissions
   });
 
   const proposals = await prisma.proposal.findMany({
     where: {
       id: {
         in: proposalIds
+      },
+      page: {
+        // Ignore proposal templates
```
Comment (mattcasey) [inline pages/api/spaces/[id]/proposals/index.ts:36]: nice

mattcasey (approved):
i wonder if we disable isCharmVerse in test, would u need to have changed the data for e2e? (still i'd like to keep both sets of tests around, and we could prefix one test space with cvt- in the domain)

PR #3283, opened 2023-12-19
Evaluations/permissions cleanup

motechFR:
### WHAT

Migrate permissions calls to permissions API

### WHY

Easier to maintain and control switching behaviour and feature parity

Also overrode the permissions API client to force calls to useProposalEvaluationPermissions in some cases


mattcasey (approved):

pages/api/pages/[id]/archive.ts
```diff
@@ -6,21 +6,14 @@ import { ActionNotPermittedError, onError, onNoMatch, requireKeys, requireUser }
 import type { ModifyChildPagesResponse } from 'lib/pages';
 import { archivePages } from 'lib/pages/archivePages';
 import { PageNotFoundError } from 'lib/pages/server';
-import { providePermissionClients } from 'lib/permissions/api/permissionsClientMiddleware';
+import { permissionsApiClient } from 'lib/permissions/api/routers';
```
Comment (mattcasey) [inline pages/api/pages/[id]/archive.ts:9]: since you got rid of /client, can you actually rename routers.ts to client.ts? that is a pretty standard file name. there is no 'routing' logic in this "client" anymore

PR #3287, opened 2023-12-20
Feat/proposal form api

msadura:


components/common/form/interfaces.ts
```diff
@@ -52,3 +52,9 @@ export type FormFieldInput = Pick<
 > & {
   options?: SelectOptionType[];
 };
+
+export type FieldAnswerInput = {
+  id?: string;
+  fieldId: string;
+  value: string | string[];
```
Devorein, line 59: The value can also be `{ content: PageContent; contentText: string;}` but I've added a type `FormFieldValue` with those unions in my PR #3293

lib/form/__tests__/upsertProposalFormAnswers.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+    const updated1 = await upsertProposalFormAnswers({ proposalId: proposal.id, formId, answers: [answer1] });
+    expect(updated1).toBeDefined();
+
+    expect(updated1).toEqual<FormFieldAnswer[]>(
+      expect.arrayContaining([
+        { ...answer1, id: expect.any(String), fieldId: fieldsInput[0].id, proposalId: proposal.id, type: 'short_text' }
+      ])
+    );
+
+    const updatedAnswer = {
```
Devorein, line 181: Shouldn't we pass the `id` in the payload since its updating an existing answer?

msadura: id doesn't really matter, because we lookup answer by fieldId + ProposalId

lib/form/upsertProposalFormAnswers.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (!stringUtils.isUUID(proposalId)) {
+    throw new InvalidInputError(`Valid proposalId is required`);
+  } else if (!stringUtils.isUUID(formId)) {
+    throw new InvalidInputError(`Valid formId is required`);
+  }
+
+  const form = await prisma.form.findUnique({ where: { id: formId }, include: { formFields: true } });
+
+  if (!form) {
+    throw new InvalidInputError(`Could not find form ${formId}`);
```
Devorein, line 24: Suggestion: This could be NotFoundError

lib/form/upsertProposalFormAnswers.ts
```diff
@@ -0,0 +1,75 @@
...
+      throw new InvalidInputError(`Could not find field ${answer.fieldId} for proposal ${proposalId}`);
+    }
+
+    if (field.required && !answer.value) {
+      throw new InvalidInputError(`Value for field ${field.name} is required`);
+    }
+  }
+
+  const hasAllRequiredAnswers = form.formFields.every(
+    (f) => !f.required || answers.some((a) => a.fieldId === f.id && !!a.value)
```
Devorein, line 43: We should update this to skip over `label` type fields, ideally it should not stored in db as required. But if it is, we need to update it from the db as there are no fe controls for making it required (as its an empty field)

lib/form/upsertProposalFormAnswers.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  if (!hasAllRequiredAnswers) {
+    throw new InvalidInputError(`All required fields must be answered`);
+  }
+
+  const res = await prisma.$transaction([
+    ...answers.map((a) => {
+      const field = form.formFields.find((f) => f.id === a.fieldId);
```
Devorein, line 52: Same case as above, I think we should filter out label so that it doesn't have any value in db. That's already done of fe so we can also skip it.

lib/proposal/form/upsertProposalFormFields.ts
```diff
@@ -0,0 +1,106 @@
...
+    throw new InvalidInputError(`No proposal found with id: ${proposalId}`);
+  }
+
+  if (!formFields.length) {
+    throw new InvalidInputError(`You need to specify at least one form field`);
+  }
+
+  const proposal = await prisma.proposal.findUnique({ where: { id: proposalId }, include: { page: true } });
+
+  if (!proposal) {
```
Devorein, line 26: We are slowly moving to the paradigm of using `findUniqueOrThrow`

lib/proposal/form/upsertProposalFormFields.ts
```diff
@@ -0,0 +1,106 @@
...
+          create: {
+            id: fieldId,
+            formId,
+            type: field.type,
+            name: field.name,
+            description: field.description as any,
+            index: field.index,
+            options: field.options,
+            private: field.private,
+            required: field.required
```
Devorein, line 80: We can forcefully make it unrequired here if its of type `label`, then we don't need to add any guards against it anywhere we are checking for required fields

lib/proposal/mapDbProposalToProposal.ts
```diff
@@ -1,33 +1,38 @@
...
 import type { ProposalWithUsers } from '@charmverse/core/proposals';
 import type { ProposalEvaluation } from '@prisma/client';
 
-import type { PopulatedEvaluation } from 'lib/proposal/interface';
-
 import { getOldProposalStatus } from './getOldProposalStatus';
 import type { ProposalWithUsersAndRubric } from './interface';
 
+type FormFieldsIncludeType = {
+  form: {
```
Devorein, line 12: We are removing `form` and only using `formFields` below, it might mislead the frontend thinking it can use `form.formFields`

Devorein (approved):

lib/proposal/mapDbProposalToProposal.ts
msadura: what is misleading? You have correct type "formFields" on the FE type

PR #3290, opened 2023-12-20
update conversion script to support permissions

mattcasey:

### DIFF
--- a/scripts/migrations/2023_11_30_convertProposals.ts
+++ b/scripts/migrations/2023_11_30_convertProposals.ts
@@ -175,5 +175,5 @@ async function convertSpaces() {
 }
 
 function isPublished(proposal: Proposal) {
-  return proposal.status !== 'draft' && proposal.status !== 'discussion';
+  return proposal.status !== 'draft';
 }
--- a/scripts/migrations/2023_11_30_convertProposals.ts
+++ b/scripts/migrations/2023_11_30_convertProposals.ts
@@ -3,8 +3,11 @@ import { prisma } from '@charmverse/core/prisma-client';
 import { getDefaultWorkflows } from 'lib/proposal/workflows/defaultWorkflows';
 import { v4 as uuid } from 'uuid';
 
+import { getDefaultFeedbackEvaluation, getDefaultPermissions } from 'lib/proposal/workflows/defaultEvaluation';
 // This script is a work in progress
 // It adds proposal steps to existing proposals
+
+const { permissions: feedbackPermissions, id, ...feedbackEvaluation } = getDefaultFeedbackEvaluation();
 
 async function convertProposals() {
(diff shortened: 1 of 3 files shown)


scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -38,21 +47,38 @@ async function convertProposals() {
               }
             });
           }
+          await tx.proposalEvaluation.create({
```
motechFR, line 50: General suggestion for preceding statement, isPublished would only be proposal.status !== 'draft'

scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -38,21 +47,38 @@ async function convertProposals() {
...
-                create: {
-                  proposalId: p.id
-                }
+                create: p.category?.proposalCategoryPermissions
+                  .filter((perm) => perm.permissionLevel === 'full_access')
+                  .map((perm) => ({
+                    proposalId: p.id,
+                    roleId: perm.roleId,
+                    systemRole: perm.spaceId ? 'space_member' : undefined
+                  }))
```
motechFR, line 78: Looks good!

scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -38,21 +47,38 @@ async function convertProposals() {
...
+              },
+              permissions: {
+                create: getDefaultPermissions()
```
motechFR, line 81: makes sense

scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -38,21 +47,38 @@ async function convertProposals() {
...
               result,
               completedAt: result ? completedAt || new Date() : null,
               proposalId: p.id,
               snapshotId: p.page?.snapshotProposalId,
               voteId: vote?.id,
               reviewers: {
-                create: {
-                  proposalId: p.id
-                }
+                create: p.category?.proposalCategoryPermissions
```
motechFR, line 72: Shouldn't this be createMany? Haven't used create before for multiple items

motechFR:
Final potential issue: Are we missing setting currentEvaluationId on the proposal? Can't see it here

scripts/migrations/2023_11_30_convertProposals.ts
mattcasey: i actually dont know i was surprised prisma lets me

scripts/migrations/2023_11_30_convertProposals.ts
mattcasey: thanks! I updated isPublished to only check for draft. THis changed over time

PR #3303, opened 2023-12-21
more changes to proposal migration

mattcasey:

### DIFF
--- a/scripts/migrations/2023_11_30_convertProposals.ts
+++ b/scripts/migrations/2023_11_30_convertProposals.ts
@@ -47,10 +47,12 @@ async function convertProposals() {
               }
             });
           }
+          const feedbackComplete = p.status !== 'draft' && p.status !== 'discussion';
           await tx.proposalEvaluation.create({
             data: {
               ...feedbackEvaluation,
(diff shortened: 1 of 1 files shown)


scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -96,10 +98,12 @@ async function convertProposals() {
               }
             });
           }
+          const feedbackComplete = p.status !== 'draft' && p.status !== 'discussion';
           await tx.proposalEvaluation.create({
             data: {
               ...feedbackEvaluation,
               index: 0,
+              result: feedbackComplete ? 'pass' : null,
```
motechFR, line 106: Suggestion: For the final evaluation; we might also want to port over the evaluation_closed logic to differentiate with evaluation_active

mattcasey: there is no way to know if it was close,d i already spok with chris and lal rubric will just revert to "open" state

mattcasey:
closed in favor of https://github.com/charmverse/app.charmverse.io/pull/3294

PR #3310, opened 2023-12-24
Fixed adding custom proposal properties

Devorein:
### WHY

1. Custom proposal properties couldn't be created before on a standalone page view. They were working from the proposals table page. The solution was a missing context provider
2. Proposal form field answers are available even when no user is logged in via public proposals table or public proposal

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -1,9 +1,9 @@
 import type { Page } from '@charmverse/core/prisma';
 import type { Theme } from '@mui/material';
-import { useMediaQuery, Divider, Box } from '@mui/material';
+import { Box, Divider, useMediaQuery } from '@mui/material';
 import dynamic from 'next/dynamic';
 import type { EditorState } from 'prosemirror-state';
-import { memo, useCallback, useEffect, useRef, useState, useMemo } from 'react';
+import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useElementSize } from 'usehooks-ts';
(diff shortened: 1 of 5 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -435,7 +436,7 @@ function DocumentPage({
   const proposalAuthors = proposal ? [proposal.createdBy, ...proposal.authors.map((author) => author.userId)] : [];
 
   return (
-    <>
+    <ProposalBlocksProvider>
```
Comment (Devorein) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:439]: We can either conditionally wrap everything with `ProposalBlockProvider` if the `page.type` is `proposal` or make it simple and add it inside `_app` context providers or just keep it here as is.

Comment (mattcasey) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:439]: lets keep it here, it doesnt belong in _app since pages like login do not need blocks

lib/proposal/form/getProposalFormAnswers.ts
```diff
@@ -27,7 +27,9 @@ export async function getProposalFormAnswers({ proposalId, userId }: { userId: s
     return [];
   }
 
-  const canViewPrivateFields = await canAccessPrivateFields({ proposalId: proposal.id, userId, proposal });
+  const canViewPrivateFields = userId
+    ? await canAccessPrivateFields({ proposalId: proposal.id, userId, proposal })
```
mattcasey, line 31: suggestion: canAccessPrivateFields could just return false if userId is undefined

mattcasey (approved):

PR #3319, opened 2023-12-27
Get nft image and name for ERC1155

valentinludu:
Show erc1155 name and image if possible.


lib/blockchain/updateLitDetails.ts
```diff
@@ -68,3 +72,24 @@ export async function updateTokenGateLitDetails<T extends LitConditions>(
 
   return updatedTokenGates;
 }
+
+function getTokenId(conditions: AccsDefaultParams) {
```
mattcasey, line 76: thanks for breaking this logic out

mattcasey (approved):

PR #3322, opened 2023-12-27
Check all users for memberships

valentinludu:
Fix role id error - There were situations where we had only roleId.
Use Promise.allSettled to see all errors and prevent breaking on the first try.


mattcasey (approved):
nice

PR #3323, opened 2023-12-27
Move proposal stepper into sidebar

mattcasey:
The major change here was to merge the horizontal proposal stepper with evaluation input in the sidebar. It simplified a good amount, but required a lot of changes at once. The sidebar is sticky, meaning you can't close it now, and we got rid of the sticky footer actions except to publish draft proposals.

image

Also, on mobile, you can now see the evaluation via tabs instead of the sidebar
image


.github/workflows/deploy_staging.yml
```diff
@@ -9,7 +9,6 @@ env:
 
 concurrency:
   group: staging-${{ github.ref }}
-  cancel-in-progress: true
```
mattcasey, line 12: if you do lots of updates, your staging env never updates because it takes 20 minutes to deploy :(

scripts/migrations/2023_11_30_convertProposals.ts
```diff
@@ -24,31 +24,36 @@ async function convertProposals() {
...
-              }
-            });
-          }
+          await tx.proposal.update({
+            where: {
+              id: p.id
+            },
+            data: {
+              status: isPublished(p) ? 'published' : undefined,
+              workflowId: workflow?.id
```
mattcasey, line 54: attempt to always have a workflowId, just so we can show the name of something in the sidebar - even though you cannot change it later

components/proposals/ProposalPage/components/ProposalProperties/ProposalPropertiesBase.tsx
```diff
@@ -455,6 +449,17 @@ export function ProposalPropertiesBase({
           </Box>
         )}
 
+        <CustomPropertiesAdapter
```
mattcasey, line 452: product wants to have all properties appear above rewards

components/proposals/ProposalPage/components/ProposalProperties/ProposalPropertiesBase.tsx
```diff
@@ -194,12 +192,8 @@ export function ProposalPropertiesBase({
...
 
   useEffect(() => {
-    if (!prevStatusRef.current && proposalStatus === 'draft') {
-      setDetailsExpanded(true);
-    }
-
-    prevStatusRef.current = proposalStatus || '';
-  }, [detailsExpanded, proposalStatus]);
+    setDetailsExpanded(proposalStatus === 'draft');
+  }, [setDetailsExpanded, proposalStatus]);
```
mattcasey, line 196: always collapse/expand the details depending on current step

components/common/BoardEditor/focalboard/src/components/cardDetail/cardDetail.scss
```diff
@@ -161,4 +124,42 @@
       }
     }
   }
+}
+
+.octo-propertyname {
```
mattcasey, line 129: now we no longer need to wrap property labels with "CardDetails". i checked and "octo-propertyname" is always inside card details, never used in, say, a table or kanban board

components/[pageId]/DocumentPage/components/Sidebar/constants.tsx
```diff
@@ -8,11 +8,6 @@ export const SIDEBAR_VIEWS = {
     tooltip: 'View evaluation',
     title: 'Evaluation'
   },
-  proposal_evaluation_settings: {
-    icon: <SvgIcon component={RiChatCheckLine} fontSize='small' sx={{ mb: '1px' }} />,
-    tooltip: 'Manage reviewers, rubric, and vote options',
-    title: 'Evaluation settings'
-  },
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/Sidebar/constants.tsx:15]: we don't need a special view, since now it is either or: we show Settings for new/draft proposals, and the regular view for everything else.

lib/proposal/interface.ts
```diff
@@ -60,6 +65,10 @@ export type ProposalFormData = {
   };
 };
 
+export type ProposalWithUsers = Omit<CoreProposalWithUsers, 'fields'> & {
```
mattcasey, line 68: now 'fields' has a type. made it a lot easier to use in the frontend. we should probably stop directly importing from core everywhere

mattcasey:
thanks @Devorein !!

responses:
1) yes, this is requested by chris since we no longer use page icons/banners for proposals, and now the title lines up with the sidebar content
2) this is probably due to the fact that web sockets are not workign on staging. i tested and it worked in my case locally
3) good point, will address this in another PR
4) fixed (will add test sin another PR)
5) fixed (made it a public endpoint)

PR #3328, opened 2023-12-29
Add remove icon in token gate review state

valentinludu:
Go to In Invites -> Add token gate
Add multiple token gates and you will see an X icon that can delete a certain item in review.

Screenshot 2023-12-29 at 16 55 01


mattcasey (approved):
Looks fine but I've never seen us use this icon. Shouldn't we be using the trash icon instead?

PR #3333, opened 2023-12-30
Edit proposal reviewers from proposals table

Devorein:


charmClient/apis/proposalsApi.ts
```diff
@@ -48,4 +49,8 @@ export class ProposalsApi {
   deleteProposalCategory(spaceId: string, categoryId: string) {
     return http.DELETE<{ ok: true }>(`/api/spaces/${spaceId}/proposal-categories/${categoryId}`);
   }
+
+  updateProposalEvaluation({ proposalId, ...payload }: UpdateEvaluationRequest) {
```
mattcasey, line 53: please do not add any more methods to charmClient. We already have `useUpdateProposalEvaluation` hook, please use that instead

components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
```diff
@@ -71,14 +71,17 @@ function SelectedOptions({
...
+  disallowEmpty = false
 }: {
   wrapColumn?: boolean;
   readOnly: boolean;
   value: SelectOptionPopulated[];
   isRequiredValue?: (option: SelectOptionPopulated) => boolean;
   onRemove: (reviewerId: string) => void;
+  disallowEmpty?: boolean;
 }) {
+  const showDeleteIcon = (disallowEmpty && value.length !== 1) || !disallowEmpty;
```
mattcasey, line 84: This is poor UX - it forces the user to add what they need before they can remove what they don't want. Instead, I think we should add a standard prop "required" and highlight it in red if it's empty, and then dont save it until there is a value

components/common/BoardEditor/components/properties/UserAndRoleSelect.tsx
```diff
@@ -188,7 +200,8 @@ export function UserAndRoleSelect<T extends { id: string; group: string } = Sele
   'data-test': dataTest,
   wrapColumn,
   type = 'roleAndUser',
-  required
+  required,
+  disallowEmpty = false
```
mattcasey, line 204: we already have a 'required' field. please use that instead, and just don't save the update  if there are no values

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -306,16 +307,45 @@ function PropertyValueElement(props: Props) {
...
+        displayType={displayType}
+        data-test='selected-reviewers'
+        readOnly={readOnly || (displayType !== 'details' && displayType !== 'table')}
+        systemRoles={[allMembersSystemRole]}
+        onChange={async (reviewers) => {
+          const proposalWithEvaluation = proposal as ProposalWithUsers & { currentEvaluationId?: string | null };
+          const evaluationId =
+            'currentEvaluationId' in proposalWithEvaluation ? proposalWithEvaluation.currentEvaluationId : undefined;
+          if (proposal && evaluationId) {
+            await charmClient.proposals.updateProposalEvaluation({
```
mattcasey, line 322: needs a try/catch in case it fails

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -306,16 +307,45 @@ function PropertyValueElement(props: Props) {
...
+  } else if (propertyTemplate.type === 'proposalReviewer') {
+    propertyValueElement = (
+      <UserAndRoleSelect
+        displayType={displayType}
+        data-test='selected-reviewers'
+        readOnly={readOnly || (displayType !== 'details' && displayType !== 'table')}
+        systemRoles={[allMembersSystemRole]}
+        onChange={async (reviewers) => {
+          const proposalWithEvaluation = proposal as ProposalWithUsers & { currentEvaluationId?: string | null };
+          const evaluationId =
```
mattcasey, line 319: this line doesn't make any sense. `proposalWithEvaluation.currentEvaluationId` is already a string or undefined. Just do:
```
const evaluationId = proposalWithEvaluation.currentEvaluationId
```

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -306,16 +307,45 @@ function PropertyValueElement(props: Props) {
...
+            await charmClient.proposals.updateProposalEvaluation({
+              reviewers: reviewers.map((reviewer) => ({
+                roleId: reviewer.group === 'role' ? reviewer.id : null,
+                systemRole: reviewer.group === 'system_role' ? (reviewer.id as ProposalSystemRole) : null,
+                userId: reviewer.group === 'user' ? reviewer.id : null
+              })),
+              proposalId: proposal.id,
+              evaluationId
+            });
+            await mutate(`/api/spaces/${board.spaceId}/proposals`);
```
mattcasey, line 331: TODO: websockets for proposal list :)

components/common/BoardEditor/focalboard/src/components/table/tableRow.tsx
```diff
@@ -69,6 +70,7 @@ type Props = {
   emptySubPagesPlaceholder?: ReactNode;
   isChecked?: boolean;
   setCheckedIds?: Dispatch<SetStateAction<string[]>>;
+  proposal?: ProposalWithUsers;
```
mattcasey, line 73: this is going to rerender every row whenever you update proposals. Instead it looks like we only need to pass in proposalId and currentEvaluationId. Please update

lib/focalboard/card.ts
```diff
@@ -21,6 +22,7 @@ export type CardPage<V = CardPropertyValue> = {
   subPages?: CardPage<V>[];
   card: Card<V>;
   page: PageMeta;
+  proposal?: ProposalWithUsers;
```
mattcasey, line 25: change this to proposalId and currentEvaluationId. These objects should be as simple as possible to avoid rerendering large tables

components/proposals/ProposalPage/components/ProposalProperties/hooks/useProposalsBoardAdapter.ts
```diff
@@ -36,11 +36,12 @@ export function useProposalsBoardAdapter() {
   const [boardProposal, setBoardProposal] = useState<BoardProposal | null>(null);
   const { space } = useCurrentSpace();
   const { membersRecord } = useMembers();
-  const { proposals } = useProposals();
+  const { proposals = [] } = useProposals();
```
mattcasey, line 39: its ususally better not to have a default. when you're workign with code its good to have to think about things loading

mattcasey (changes requested):
left comments, primarily i don't think we should pass the entire proposal object around when we only need the ids

lib/focalboard/card.ts
mattcasey: In fact, I think you may only need to add currentEvaluationId, since PageMeta already includes the proposalId?

charmClient/apis/proposalsApi.ts
Devorein: Right, updated. I will use the hooks from now on.

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Trimmed the proposal object to only have currentEvaluationId and id

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Added a task card in backlog

lib/focalboard/card.ts
Devorein: Yep good idea, updated.

mattcasey (approved):

lib/focalboard/card.ts
mattcasey: @Devorein not sure you actually updated this, 'proposal' is still an object, so it will re-render all the time still. The problem is we're recreating the 'proposal' property every time proposals are updated

mattcasey: I don't think the update really addresses what I meant, 'proposal' should not be an object we re-create all the time. but we're doing it for the card too... the useProposalsBoardAdapter hook is just not set up correctly

PR #3345, opened 2024-01-02
Carry over more values from the original workflow when selecting a template

mattcasey:
Carries over some updated properties from the workflow, like step order

Also removed one more isCharmVerse check


components/proposals/ProposalPage/NewProposalPage.tsx
```diff
@@ -253,6 +208,52 @@ export function NewProposalPage({
...
+          index,
+          reviewers: existingStep?.reviewers || [],
+          rubricCriteria: existingStep?.rubricCriteria || ([] as ProposalRubricCriteriaWithTypedParams[]),
+          title: evaluation.title,
+          type: evaluation.type,
+          result: null,
+          voteSettings: existingStep?.voteSettings,
+          permissions: evaluation.permissions as ProposalEvaluationPermission[]
+        };
+      })
```
motechFR, line 241: I recommend extending the assertions in this E2E test to make sure the settings are ported over correctly

__e2e__/proposals/proposalEvaluationCreation.spec.ts

mattcasey: Interestingly, the e2e test broke cuz i was not updating it properly :)

motechFR: There were definitely a few places we could test. Reason I went with that e2e test originally, is there's so many moving pieces, it's the one place we can ensure the feature will work

PR #3346, opened 2024-01-02
Removed Share tab for proposal page header

Devorein:
### WHY

image


motechFR (approved):

mattcasey (approved):
approved but i dont think we need the "Publish" tab anymore

PR #3348, opened 2024-01-03
Feat/saving snapshot proposal expiry

msadura:
### WHAT

- saving snapshot expiry date when creating new proposal
- remove snapshot expiry field from proposal and use one from evaluation

NOTE: tests are failing jsut due to wrong types in permissions. WIll be fixed once core pr is merged

### DIFF
--- a/background/tasks/updateProposalStatus/task.ts
+++ b/background/tasks/updateProposalStatus/task.ts
@@ -10,9 +10,13 @@ export async function task() {
     const proposalsToUpdate = await prisma.proposal.findMany({
       where: {
         status: 'vote_active',
-        snapshotProposalExpiry: {
-          not: null,
-          lte: new Date()
+        evaluations: {
+          some: {
+            snapshotExpiry: {
+              not: null,
+              lte: new Date()
+            }
+          }
         }
       }
     });
--- a/background/tasks/updateProposalStatus/task.ts
+++ b/background/tasks/updateProposalStatus/task.ts
@@ -7,37 +7,6 @@ export async function task() {
   log.debug('Running Proposal status cron job');
 
   try {
-    const proposalsToUpdate = await prisma.proposal.findMany({
-      where: {
-        status: 'vote_active',
-        evaluations: {
-          some: {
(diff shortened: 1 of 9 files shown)


pages/api/proposals/[id]/snapshot.ts
```diff
@@ -44,12 +46,15 @@ async function recordSnapshotInfo(req: NextApiRequest, res: NextApiResponse<Page
     throw error;
   }
 
+  const snapshotProposal = snapshotProposalId ? await getSnapshotProposal(snapshotProposalId) : null;
```
Comment (msadura) [inline pages/api/proposals/[id]/snapshot.ts:49]: this is code responsible for saving snapshotExpiry on evaluation

background/tasks/updateProposalStatus/task.ts
```diff
@@ -10,9 +10,13 @@ export async function task() {
     const proposalsToUpdate = await prisma.proposal.findMany({
       where: {
         status: 'vote_active',
-        snapshotProposalExpiry: {
-          not: null,
-          lte: new Date()
+        evaluations: {
```
mattcasey, line 13: we no longer use "vote_closed" on proposal, so we should update the evaluation. instead with "result": "pass/fail"

scripts/migrations/2023_12_29_convertProposals.ts
```diff
@@ -72,7 +72,7 @@ async function convertProposals({ offset = 0 }: { offset?: number } = {}) {
           type: vote?.type || 'Approval',
           options: vote?.voteOptions.map((o) => o.name) || ['Yes', 'No', 'Abstain'],
           maxChoices: vote?.maxChoices || 1,
-          publishToSnapshot: !!p.snapshotProposalExpiry || !!p.page?.snapshotProposalId,
+          publishToSnapshot: !!p.page?.snapshotProposalId,
```
mattcasey, line 75: this code was already run, we should write a new script that only migrates proposals with snapshot

mattcasey (approved):

mattcasey (approved):

PR #3349, opened 2024-01-03
Updated proposal notification

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/update-notifications-18889742982561852)


lib/notifications/constants.ts
```diff
@@ -12,11 +12,12 @@ export const bountyNotificationTypes = [
...
-  'reviewed',
-  'needs_review',
+  'review_required',
   'vote',
-  'evaluation_active',
-  'evaluation_closed'
+  'vote_closed',
+  'reward_published',
+  'proposal_passed',
+  'step_passed',
```
mattcasey, line 21: There's something weird about these events. It should either be vague about content, and we just have step_passed, step_failed, and step_started, or we should not have 'step_failed' or 'step_passed' at all and each proposal type should have its own events. But I would definitely prefer we just have a few events, and the template or whatever downstream figures out _how to express it_ and _who should receive it_ based on evaluation type

I think we could just reduce the types of events further, maybe its just step_started and step_closed? and you can figure out based on the type and 

if we have "step_passed", then i dont think we should have any other alternatives based on type, so it should also replace "proposal_passed" and "reward_published".

pages/api/proposals/[id]/publish.ts
```diff
@@ -36,10 +38,30 @@ async function publishProposalStatusController(req: NextApiRequest, res: NextApi
     },
     select: {
       id: true,
+      proposal: {
+        select: {
+          evaluations: {
```
Comment (mattcasey) [inline pages/api/proposals/[id]/publish.ts:43]: sort the evaluations

lib/notifications/constants.ts
mattcasey: change step_failed to proposal_failed

mattcasey:
Please check both view and comment permission when sending event for `discussion`

background/tasks/updateVotesStatus/updateVoteStatus.ts
```diff
@@ -97,22 +97,15 @@ const updateVoteStatus = async () => {
...
 
   await Promise.all([
-    ...rejectedVotes.map((vote) => {
-      if (vote.pageId) {
+    [...rejectedVotes, ...passedVotes].map((vote) => {
+      const evaluation = passedEvaluations.find((e) => e.voteId === vote.id);
+      if (vote.pageId && evaluation) {
         return publishProposalEvent({
-          scope: WebhookEventNames.ProposalFailed,
+          scope: WebhookEventNames.ProposalStatusChanged,
```
mattcasey, line 104: use ProposalFailed or ProposalPassed if this is the last step

lib/notifications/constants.ts
```diff
@@ -12,11 +12,13 @@ export const bountyNotificationTypes = [
...
-  'start_review',
   'start_discussion',
-  'reviewed',
-  'needs_review',
+  'review_required',
   'vote',
-  'evaluation_active',
-  'evaluation_closed'
+  'vote_passed',
+  'vote_failed',
```
mattcasey, line 19: vote_failed === proposal_failed

mattcasey (approved):

PR #3353, opened 2024-01-03
make labels h1 size for form fields

mattcasey:


components/common/form/fields/FieldWrapper.tsx
```diff
@@ -20,13 +20,29 @@ type Props = {
...
+  inline,
+  sx,
+  children
+}: {
+  inline?: boolean;
+  sx?: SxProps;
+  children: ReactNode;
+}) {
+  return (
+    <Box flex={1} flexDirection={{ xs: 'column', sm: inline ? 'row' : 'column' }} display='flex' sx={sx}>
```
mattcasey, line 33: I removed "my={1}" here, and was able to completely remove this property "fieldWrapperSx" which was only being set to remove it. Outside padding is layout and should be handled outside of common components

PR #3366, opened 2024-01-04
Updated Proposal Status and Step columns

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/new-proposal-table-columns-5493120085075087)


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -315,8 +320,8 @@ function PropertyValueElement(props: Props) {
           !proposal ||
           readOnly ||
           (displayType !== 'details' && displayType !== 'table') ||
-          proposal.status === 'draft' ||
-          proposal.currentEvaluation?.type === 'feedback' ||
+          proposal.currentEvaluation?.step === 'draft' ||
```
mattcasey, line 323: 'draft' is not a current evaluation, i would like to not introduce that concept

lib/proposal/getProposalEvaluationStatus.ts
```diff
@@ -0,0 +1,74 @@
+import type { ProposalEvaluation, ProposalStatus } from '@charmverse/core/prisma';
+import { getCurrentEvaluation } from '@charmverse/core/proposals';
+
+import type { ProposalEvaluationStatus } from './interface';
+
+/**
+ * find the first evaluation that does not have a result
+ * */
+export function getOldProposalStatus({
```
mattcasey, line 9: why do we still need this?

lib/proposal/interface.ts
```diff
@@ -60,7 +64,17 @@ export type ProposalWithUsersLite = ProposalWithUsers & {
...
-  currentEvaluation?: Pick<ProposalEvaluation, 'title' | 'type'>;
+  evaluations: {
+    title: string;
+    type: ProposalEvaluationType;
+    id: string;
+    result: ProposalEvaluationResult | null;
+    index: number;
+  }[];
+  currentEvaluation?: Pick<ProposalEvaluation, 'title'> & {
+    step: ProposalEvaluationStep;
```
mattcasey, line 75: step does not belong in currentEvaluation, because rewards and draft is not an evaluation

mattcasey (changes requested):
'currentEvaluation' already means something in our codebase, it should not be used for draft/rewards. These two statuses are special cases that should be handled with some other model based on the context, so for example we could just have "currentStep" and "currentStatus" as fields, but dont use the word 'evaluation'

lib/proposal/mapDbProposalToProposal.ts
```diff
@@ -75,22 +81,50 @@ export function mapDbProposalToProposalLite({
...
+            }
+          : {
+              title: currentEvaluation.title,
+              step: currentEvaluation.type,
+              status: evaluationStatus
+            }
+        : undefined,
     currentEvaluationId: proposal.status !== 'draft' && proposal.evaluations.length ? currentEvaluation?.id : undefined,
-    evaluationType: evaluationWithOldType?.type || proposal.evaluationType,
+    evaluationType: currentEvaluation?.type || proposal.evaluationType,
```
mattcasey, line 123: do not use proposal.evaluationType

lib/proposal/mapDbProposalToProposal.ts
```diff
@@ -75,22 +81,50 @@ export function mapDbProposalToProposalLite({
...
+    })),
     permissions,
-    currentEvaluation: currentEvaluation
-      ? {
-          title: currentEvaluation.title,
-          type: currentEvaluation.type
-        }
-      : undefined,
+    currentEvaluation:
+      proposal.status === 'draft'
```
mattcasey, line 103: draft state has no currentEvaluation

mattcasey (approved):

PR #3367, opened 2024-01-04
Fix vote creation logic so duplicates are assessed at evaluation level

motechFR:
### WHAT

Deduplicate vote creation

### WHY

We can have multiple proposal votes as long as it's only 1 per evaluation

### DIFF
--- a/lib/votes/createVote.ts
+++ b/lib/votes/createVote.ts
@@ -25,22 +25,22 @@ export async function createVote(vote: VoteDTO & { spaceId: string }): Promise<E
     context
   } = vote;
 
-  if (pageId) {
-    const page = await prisma.page.findUnique({
+  if (pageId && evaluationId) {
+    const page = await prisma.page.findUniqueOrThrow({
       where: {
         id: pageId
       },
       select: {
-        votes: true
+        proposal: {
+          select: {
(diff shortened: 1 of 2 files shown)


lib/votes/createVote.ts
```diff
@@ -25,22 +25,22 @@ export async function createVote(vote: VoteDTO & { spaceId: string }): Promise<E
...
       }
     });
 
-    if (!page) {
-      throw new PageNotFoundError(pageId);
-    }
-
-    if (context === 'proposal' && page.votes.some((v) => v.context === 'proposal')) {
-      throw new DuplicateDataError('A proposal vote has already been created for this page.');
+    if (page.proposal?.evaluations.find((ev) => ev.id === evaluationId)?.voteId) {
```
mattcasey, line 42: can we just query the proposalEvaluation directly instead?

mattcasey (approved):
approved but the query could be optimized

lib/votes/createVote.ts
mattcasey: i don't think it matters, but one new change is that we won't check if a page exists anymore unless it's a proposal

mattcasey (approved):

PR #3368, opened 2024-01-04
allow admin to edit all settings even when using a template

mattcasey:


components/proposals/ProposalPage/components/EvaluationSettingsSidebar/components/EvaluationStepSettings.tsx
```diff
@@ -25,12 +26,13 @@ type Props = {
 };
 
 export function EvaluationStepSettings({ evaluation, evaluationTemplate, isReviewer, onChange, readOnly }: Props) {
+  const isAdmin = useIsAdmin();
```
mattcasey, line 29: basically, 'readOnly' relies on the computed permissions, so admin is already handled. but we need to consider admin when considering template logic

motechFR (approved):
Tested & approved! (admin and non admin user)

PR #3369, opened 2024-01-04
Script: Add reviewers to old proposal evaluations

mattcasey:

### DIFF
--- a/scripts/migrations/2024_01_04_add_reviewers.ts
+++ b/scripts/migrations/2024_01_04_add_reviewers.ts
@@ -1,0 +1,49 @@
+import { prisma } from '@charmverse/core/prisma-client';
+
+async function migrate() {
+  const evaluations = await prisma.proposalEvaluation.findMany({
+    where: {
+      proposal: {
+        status: 'published',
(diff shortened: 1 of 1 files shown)


scripts/migrations/2024_01_04_add_reviewers.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+
+  console.log(`Found ${evaluations.length} evaluations`);
+
+  for (let evaluation of evaluations) {
+    const proposalReviewers = evaluation.proposal.reviewers;
+    if (proposalReviewers.length > 0) {
+      await prisma.proposalReviewer.createMany({
+        data: proposalReviewers.map((reviewer) => ({
+          ...reviewer,
```
motechFR, line 40: I think we need to drop the ID from this spread

motechFR (approved):

scripts/migrations/2024_01_04_add_reviewers.ts
mattcasey: good point!

PR #3373, opened 2024-01-04
Fix state for proposals

motechFR:
### WHAT

Fix custom properties on proposals

### WHY

So that they work


components/proposals/ProposalPage/components/ProposalProperties/CustomPropertiesAdapter.tsx
```diff
@@ -25,7 +26,10 @@ export function CustomPropertiesAdapter({ proposal, onChange, readOnly, readOnly
   const mutator = usePropertiesMutator({ proposal, onChange });
 
   useEffect(() => {
-    setBoardProposal(proposal);
+    setBoardProposal({
+      ...proposal,
+      id: proposal.id ?? pageId
```
mattcasey, line 31: pageId !== proposal id, is there another option?

mattcasey: [edit] i noticed safwan recently added some more code where we use page id for proposal to useProposalsBoardAdapter. This is all broken as far as i can tell

motechFR: Fixed logic to pass down explicit proposal ID

PR #3375, opened 2024-01-05
allow duplicating a template

mattcasey:


components/proposals/components/NewProposalButton.tsx
```diff
@@ -78,6 +79,10 @@ export function NewProposalButton() {
     navigateToSpacePath(`/proposals/new`, { template: pageId });
   }
 
+  function duplicateTemplate(pageId: string) {
+    navigateToSpacePath(`/proposals/new`, { type: 'proposal_template', template: pageId });
```
mattcasey, line 83: tested and it works! why not? :)

PR #3378, opened 2024-01-05
Clean search-domain API for faster loads

valentinludu:
Clean `search-domain` and `search-name` API for faster loads.

Conclusions:
No token gates will be returned from space endpoints.
Places where we need token gates, will be loaded from the `token-gates` endpoint


msadura (approved):

mattcasey (changes requested):
can the frontend be changed so that we dont mutate the state of another api request? this can be overridden

valentinludu:
Done!

> can the frontend be changed so that we dont mutate the state of another api request? this is not foolproof and can be overridden

mattcasey (changes requested):
We still are kind of combining space and token gates into one object with i think will result in bugs. I would question if we even nee to return tokenGates in the spaces context anymore, and could get rid of the SpaceWithGates interface. This would highlight some other areas we need to address including these components that use `SpaceAccessGate: SpaceAccessGateWithSearch and RewardSignupButton. It seems like SpaceAccessGate ought to be able to just query token gates on its own?

charmClient/hooks/tokenGates.ts
```diff
@@ -9,8 +9,8 @@ import type { TokenGate, TokenGateWithRoles, TokenGateConditions } from 'lib/tok
 
 import { useDELETE, useGET, usePOST, usePUT } from './helpers';
 
-export function useGetTokenGates(spaceId: string) {
-  return useGET<TokenGateWithRoles[]>(spaceId ? '/api/token-gates' : null, { spaceId });
+export function useGetTokenGates(spaceId: string, spaceDomain?: string) {
```
mattcasey, line 12: i dont see a need to also query by spacedomain, please remove this and only use spaceId

mattcasey (changes requested):
I think there are other places where we get token gates, also please see if you can kill the SpaceWIthGates interface

valentinludu:
Killed it!
We did not need to load a space with gates anywhere.
The only place we need token gates list is the join page. Now, the search-domain and search-name responses are lighter and faster!

> I think there are other places where we get token gates, also please see if you can kill the SpaceWIthGates interface

charmClient/hooks/tokenGates.ts
valentinludu: Done

mattcasey (approved):

mattcasey (commented):
cleanup looks good

PR #3390, opened 2024-01-06
Proposal status and step column extras

Devorein:
### WHAT

1. Move to next step through step column
2. Multi-select for status 
3. Use the proposal status values for filter (and not proposal result)
4. Allow user to set step to "In Progress" for rubric and review
5. When Rewards is published, show the previous Step and Status
6. Disable the Status/Step/Reviewer options when multiple rows are selected but not all vali

### WHY

[Card](https://app.charmverse.io/charmverse/proposal-step-and-status-columns-extra-6382801826497115)


charmClient/apis/proposalsApi.ts
```diff
@@ -33,4 +34,20 @@ export class ProposalsApi {
   updateProposalEvaluation({ proposalId, ...payload }: UpdateEvaluationRequest) {
     return http.PUT(`/api/proposals/${proposalId}/evaluation`, payload);
   }
+
+  submitEvaluationResult({ proposalId, ...payload }: Omit<ReviewEvaluationRequest, 'decidedBy'>) {
+    return http.PUT(`/api/proposals/${proposalId}/submit-result`, payload);
+  }
+
+  updateProposalStatusOnly({ proposalId, ...payload }: { proposalId: string; newStatus: 'draft' | 'published' }) {
+    return http.PUT(`/api/proposals/${proposalId}/status-only`, payload);
```
mattcasey, line 43: this endpoint doesn't exist anymore, please use /publish and rename the method to match usePublishProposal

components/common/BoardEditor/components/properties/TagSelect/TagSelect.tsx
```diff
@@ -186,6 +186,7 @@ export function TagSelect({
   if (!isOpened) {
     return (
       <SelectPreviewContainer
+        className='SelectPreviewContainer'
```
mattcasey, line 189: using classes is an anti-pattern, which we should try really hard to avoid in our shared libs. please explain why this is needed for all instances of TagSelect, and vs using styled components or wrapping TagSelect where you need it with custom styles?

components/common/BoardEditor/focalboard/src/components/kanban/kanbanColumnHeader.tsx
```diff
@@ -13,8 +13,8 @@ import { proposalPropertyTypesList } from 'lib/focalboard/board';
 import type { BoardView } from 'lib/focalboard/boardView';
 import type { Card } from 'lib/focalboard/card';
 import { Constants } from 'lib/focalboard/constants';
-import { PROPOSAL_RESULT_LABELS } from 'lib/focalboard/proposalDbProperties';
-import type { ProposalEvaluationResultExtended } from 'lib/proposal/interface';
+import { PROPOSAL_STATUS_LABELS } from 'lib/focalboard/proposalDbProperties';
```
mattcasey, line 16: EVALUATION_STATUS_LABELS could be a better alternative, so its not confused with the old status

components/proposals/components/ProposalStepSelect.tsx
```diff
@@ -61,48 +82,42 @@ export function ProposalStepSelect({
...
+
     _options.forEach((option, index) => {
       option.disabled =
-        index >= currentEvaluationIndex ||
+        index === currentEvaluationIndex ||
         index < currentEvaluationIndex - 1 ||
-        (currentEvaluationStep === 'rewards' && currentEvaluationResult === 'pass');
+        index > currentEvaluationIndex + 1 ||
+        // If we are on the vote step, then we can only go back to the previous step
+        (currentEvaluationStep === 'vote' ? index >= currentEvaluationIndex : false);
```
mattcasey, line 92: the vote result must also be null, are we checking that?

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+              currentEvaluationStep,
+              result: 'pass'
+            });
+          } else if (evaluationId !== 'draft') {
+            await charmClient.proposals.goBackToEvaluationStep({
+              evaluationId,
+              proposalId
+            });
+          } else {
+            await charmClient.proposals.updateProposalStatusOnly({
```
mattcasey, line 94: should use goBackToStep. we no longer use "updateProposalStatusOnly" anywhere except one test file

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+  async function batchUpdateProposalSteps(
+    proposalsData: { evaluationId: string; proposalId: string; currentEvaluationStep: ProposalEvaluationStep }[],
+    moveForward: boolean
+  ) {
+    if (space) {
+      const { id: spaceId } = space;
+      try {
+        for (const { currentEvaluationStep, evaluationId, proposalId } of proposalsData) {
+          if (moveForward) {
+            await updateProposalStatus({
```
mattcasey, line 83: suggestion: rename this to "moveStatusForward" and below make a "moveStatusBackward" function, so that there's a consistent abstraction

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
mattcasey, line 1: better name: "useBatchUpdateProposalStatusOrStep" and then it should just export "updateStatuses" and "updateSteps"

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+    currentEvaluationStep: ProposalEvaluationStep;
+    result: ProposalEvaluationResult | null;
+  }) {
+    const { proposalId, evaluationId } = proposalData;
+    if (currentEvaluationStep === 'rewards' && result === 'pass') {
+      await charmClient.proposals.createProposalRewards({
+        proposalId
+      });
+    } else if (currentEvaluationStep === 'draft' && result === 'pass') {
+      await charmClient.proposals.updateProposalStatusOnly({
```
mattcasey, line 28: should call publishProposal instead

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+        proposalId,
+        newStatus: 'published'
+      });
+    } else if (evaluationId) {
+      if (result === null) {
+        await charmClient.proposals.goBackToEvaluationStep({
+          proposalId,
+          evaluationId
+        });
+      } else {
```
mattcasey, line 38: This function seems to be doing way too much. I think it could just do "moveStatusForawrd" logic, except for this one case here where we move an evaluation back if result is !== null. That's too confusing and arbitrary

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+        proposalId
+      });
+    } else if (currentEvaluationStep === 'draft' && result === 'pass') {
+      await charmClient.proposals.updateProposalStatusOnly({
+        proposalId,
+        newStatus: 'published'
+      });
+    } else if (evaluationId) {
+      if (result === null) {
+        await charmClient.proposals.goBackToEvaluationStep({
```
mattcasey, line 34: maybe a comment here could be good, like "unset the result of the current evaluation". when reading this at first i thought this case was about moving the step bckwards (i know that's partly myself to blame, and i'm open to other names for the method :)

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
```diff
@@ -0,0 +1,111 @@
...
+      const { id: spaceId } = space;
+      try {
+        for (const proposalData of proposalsData) {
+          await updateProposalStatus({
+            currentEvaluationStep,
+            proposalData,
+            result
+          });
+        }
+        await mutate(`/api/spaces/${spaceId}/proposals`);
```
mattcasey, line 67: suggestion: for the first time i feel like there's a good argument to have a success message, since it's not very obvious the table gets updated

components/common/BoardEditor/focalboard/src/components/kanban/kanbanColumnHeader.tsx
Devorein: Sounds good, actually. Its confusing otherwise.

components/proposals/components/ProposalStepSelect.tsx
Devorein: Now we do. Thnx for catching that.

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
Devorein: Updated. Thanks.

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
Devorein: Sounds good. Updated.

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
Devorein: Updated. All instances of `updateProposalStatusOnly` has been removed

components/proposals/hooks/useProposalUpdateStatusAndStep.ts
Devorein: But the status and step columns do get updated in the table as soon as we fetch the new proposals 🤔

PR #3403, opened 2024-01-09
Enable 2fa in Space Overview

valentinludu:
Main features:
- Admin can go to Settings -> Space Overview and turn on 2fa.
- Any user can go to Settings -> My Account Two factor auth section and start the process of setting up 2fa with their preferred authenticator app.
- Reset user recovery code
- Check the user qr code again

Keep in mind:
- Everything is under feature flag and not generally available
- Only switch for admins to enable 2fa on their space. In the next PR users will be forced to setup 2fa
- Users that setup half of the flow and do not finish it can, will see the same qr code.

Extra:
- Unit tests: for each method on each endpoint
- Storybook: on Settings -> My Accoount + Modals
- Reusable encryption/decryption methods with different keys


mattcasey:
The copy icon is not lined up here: 
image. I think it's the margin you're adding?

mattcasey:
I was able to set up OTP on my phone using authy, it even has our logo.. pretty nice. I don't know if i can test login using this yet

mattcasey (approved):

PR #3406, opened 2024-01-09
fix rewards application dialog not appearing when user wants to apply

mattcasey:

### DIFF
--- a/components/[pageId]/EditorPage/EditorPage.tsx
+++ b/components/[pageId]/EditorPage/EditorPage.tsx
@@ -5,7 +5,6 @@ import { useCallback, useEffect } from 'react';
 
 import { trackPageView } from 'charmClient/hooks/track';
 import ErrorPage from 'components/common/errors/ErrorPage';
-import { PageDialogGlobal } from 'components/common/PageDialog/PageDialogGlobal';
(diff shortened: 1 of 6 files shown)


components/common/PageLayout/PageLayout.tsx
```diff
@@ -225,6 +226,7 @@ function PageLayout({ children }: PageLayoutProps) {
               <PageContainer>
                 <HeaderSpacer />
                 {children}
+                <PageDialogGlobal />
```
mattcasey, line 229: i moved the global provider here, instead of .. pretty much all the children pages that use this component

components/[pageId]/EditorPage/EditorPage.tsx
```diff
@@ -126,9 +125,6 @@ export function EditorPage({ pageId: pageIdOrPath }: { pageId: string }) {
       return (
         <Box flexGrow={1} minHeight={0} /** add minHeight so that flexGrow expands to correct heigh */>
           <DocumentPage page={page} refreshPage={refreshPage} readOnly={readOnly} savePage={savePage} enableSidebar />
-
-          {/* needed to handle applications for rewards */}
-          <PageDialogGlobal />
```
Comment (mattcasey) [inline components/[pageId]/EditorPage/EditorPage.tsx:131]: this was added for DocumentPage but not DatabasePage

PR #3409, opened 2024-01-09
Fix/proposal evaluation

motechFR:
### WHAT

Use same calculation in UI as backend

### WHY

For proposal refresh, use evaluations logic in proposals as datasource


mattcasey (approved):
nice thanks

PR #3410, opened 2024-01-09
fix api docs for creting a comment

mattcasey:


motechFR (approved):
Approving with caveat to double check this is indeed what's currently returned

mattcasey:
> Approving with caveat to double check this is indeed what's currently returned

this is the input, not the output

PR #3419, opened 2024-01-10
Remove duplicate notification migration script

Devorein:

### DIFF
--- a/scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
+++ b/scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
@@ -9,21 +9,18 @@ async function removeDuplicateNotifications() {
     }
   })
 
-  const userIdCreatedAtNotificationsRecord: Record<string, typeof notificationsMetadata[0][]> = {};
+  const userIdCreatedAtNotificationsSet: Set<string> = new Set();
+  const notificationMetadataIds: string[] = [];
   notificationsMetadata.forEach(notificationMetadata => {
     const { userId, createdAt } = notificationMetadata;
     const key = `${userId}.${createdAt}`
-    if (!userIdCreatedAtNotificationsRecord[key]) {
-      userIdCreatedAtNotificationsRecord[key] = [];
+
+    if (userIdCreatedAtNotificationsSet.has(key)) {
+      notificationMetadataIds.push(notificationMetadata.id);
+    } else {
+      userIdCreatedAtNotificationsSet.add(key);
     }
-
-    userIdCreatedAtNotificationsRecord[key].push(notificationMetadata);
   })
-
-  const notificationMetadataIds = Object.values(userIdCreatedAtNotificationsRecord)
-    .filter(notifications => notifications.length > 1)
-    .map(notifications => notifications.slice(1).map(notification => notification.id).flat())
-    .flat();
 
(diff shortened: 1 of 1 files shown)


scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
```diff
@@ -0,0 +1,37 @@
+import { prisma } from "@charmverse/core/prisma-client";
+
+async function removeDuplicateNotifications() {
+  const notificationsMetadata = await prisma.userNotificationMetadata.findMany({
```
mattcasey, line 4: we have 27k records, just to save some time i'd probably filter for anything created from this year only

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  const userIdCreatedAtNotificationsRecord: Record<string, typeof notificationsMetadata[0][]> = {};
+  notificationsMetadata.forEach(notificationMetadata => {
+    const { userId, createdAt } = notificationMetadata;
+    const key = `${userId}.${createdAt}`
+    if (!userIdCreatedAtNotificationsRecord[key]) {
+      userIdCreatedAtNotificationsRecord[key] = [];
+    }
+
+    userIdCreatedAtNotificationsRecord[key].push(notificationMetadata);
```
mattcasey, line 20: suggestion: only append to the list if it exists. that way you only end up with a list of ids to delete. you could also just keep a map of whats been sent, and a single list of ids to delete.

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
```diff
@@ -0,0 +1,37 @@
...
+  })
+
+  const notificationMetadataIds = Object.values(userIdCreatedAtNotificationsRecord)
+    .filter(notifications => notifications.length > 1)
```
mattcasey, line 24: at the moment this filter does nothing ,right? every list has at least one element

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
```diff
@@ -0,0 +1,37 @@
...
+    .map(notifications => notifications.slice(1).map(notification => notification.id).flat())
+    .flat();
+
+  await prisma.userNotificationMetadata.deleteMany({
```
mattcasey, line 28: i think we also need to somehow find the proposalNotifications that link to these?

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
Devorein: this removes all the list where there are more than one member so that we can keep just a single one.

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
Devorein: Why do we need to do that?

scripts/migrations/2024_01_10_removeDuplicateNotifications.ts
Devorein: Good point. That would be easier. Updated.

PR #3421, opened 2024-01-10
Split structured proposal form migration script

Devorein:

### DIFF
--- a/scripts/migrations/2024_01_10_splitStructuredProposalForms.ts.ts
+++ b/scripts/migrations/2024_01_10_splitStructuredProposalForms.ts.ts
@@ -1,0 +1,112 @@
+import { prisma } from "@charmverse/core/prisma-client";
+import { SelectOptionType } from 'components/common/form/fields/Select/interfaces';
+import { createForm } from 'lib/form/createForm';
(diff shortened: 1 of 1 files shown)


scripts/migrations/2024_01_10_splitStructuredProposalForms.ts.ts
```diff
@@ -0,0 +1,114 @@
...
+              data: {
+                fieldId: newFormFieldIds[index]
+              }
+            })
+          ))).flat()
+        ])
+      }
+      existingFormIds.add(proposalTemplate.formId)
+    } catch (err) {
+      console.log(`Failed to split proposal template ${proposalTemplate.id}`)
```
mattcasey, line 106: this try/catch should probably occure outside the for loop, right? do u really want to continue if you tried to make a form but failed

mattcasey (approved):

scripts/migrations/2024_01_10_splitStructuredProposalForms.ts.ts
Devorein: Fair point, I will move it outside just to be safe

PR #3422, opened 2024-01-10
Script for resyncing boards with proposals as datasource

motechFR:
### WHAT

Bring boards back in sync

### WHY

Fix old permission creation


scripts/resyncBoardPermissions.ts
```diff
@@ -0,0 +1,97 @@
...
+        path: ['sourceType'],
+        equals: 'proposals'
+      }
+    }
+  });
+
+  const pages = await prisma.page.findMany({
+    where: {
+      id: {
+        in: proposalBoards.map(board => board.id)
```
mattcasey, line 21: i believe you could use `boardId` but i know this probably also works 🤣

mattcasey (approved):

PR #3428, opened 2024-01-11
Fix archive proposal action

Devorein:
### WHY

[Discussion](https://discord.com/channels/894960387743698944/1194431225348239520)


motechFR (approved):
This looks good, can you please add an E2E test?

Ideally a non admin member who is the proposal author:
- Create proposal in db directly
- Archive proposal
- Expect next step button to be disabled
// Check proposal is archived in DB

- Unarchive proposal
- Try moving proposal forward
// Check new current evaluation and also that archived is false

PR #3432, opened 2024-01-12
Enable updated column on proposals page and make sure we update the page

mattcasey:
We want to show 'updatedAt' on proposals, we also want it to be updated whenever anything changes on any of the sub-assets of the page


Devorein (approved):

mattcasey:
Thanks @Devorein it looks like i missed two things:
1) we re-generated default properties on proposal board all the time, you can't delete them
2) we had some hidden logic on which columns can be deleted, which i now just check for "__". this code is crazy complicated, lots of magic and i want to avoid so much configuration overhead. this is a pattern we'll be reusing and its not even consistent between rewards and proposals

Devorein (approved):

PR #3433, opened 2024-01-12
Remove extra logic to detect linked page

Devorein:
### WHY

1. Remove extra logic to detect linked page
2. Store applied page diffs to document


mattcasey (approved):
thanks!

PR #3439, opened 2024-01-12
add an info alert when looking at a page that is synced elsewhere

mattcasey:
Users dont know if they're looking a t the original or not

image


components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:1]: I moved all the banners inside of document-print-container, to fix a double-scrollbar issue in dialog mode:
image


It doesn't seem to have any adverse effects. The banners scroll with the page, but i think we always want that anyway

PR #3444, opened 2024-01-15
Fix proposal as a source board evaluation properties 

Devorein:
### WHY

image


components/common/BoardEditor/focalboard/csvExporter/__tests__/csvExporter.spec.ts
```diff
@@ -1,15 +1,17 @@
-import type { Block } from '@charmverse/core/prisma';
-import { v4 as uuid } from 'uuid';
+import type { Block, ProposalEvaluation } from '@charmverse/core/prisma';
+import { v4 as uuid, v4 } from 'uuid';
```
valentinludu, line 2: I don't think you need `v4` twice. Probably should stick with `v4 as uuid` naming since is more clear what it is.

PR #3447, opened 2024-01-16
Update proposal as a source board evaluation properties

Devorein:

### DIFF
--- a/__e2e__/database/databaseProposalsAsSource.spec.ts
+++ b/__e2e__/database/databaseProposalsAsSource.spec.ts
@@ -87,9 +87,10 @@ test.describe.serial('Database with proposals as datasource', async () => {
 
     await pagesSidebar.pagesSidebarSelectAddDatabaseButton.click();
 
-    await databasePage.page.waitForTimeout(500);
-
-    await expect(pagesSidebar.databasePage).toBeVisible();
+    await pagesSidebar.databasePage.waitFor({
+      state: 'visible',
+      timeout: 5000
+    });
 
     // Initialise the new database
     await expect(databasePage.selectProposalsAsSource()).toBeVisible();
--- a/__e2e__/database/databaseProposalsAsSource.spec.ts
+++ b/__e2e__/database/databaseProposalsAsSource.spec.ts
@@ -87,6 +87,8 @@ test.describe.serial('Database with proposals as datasource', async () => {
 
     await pagesSidebar.pagesSidebarSelectAddDatabaseButton.click();
 
+    await databasePage.page.waitForTimeout(500);
+
     await expect(pagesSidebar.databasePage).toBeVisible();
 
     // Initialise the new database
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
@@ -26,7 +26,7 @@ import {
   makeSelectViewCardsSortedFilteredAndGrouped,
   sortCards
 } from 'components/common/BoardEditor/focalboard/src/store/cards';
-import { useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
+import { useAppDispatch, useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
 import { Button } from 'components/common/Button';
 import LoadingComponent from 'components/common/LoadingComponent';
 import { webhookEndpoint } from 'config/constants';
@@ -45,6 +45,7 @@ import { CardFilter } from 'lib/focalboard/cardFilter';
 
 import mutator from '../mutator';
 import { addCard as _addCard, addTemplate } from '../store/cards';
+import { initialDatabaseLoad } from '../store/databaseBlocksLoad';
 import { updateView } from '../store/views';
 import { Utils } from '../utils';
 
@@ -103,7 +104,7 @@ function CenterPanel(props: Props) {
   const { pages, refreshPage } = usePages();
   const { membersRecord } = useMembers();
   const localViewSettings = useLocalDbViewSettings(activeView?.id);
-
+  const dispatch = useAppDispatch();
   const isEmbedded = !!props.embeddedBoardPath;
   const boardPageType = boardPage?.type;
   // for 'linked' boards, each view has its own board which we use to determine the cards to show
@@ -494,11 +495,20 @@ function CenterPanel(props: Props) {
   );
 
   // refresh proposals as a source
+  const pageId = props.page?.id;
   useEffect(() => {
-    if (currentRootPageId && activeBoard?.fields.sourceType === 'proposals' && activeBoard?.id === currentRootPageId) {
-      updateProposalSource({ pageId: currentRootPageId });
+    if (
+      pageId &&
+      currentRootPageId &&
+      activeBoard?.fields.sourceType === 'proposals' &&
+      activeBoard?.id === currentRootPageId
+    ) {
+      updateProposalSource({ pageId: currentRootPageId }).then(() => {
+        // Refetch database after updating proposal source board, otherwise the UI will be out of sync
+        dispatch(initialDatabaseLoad({ pageId }));
+      });
     }
-  }, [currentRootPageId, activeBoard?.id]);
+  }, [currentRootPageId, activeBoard?.id, pageId]);
 
   const isLoadingSourceData = !activeBoard && (!views || views.length === 0);
   const readOnlyTitle = activeBoard?.fields.sourceType === 'proposals';
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
@@ -494,21 +494,14 @@ function CenterPanel(props: Props) {
     (_url, { arg }: Readonly<{ arg: { pageId: string } }>) => charmClient.updateProposalSource(arg)
   );
 
-  // refresh proposals as a source
-  const pageId = props.page?.id;
   useEffect(() => {
-    if (
-      pageId &&
-      currentRootPageId &&
-      activeBoard?.fields.sourceType === 'proposals' &&
-      activeBoard?.id === currentRootPageId
-    ) {
+    if (currentRootPageId && activeBoard?.fields.sourceType === 'proposals' && activeBoard?.id === currentRootPageId) {
       updateProposalSource({ pageId: currentRootPageId }).then(() => {
         // Refetch database after updating proposal source board, otherwise the UI will be out of sync
-        dispatch(initialDatabaseLoad({ pageId }));
+        dispatch(initialDatabaseLoad({ pageId: currentRootPageId }));
       });
     }
-  }, [currentRootPageId, activeBoard?.id, pageId]);
+  }, [currentRootPageId, activeBoard?.id]);
 
   const isLoadingSourceData = !activeBoard && (!views || views.length === 0);
   const readOnlyTitle = activeBoard?.fields.sourceType === 'proposals';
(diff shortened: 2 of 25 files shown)


__e2e__/database/databaseProposalsAsSource.spec.ts
```diff
@@ -87,6 +87,8 @@ test.describe.serial('Database with proposals as datasource', async () => {
 
     await pagesSidebar.pagesSidebarSelectAddDatabaseButton.click();
 
+    await databasePage.page.waitForTimeout(500);
```
mattcasey, line 90: ideally you could just wait for the network request to succeed using waitForResponse or something else

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -494,11 +495,20 @@ function CenterPanel(props: Props) {
...
-    if (currentRootPageId && activeBoard?.fields.sourceType === 'proposals' && activeBoard?.id === currentRootPageId) {
-      updateProposalSource({ pageId: currentRootPageId });
+    if (
+      pageId &&
+      currentRootPageId &&
+      activeBoard?.fields.sourceType === 'proposals' &&
+      activeBoard?.id === currentRootPageId
+    ) {
+      updateProposalSource({ pageId: currentRootPageId }).then(() => {
+        // Refetch database after updating proposal source board, otherwise the UI will be out of sync
```
mattcasey, line 507: thanks for adding a comment

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -145,13 +139,6 @@ function PropertyValueElement(props: Props) {
   } = props;
   const { trigger } = useUpdateProposalEvaluation({ proposalId: proposal?.id });
   const { showMessage } = useSnackbar();
-  const { rubricProposalIdsWhereUserIsEvaluator, isLoading: isLoadingReviewerStatus } =
```
mattcasey, line 148: i ran into this one recently, are we still filtering by permissions someplace else?

lib/focalboard/createCardsFromProposals.ts
```diff
@@ -130,6 +130,16 @@ export async function createCardsFromProposals({
     }
   });
 
+  const rubricEvaluationTitles: string[] = [];
```
mattcasey, line 133: this doesnt seem to get used?

lib/focalboard/generateResyncedProposalEvaluationForCard.ts
```diff
@@ -1,51 +1,52 @@
...
-    });
-
-    const uniqueReviewers = Object.keys(reviewersResults);
-
-    if (databaseProperties.proposalEvaluationAverage) {
-      cardProperties[databaseProperties.proposalEvaluationAverage.id] = allScores.average ?? '';
-    }
+  step: { id: string; title: string };
+}): Record<string, CardPropertyValue> {
+  const cardProperties = JSON.parse(JSON.stringify(properties)) as Record<string, CardPropertyValue>;
```
mattcasey, line 20: This is an extremely inefficient way to create a clone of something, is it really necessary?

lib/focalboard/generateResyncedProposalEvaluationForCard.ts
```diff
@@ -1,51 +1,52 @@
...
-import type { Block } from './block';
-import type { Card } from './card';
-import type { ExtractedDatabaseProposalProperties } from './extractDatabaseProposalProperties';
+import type { IPropertyTemplate } from './board';
+import type { CardPropertyValue } from './card';
 
 export function generateResyncedProposalEvaluationForCard({
-  cardProps,
-  databaseProperties,
+  properties,
```
mattcasey, line 8: this function should not be used to mutate 'properties' i dont know why we're cloning it each time,  you could just pass in an empty object to start.

But I think it would be much better if it simply returned the props that need to be set, rather than have it update the properties value that is passed in. It would lead to better calling code, its very hard to understand what's going on when you are constantly re-assigning a value from inside of a loop

lib/focalboard/generateResyncedProposalEvaluationForCard.ts
mattcasey: The way this method is called i don't think we need to do any kind of cloning of the properties in the first place. We always re-assign the value that's passed in with the result.

But a better way to rewrite this would be to only return the properties you want to change, and merge them in the calling code. Taht would be more transparent as to what's going on

lib/focalboard/setDatabaseProposalProperties.ts
```diff
@@ -56,25 +56,31 @@ export async function setDatabaseProposalProperties({
     select: {
       evaluations: {
         select: {
```
mattcasey, line 58: make sure to order evaluations correctly

lib/focalboard/setDatabaseProposalProperties.ts
```diff
@@ -139,9 +147,44 @@ export function getBoardProperties({
     boardProperties.push(stepProp);
   }
 
-  const evaluatedByProp = generateUpdatedProposalEvaluatedByProperty({ boardProperties });
-  const evaluationTotalProp = generateUpdatedProposalEvaluationTotalProperty({ boardProperties });
-  const evaluationAverageProp = generateUpdatedProposalEvaluationAverageProperty({ boardProperties });
+  for (const rubricStepTitle of rubricStepTitles) {
+    const evaluatedByProp = boardProperties.find(
+      (p) => p.type === 'proposalEvaluatedBy' && p.name === `${rubricStepTitle} (Evaluation reviewers)`
```
mattcasey, line 152: suggestion: save p.title in addition to p.name for this type of logic"(Evaluation reviewers)" seems pretty arbitrary - how hard will it be to change in the future?

scripts/migrations/2024_01_16_deleteProposalEvaluationProperties.ts
```diff
@@ -0,0 +1,43 @@
...
+  for (const proposalBoardBlock of proposalBoardBlocks) {
+    try {
+      await prisma.block.update({
+        where: {
+          id: proposalBoardBlock.id
+        },
+        data: {
+          fields: {
+            ...proposalBoardBlock.fields as unknown as BoardFields,
+            cardProperties: ((proposalBoardBlock.fields as unknown as BoardFields).cardProperties ?? []).filter(boardCardProperty => boardCardProperty.type !== "proposalEvaluatedBy" && boardCardProperty.type !== "proposalEvaluationAverage" && boardCardProperty.type !== "proposalEvaluationTotal")
```
mattcasey, line 30: im afraid to ask but do we need or should we remove the properties on individual cards?

lib/focalboard/setDatabaseProposalProperties.ts
```diff
@@ -139,9 +147,44 @@ export function getBoardProperties({
     boardProperties.push(stepProp);
   }
 
-  const evaluatedByProp = generateUpdatedProposalEvaluatedByProperty({ boardProperties });
-  const evaluationTotalProp = generateUpdatedProposalEvaluationTotalProperty({ boardProperties });
-  const evaluationAverageProp = generateUpdatedProposalEvaluationAverageProperty({ boardProperties });
+  for (const rubricStepTitle of rubricStepTitles) {
```
mattcasey, line 150: for consistency sake and readability it might make sense to move this logic into its own method

lib/focalboard/setDatabaseProposalProperties.ts
```diff
@@ -249,21 +268,34 @@ function generateUpdatedProposalStepProperty({
...
+  } else {
+    const existingOptions = existingProposalStepProperty.options.map((opt) => opt.value);
+    const newOptions = evaluationStepTitles.filter((title) => !existingOptions.includes(title));
+    return {
+      ...existingProposalStepProperty,
+      options: [
+        ...existingProposalStepProperty.options,
+        ...newOptions.map((title) => ({
+          color: 'propColorGray',
+          id: title,
```
mattcasey, line 293: id should be random shouldnt it? not a big deal probably

lib/focalboard/setDatabaseProposalProperties.ts
```diff
@@ -249,21 +268,34 @@ function generateUpdatedProposalStepProperty({
...
         id: title,
         value: title
       }))
-    })
-  };
-
-  return proposalStepProp;
+    };
+  } else {
+    const existingOptions = existingProposalStepProperty.options.map((opt) => opt.value);
```
mattcasey, line 285: why are we going to some extra effort to include existing options if we have the updated list of step titles?

lib/focalboard/updateCardsFromProposals.ts
```diff
@@ -140,19 +130,46 @@ export async function updateCardsFromProposals({
...
+  });
+
+  const boardFieldsProperties = (board.fields as unknown as BoardFields)?.cardProperties ?? [];
+  const newBoardProposalEvaluationPropertyIds = (
+    ((updatedBoardBlock.fields as unknown as BoardFields)?.cardProperties ?? []).filter((property) => {
+      return (
+        (property.type === 'proposalEvaluatedBy' ||
+          property.type === 'proposalEvaluationAverage' ||
+          property.type === 'proposalEvaluationTotal') &&
+        !boardFieldsProperties.find((boardProperty) => boardProperty.id === property.id)
```
mattcasey, line 161: use .some() for boolean results instead of .find()

__e2e__/database/databaseProposalsAsSource.spec.ts
Devorein: Made it more robust by making use of `waitFor` with checking visible state.

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
Devorein: yep you're correct. Even though both are same, `currentRootPageId` is already present there.

lib/focalboard/createCardsFromProposals.ts
Devorein: Great catch! Removed.

lib/focalboard/setDatabaseProposalProperties.ts
Devorein: Good catch again! Went through all the evaluations and made sure they are ordered in asc order based on their index

lib/focalboard/setDatabaseProposalProperties.ts
Devorein: Sure

lib/focalboard/setDatabaseProposalProperties.ts
Devorein: Correct, but there is no way we will have duplicate values here thus I went with having a title for the id. Plus there are already options in other boards with title as id so might need another migration to fix those.

lib/focalboard/setDatabaseProposalProperties.ts
Devorein: That is a good point, I thought to keep the options in the same order as before, but nobody would notice and it would probably be in the same order as before.

scripts/migrations/2024_01_16_deleteProposalEvaluationProperties.ts
Devorein: No they won't be removed, the visiblePropertyIds in the views won't get removed either. But it doesn't matter since they won't be displayed in the UI

lib/focalboard/generateResyncedProposalEvaluationForCard.ts
Devorein: It actually won't cause any issues if we mutate the original properties object directly, if we go with the second approach we need to make sure it's being merged with the field properties every time.

components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
Devorein: Confirmed with Chris and he mentioned its no longer necessary.

lib/focalboard/setDatabaseProposalProperties.ts
Devorein: Updated.

PR #3449, opened 2024-01-16
Space template proposals adjustments

motechFR:
### WHAT

Pass in workflows into importWorkspace pages to limit side effects

Fix copying the proposal evaluation data

### WHY

Easier maintainability. importWorkspacePages is already huge.


lib/templates/exportWorkspacePages.ts
```diff
@@ -173,7 +175,8 @@ export async function exportWorkspacePages({
           evaluations: {
             include: {
               reviewers: true,
-              rubricCriteria: true
+              rubricCriteria: true,
+              permissions: true
```
mattcasey, line 179: ah good catch, needed these for templates

lib/templates/importSpaceSettings.ts
```diff
@@ -90,9 +92,50 @@ export async function importSpaceSettings({
...
+        evaluations: workflow.evaluations.map((_eval) => {
+          const typedEval = _eval as WorkflowEvaluationJson;
+          const permissions = typedEval.permissions
+            .map((perm) => {
+              return {
+                operation: perm.operation,
+                roleId: perm.roleId ? oldNewRoleIdHashMap[perm.roleId] : undefined,
+                systemRole: perm.systemRole,
+                // Ignore user ids
+                userId: undefined
```
mattcasey, line 124: don't think you can just ignore it, we need to filter these out

mattcasey (approved):

PR #3451, opened 2024-01-16
Add missing confirmations for proposals

mattcasey:
- when changing status/step from the proposals table
- added confirmation when submitting rubric reviews

### DIFF
--- a/charmClient/apis/proposalsApi.ts
+++ b/charmClient/apis/proposalsApi.ts
@@ -47,7 +47,7 @@ export class ProposalsApi {
     return http.POST(`/api/proposals/${proposalId}/rewards`);
   }
 
-  goBackToEvaluationStep({ proposalId, ...payload }: { proposalId: string; evaluationId?: string }) {
+  goBackToStep({ proposalId, ...payload }: { proposalId: string; evaluationId?: string }) {
     return http.PUT(`/api/proposals/${proposalId}/back-to-step`, payload);
   }
 }
--- a/components/proposals/hooks/useBatchUpdateProposalStatusOrStep.ts
+++ b/components/proposals/hooks/useBatchUpdateProposalStatusOrStep.ts
@@ -2,12 +2,14 @@ import type { ProposalEvaluationResult } from '@charmverse/core/prisma-client';
 import { mutate } from 'swr';
 
 import charmClient from 'charmClient';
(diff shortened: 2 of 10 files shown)


components/proposals/hooks/useBatchUpdateProposalStatusOrStep.ts
```diff
@@ -86,7 +107,7 @@ export function useBatchUpdateProposalStatusOrStep() {
               result: 'pass'
             });
           } else {
-            await charmClient.proposals.goBackToEvaluationStep({
+            await charmClient.proposals.goBackToStep({
```
mattcasey, line 110: renamed this to be standard with other places we just call it goBackToStep

charmClient/apis/proposalsApi.ts
```diff
@@ -47,7 +47,7 @@ export class ProposalsApi {
     return http.POST(`/api/proposals/${proposalId}/rewards`);
   }
 
-  goBackToEvaluationStep({ proposalId, ...payload }: { proposalId: string; evaluationId?: string }) {
+  goBackToStep({ proposalId, ...payload }: { proposalId: string; evaluationId?: string }) {
```
mattcasey, line 50: renamed this to be standard with other places we just call it goBackToStep

components/proposals/hooks/useBatchUpdateProposalStatusOrStep.ts
```diff
@@ -31,11 +33,14 @@ export function useBatchUpdateProposalStatusOrStep() {
       // the evaluationId is the current evaluation id,
       // technically its not going back to the previous step, but only the result is being unset
       if (result === null) {
-        await charmClient.proposals.goBackToEvaluationStep({
+        await charmClient.proposals.goBackToStep({
           proposalId,
           evaluationId
         });
       } else {
+        const { confirmed } = await showConfirmation({
```
Devorein: Is this required? It's blocking from changing the status

Devorein (approved):

components/proposals/hooks/useBatchUpdateProposalStatusOrStep.ts
Devorein: Made a few more changes to keep the confirmation message consistent with what we show in the evaluation sidebar

PR #3452, opened 2024-01-17
Check permissions for each evaluation step

mattcasey:
Bug fix: We need to check your reviewer permissions for each evaluation step, not just the current one

I also included some more cleanup, removing some fields on ProposalWithUserAndRubric that we no longer need.


components/proposals/hooks/useProposals.tsx
```diff
@@ -59,27 +59,27 @@ export function ProposalsProvider({ children }: { children: ReactNode }) {
     [mutateProposals]
   );
 
-  const refreshProposal = useCallback(
-    async (proposalId: string) => {
-      const proposal = await charmClient.proposals.getProposal(proposalId);
-      mutateProposals((data) => {
-        const proposalList = data ?? [];
-        const proposalIndex = proposalList.findIndex((p) => p.id === proposalId);
+  // const refreshProposal = useCallback(
```
mattcasey: removed this, since it couples the result of getting an individual proposal with the types we use for proposal list for just a little extra optimization

pages/api/proposals/[id]/index.ts
```diff
@@ -62,35 +63,33 @@ async function getProposalController(req: NextApiRequest, res: NextApiResponse<P
   if (!proposal) {
     throw new NotFoundError();
   }
-  const proposalPermissions = await permissionsApiClient.proposals.computeProposalPermissions({
+  const permissionsByStep = await permissionsApiClient.proposals.computeAllProposalEvaluationPermissions({
```
Comment (mattcasey) [inline pages/api/proposals/[id]/index.ts:66]: get permissions for all steps, not just the current one

PR #3461, opened 2024-01-18
Additional blocks count script

Devorein:
### WHY

Additional space block quotas to spaces


mattcasey (approved):
looks good, comments:
- generally i see us put the configuration at the top, outside the function so it's easy to find
- instead of createMany, shouldn't it be create? dunno if it matters

PR #3481, opened 2024-01-22
Log all errors for magic links

valentinludu:


components/invite/page/hooks/useMagicLink.ts
```diff
@@ -25,9 +25,9 @@ export function useMagicLink({ error: ssrError }: { error?: 'error_invalid_page_
         if ((error as any)?.code === 'auth/invalid_email') {
           setStatus('error_invalid_email');
         } else {
-          log.error('Error requesting firebase magic link', { error });
           setStatus('error');
         }
+        log.error('Error requesting firebase magic link', { error });
```
mattcasey, line 30: please make the other case a warning, errors are things that we should fix

components/login/components/LoginButton.tsx
```diff
@@ -142,6 +143,7 @@ function LoginHandler(props: DialogProps) {
         onClose();
         setLoginMethod(null);
       } catch (err) {
+        log.error(`Error sending magic link in login button.`, { error: err });
```
mattcasey, line 146: this could probably be in handleLoginError, so it handles other places too

components/settings/account/components/NewIdentityModal.tsx
```diff
@@ -68,14 +69,14 @@ export function NewIdentityModal({ isOpen, onClose }: Props) {
...
       sendingMagicLink.current = true;
-      // console.log('Handling magic link request');
       try {
         await requestMagicLinkViaFirebase({ email, connectToExistingAccount: true });
         showMessage(`Magic link sent. Please check your inbox for ${email}`, 'success');
         onClose();
         setIdentityToAdd(null);
       } catch (err) {
         showMessage((err as any).message ?? 'Something went wrong', 'error');
+        log.error(`Error sending magic link in identity modal for user ${user?.id}.`, { error: err });
```
mattcasey, line 79: you dont need to include user.id on the frontend, it's added at a global level. But if u did include userId you should add it as an attribute, like error

mattcasey (approved):
left some suggestions

components/invite/page/hooks/useMagicLink.ts
valentinludu: Done

components/login/components/LoginButton.tsx
valentinludu: Done

components/settings/account/components/NewIdentityModal.tsx
valentinludu: Done

PR #3485, opened 2024-01-23
Feat/external credentials

motechFR:
### WHAT

Load external credentials into the app

### WHY

So users can see the universe of credentials available


background/ceramicServer/run-deploy.ts
```diff
@@ -23,7 +23,16 @@ async function bootstrapGqlServer() {
...
+  log.info('[server] Exiting process...');
+  process.exit(0);
+}
+
+process.on('SIGINT', cleanup);
+process.on('SIGTERM', cleanup);
+
 process.on('SIGTERM', async () => {
-  (await server).stop();
   log.info('GraphQL Composedb server stopped');
```
mattcasey, line 37: i think this callback is redundant now

components/settings/credentials/hooks/credentialHooks.ts
```diff
@@ -2,10 +2,10 @@ import type { CredentialTemplate } from '@charmverse/core/prisma-client';
 
 import type { MaybeString } from 'charmClient/hooks/helpers';
 import { useGET } from 'charmClient/hooks/helpers';
-import type { PublishedSignedCredential } from 'lib/credentials/queriesAndMutations';
+import type { EASAttestationFromApi } from 'lib/credentials/external/getExternalCredentials';
 
 export function useGetUserCredentials({ userId }: { userId: MaybeString }) {
-  return useGET<PublishedSignedCredential[]>(userId ? `/api/credentials` : null, { userId });
+  return useGET<EASAttestationFromApi[]>(userId ? `/api/credentials` : null, { userId });
```
mattcasey, line 8: these hooks seem like they belong in a charmClient/hooks/ceramic.ts file. That's where we put all our calls to the api

components/members/components/MemberProfile/components/UserCredentials/UserCredentialRow.tsx
```diff
@@ -1,23 +1,39 @@
...
-import type { PublishedSignedCredential } from 'lib/credentials/queriesAndMutations';
+import { useGetEnsName } from 'hooks/useGetEnsName';
+import type { EASAttestationFromApi } from 'lib/credentials/external/getExternalCredentials';
+import { trackedSchemas } from 'lib/credentials/external/schemas';
+import { shortWalletAddress } from 'lib/utilities/blockchain';
+
+export function UserCredentialRow({ credential }: { credential: EASAttestationFromApi }) {
+  const logo = credential.type === 'internal' ? '/images/logo_black_transparent.64.png' : '/images/logos/eas.png';
+
+  const { ensname } = useGetEnsName({ address: credential.attester });
```
mattcasey, line 15: this hook requires the user to have a connected wallet to work. i'm not sure but i think it requires the user to be on mainnet too? we should probably be getting the ENS on the backend.. and store it so we dont have to make calls all the time, but that part is not really an issue at the moment.

lib/credentials/getAllUserCredentials.ts
```diff
@@ -0,0 +1,40 @@
...
+    })
+    .then((_wallets) => _wallets.map((w) => w.address));
+
+  if (!wallets.length) {
+    return [];
+  }
+
+  const allCredentials = await Promise.all([
+    getCharmverseCredentialsByWallets({ wallets }),
+    getAllOnChainAttestations({ wallets: testWallets })
```
mattcasey, line 36: did you mean to leave testWallets in here?

lib/credentials/templates.ts
```diff
@@ -1,11 +1,10 @@
-import type { CredentialTemplate, CredentialEventType, Space } from '@charmverse/core/prisma-client';
+import type { CredentialTemplate } from '@charmverse/core/prisma-client';
 import { prisma } from '@charmverse/core/prisma-client';
 import { stringUtils } from '@charmverse/core/utilities';
+import { urPK } from '@mui/x-date-pickers';
```
mattcasey, line 4: looks like an accidental import

mattcasey (approved):
left comments. .i think we can save loading ens on the backend for later but if we dont do that i want to make sure chris/alex are OK with not always loading ENS

PR #3487, opened 2024-01-23
only create credentials wallet when we need it

mattcasey:

### DIFF
--- a/lib/credentials/queriesAndMutations.ts
+++ b/lib/credentials/queriesAndMutations.ts
@@ -7,8 +7,6 @@ import { credentialsWalletPrivateKey, graphQlServerEndpoint } from 'config/const
 
 import type { ProposalCredential } from './schemas';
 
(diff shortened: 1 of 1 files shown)


motechFR (approved):
We need an exit clause if not exists here too

l:178 - lib/credentials/issueProposalCredentialsIfNecessary.ts

mattcasey:
> We need an exit clause if not exists here too
> 
> l:178 - lib/credentials/issueProposalCredentialsIfNecessary.ts

Not really sure what u mean by this, i think that code triggers something where we need the wallet key? but it wont be triggered unless i'm using credentials, afaict so i will leaveit alone for now. I *do* think its fine to throw an error if an endpoint is being called that requires the env var to exist for the main process to succeed. BUt in this case, i was trying to create a proposal and had no credentials configured either in my space or for that proposal

PR #3488, opened 2024-01-23
Feat/db table polish

msadura:
Polished DB tables (including proposals, rewards and all dbs in CV)
- fixed rows length to be consistent when in non-wrapping mode
- table does not jump when fields open
- all the "editable" cells in table have convenient popup inputs + all different kind of selects look similar

Screenshot 2024-01-23 at 23 58 37
Screenshot 2024-01-23 at 23 58 20
Screenshot 2024-01-23 at 23 58 51


mattcasey (approved):

mattcasey:
so much better, thank you

PR #3492, opened 2024-01-24
Add the option to set the template for proposal milestones (rewards)

mattcasey:
i want to see if i can also fix the issue of selecting a template for milestones (currently broken in prod)

Settings appear at the bottom of teh sidebar:
image


msadura (approved):
Looks good, I think it is fair enough to keep rewardsTemplateId in fields

PR #3493, opened 2024-01-24
Relation property for boards

Devorein:
### WHY

image

image

image

### TODO

- [ ] Handle syncing card values with connected board
- [ ] Connecting with rewards & proposals boards
- [ ] Adding relation property to connected board (This is done but without the two above it is incomplete)


msadura:
Screenshot 2024-01-24 at 12 43 56
 
this looks bad and wrong. Icon does not fit default row height and extends row. I've spend over a day trying to fix how the table looks like and before I merged my pr we have another weird looking component. Can we start looking into this in more holistic way and note how the whole tale behaves ratheer than adding yet another copy-pasted + customized autocomplete tat looks wrong?

msadura:
I would recommend to reuse popup field I intruduced in https://github.com/charmverse/app.charmverse.io/pull/3488 or if your pr will be merged first I will fix it and make cosistent with rest of table fields

hooks/useRelationProperties.ts
```diff
@@ -0,0 +1,29 @@
+import { useMemo } from 'react';
+
+import { makeSelectBoard } from 'components/common/BoardEditor/focalboard/src/store/boards';
+import { useAppSelector } from 'components/common/BoardEditor/focalboard/src/store/hooks';
+import { getRelationPropertiesCardsRecord } from 'lib/focalboard/getRelationPropertiesCardsRecord';
+import type { PublicPageResponse } from 'lib/pages';
+
+import { usePages } from './usePages';
+
+export function useRelationProperties({ page, boardId }: { boardId?: string; page?: PublicPageResponse['page'] }) {
```
mattcasey, line 10: doesn't look like we ever actually pass in boardId

components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -65,9 +66,17 @@ export interface DocumentPageProps {
   readOnly?: boolean;
   insideModal?: boolean;
   enableSidebar?: boolean;
+  relationPropertiesCardsRecord?: PageListItemsRecord;
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:69]: there's no reason to pass this in from outside CardDetailProperties as far as i can tell. It would greatly simplify if you can just call the hook inside of that

lib/focalboard/getRelationPropertiesCardsRecord.ts
```diff
@@ -0,0 +1,46 @@
...
+import { isTruthy } from 'lib/utilities/types';
+
+import type { Board } from './board';
+
+export function getRelationPropertiesCardsRecord({
+  activeBoard,
+  pages
+}: {
+  activeBoard: Board;
+  pages:
```
mattcasey, line 14: please force pages to be a certain structure. this overcomplicates the logic inside

components/common/BoardEditor/focalboard/src/mutator.ts
```diff
@@ -651,6 +651,19 @@ export class Mutator {
     return this.updateBlocks(newBlocks, oldBlocks, description);
   }
 
+  async updateProperty(board: Board, propertyId: string, updatedProperty: IPropertyTemplate) {
+    const newBoard = createBoard({ block: board });
+    const cardPropertyIndex = board.fields.cardProperties.findIndex((o: IPropertyTemplate) => o.id === propertyId);
+
+    if (cardPropertyIndex === -1) {
+      return;
```
mattcasey, line 659: should throw an error

components/common/BoardEditor/focalboard/src/widgets/menu/RelationPropertyMenu.tsx
```diff
@@ -0,0 +1,204 @@
...
+}) {
+  const { pages } = usePages();
+  const selectedPage = relationData ? pages[relationData.boardId] : null;
+
+  if (!selectedPage || !relationData) {
+    return null;
+  }
+
+  return (
+    <Box sx={sx}>
```
mattcasey, line 117: MenuItem should be the imediate child of Menu, this Box messes with the normal MUI css

components/common/BoardEditor/focalboard/src/widgets/menu/RelationPropertyMenu.tsx
```diff
@@ -0,0 +1,204 @@
...
+          setShowSelectDatabaseMenu?.(true);
+        }}
+        disabled={!setShowSelectDatabaseMenu}
+      >
+        <Typography color={setShowSelectDatabaseMenu ? '' : 'secondary'} mr={3}>
+          Related to
+        </Typography>
+        <Stack
+          sx={{
+            flexDirection: 'row',
```
mattcasey, line 129: these are props on Stack, no need for sx prop

theme/focalboard/focalboard.main.scss
```diff
@@ -14,7 +14,6 @@
   flex-grow: 1;
 
   * {
-    box-sizing: border-box;
```
mattcasey, line 17: +1 but make sure this doesnt break anything and all the whitespace is still good on databases... this is a huge change

pages/api/blocks/index.ts
```diff
@@ -257,18 +257,186 @@ async function createBlocks(req: NextApiRequest, res: NextApiResponse<Omit<Block
...
+      const dbBlock = dbBlocks.find((b) => b.id === block.id);
+      const spaceId = block.spaceId ?? dbBlock?.spaceId;
       const { error } = await hasAccessToSpace({ userId: req.session.user.id, spaceId });
       if (error) {
         throw new UnauthorisedActionError();
       }
     })
   );
 
+  // const boardBlocks = blocks.filter((block) => block.type === 'board');
```
mattcasey, line 297: can we just delete this?

hooks/useRelationProperties.ts
Devorein: Good catch, I used it in some places then removed it but forgot to remove it from this hook

lib/focalboard/getRelationPropertiesCardsRecord.ts
Devorein: Good point, removed.

components/common/BoardEditor/focalboard/src/widgets/menu/RelationPropertyMenu.tsx
Devorein: fair point, updated

theme/focalboard/focalboard.main.scss
Devorein: I've tested it out pretty much everywhere, haven't noticed anything obvious that stood out. Should be good to go.

pages/api/blocks/index.ts
Devorein: Yep, this has been deleted and moved to a separate endpoint.

PR #3499, opened 2024-01-24
Make loading more fault tolerant if ceramic is down

motechFR:
### WHAT

If Ceramic is down, shouldn't affect EAS and vice-versa


mattcasey (approved):
thx

PR #3503, opened 2024-01-24
Setup ceramic staging instance

motechFR:
### WHAT

Setup staging env for standalone ceramic node


.ebstalk.apps.env/ceramic.env
```diff
@@ -0,0 +1,10 @@
+# Format of this file is <env_var>=<value> key/value pair
+
+DATABASE_URL="{{pull:secretsmanager:/io.cv.app/prd/db:SecretString:database_url}}"
+DD_API_KEY="{{pull:secretsmanager:/io.cv.app/shared/datadog:SecretString:dd_api_key}}"
+DD_AGENT_HOST="datadog-agent"
```
mattcasey, line 5: i don't think we need datadog running on this server

.ebextensions_ceramic/01_env_health_ignore_400.config
```diff
@@ -0,0 +1,13 @@
+option_settings:
```
mattcasey, line 1: please only add the configs that are necessary. we added some of these specifcally for webapp, they might be helpful but i would like to be sure they're worth it

mattcasey: i'm not sure if we need this one, filestystem.config, autoscaling.config, or targetgrp.config.

mattcasey (changes requested):
few minor tweaks otherwise looks good? do we need any special config in the webapp to talk to this instance?

mattcasey (approved):

PR #3504, opened 2024-01-25
Bulk archive proposals

mattcasey:
One thing i wasnt sure how to do is have the dropdown close when you click "Archive or "Unarchive". It seems for all other cases we want to leave the dropdown open? Feels a little weird that we dont hide it sometimes tho


components/common/BoardEditor/focalboard/src/components/viewHeader/viewHeaderRowsMenu/components/PropertyMenu.tsx
```diff
@@ -72,33 +63,25 @@ export function PropertyMenu({
...
       </Tooltip>
 
       <Menu
         anchorEl={ref.current}
         open={popupState.isOpen}
-        style={{
-          position: 'relative',
-          top: -32,
-          height: '100%'
-        }}
```
mattcasey, line 87: I fixed the dropdowns on Firefox by removing this rule. I think it looks better having the menu below the element, anyway

Devorein (approved):

components/proposals/components/ProposalsHeaderRowsMenu.tsx
```diff
@@ -0,0 +1,271 @@
...
+  const { isStepDisabled, isStatusDisabled, isReviewersDisabled } = useMemo(() => {
+    const checkedPages = checkedIds.map((id) => pages[id]).filter(isTruthy);
+    const firstProposal = proposalsMap[checkedPages[0]?.proposalId ?? ''];
+    const _isReviewerDisabled = checkedPages.some((checkedPage) => {
+      const proposal = proposalsMap[checkedPage.proposalId ?? ''];
+      return (
+        !proposal ||
+        proposal.archived ||
+        proposal.currentStep.step === 'draft' ||
+        proposal.currentStep.step === 'feedback'
```
Devorein, line 198: This should be removed since we allow updating reviewers on feedback

Devorein (changes requested):

components/proposals/components/ProposalsHeaderRowsMenu.tsx
mattcasey: thats a different PR? :)

PR #3509, opened 2024-01-25
Set primary wallet & use during credential issue process

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/set-primary-credential-wallet-in-my-account-919083490366847)


pages/api/wallets/primary.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler
+  .use(requireUser)
+  .get(getPrimaryWalletController)
+  .use(requireKeys(['walletId']))
+  .put(setPrimaryWalletController);
+
+async function getPrimaryWalletController(req: NextApiRequest, res: NextApiResponse<UserWallet | null>) {
```
motechFR, line 17: Suggestion: Let's have this under /api/profile/primary-wallet

pages/api/wallets/primary.ts
```diff
@@ -0,0 +1,63 @@
...
+  const userId = req.session.user.id;
+
+  const primaryWallet = await prisma.userWallet.findFirst({
+    where: {
+      userId,
+      primary: true
```
motechFR, line 23: Regarding the model, we might end up with multiple primary wallets accidentally.

Instead, how about a primaryWallet on the User model which is a relation to UserWallet, and Cascade setNull ?

pages/api/wallets/primary.ts
Devorein: Good point, updated

pages/api/wallets/primary.ts
Devorein: Very good idea. Updated app code and core schema

PR #3510, opened 2024-01-25
add ability to add a workflow or reward template from proposal page

mattcasey:
Added "+ Add new" option to workflow and reward template dropdowns when editing a proposal:

image

Eventually the plan is to show the reward template modal from here, with some pre-defined defaults. but for now it just sends the user to the rewards page


components/common/BoardEditor/components/properties/TagSelect/TagSelect.tsx
```diff
@@ -201,7 +206,7 @@ export function TagSelect({
       forcePopupIcon={false}
       displayType={displayType}
       fluidWidth={fluidWidth}
-      popupField={displayType === 'table'}
+      disablePopper={showInsidePopup}
```
mattcasey, line 209: i thot this was name a bit more informative than 'popupField'

PR #3513, opened 2024-01-25
allow converting posts and pages to proposals

mattcasey:


lib/permissions/forum/__tests__/computePostPermissions.spec.ts
```diff
@@ -205,89 +204,6 @@ describe('computePostPermissions - with editable by author policy', () => {
   });
 });
 
-describe('computePostPermissions - with proposal permission filtering policy', () => {
```
mattcasey, line 208: chris was ok removing this additional logic about converted posts

PR #3523, opened 2024-01-26
Fix Proposal rewards override

valentinludu:
We need to make sure that we compare valid draftIds. 
In some cases draftId is an empty string.


mattcasey:
draftId should always be a value, afaict.

PR #3530, opened 2024-01-28
Editor context

mattcasey:


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -449,30 +383,102 @@ function DocumentPage({ insideModal = false, page, savePage, readOnly = false, e
...
+                    disabledViews={isStructuredProposal ? ['suggestions'] : []}
+                    onChangeWorkflow={onChangeWorkflow}
+                    onChangeRewardTemplate={onChangeRewardTemplate}
+                    isProposalTemplate={page.type === 'proposal_template'}
+                  />
+                )}
+              </CardPropertiesWrapper>
               {proposal && proposal.formId ? (
-                <>
-                  {documentPageContent}
```
Comment (valentinludu) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:454]: Before `documentPageContent` was rendered only if `proposal && proposal.formId` was defined. And now it has moved up without this check.
Don't you need to check proposal is defined before rendering the content you added at line 386 ?

components/proposals/ProposalPage/NewProposalPage.tsx
```diff
@@ -417,40 +326,93 @@ export function NewProposalPage({
...
-                            return {
-                              ...formAnswer,
-                              value: updatedFormField?.value ?? formAnswer.value
-                            };
-                          })
-                        });
-                        onFormChange(updatedFormFields);
-                      }}
-                      formFields={proposalFormFields}
+              <PageTitleInput
```
valentinludu, line 329: Is it fine that you don't check for `isStructured` to be defined anymore?
Before the code was checking for isStructured to be defined.

valentinludu (approved):
Tested, the behaviour remained the same.
The code looks good.

Left 2 questions.

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (mattcasey) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:454]: documentPageContent should always be rendered

components/proposals/ProposalPage/NewProposalPage.tsx
mattcasey: same comment ... we always rendered this content, actually. the logic was just to put it one of two places

PR #3534, opened 2024-01-29
Fix build cache

motechFR:
### WHAT

Update build caching approach

### WHY

For some reason, we're not getting the most up to date ceramic file


.ebstalk.apps.env/ceramic.env
```diff
@@ -1,6 +1,6 @@
 # Format of this file is <env_var>=<value> key/value pair
 
-DATABASE_URL="{{pull:secretsmanager:/io.cv.app/prd/db:SecretString:database_url}}"
+# DATABASE_URL="{{pull:secretsmanager:/io.cv.app/prd/db:SecretString:database_url}}"
```
mattcasey, line 3: does this need to be uncommented?

mattcasey (approved):

.ebstalk.apps.env/ceramic.env
motechFR: Was testing if it'll work without. It did so I'm going to delete this entirely since this server doesn't need the db secret

PR #3538, opened 2024-01-29
fix for custom domain logic

mattcasey:
Any page that starts with "u" in the title or any other page will not work on custom domains, because we think it's a public page.

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -12,11 +12,11 @@ const PUBLIC_FILE = /\.(.*)$/; // Files
 
 const FORCE_SUBDOMAINS = process.env.FORCE_SUBDOMAINS === 'true';
 
-export async function middleware(req: NextRequest) {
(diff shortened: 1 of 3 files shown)


middleware.ts
```diff
@@ -12,11 +12,11 @@ const PUBLIC_FILE = /\.(.*)$/; // Files
 
 const FORCE_SUBDOMAINS = process.env.FORCE_SUBDOMAINS === 'true';
 
-export async function middleware(req: NextRequest) {
-  if (isTestEnv) {
-    // Skip middleware in tests
-    return;
-  }
+export function middleware(req: NextRequest) {
+  // if (isTestEnv) {
```
mattcasey, line 16: commenting to see if this is necesary. if it is, i'll just workaround it

msadura (commented):
Woah awesome to have middleware test 😮

PR #3539, opened 2024-01-29
Feat/charms logic

msadura:
### WHAT

Base methods for generating user wallets and operating charms:
- adding charms - "generating" new charms (by specific user acitons)
- transferring charms - initially from users to spaces
- spending charms - removing charms from space's balance by admins (as a result of purchasing blocks)

core: https://github.com/charmverse/core/pull/224


lib/charms/getCharmTx.ts
mattcasey, line 1: suggestion: abbreviations and acronyms are easy to misinterpret... instead of using the word "charmTx" everywhere, what about "Transaction" (ie getTransaction, TransactionResult, etc)? I think charm's implied as it's already in a lib/charms folder.

lib/charms/__tests__/transferCharms.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { InvalidInputError } from '@charmverse/core/errors';
```
mattcasey, line 1: 🙌 tests :)

lib/google/loginWithGoogle.ts
```diff
@@ -92,6 +93,9 @@ export async function loginWithGoogle({
         },
         include: sessionUserRelations
       });
+
+      await generateCharmWallet({ userId: user.id });
```
mattcasey, line 97: I think this is very hard, maybe impossible, to ensure we already create a wallet whenever a user is added to the system.  They can be added via the public API, or invited as guests. How will we protect against that?

Alternatively, can I suggest we make the charm lib methods create wallets as needed?

mattcasey (changes requested):
I'd like to discuss making it so that the Charms system does not require charm wallets to exist beforehand.. it seems easier and safer to just create them as needed

msadura:
> I'd like to discuss making it so that the Charms system does not require charm wallets to exist beforehand.. it seems easier and safer to just create them as needed

Actually it is not needed and I can remove automatic creation with user profile. `getUserOrSpaceWallet` method will automatically create a wallet if user does not have one, so it can be created only when actual transaction happens

mattcasey (approved):

PR #3542, opened 2024-01-30
Fetch & display gitcoin offchain stamps

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/pull-in-credentials-from-gitcoin-api-24745322138210812)


components/members/components/MemberProfile/components/UserCredentials/UserCredentialRow.tsx
```diff
@@ -5,10 +5,18 @@ import Image from 'next/image';
 
 import Link from 'components/common/Link';
 import type { EASAttestationFromApi } from 'lib/credentials/external/getExternalCredentials';
+import type { TrackedSchemaParams } from 'lib/credentials/external/schemas';
 import { trackedSchemas } from 'lib/credentials/external/schemas';
 
 export function UserCredentialRow({ credential }: { credential: EASAttestationFromApi }) {
-  const schemaInfo = trackedSchemas[credential.chainId]?.find((s) => s.schemaId === credential.schemaId);
+  const schemaInfo = credential.chainId
+    ? trackedSchemas[credential.chainId]?.find((s) => s.schemaId === credential.schemaId)
```
Devorein, line 13: Since these are offchain credentials I made chainId nullable, not sure if that's the right approach @motechFR

components/members/components/MemberProfile/components/UserCredentials/UserCredentialRow.tsx
```diff
@@ -59,11 +63,13 @@ export function UserCredentialRow({ credential }: { credential: EASAttestationFr
           <Chip variant='outlined' key={field.name} label={field.value} />
         ))}
       </Grid>
-      <Grid item xs={2} display='flex' justifyContent='flex-end' pr={2}>
-        <Link href={credential.verificationUrl} external target='_blank'>
-          <LaunchIcon sx={{ alignSelf: 'center' }} />
-        </Link>
-      </Grid>
+      {credential.verificationUrl && (
```
Devorein, line 66: Couldn't figure out how to show the stamps as EAS attestations thus also made this nullable

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+};
+
+const GITCOIN_SCORER_BASE_URL = 'https://api.scorer.gitcoin.co';
+
+const GITCOIN_API_HEADERS = {
+  'Content-Type': 'application/json',
+  'X-Api-Key': process.env.GITCOIN_API_KEY
+};
+
+const GITCOIN_SCORER_ID = 6441;
```
Devorein, line 67: The scorer ID can be exposed as It can't be used without the API token

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+  name: string;
+  description: string;
+  hash: string;
+};
+
+export type GitcoinStamp = {
+  version: string;
+  credential: Credential;
+  metadata: Metadata;
+};
```
Devorein, line 52: There are a lot of fields which aren't present here that we expect like issuerId, attestationId, schemaId & chainId (these are offchain so it could be nullable)

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+
+type GetAddressesGitcoinPassportScoresResponse = {
+  items: ScoreItem[];
+  count: number;
+};
+
+async function getGitcoinPassportScores(wallets: string[]) {
+  try {
+    const response = await http.GET<GetAddressesGitcoinPassportScoresResponse>(
```
Devorein, line 111: In all cases the scorer returns empty values for my address even though I have gitcoin passport enabled

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+}: {
+  wallets: string[];
+}): Promise<EASAttestationFromApi[]> {
+  const gitcoinStamps = (await Promise.all(wallets.map((wallet) => getAddressGitcoinStamps({ wallet }).then()))).flat();
+  const gitcoinPassportScores = await getGitcoinPassportScores(wallets);
+  return gitcoinStamps.map((stamp) => {
+    const gitcoinPassportScore = gitcoinPassportScores.find((score) => score.address === stamp.recipient);
+
+    const mappedData: EASAttestationFromApi = {
+      id: stamp.metadata.platform.id, // This is not the stamp id, the api doesn't return the stamp id
```
Devorein, line 143: This is not the attestation id, its the platform id which is not guaranteed to be unique everytime. I couldn't find the attestation id from the response

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+      content: {
+        title: 'Gitcoin Passport',
+        organization: 'Gitcoin',
+        fields: gitcoinPassportScore ? [{ name: 'passport_score' }] : [],
+        passport_score: gitcoinPassportScore ? `Passport Score: ${gitcoinPassportScore.evidence.rawScore}` : null
```
Devorein, line 148: There is a point associated with each stamp, should this be the point or the passport score of the user? If so there will be multiple stamps with the same passport score @motechFR

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+      },
+      attester: stamp.credential.issuer, // This is in did:key:{publicKey} format not a address
```
Devorein, line 150: This is not an address, do we need to get the address from the public key? @motechFR

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+      recipient: stamp.recipient,
+      schemaId: '0xd7b8c4ffa4c9fd1ecb3f6db8201e916a8d7dba11f161c1b0b5ccf44ceb8e2a39', // Optimism attestation schema id: https://docs.passport.gitcoin.co/building-with-passport/smart-contracts/contract-reference#eas-schema,
+      timeCreated: new Date(stamp.credential.issuanceDate).getTime(),
+      chainId: null, // Optimism or Linea chain id
```
Devorein, line 154: These are offchain stamps thus chainId is kept null

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,162 @@
...
+      type: 'external',
+      verificationUrl: null,
```
Devorein, line 156: Not sure how to get the verificationUrl since we don't have the attestation id

Devorein:
image
Currently the UI looks like this, shouldn't we be showing the stamp logo as well as the title of the stamp?

lib/credentials/external/schemas.ts
```diff
@@ -16,12 +16,13 @@ export type TrackedSchemaParams = {
   iconUrl: string;
   // Some schemas have unfriendly schema keys or values that aren't useful by themselves.
   // For simplicity, we render only the value, but can optionally tweak the rendered value if required
-  fields?: { name: string; mapper?: (val: any) => string }[];
+  fields?: { name: string; mapper?: (val: any) => string | undefined }[];
```
Devorein, line 19: If it returns falsy value we filter it out, required since passport_score could be null

motechFR, line 19: We should remove any changes made in lib/credenials/external that relate to Gitcoin since the data source is entirely separate

PR #3551, opened 2024-01-30
fix Save button not allowing user to remove form fields questions

mattcasey:

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -21,7 +21,7 @@ import type { ConnectionEvent } from 'components/common/CharmEditor/components/f
 import { focusEventName } from 'components/common/CharmEditor/constants';
 import { FormFieldsEditor } from 'components/common/form/FormFieldsEditor';
 import { EvaluationSidebar } from 'components/proposals/ProposalPage/components/EvaluationSidebar/EvaluationSidebar';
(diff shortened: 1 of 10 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -472,7 +472,7 @@ function DocumentPage({ insideModal = false, page, savePage, readOnly = false, e
                       refreshProposal={refreshProposal}
                     />
                   ) : (
-                    <ProposalFormFieldInputs
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:475]: renamed FormFieldInputs to FormFieldsInput, which is more clear and simila to "FormFieldsEditor"

components/common/form/FormFieldsEditor.tsx
```diff
@@ -48,27 +48,60 @@ export function FormFieldsEditor({
...
-  const { trigger } = useUpdateProposalFormFields({ proposalId });
-  const saveFormFields = async () => {
-    await trigger({ formFields });
-    refreshProposal?.();
-  };
+  const { trigger, isMutating } = useUpdateProposalFormFields({ proposalId });
+  const { showMessage } = useSnackbar();
+  const [isFormDirty, setIsFormDirty] = useState(false);
+
+  async function saveFormFields() {
```
mattcasey, line 55: moved this logic out of the base component since it didn't need to be there

PR #3553, opened 2024-01-30
Fix flow for joining space in new proposals page

motechFR:
### WHAT

Handle the redirect lifcyle directly from within the new proposal page


components/common/RouteGuard.tsx
```diff
@@ -14,18 +13,28 @@ import { redirectToAppLogin, shouldRedirectToAppLogin } from 'lib/utilities/brow
...
+  '/',
+  'share',
+  'api-docs',
+  'u',
+  'join',
+  'invite',
+  'authenticate',
+  'test',
+  'permalink',
+  'proposals/new'
```
mattcasey, line 26: Incuding a "/" won't work for custom domains - see middleware.ts. I think there may be other places where we are only looking at the first part of the path

mattcasey: But i dont think it belongs in this list anyway. we still want users to have to log in

components/common/RouteGuard.tsx
```diff
@@ -67,7 +79,9 @@ export default function RouteGuard({ children }: { children: ReactNode }) {
     // special case, when visiting main app url on space subdomain
     const isSpaceSubdomainPath = firstPathSegment === '/' && !!_spaceDomain;
     // visiting page that shoould be alway available to logged in users
-    const isAvailableToLoggedInUsers = publicLoggedInPages.some((basePath) => firstPathSegment === basePath);
+    const isAvailableToLoggedInUsers = publicLoggedInPages.some(
```
mattcasey, line 82: you're using completely different logic here by checking the router.pathname, i dont think it belongs in "publicLoggedInPages" config.

pages/[domain]/proposals/new.tsx
```diff
@@ -11,6 +16,30 @@ export default function PageView() {
...
+
+  const { verifyCustomJoin } = useCustomJoinSpace();
+
+  const { isLoaded: isUserLoaded, user } = useUser();
+  const { isLoading: isCurrentSpaceLoaded, space } = useCurrentSpace();
+
+  async function tryJoining() {
+    await verifyCustomJoin({
+      proposalTemplate: selectedTemplate as string
+    }).catch(() => router.push(`/join?domain=${encodeURIComponent(space?.domain as string)}`));
```
Comment (mattcasey) [inline pages/[domain]/proposals/new.tsx:30]: so this handles the case if the space doesn't have public templates right?

mattcasey (approved):
looks good, i still think it'd be cleaner to just put all this logic in serversideprops. Was there a reason we don't do that? users will get a flicker of an empty proposals template while they join the space

mattcasey (changes requested):
i dont think it needs to be added to publicPages, and right now all the words in that list need to not include a "/"

mattcasey (approved):

PR #3555, opened 2024-01-31
Favorite credentials

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/set-up-to-5-credentials-as-favourite-9152370013517406)


lib/credentials/external/getExternalCredentials.ts
```diff
@@ -62,6 +69,57 @@ const GET_EXTERNAL_CREDENTIALS = gql`
   }
 `;
 
+const GET_EXTERNAL_CREDENTIAL = gql`
```
motechFR, line 72: GET_CREDENTIAL

lib/credentials/external/getExternalCredentials.ts
```diff
@@ -62,6 +69,57 @@ const GET_EXTERNAL_CREDENTIALS = gql`
...
+      content: JSON.parse(attestation.decodedDataJson).reduce((acc: any, val: SchemaDecodedItem) => {
+        acc[val.name] = val.value.value;
+        return acc;
+      }, {} as any),
+      timeCreated: attestation.timeCreated * 1000,
+      verificationUrl: getOnChainAttestationUrl({ chainId, attestationId: attestation.id })
+    } as EASAttestationFromApi;
+  } catch (_) {
+    return null;
+  }
```
motechFR, line 120: Lil' cleanup, can we get a reusable utility method in this file please

pages/api/credentials/favorites.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (!payload.attestationId && !payload.issuedCredentialId) {
+    throw new InvalidInputError('Either attestationId or issuedCredentialId must be provided');
+  }
+  const favoriteCredential = await prisma.favoriteCredential.create({
+    data: {
+      attestationId: payload.attestationId,
+      chainId: payload.chainId,
+      issuedCredentialId: payload.issuedCredentialId
+    }
+  });
```
motechFR, line 34: Let's make this an upsert instead.

If you define the correct compound indexes on the schema, you'll be able to pass those params to the where clause

See this [comment](https://github.com/charmverse/core/pull/229/files#r1474707880) on core

lib/credentials/external/getExternalCredentials.ts
Devorein: We got rid of this part of the code, so this mapper is used only in a single place.

PR #3562, opened 2024-01-31
Save proposal template with a reward

motechFR:
### WHAT

Fix the proposal template so it can be saved as an reward, without immediately assigning the submitter

If assignee is changed, handle that. If not available, fall back to proposal authors

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -412,6 +412,7 @@ function DocumentPage({ insideModal = false, page, savePage, readOnly = false, e
                     proposalPage={page}
                     proposal={proposal}
                     refreshProposal={refreshProposal}
+                    isProposalTemplate={page.type === 'proposal_template'}
                   />
                 )}
                 {reward && (
--- a/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/AttachRewardButton.tsx
+++ b/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/AttachRewardButton.tsx
@@ -6,10 +6,12 @@ import type { UpdateableRewardFields } from 'lib/rewards/updateRewardSettings';
 
 export function AttachRewardButton({
   createNewReward,
-  variant = 'card_property'
+  variant = 'card_property',
+  isProposalTemplate
 }: {
   createNewReward: VoidFunction;
   variant?: 'solid_button' | 'card_property';
+  isProposalTemplate?: boolean;
 }) {
   const { getFeatureTitle } = useSpaceFeatures();
   if (variant === 'card_property') {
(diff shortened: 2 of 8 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -412,6 +412,7 @@ function DocumentPage({ insideModal = false, page, savePage, readOnly = false, e
                     proposalPage={page}
                     proposal={proposal}
                     refreshProposal={refreshProposal}
+                    isProposalTemplate={page.type === 'proposal_template'}
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:415]: since we're passing in 'page' already, please just use that. its harder when we add lots of props

lib/proposal/createRewardsForProposal.ts
```diff
@@ -66,11 +66,18 @@ export async function createRewardsForProposal({ proposalId, userId }: { userId:
...
+      proposalId: proposal.id
+    }
+  });
+
   let rewardsToCreate = [...pendingRewards];
   const rewardsPromises = rewardsToCreate.map(async ({ page, reward, draftId }) => {
     // create reward
     const { createdPageId } = await createReward({
       ...reward,
+      assignedSubmitters: reward.assignedSubmitters?.length ? reward.assignedSubmitters : authors.map((a) => a.userId),
```
mattcasey, line 80: as mentioned before, this should be done in applyTemplate, NOT in this endpoint

components/rewards/components/RewardProperties/RewardPropertiesForm.tsx
```diff
@@ -302,13 +305,12 @@ export function RewardPropertiesForm({
...
               />
             </Box>
-
             <Box display='flex' height='fit-content' flex={1} className='octo-propertyrow'>
               <PropertyLabel readOnly highlighted>
                 Application Type
               </PropertyLabel>
               <RewardApplicationType
-                readOnly={readOnlyApplicationType}
+                readOnly={readOnlyApplicationType || isProposalTemplate}
```
mattcasey, line 313: blease include "isProposalTemplate" when we define "readOnlyApplicationType" instead of adding an exception down here

components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/AttachRewardButton.tsx
```diff
@@ -29,17 +31,19 @@ export function AttachRewardButton({
...
   rewardValues: UpdateableRewardFields;
+  isProposalTemplate?: boolean;
 }) {
   let disabledTooltip: string | undefined;
   if (!newPageValues?.title) {
     disabledTooltip = 'Page title is required';
   } else if (!rewardValues.reviewers?.length) {
     disabledTooltip = 'Reviewer is required';
-  } else if (rewardValues.assignedSubmitters && rewardValues.assignedSubmitters.length === 0) {
+  } else if (rewardValues.assignedSubmitters && rewardValues.assignedSubmitters.length && !isProposalTemplate) {
```
mattcasey, line 46: assigned submitters should always be required

motechFR: The problem is assigned submitters will never have an actual value when using a proposal template since it wouldn't make sense

PR #3567, opened 2024-02-01
Gitcoin passport score in credentials

Devorein:
### WHY

image
image


components/members/components/MemberProfile/components/UserCredentials/UserCredentialRow.tsx
```diff
@@ -4,25 +4,35 @@ import Box from '@mui/material/Box';
...
           iconUrl: credential.iconUrl ?? '/images/logo_black_lightgrey.png',
           attestationContent: [{ name: 'status', value: credential.content.status }]
         }
+      : credential.type === 'gitcoin'
+      ? {
+          title: 'Gitcoin Passport Score',
+          subtitle: 'Gitcoin',
+          iconUrl: '/images/logos/Gitcoin_Passport_Logomark_SeaFoam.svg',
+          attestationContent: [
+            { name: 'Passport Score', value: `Passport Score: ${credential.content.passport_score?.toFixed(2)}` }
```
motechFR, line 33: Let's remove  "Passport Score" from the label here since the name of the credential is already Passport Score

lib/credentials/external/getOnchainCredentials.ts
```diff
@@ -44,8 +44,8 @@ export type EASAttestationFromApi<T = any> = {
   schemaId: string;
   timeCreated: number;
   chainId: ExternalCredentialChain | EasSchemaChain;
-  type: 'external' | 'internal';
-  verificationUrl: string;
+  type: 'onchain' | 'charmverse' | 'gitcoin';
```
motechFR, line 47: Nice, good change!

lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -0,0 +1,94 @@
...
+  evidence: Evidence | null;
+  error: string;
+  stamp_scores: Record<string, unknown>;
+};
+
+async function getGitcoinPassportScores(wallets: string[]) {
+  try {
+    const scoreItems = await Promise.all(
+      wallets.map((wallet) =>
+        http.POST<ScoreItem>(
```
motechFR, line 39: Let's avoid refreshing scores all the time. Let's do a GET, and only call the POST if we don't get back a result

motechFR (approved):

components/members/components/MemberProfile/components/UserCredentials/UserCredentialRow.tsx
Devorein: Good point. Updated

lib/credentials/getGitcoinCredentialsByWallets.ts
Devorein: Good point. Updated.

PR #3568, opened 2024-02-01
Use 'Untitled' for undefined page titles

Devorein:


components/common/BoardEditor/components/properties/PagesAutocomplete.tsx
```diff
@@ -148,7 +148,7 @@ function PagesAutocompleteBase({
         forcePopupIcon={false}
         fullWidth
         getOptionLabel={(option) => {
-          return option.title.toLocaleLowerCase() || 'untitled';
+          return (option.title || 'Untitled').toLocaleLowerCase() || 'untitled';
```
mattcasey, line 151: isn't it weird that 'untitled' appears lowercase?

mattcasey (approved):

components/common/BoardEditor/components/properties/PagesAutocomplete.tsx
Devorein: Good catch, I always thought this was used for searching the dropdown items but its not required since we are using renderOption

PR #3574, opened 2024-02-02
Better error handling and logs for fetching gitcoin passport score

Devorein:


lib/credentials/getGitcoinCredentialsByWallets.ts
```diff
@@ -32,49 +32,47 @@ type ScoreItem = {
...
-          }
+  const scoreItems = await Promise.all(
+    wallets.map(async (wallet) => {
+      try {
+        const score = await http
+          .GET<ScoreItem>(`${GITCOIN_SCORER_BASE_URL}/registry/score/${GITCOIN_SCORER_ID}/${wallet}`, undefined, {
+            credentials: 'omit',
+            headers: GITCOIN_API_HEADERS
+          })
+          .catch(() => null);
```
motechFR, line 46: We need error logging here.

If we are not throwing an error for later handling, then we should be logging it. Otherwise we won't be able to diagnose what's wrong in prod

PR #3581, opened 2024-02-02
UI/editing rewards

motechFR:
### WHAT

Pass readonly state to proposal rewards

### DIFF
--- a/components/[pageId]/DocumentPage/components/ProposalProperties.tsx
+++ b/components/[pageId]/DocumentPage/components/ProposalProperties.tsx
@@ -1,3 +1,4 @@
+import { isProposalAuthor } from '@charmverse/core/permissions';
 import { Box } from '@mui/material';
 
 import { useUpdateProposal } from 'charmClient/hooks/proposals';
@@ -6,6 +7,7 @@ import type { ProposalPropertiesInput } from 'components/proposals/ProposalPage/
 import { ProposalPropertiesBase } from 'components/proposals/ProposalPage/components/ProposalProperties/ProposalPropertiesBase';
 import { useIsAdmin } from 'hooks/useIsAdmin';
 import { useSnackbar } from 'hooks/useSnackbar';
+import { useUser } from 'hooks/useUser';
 import type { PageWithContent } from 'lib/pages';
 import type { ProposalWithUsersAndRubric } from 'lib/proposal/interface';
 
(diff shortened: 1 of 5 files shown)


components/[pageId]/DocumentPage/components/ProposalProperties.tsx
```diff
@@ -30,11 +32,14 @@ export function ProposalProperties({
   const sourceTemplate = useProposalTemplateById(proposal?.page?.sourceTemplateId);
 
   const isAdmin = useIsAdmin();
+  const { user } = useUser();
 
   // further restrict readOnly if user cannot update proposal properties specifically
   const readOnlyProperties = readOnly || !proposal?.permissions.edit;
   const readOnlySelectedCredentialTemplates =
     readOnly || !proposal?.permissions.edit || (sourceTemplate?.selectedCredentialTemplates && !isAdmin);
+  const readOnlyRewards =
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/components/ProposalProperties.tsx:41]: lets just use proposal permissions

components/common/PageDialog/components/NewDocumentPage.tsx
```diff
@@ -60,7 +62,7 @@ export function NewDocumentPage({
               icon={newPageValues.icon || null}
               updatedAt={new Date().toString()}
               title={newPageValues.title || ''}
-              readOnly={false}
+              readOnly={!!readOnly}
```
mattcasey, line 65: this is NewDocumentPage, it should never be readonly

components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewards.tsx
```diff
@@ -117,8 +118,6 @@ export function ProposalRewards({
   }
 
   function editReward({ reward, page, draftId }: ProposalPendingReward) {
-    if (readOnly) return;
```
mattcasey, line 120: so the original logic was not to let the user view the reward if readOnly... why don't we just keep that logic? it is safer

components/rewards/components/RewardProperties/RewardPropertiesForm.tsx
```diff
@@ -235,7 +235,7 @@ export function RewardPropertiesForm({
                   <Box display='flex' flex={1}>
                     <TemplateSelect
                       options={rewardTemplates.map((rewardTemplate) => rewardTemplate.page)}
-                      disabled={readOnlyTemplate}
+                      disabled={readOnlyTemplate || readOnly}
```
mattcasey, line 238: if we have a "readOnlyTemplate" var, can u make sure that var is correct, rather than adding another "or" condition here?

mattcasey, line 238: This looks like we're not setting readOnlyTemplate correctly in the parent components

PR #3584, opened 2024-02-02
Fix for adding rewards to a proposal

mattcasey:
I tried to add a reward to a proposal but it was telling me I needed to assign reviewers. I think checking for isProposalTemplate was added originally because we would make an exception and save them on the backend.

### DIFF
--- a/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewards.tsx
+++ b/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewards.tsx
@@ -245,20 +245,22 @@ export function ProposalRewards({
                         </Grid>
                       </Hidden>
 
-                      <Grid item xs={4} lg={2}>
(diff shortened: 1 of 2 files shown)


components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewards.tsx
```diff
@@ -245,20 +245,22 @@ export function ProposalRewards({
...
-                            disabled={readOnly}
-                          >
-                            <Edit color='secondary' fontSize='small' />
-                          </IconButton>
-                          <IconButton size='small' onClick={() => onDelete(draftId)} disabled={readOnly}>
-                            <Delete color='secondary' fontSize='small' />
-                          </IconButton>
-                        </Stack>
-                      </Grid>
+                      {!readOnly && (
```
mattcasey, line 248: also, we want to hide the edit/delete icons for rewards if u do not have permission to edit them

PR #3587, opened 2024-02-05
Rendering Farcaster frame on charmeditor

Devorein:
### WHY

image
image


components/common/CharmEditor/components/FarcasterFrame.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+import type { BaseRawNodeSpec } from './@bangle.dev/core/specRegistry';
+import BlockAligner from './BlockAligner';
+import { MediaSelectionPopup } from './common/MediaSelectionPopup';
+import { MediaUrlInput } from './common/MediaUrlInput';
+import type { CharmNodeViewProps } from './nodeView/nodeView';
+
+export function farcasterFrameSpec() {
+  const spec: BaseRawNodeSpec = {
+    name: 'farcasterFrame',
```
Devorein, line 17: Intentionally appended farcaster as frame is too ambigious

charmClient/hooks/farcaster.ts
```diff
@@ -0,0 +1,7 @@
+import type { Frame } from 'frames.js';
+
+import { useGET } from './helpers';
+
+export function useGetFarcasterFrame(frameUrl?: string) {
+  return useGET<Frame | null>(frameUrl ? '/api/farcaster/get-frame' : null, { frameUrl });
```
mattcasey, line 6: i think "get-" in the pathname is already implied, for ex we don't use "api/pages/get-page"

components/common/CharmEditor/components/FarcasterFrame.tsx
mattcasey: that's a good idea! Suggestion: should we start using snake case for components? I know we have a mix of both, but snake case is our preferred format for enums

components/common/CharmEditor/components/FarcasterFrame.tsx
```diff
@@ -0,0 +1,119 @@
...
+  };
+  return spec;
+}
+
+export function FarcasterFrame({ selected, attrs, node, deleteNode, updateAttrs, readOnly }: CharmNodeViewProps) {
+  const { data: farcasterFrame, isLoading } = useGetFarcasterFrame(attrs.src);
+
+  if (isLoading) {
+    return (
+      <Box my={5}>
```
mattcasey, line 47: you probably just want to set minHeight on LoadingComponent, so that the loading cursor appears in the middle of whitespace and not at the bottom?

components/common/CharmEditor/components/FarcasterFrame.tsx
```diff
@@ -0,0 +1,119 @@
...
+              />
+            ]
+          ]}
+        />
+      </MediaSelectionPopup>
+    );
+  }
+
+  if (!farcasterFrame) {
+    return null;
```
mattcasey, line 92: when does this case actually occur? should we just show LoadingComponent instead of nothing?

components/common/CharmEditor/components/inlinePalette/editorItems/embed.tsx
```diff
@@ -55,6 +56,35 @@ function iframeEmbedType(type: EmbedType): PaletteItemTypeNoGroup {
   };
 }
 
+function handleCustomEmbed({
```
mattcasey, line 59: nice cleanup

components/common/CharmEditor/components/FarcasterFrame.tsx
```diff
@@ -0,0 +1,119 @@
...
+import LoadingComponent from 'components/common/LoadingComponent';
+import MultiTabs from 'components/common/MultiTabs';
+
+import type { BaseRawNodeSpec } from './@bangle.dev/core/specRegistry';
+import BlockAligner from './BlockAligner';
+import { MediaSelectionPopup } from './common/MediaSelectionPopup';
+import { MediaUrlInput } from './common/MediaUrlInput';
+import type { CharmNodeViewProps } from './nodeView/nodeView';
+
+export function farcasterFrameSpec() {
```
mattcasey, line 15: It is extremely important that specs need to live in their own file outside of React. Otherwise we have to tranpsile React packages when trying to use prosemirror on the backend.
Please break this out into a "farcasterFrame" folder with spec and component files.

mattcasey (changes requested):
Node specs must be in their own file. EVerything else is a suggestion

components/common/CharmEditor/components/FarcasterFrame.tsx
Devorein: I don't think so atleast not now since we already have `cryptoPrice`

pages/api/farcaster/get-frame.ts
```diff
@@ -0,0 +1,29 @@
...
+import { onError, onNoMatch, requireKeys } from 'lib/middleware';
+import { withSessionRoute } from 'lib/session/withSession';
+
+const handler = nc<NextApiRequest, NextApiResponse>({ onError, onNoMatch });
+
+handler.use(requireKeys([{ key: 'frameUrl', valueType: 'string' }], 'query')).get(extractFarcasterMetadataFromUrl);
+
+async function extractFarcasterMetadataFromUrl(req: NextApiRequest, res: NextApiResponse<Frame | null>) {
+  const frameUrl = req.query.frameUrl as string;
+  const unparsedFrameHtml = await http.GET<string>(frameUrl);
```
mattcasey, line 16: This code's a little dangerous because we are letting useres have our server call any endpoint. Some security suggestions from Farcaster:

```
    Proxy image requests to prevent frame servers from tracking users.
    Sanitize redirect URLs to ensure they start with http:// or https://.
    Only accept data URIs if they are images.
    Avoid rendering SVGs as they may contain executable code.
    ```

components/common/CharmEditor/components/FarcasterFrame.tsx
Devorein: This could happen if the URL provided is not following the metadata structure required by Farcaster. We can't validate whether the URL is correct or not either without fetching and inspecting the metadata. We could however show an error message and request the user to provide the URL again. @mattcasey

components/common/CharmEditor/components/FarcasterFrame.tsx
Devorein: yikes sure, we still have the spaces on the same as the component for the cryptoPrice so I will move that too.

components/common/CharmEditor/components/FarcasterFrame.tsx
Devorein: Yeah thats fair, I actually copied it from cryptoPrice, I will make the changes there as well

mattcasey (approved):

PR #3599, opened 2024-02-06
track forum and proposal events in the public api

mattcasey:

### DIFF
--- a/lib/forums/posts/voteForumComment.ts
+++ b/lib/forums/posts/voteForumComment.ts
@@ -41,17 +41,13 @@ export async function voteForumComment({
       }
     });
 
-    const category = post.category;
-
-    if (category) {
-      trackUserAction(upvoted ? 'upvote_comment' : 'downvote_comment', {
-        resourceId: commentId,
-        spaceId: post.spaceId,
-        userId,
-        categoryName: category.name,
-        postId: post.id
-      });
-    }
+    trackUserAction(upvoted ? 'upvote_comment' : 'downvote_comment', {
+      resourceId: commentId,
+      spaceId: post.spaceId,
+      userId,
+      categoryName: post.category.name,
+      postId: post.id
+    });
 
     const commentVote = await prisma.postCommentUpDownVote.upsert({
       create: {
--- a/pages/api/v1/forum/comments/[commentId]/index.ts
+++ b/pages/api/v1/forum/comments/[commentId]/index.ts
@@ -2,17 +2,15 @@ import { log } from '@charmverse/core/log';
 import { prisma } from '@charmverse/core/prisma-client';
 import type { NextApiRequest, NextApiResponse } from 'next';
 
+import { deletePostComment } from 'lib/forums/comments/deletePostComment';
 import { requireKeys } from 'lib/middleware';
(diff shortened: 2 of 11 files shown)


lib/forums/posts/voteForumComment.ts
```diff
@@ -41,17 +41,13 @@ export async function voteForumComment({
...
-    if (category) {
-      trackUserAction(upvoted ? 'upvote_comment' : 'downvote_comment', {
-        resourceId: commentId,
-        spaceId: post.spaceId,
-        userId,
-        categoryName: category.name,
-        postId: post.id
-      });
-    }
+    trackUserAction(upvoted ? 'upvote_comment' : 'downvote_comment', {
```
motechFR, line 44: Edge case: I think we can have null, which removes the upvote, but doesn't downvote

pages/api/v1/forum/comments/[commentId]/index.ts
```diff
@@ -29,8 +27,15 @@ handler.delete(deletePostComment).put(requireKeys(['contentMarkdown'], 'body'),
...
  *
  */
-async function deletePostComment(req: NextApiRequest, res: NextApiResponse) {
+
+const handler = superApiHandler();
+handler.delete(deletePostComment).put(requireKeys(['contentMarkdown'], 'body'), updatePostCommentEndpoint);
+
+async function updatePostCommentEndpoint(req: NextApiRequest, res: NextApiResponse) {
   // This should never be undefined, but adding this safeguard for future proofing
+  if (!req.query.commentId) {
```
Comment (motechFR) [inline pages/api/v1/forum/comments/[commentId]/index.ts:36]: We could have this as additional middleware requireKeys(['commentId'], 'query')

pages/api/v1/forum/comments/[commentId]/index.ts
```diff
@@ -44,17 +49,12 @@ async function deletePostComment(req: NextApiRequest, res: NextApiResponse) {
...
   });
 
-  if (result) {
-    await prisma.postComment.delete({
-      where: {
-        id: result.id
-      }
-    });
-  }
+  await deletePostComment({ commentId: result.id, userId: result.createdBy });
```
Comment (motechFR) [inline pages/api/v1/forum/comments/[commentId]/index.ts:57]: Are we getting userId from API? If so we might want to pass that here instead

pages/api/v1/proposals/[proposalId]/comments/[commentId]/vote.ts
```diff
@@ -76,7 +78,8 @@ async function voteComment(req: NextApiRequest, res: NextApiResponse<PublicApiPr
       }
     },
     select: {
-      id: true
+      id: true,
```
Comment (motechFR) [inline pages/api/v1/proposals/[proposalId]/comments/[commentId]/vote.ts:81]: Suggestion; We can refactor this logic like you did with forum posts endpoint

motechFR (approved):

lib/forums/posts/voteForumComment.ts
mattcasey: thanks, you are right, i think this was an existing bug

pages/api/v1/forum/comments/[commentId]/index.ts
Comment (mattcasey) [reply pages/api/v1/forum/comments/[commentId]/index.ts:57]: no, we dont ask for userId here

PR #3602, opened 2024-02-06
Replace Try charmverse with "Built with..." and fix the link for custom domains

mattcasey:
image


valentinludu (approved):
Tested on Mobile and desktop, FIrefox and Chrome.

Looks good!

PR #3605, opened 2024-02-07
Remove use of ProposalWithUsers from core

mattcasey:
Dis-entangling the use of a core type from the interface we use in the webapp api.


lib/form/__tests__/upsertProposalFormAnswers.spec.ts
```diff
@@ -10,7 +9,7 @@ import { generateUserAndSpace } from 'testing/setupDatabase';
 import { generateProposal } from 'testing/utils/proposals';
 
 describe('upsertFormAnswers', () => {
-  let proposal: ProposalWithUsers;
+  let proposal: Awaited<ReturnType<typeof generateProposal>>;
```
mattcasey, line 12: i've updated core to export this type specifically, but this type extraction works for now

PR #3606, opened 2024-02-07
Update mixpanel space profiles with block counts

msadura:
### WHAT

Update mixpanel space profiles with block counts  in a cron job. It uses mixpanel batch endpoint to update 1000 spaces at time

### DIFF
--- a/background/cron.ts
+++ b/background/cron.ts
@@ -1,6 +1,8 @@
 import { log } from '@charmverse/core/log';
 import cron from 'node-cron';
 
+import { updateSpacesMixpanelProfilesTask } from 'background/tasks/updateSpacesMixpanelProfilesTask';
+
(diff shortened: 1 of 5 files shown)


mattcasey (approved):

background/cron.ts
```diff
@@ -33,6 +35,7 @@ cron.schedule('*/30 * * * *', refreshBountyApplications);
 
 // Count blocks in all spaces
 cron.schedule('*/30 * * * *', countAllSpacesBlocksTask);
+cron.schedule('*/30 * * * *', updateSpacesMixpanelProfilesTask);
```
mattcasey, line 38: have this run once a day, no need to update so often

mattcasey (approved):

PR #3607, opened 2024-02-07
Fix joining spaces from proposal template for logged in users

motechFR:
### WHAT

Handle case where user is already logged in but not yet a space member

### DIFF
--- a/components/common/RouteGuard.tsx
+++ b/components/common/RouteGuard.tsx
@@ -1,10 +1,12 @@
 import type { UrlObject } from 'url';
 
 import { log } from '@charmverse/core/log';
+import { space } from '@lens-protocol/metadata/*';
 import { useRouter } from 'next/router';
 import type { ReactNode } from 'react';
 import { useEffect, useRef, useState } from 'react';
 
+import charmClient from 'charmClient';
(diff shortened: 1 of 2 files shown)


pages/[domain]/proposals/new.tsx
```diff
@@ -27,6 +27,13 @@ export const getServerSideProps: GetServerSideProps = withSessionSsr(async (cont
     });
 
     if (space) {
+      const redirect = {
+        redirect: {
+          destination: space.publicProposals ? `/${domainToUse}/proposals` : `/join?domain=${domainToUse}`,
```
Comment (mattcasey) [inline pages/[domain]/proposals/new.tsx:32]: I think if we just hardcode the domain like this, it will break custom odmains

mattcasey (changes requested):
Can we make this work without adding more context/data inside of RouteGuard or new redirect logic in /new. Needs tests for sure

components/common/RouteGuard.tsx
```diff
@@ -1,10 +1,12 @@
 import type { UrlObject } from 'url';
 
 import { log } from '@charmverse/core/log';
+import { space } from '@lens-protocol/metadata/*';
```
msadura, line 4: probably some mistake?

PR #3609, opened 2024-02-07
Sticky save space settings

valentinludu:
Space Settings changes:
- Have only 1 Save button and be sticky
- All input fields should be inside 1 form


msadura:
not really important but I do not see the reason why it should scroll with the page at the bottom

Screenshot 2024-02-08 at 14 03 04

msadura:
It is more of a product thing, but in my opinion it should also have cancel / discard button that would revert all the changes you made

msadura:
Might not be related to this pr, but if you change some settings without saving and change tab in sidebar they will be reverted without a warning

valentinludu:
PR updated after the feedback.
Sticky footer looks like other sticky footers in app. 

For the cancel button and preventing of the tab switch, I will ask Chris.
In the meantime let's merge this PR and will circle back once I have a response from Chris.

msadura (approved):

PR #3613, opened 2024-02-08
Readonly public interaction with farcaster frame

Devorein:

### DIFF
--- a/lib/farcaster/getFarcasterProfile.ts
+++ b/lib/farcaster/getFarcasterProfile.ts
@@ -1,8 +1,4 @@
-import useSWR from 'swr';
-
 import * as http from 'adapters/http';
-
-import { useUser } from './useUser';
 
 export type FarcasterProfile = {
   body: {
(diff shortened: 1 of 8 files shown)


mattcasey (approved):

mattcasey (approved):
approved but this is confusing how we end up w/the profile and loks like it would be hard to debug. why do we sometimes rely on a farcasterUser, sometimes a logged in user? it seems like there should be one source of truth, one direction of dependencies... or maybe just at least have only one hook. Why doesnt useFarcasterUser just get the profile too, do we actually need them to be separate?

lib/farcaster/getFarcasterProfile.ts
```diff
@@ -0,0 +1,55 @@
...
+      `${profileApiUrl}?connected_address=${wallet}`,
+      {},
+      {
+        credentials: 'omit'
+      }
+    );
+    if (!_farcasterProfile) {
+      [_farcasterProfile] = await http.GET<FarcasterProfile[]>(
+        `${profileApiUrl}?address=${wallet}`,
+        {},
```
mattcasey, line 35: you dont need to include this extra argument

Devorein: Thats actually required because of `credentials: 'omit'`. Otherwise the endpoint returns a CORS error

PR #3618, opened 2024-02-08
User and wallet cleanup

valentinludu:
Use SWR and less states and effects


components/_app/Web3ConnectionManager/Web3ConnectionManager.tsx
```diff
@@ -8,9 +8,7 @@ const Web3Connection = createContext({
   isWalletSelectorModalOpen: false,
   connectWallet: () => {},
   closeWalletSelectorModal: () => {},
-  triedEager: false,
-  isConnectingIdentity: false,
```
valentinludu, line 12: setIsConnectingIdentity was not used at all.
isConnectingIdentity was always false.

Below there is a comment about using this only for unstoppable domains, but we don't have that anymore.

hooks/useUser.tsx
```diff
@@ -54,21 +49,12 @@ export function UserProvider({ children }: { children: ReactNode }) {
...
-  }, []);
-
-  useEffect(() => {
-    if (user?.deletedAt) {
-      charmClient.logout().then(() => {
-        window.location.href = window.location.origin;
-      });
-    }
-  }, [user]);
+  const updateUser = (updatedUser: Partial<LoggedInUser>) => refreshUser(updatedUser);
```
valentinludu, line 52: These 2 functions were left here just for legacy. We are using them in a lot of places but in reality we only need 1.

hooks/useWeb3Account.tsx
```diff
@@ -204,7 +209,7 @@ export function Web3AccountProvider({ children }: { children: ReactNode }) {
       const storedWalletSignature = getStoredSignature(account);
       setSignature(account, storedWalletSignature);
     }
-    // Case 3: user is switching wallets
+    // Case 2: user is switching wallets
```
mattcasey, line 212: this is still case 3 isnt it?

hooks/useWeb3Account.tsx
```diff
@@ -266,29 +262,27 @@ export function Web3AccountProvider({ children }: { children: ReactNode }) {
...
+        {
+          onSuccess: async (updatedUser) => {
+            logoutWallet();
+
+            setLitAuthSignature(null);
+            setLitProvider(null);
+            setStoredAccount(null);
+            updateUser(updatedUser);
+            activeConnector?.disconnect();
+            await mutate((key) => typeof key === 'string' && key.startsWith(`/nfts/${updatedUser?.id}`));
```
mattcasey, line 280: i'm just curious whats the difference with this callback method vs how we called mutate before?

mattcasey (approved):
Great cleanup! Thank you

hooks/useWeb3Account.tsx
valentinludu: yes.

I renamed it back

hooks/useWeb3Account.tsx
valentinludu: From official docs, this is how we need to mutate the global cache key.
I tried before the old approach and it did not work.

PR #3619, opened 2024-02-08
Improve farcaster frames security mechanisms

Devorein:

### DIFF
--- a/pages/api/farcaster/frame-action.ts
+++ b/pages/api/farcaster/frame-action.ts
@@ -29,6 +29,26 @@ export type FrameActionRequest = {
 
 const requestTimeout = 7500;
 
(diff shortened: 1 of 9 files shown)


mattcasey (approved):

pages/api/farcaster/frame-action.ts
```diff
@@ -81,6 +105,10 @@ async function getNextFrame(req: NextApiRequest, res: NextApiResponse<FrameActio
     throw new InvalidInputError('Invalid Farcaster frame URL');
   }
 
+  if (pageId) {
+    await trackFarcasterFrameInteractionEvent({ pageId, userId });
```
mattcasey, line 109: we usually do not block the response when tracking events. any errors should be try/catched so they dont break the request

Devorein: good point. I will update it.

PR #3623, opened 2024-02-08
use correct tenant urls and user ids for Summon API

mattcasey:
We shouldn't need a migration, at least right away. The cron sync job should retrieve and update spaceRoles automatically. I'll make another PR to remove xpsEngineId on user table later


testing/mocks/user.ts
```diff
@@ -37,5 +36,5 @@ export function createMockUser(user?: Partial<LoggedInUser>): LoggedInUser {
     googleAccounts: [],
     verifiedEmails: [],
     ...user
-  };
+  } as LoggedInUser;
```
mattcasey, line 39: this seems to allow us to make changes to the schema without having to add empty null values all the time

PR #3626, opened 2024-02-09
Added script to duplicate all rubric criteria from a evaluation step …

Devorein:
…to other templates

### DIFF
--- a/scripts/duplicateRubricCriterias.ts
+++ b/scripts/duplicateRubricCriterias.ts
@@ -5,12 +5,9 @@ import { Prisma, prisma } from '@charmverse/core/prisma-client';
  */
 
 async function duplicateRubricCriterias() {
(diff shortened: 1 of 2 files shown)


scripts/duplicateRubricCriterias.ts
```diff
@@ -0,0 +1,62 @@
...
+    "e3edf64f-c6e0-4958-93ca-a3775d1ffab2",
+    "976f70f3-94c8-4526-987e-55fc18071b6a",
+    "8b2a4b4d-2f8e-4ece-aa34-5625b484a4b9",
+    "4e261347-dc04-49b7-979e-df9b936d882c"
+  ]
+
+  const proposalTemplate = await prisma.proposal.findFirstOrThrow({
+    where: {
+      page: {
+        id: sourceProposalTemplateId
```
motechFR, line 19: Note, make sure page id and proposal id are the same. We should avoid relying on pageId === proposalId

motechFR (approved):

PR #3627, opened 2024-02-09
disable flaky test

mattcasey:
Flaky example: https://github.com/charmverse/app.charmverse.io/actions/runs/7845360310/job/21410423938


valentinludu:
I did a fix in my PR. Ofc, you can add it here if you want.
https://github.com/charmverse/app.charmverse.io/pull/3618/commits/d9d95763562bc0ee531a151181c474adbc5184a9

mattcasey:
thanks for fixing this @valentinludu ! it seemed like it would succeed after retries, i am not sure why

PR #3638, opened 2024-02-11
fix pasting markdown form VS Code

mattcasey:
To reproduce:
Type some markdown in VS Code, copy it and try pasting it in CharmEditor. We do not pick up any styles. But it works if you copy the markdown from most other editorsl

I figured out that, for some reason, VS Code doesn't trigger the regular method that parses markdown text. I just tried lots of methods and ideas to figure out how to get this to work.

* I also moved prosemirror/plugins/markdown to prosemirror/markdown, since the methods in that folder are not a plugin


components/common/CharmEditor/components/markdown/markdown.plugins.ts
```diff
@@ -36,6 +42,30 @@ export function plugins() {
             dom.innerHTML = md.render(str);
           }
           return parser.parseSlice(dom, { preserveWhitespace: true, context: $context });
+        },
+        // handle paste from VS Code, which does not trigger clipboardTextParser for some reason
+        handlePaste: (view: EditorView, rawEvent: ClipboardEvent) => {
```
mattcasey, line 47: added this method "handlePaste" to capture HTML. clipboardTextParser() doesn't get called when pasting from VS Code

Devorein (approved):

Devorein:
Works perfectly with the core readme.md file but there are issues with app readme.md file weirdly due to a single line (line 302). Copy/pasting every other line works perfectly fine except for that line which throws an `Empty text nodes are not allowed` error. Try copy/pasting just that single line and you will trigger it as well. I haven't been able to figure out the root cause

PR #3648, opened 2024-02-14
Syncing relation property card values

Devorein:
…erties### WHAT

### WHY

- [x] Implement Show on {board} toggle
- [x] Handle syncing card values with connected board
- [x] Deleting relation property on related and source board
- [x] Batch update relation properties
- [x] Real-time updates for syncing card values


motechFR:
UX Feedback: When editing property to show in related board, default "Related property on Untitled" to "Related property on <db_name>"


Bug: Needed to add opposite relation for this to work

components/common/BoardEditor/focalboard/src/widgets/propertyTypes.tsx
```diff
@@ -38,6 +40,7 @@ export function PropertyTypes({
     name?: IPropertyTemplate['name'];
   }) => void;
   isMobile?: boolean;
+  boardType?: 'proposals' | 'rewards';
```
motechFR, line 43: Suggestion, also add 'database' and default the value to database, OR add a comment explaining what this prop is for (above in JSDOC so it shows up to component consumers)

mattcasey (approved):
Amazing.. hard to review all the code

PR #3657, opened 2024-02-15
Feat/frame mint

msadura:
### WHAT

Minting directly from frames

### DIFF
--- a/components/common/CharmEditor/components/farcasterFrame/components/FarcasterButton.tsx
+++ b/components/common/CharmEditor/components/farcasterFrame/components/FarcasterButton.tsx
@@ -1,0 +1,17 @@
+import styled from '@emotion/styled';
+
+import { Button } from 'components/common/Button';
+
+export const FarcasterButton = styled(Button)(({ theme, disabled, loading }) => ({
+  width: '100%',
+  border:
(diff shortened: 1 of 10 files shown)


components/common/CharmEditor/components/farcasterFrame/components/FarcasterButton.tsx
```diff
@@ -0,0 +1,17 @@
+import styled from '@emotion/styled';
```
msadura, line 1: extracted component

hooks/useReservoir.ts
mattcasey, line 1: i wonder if it'd be better to put these hooks in the farcaster component. no big deal tho

components/common/CharmEditor/components/farcasterFrame/components/FarcasterMintButton.tsx
mattcasey, line 1: these should be in a components folder, since they're children of the NodeView.

mattcasey (approved):

hooks/useReservoir.ts
msadura: I can, although this api can be used for any nft-related activity, not related to farcaster

PR #3668, opened 2024-02-17
Proposal note prop

mattcasey:
The new inputs were breaking the layout of the table. I tried to reuse the ProposalNotesLink component but the wrapping element is what's breaking things, so it was easier to copy/paste

image

### DIFF
--- a/components/common/BoardEditor/components/properties/ProposalNotesLink.tsx
+++ b/components/common/BoardEditor/components/properties/ProposalNotesLink.tsx
@@ -2,29 +2,29 @@ import styled from '@emotion/styled';
 import type { MouseEvent } from 'react';
 
 import { useGetOrCreateProposalNotesId } from 'charmClient/hooks/proposals';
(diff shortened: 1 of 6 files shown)


components/common/BoardEditor/components/properties/ProposalNotesLink.tsx
```diff
@@ -0,0 +1,52 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
+
+export function ProposalNotesLink({ pageId }: { pageId: string }) {
+  const { showPage } = usePageDialog();
+  const { showError } = useSnackbar();
+  const { pages } = usePages();
+  const syncedPageId = pages[pageId]?.syncWithPageId;
```
mattcasey, line 30: this value seems like it should just be on the card somehow

Devorein (approved):

PR #3677, opened 2024-02-20
Show warning before reordering cards in board with sort

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/if-user-tries-to-manually-sort-when-sort-is-on-ask-if-they-want-to-reset-sort-3380296860939367)

### DIFF
--- a/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
+++ b/components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
@@ -711,7 +711,6 @@ function CenterPanel(props: Props) {
                 {activeBoard && activeView?.fields.viewType === 'gallery' && (
                   <Gallery
                     board={activeBoard}
-                    cardPages={cardPages}
                     cards={cards}
                     activeView={activeView}
                     readOnly={props.readOnly}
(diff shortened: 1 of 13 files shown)


components/common/BoardEditor/focalboard/src/components/gallery/gallery.tsx
```diff
@@ -21,36 +23,71 @@ type Props = {
...
+
+    if (hasSort) {
+      const { confirmed } = await showConfirmation({
+        message: 'Would you like to remove sorting?'
+      });
+
+      if (confirmed && localViewSettings) {
+        await mutator.changeViewSortOptions(activeView.id, activeView.fields.sortOptions, []);
+        localViewSettings.setLocalSort(null);
+        changeCardOrder = true;
```
mattcasey, line 64: instead of using a variable, i think the code should just return

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
```diff
@@ -711,6 +711,7 @@ function CenterPanel(props: Props) {
                 {activeBoard && activeView?.fields.viewType === 'gallery' && (
                   <Gallery
                     board={activeBoard}
+                    cardPages={cardPages}
```
mattcasey, line 714: can we not start passing this down too? our DBs are already very slow

mattcasey (approved):

components/common/BoardEditor/focalboard/src/components/centerPanel.tsx
Devorein: Updated.

PR #3678, opened 2024-02-20
Feat/tokengate v2

valentinludu:


components/common/TokenGateModal/utils/getTokensUnifiedAccessControlConditions.ts
```diff
@@ -1,46 +1,38 @@
...
-        }
+        type: 'ERC20',
+        tokenIds: [],
+        quantity: amount
       }
     ];
   }
 
-  if (check === 'token') {
+  if (check === 'token' && contract) {
```
Devorein, line 26: contract is not used in `contractAddress` section. Not sure we should be adding it to the condition

lib/blockchain/readContract.ts
Devorein, line 1: This doesn't seem to be used anywhere

Devorein (approved):

Devorein (approved):

components/common/TokenGateModal/utils/getTokensUnifiedAccessControlConditions.ts
valentinludu: Good point. No we don't need it.

lib/blockchain/readContract.ts
valentinludu: Deleted!

PR #3682, opened 2024-02-21
Fix member card and add browser tests

motechFR:
### WHAT

Don't show props in member card if they're empty


mattcasey (approved):

mattcasey:
re-gen the browser snapshots: `npm run test:browser:update`

PR #3684, opened 2024-02-21
Delete archival space block counts

msadura:
### WHAT

https://app.charmverse.io/charmverse/slow-queries-for-charms-7597386042394729


mattcasey (changes requested):
can we have a test on this one? it's a background job, and possible we never even know if it's deleting the wrong thing. There's already a test file.. all u gotta do is add one that has a block count from < 1 week ago, check that there's only one existing after you call countAllSpacesBlocksTask

PR #3695, opened 2024-02-22
Optimized query to update cards from proposals

Devorein:

### DIFF
--- a/lib/focalboard/updateCardsFromProposals.ts
+++ b/lib/focalboard/updateCardsFromProposals.ts
@@ -1,4 +1,10 @@
-import type { Block, Page, ProposalRubricCriteria, ProposalRubricCriteriaAnswer } from '@charmverse/core/prisma-client';
+import type {
+  Block,
+  Page,
+  Prisma,
+  ProposalRubricCriteria,
+  ProposalRubricCriteriaAnswer
+} from '@charmverse/core/prisma-client';
 import { prisma } from '@charmverse/core/prisma-client';
 import _ from 'lodash';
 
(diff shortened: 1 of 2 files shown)


lib/focalboard/updateCardsFromProposals.ts
```diff
@@ -27,6 +33,93 @@ import { generateResyncedProposalEvaluationForCard } from './generateResyncedPro
...
+  cursor?: string | null;
+  spaceId: string;
+}) {
+  const batchSize = 100;
+  const proposalPages = await prisma.page.findMany({
+    where: {
+      spaceId,
+      type: 'proposal',
+      proposal: {
+        status: {
```
mattcasey, line 99: might be more optimal to say status: "published"

Devorein: fair point, updated.

lib/focalboard/updateCardsFromProposals.ts
```diff
@@ -129,33 +125,35 @@ export async function updateCardsFromProposals({
...
-    select: {
-      permissions: true
-    }
-  });
-
-  const proposalBoardBlock = (await prisma.proposalBlock.findUnique({
-    where: {
-      id_spaceId: {
-        id: DEFAULT_BOARD_BLOCK_ID,
+  const [board, rootPagePermissions, proposalBoardBlock] = await prisma.$transaction([
```
mattcasey, line 128: do these in parallel

lib/focalboard/updateCardsFromProposals.ts
```diff
@@ -69,12 +69,7 @@ const pageSelectObject = {
           id: true
         }
       },
-      fields: true,
-      form: {
-        select: {
-          id: true
-        }
-      }
+      fields: true
```
mattcasey, line 72: we didnt need to "JOIN" the form table, we alerady get the formId column

Devorein: good catch

PR #3696, opened 2024-02-22
Token Gate Cron job fix

valentinludu:
Changes:
- fixes for readContract method
- Better logging when a condition fails
- Revert alchemy API's. Sometimes they were failing. Better to rely on Ankr or the public rpcs
- On error we return true (validation) so we don't kick out the user just because an API call failed.

### DIFF
--- a/scripts/litTokenGateAdaptor.ts
+++ b/scripts/litTokenGateAdaptor.ts
@@ -47,7 +47,7 @@ export function transformToTokenGateCondition(conditions: JsonAccsRequest): Toke
       ) {
         return {
           contractAddress: condition.contractAddress,
-          type: condition.standardContractType,
+          type: 'ERC1155',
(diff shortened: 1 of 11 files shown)


mattcasey (approved):

scripts/litTokenGateAdaptor.ts
```diff
@@ -158,7 +158,7 @@ export function transformToTokenGateCondition(conditions: JsonAccsRequest): Toke
       } else if (condition.standardContractType === 'CASK') {
         return {
           contractAddress: condition.contractAddress,
-          type: condition.standardContractType,
+          type: 'CASK',
```
mattcasey, line 161: we can get rid of Cask right?

valentinludu: yes

PR #3700, opened 2024-02-22
Cleanup: only send meta data forward for proposal templates

mattcasey:
Given that we get all the forms and evaluation information currently, we're grabbing an enormous amount of data for spaces like OP that have 10-15 proposal templates (1 MB in size!)


motechFR (approved):
Looks good, especially switching around to get the bare minimum page data, and then load the template on the fly

PR #3701, opened 2024-02-23
Batch update properties in rewards page

Devorein:
### WHAT

* Multi-select rows in the rewards table
* Update reward properties from the table view
* Drag and drop + row action for rewards row


components/common/BoardEditor/focalboard/src/components/propertyValueElement.tsx
```diff
@@ -179,14 +189,123 @@ function PropertyValueElement(props: Props) {
...
+    const reviewers = (card.fields.properties[REWARD_REVIEWERS_BLOCK_ID] as unknown as RewardReviewer[]) ?? [];
+    propertyValueElement = (
+      <UserAndRoleSelect
+        displayType={displayType}
+        readOnly={readOnly || !isAdmin}
+        onChange={(options) => {
+          if (!reward) {
+            return;
+          }
+          const reviewerOptions = options.filter(
```
valentinludu, line 220: You can use `.filter((option): option is RewardReviewer => )` instead of aserting

components/common/BoardEditor/focalboard/src/components/viewHeader/ViewHeaderRowsMenu/components/RewardsDueDatePropertyTemplateMenu.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const propertyValue = cards[0].fields.properties[propertyTemplate.id] || '';
+  const rewardId = pages[cards[0]?.id]?.bountyId;
+
+  if (!rewardId) {
+    return null;
+  }
+
+  return (
+    <PropertyMenu lastChild={lastChild} propertyTemplate={propertyTemplate}>
+      {() => {
```
valentinludu, line 32: Since you don't have any properties you need here, I think you can just render the <Box>...., without this part `{() => {`

valentinludu (approved):
Tested and works great on desktop ad mobile.

Left a couple of small possible tweaks.

components/rewards/components/RewardsHeaderRowsMenu.tsx
```diff
@@ -26,16 +35,148 @@ export function RewardsHeaderRowsMenu({
...
+      message: `Are you sure you want to mark ${checkedRewards.length} ${
+        checkedRewards.length > 1 ? 'rewards' : 'reward'
+      } as complete?`
+    });
+
+    if (!confirmed) {
+      return;
+    }
+    for (const reward of checkedRewards) {
+      await charmClient.rewards.closeReward(reward.id);
```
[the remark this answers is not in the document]
Devorein: It's a bit challenging since we are marking each reward as either paid or complete. We need to call the trigger method from useSWR in each loop iteration, which would cause the disable/isMutating value to shift between on and off periodically, leading to a bit of confusing UX. I went with a state variable to keep track of the mutating state for all the rewards and disabled the button while that was happening

PR #3708, opened 2024-02-26
Create reward templates from proposal page

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/create-reward-templates-from-proposal-page-7711943658318645)

image


mattcasey (approved):
nice, an e2e test would be good

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/RewardSettings.tsx
```diff
@@ -40,6 +74,33 @@ export function RewardSettings({ value, onChange }: RewardSettingsProps) {
         </Typography>
       </FormLabel>
 
+      <NewPageDialog
```
mattcasey, line 77: we should put modals, etc. at the very bottom of the JSX.. makes it easier to find the UI that user sees first

PR #3709, opened 2024-02-26
Combine Identity + connected Accounts settings

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/combine-identity-connected-accounts-settings-3121252963512282)

image


valentinludu (approved):
Looks good!

mattcasey (changes requested):
I think Select and "Set as primary" are redundant right now. Don't care which one we should keep, but i like having a Select button.
- the "..." should never be disabled. either we should completely hide it or let you open it with options disabled+tooltips
- disabled items in the context menu should have tooltips why (i know we dont have anything on them in production either)

image

mattcasey:
Another bug, I think: how can one option be Primary but another is Selected?
image

PR #3714, opened 2024-02-27
Remove Profile form from Settings modal

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/remove-profile-form-from-settings-modal-006326267268503605)


components/_app/GlobalComponents.tsx
```diff
@@ -84,6 +85,7 @@ export function GlobalComponents() {
     <>
       <HexagonalAvatarMask id='hexagon-avatar' />
       <MemberProfileDialogGlobal />
+      <EditMemberProfileDialogGlobal />
```
mattcasey, line 88: can you make this part of te same modal? when i toggle between them there is a flicker of the backdrop. it's the view/edit of the profile so doesnt seem like we need two modals

mattcasey (changes requested):
can you make this part of the same modal as MemberProfile? when i toggle between them there is a flicker of the backdrop. it's the view/edit of the profile so doesnt seem like we need two modals

components/common/PageLayout/components/Sidebar/components/SidebarSubmenu.tsx
```diff
@@ -153,23 +152,27 @@ export default function SidebarSubmenu({
...
-        >
-          <UserDisplay userId={user?.id} hideName gridColumn='1' gridRow='1/3' />
-          <Typography variant='body2' noWrap>
-            {user?.username}
-          </Typography>
-          <Typography variant='body2' color='secondary'>
-            My Profile
-          </Typography>
-        </MenuItem>
+        {user && (
```
mattcasey, line 155: why do we have to check if user exists?

Devorein: Updated

components/members/components/MemberDirectoryGalleryView.tsx
```diff
@@ -68,7 +68,7 @@ function MemberDirectoryGalleryCard({
 
   function onClickEdit(e: MouseEvent<HTMLElement>) {
     e.stopPropagation();
-    openSettings('profile');
+    setIsOpen(true);
```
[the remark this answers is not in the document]
Devorein: Good point, updated in the refactor

PR #3715, opened 2024-02-27
Duplicated page doesn't show in parent page

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/duplicated-page-doesn-t-show-in-parent-page-869352925746254)

### DIFF
--- a/lib/websockets/spaceEvents.ts
+++ b/lib/websockets/spaceEvents.ts
@@ -257,6 +257,63 @@ export class SpaceEventHandler {
         });
         this.sendError(errorMessage);
       }
+    } else if (message.type === 'page_duplicated') {
+      const duplicatedPage = await prisma.page.findUniqueOrThrow({
+        where: {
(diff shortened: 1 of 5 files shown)


lib/websockets/spaceEvents.ts
```diff
@@ -257,6 +257,63 @@ export class SpaceEventHandler {
         });
         this.sendError(errorMessage);
       }
+    } else if (message.type === 'page_duplicated') {
```
mattcasey, line 260: suggestion: move these blocks of code related to page-syncing into a separate place and call them from documentEvents/spaceEvents, to reduce their size

mattcasey (approved):

lib/websockets/spaceEvents.ts
Devorein: They depend a lot on internal methods + `this` thus it will be a bit difficult. Skipping for now.

PR #3719, opened 2024-02-27
Undo deleted pages brings back no access pages

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/cannot-undelete-page-results-in-no-access-53334147529846)


mattcasey (approved):

components/common/CharmEditor/components/fiduswriter/ws.ts
```diff
@@ -337,12 +337,20 @@ export class WebSocketConnector {
         // message is empty
         return;
       }
+      const historyPlugin = this.editor.view.state.plugins.find((plugin) => (plugin as any).key === 'history$');
```
mattcasey, line 340: can you move this logic to the constructor of the client? Its a tad expensive and this is a function that gets called as you are typing

Devorein: Good idea! Updated and its working 🚀

PR #3727, opened 2024-02-27
fix attribution of submitters

mattcasey:
Somehow, we ended up with roleIds inside of assignedSubmitters in a pending reward on proposals. This seems to be possible if you change the application type as an admin, but not sure if thats actually what happened


components/rewards/components/RewardProperties/RewardPropertiesForm.tsx
```diff
@@ -89,7 +89,7 @@ export function RewardPropertiesForm({
   const readOnlyReviewers = !isAdmin && (readOnly || !!template?.reward.reviewers?.length);
   const readOnlyDueDate = !isAdmin && (readOnly || !!template?.reward.dueDate);
   const readOnlyApplicationType =
-    (!isAdmin && (readOnly || !!forcedApplicationType || !!template)) || !!isProposalTemplate;
+    !!forcedApplicationType || (!isAdmin && (readOnly || !!template)) || !!isProposalTemplate;
```
mattcasey, line 92: doesnt make sense for admins to be able to override this setting

PR #3730, opened 2024-02-28
Siwe v2 + remove wallet storage

valentinludu:
### WHAT
- Upgrader de siwe 2
- Removed signature storage
- added @spruceid/siwe-parser because siwe v2 doesn't want to work without it
- remove all remaining lit protocol vars
- verify token gates without user wallet connected if user is loggedin with another form of identity

### WHY
We needed to get rid of lit-siwe and lit stored signature

### TEST CASES
- login
- logout
- reward signup button
- add a new wallet in my account
- join page when a user is not loggedin
- invite page when a user is not logged in
- when logged in, switch wallets for users that have multiple wallets in their account
- when logged in, switch wallet with a new one


mattcasey (approved):

mattcasey (approved):
lots of changes, but tested and seems to work well

PR #3733, opened 2024-02-29
Guild xyz updates

valentinludu:
The main focus here is to store only the guild id
+
Fix import guild roles


mattcasey (approved):
thanks for the cleanup

PR #3741, opened 2024-03-04
Fixed deleting page node in a column deletes whole column node

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/page-1550658364736237)


mattcasey (approved):
+100 thanks for adding a test

PR #3742, opened 2024-03-04
Whitespace for proposal board with description

Devorein:
### WHY
[Conversation](https://discord.com/channels/894960387743698944/1213099680108650556)

Before
image

Now
image


mattcasey (approved):
approved, but its weird to me that you have to set marginLeft on charm editor. i think if the row toggles are not enabled, the editor component should handle that automatically

PR #3746, opened 2024-03-04
Hats Protocol - Support arb and base

valentinludu:


connectors/chains.ts
```diff
@@ -531,11 +531,11 @@ export const ercSupportedChains = RPCList.filter((chain) => !!chain.alchemyUrl |
 );
 
 export const daoChains = RPCList.filter((chain) =>
-  ['ethereum', 'arbitrum', 'optimism', 'polygon', 'gnosis'].includes(chain.chainName.toLowerCase())
+  ['ethereum', 'arbitrum one', 'optimism', 'polygon', 'gnosis'].includes(chain.chainName.toLowerCase())
 );
 
 export const hatsProtocolChains = RPCList.filter((chain) =>
-  ['ethereum', 'arbitrum', 'optimism', 'polygon', 'gnosis', 'ethereum - sepolia', 'celo'].includes(
+  ['ethereum', 'arbitrum one', 'optimism', 'polygon', 'gnosis', 'ethereum - sepolia', 'celo', 'base'].includes(
```
motechFR, line 538: we could use the shortname instead as per EIP-99

Here's the ref
https://github.com/ethereum-lists/chains/tree/master/_data/chains

valentinludu: Good idea. Updated!

mattcasey (approved):

PR #3749, opened 2024-03-04
allow space admins to enable testnets

mattcasey:


connectors/chains.ts
```diff
@@ -458,21 +451,46 @@ const RPC: Record<string, IChainDetails> = {
   }
 } as const;
 
+export const daoChains: IChainDetails['shortName'][] = ['eth', 'arb1', 'oeth', 'matic', 'gno'];
+
+export const hatsProtocolChains: IChainDetails['shortName'][] = [
```
mattcasey, line 456: i think it would be better to use wagmi's ids in the future, that way they are hard-coded and/or we can have Typescript support on them

components/common/TokenGateModal/components/TokenGateUnlockProtocol.tsx
```diff
@@ -12,13 +13,15 @@ export function TokenGateUnlockProtocol() {
     register,
     formState: { errors }
   } = useFormContext<FormValues>();
+  const { space } = useCurrentSpace();
+  const chains = getChainList({ enableTestnets: !!space?.enableTestnets }).filter((chain) => !!chain.hypersubNetwork);
```
valentinludu, line 17: This should be `!!chain.unlockNetwork`

valentinludu (approved):
Looks good.

PR #3758, opened 2024-03-06
Show clicked related pages in modal

Devorein:
### WHY

1. Clicking on related cards inside a modal should open the clicked card in the same modal
2. Show clicked related card in a modal/new page for public databases


motechFR (approved):
Tested both flows, looks good

PR #3760, opened 2024-03-06
Improved layout for user identities section

Devorein:

### DIFF
--- a/components/settings/account/AccountSettings.tsx
+++ b/components/settings/account/AccountSettings.tsx
@@ -16,7 +16,7 @@ export function AccountSettings() {
   return (
     <>
       <Legend>My Account</Legend>
-      <DialogTitle>Select a public identity</DialogTitle>
+      <DialogTitle>Select your identity</DialogTitle>
       <UserIdentities />
       <Box sx={{ mb: 2 }} />
       <AccountPreferences />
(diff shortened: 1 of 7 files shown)


components/settings/profile/components/Integration.tsx
```diff
@@ -51,22 +44,61 @@ function Integration(props: IntegrationProps) {
...
+          }}
+          gap={{
+            xs: 0.5
+          }}
+          alignItems={{
+            xs: 'flex-start',
+            md: 'center'
+          }}
+        >
+          <Tooltip title={secondaryUserName ?? ''}>
```
mattcasey, line 61: you should wrap this inside a box, otherwise the tooltip right now does not appear directly over the name

mattcasey: image

mattcasey (approved):

components/settings/account/AccountSettings.tsx
```diff
@@ -16,7 +16,7 @@ export function AccountSettings() {
   return (
     <>
       <Legend>My Account</Legend>
-      <DialogTitle>Select a public identity</DialogTitle>
+      <DialogTitle>Select your identity</DialogTitle>
```
mattcasey, line 19: DialogTitle is an odd component to use here, it's not a dialog title and we don't use this font style anyplace else in settings.  My suggestion would be to use InputLabel instead

components/settings/profile/components/Integration.tsx
Devorein: Updated. Nice catch!

components/settings/account/AccountSettings.tsx
Devorein: Updated

PR #3765, opened 2024-03-07
Allow copying rubric criteria from templates

mattcasey:

### DIFF
--- a/lib/proposals/rubric/getRubricTemplates.ts
+++ b/lib/proposals/rubric/getRubricTemplates.ts
@@ -28,12 +28,14 @@ export async function getRubricTemplates({ spaceId }: { spaceId: string }): Prom
     }
   });
   return proposals.flatMap((p) => {
-    return (p.proposal?.evaluations || []).map((e) => {
-      return {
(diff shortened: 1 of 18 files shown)


lib/proposals/rubric/getRubricTemplates.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function getRubricTemplates({ spaceId }: { spaceId: string }): Promise<RubricTemplate[]> {
+  const proposals = await prisma.page.findMany({
+    where: {
+      spaceId,
+      type: 'proposal_template'
+    },
+    include: {
+      proposal: {
+        include: {
+          evaluations: {
```
Devorein, line 20: Is there any specific reason why we aren't adding `where: {type: "rubric"}` here rather than filtering it later?

Devorein (approved):
Works great, left a few suggestions

lib/proposals/rubric/getRubricTemplates.ts
mattcasey: good idea(s), thanks!

PR #3770, opened 2024-03-07
use one playwright worker in CI

mattcasey:
I noticed tests were timing out, and only working due to retries. But just today one test is failing very often


__e2e__/publicPage.spec.ts
```diff
@@ -23,10 +23,10 @@ const test = base.extend<Fixtures>({
   databasePage: ({ page }, use) => use(new DatabasePage(page))
 });
 
-test.beforeAll(async () => {
-  // Set headless to false in chromium.launch to visually debug the test
-  browser = await chromium.launch({});
+test.beforeAll(async ({ browser: _browser }) => {
+  browser = _browser;
```
mattcasey, line 27: removed memory-intensive op of opening a new browser instance

motechFR (approved):

PR #3773, opened 2024-03-08
Drop Goerli support

valentinludu:
Goerli will not be supported in the next few months by our RPC providers.


connectors/chains.ts
```diff
@@ -167,19 +166,6 @@ export const RPC: Record<string, IChainDetails> = {
     shortName: 'base-sepolia',
     testnet: true
   },
-  BASE_TESTNET: {
```
valentinludu, line 170: I dropped this one completely since no one was using it as a token gate chain.

Devorein (approved):

PR #3781, opened 2024-03-11
Manage identities in Join Page

valentinludu:
Users can check and add identities in Join Page


mattcasey (approved):

mattcasey:
surprised it was so simple

PR #3782, opened 2024-03-11
Proposal token voting

Devorein:


lib/blockchain/getTokenAmountOnBlockNumber.ts
```diff
@@ -0,0 +1,109 @@
+import { log } from '@charmverse/core/log';
+
+import * as http from 'adapters/http';
+
+import { getTokenDecimals } from './getTokenDecimals';
+
+const SNAPSHOT_SCORE_URL = 'https://score.snapshot.org';
```
mattcasey, line 7: please add a comment about where this url came from, even if its just "endpoint used by Snapshot app"

lib/blockchain/getTokenSupplyAmount.ts
```diff
@@ -0,0 +1,29 @@
...
+  chainId,
+  tokenContractAddress
+}: {
+  tokenContractAddress: string;
+  chainId: number;
+}) {
+  const tokenDecimal = await getTokenDecimals({ chainId, tokenContractAddress });
+  const chain = getChainById(chainId);
+  if (!chain || tokenDecimal === null) {
+    return null;
```
mattcasey, line 16: this should probably throw an error?

lib/proposals/createProposal.ts
```diff
@@ -139,6 +140,12 @@ export async function createProposal({
     proposalFormId = await createForm(formFields);
   }
 
+  const evaluationsWithBlockNumber = await getVoteEvaluationStepsWithBlockNumber({
+    evaluations: evaluations.map((evaluation, index) => ({ ...evaluation, id: evaluationIds[index] })),
+    isDraft: !!isDraft,
+    proposalType: pageProps.type ?? 'page'
```
mattcasey, line 146: "page" does not make sense for a proposalType.

scripts/migrations/2024_03_12_updateVoteStrategy.ts
```diff
@@ -0,0 +1,55 @@
...
+      voteSettings: true
+    }
+  })
+
+  let current = 0;
+
+  for (const voteProposalEvaluation of voteProposalEvaluations) {
+    try {
+      const vote = voteProposalEvaluation.vote;
+      if (vote && (voteProposalEvaluation.voteSettings as any).publishToSnapshot) {
```
mattcasey, line 24: please set strategy on all vote evaluations

lib/blockchain/getTokenAmountOnBlockNumber.ts
```diff
@@ -0,0 +1,109 @@
+import { log } from '@charmverse/core/log';
```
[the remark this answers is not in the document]
Devorein: Moved there, coincidentally there is a `getVotingPower` file in there, but it requires an actual snapshot proposal id so its tightly integrated with snapshot. Kept the same name as before.

lib/blockchain/getTokenAmountOnBlockNumber.ts
Devorein: Added.

lib/blockchain/getTokenSupplyAmount.ts
Devorein: Yeah sure, added a log to keep track of the chain id and token address combo

lib/proposals/createProposal.ts
Devorein: yeah that sounds good, made pageType optional. I think I deliberately added page so that it returns the evaluations as is but its redundant.

scripts/migrations/2024_03_12_updateVoteStrategy.ts
Devorein: Nice catch. Updated.

PR #3783, opened 2024-03-11
Fix gitcoin token gate crashes

valentinludu:


mattcasey (approved):

lib/tokenGates/humanizeConditions.ts
```diff
@@ -27,7 +27,7 @@ export function humanizeConditionsData(conditions: TokenGate['conditions']): Hum
     const chainDetails = getChainById(Number(acc.chain));
     const tokenSymbol = chainDetails?.viem.nativeCurrency.symbol;
     const quantity = acc.quantity;
-    const balance = Number(quantity) ? formatEther(BigInt(quantity)) : quantity;
+    const balance = Number(quantity) && Number(quantity) % 1 === 0 ? formatEther(BigInt(quantity)) : quantity;
```
mattcasey, line 30: maybe add a comment why we need this check. I want to point out we do have a utility "isNumber" inside of lib/utils/number. what would this check be? "isBigIntString"?

PR #3790, opened 2024-03-13
Fix loading polygon

motechFR:
### WHAT

Readded alchemy url so we can pull polygon NFTs


valentinludu (approved):

motechFR:
Closing this as fix included in #3791

PR #3793, opened 2024-03-13
Rewards submissions public endpoint

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/api-for-rewards-5303771278893195)


pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+        createdAt: submission.createdAt,
+        rewardId: submission.bountyId
+      };
+
+      const isSubmission = submission.bounty.approveSubmitters ? submissionStatuses.includes(submission.status) : false;
+      return {
+        ...(isSubmission
+          ? {
+              submission: application,
+              application: null
```
mattcasey, line 163: we should still return the application that came w/the submission

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  return res.status(200).json(
+    submissions.map((submission) => {
+      const application = {
+        id: submission.id,
+        createdAt: submission.createdAt,
+        rewardId: submission.bountyId
+      };
+
+      const isSubmission = submission.bounty.approveSubmitters ? submissionStatuses.includes(submission.status) : false;
```
mattcasey, line 158: where is this logic from? (application vs submission). is this also used in charmverse somewhere/could we have a standard method?

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,176 @@
...
+      status: true,
+      applicant: {
+        select: userProfileSelect
+      }
+    }
+  });
+
+  return res.status(200).json(
+    submissions.map((submission) => {
+      const application = {
```
mattcasey, line 152: These 3 fields are actually metadata that apply to both application and submission, so i think they just belong at the top. But we're missing the content of applications and submissions which is important. What do you think about this  structure?
```
{
  id: string
  createdAt: string
  author
  status: ApplicationStatus
  submission?: { content: { text, markdown } }
  application?: { content: { text } }
}
```
- I'm open to just always returning an empty string for submission and application too. might be simpler to consume
- I think it makes sense to return 'status' but we probably need to clean the values up a bit, for instance "inProgress" should be mapped to "in_progress," and maybe "in_review" instead of just "review"

pages/api/v1/rewards/submissions.ts
mattcasey: This was my first comment, after making the comment above, maybe we just always return application/submission but also include the status in case they want it.

lib/public-api/getPageApi.ts
```diff
@@ -44,6 +49,7 @@ export type PublicApiPage = {
   };
   title: string;
   url: string;
+  author: UserProfile;
```
mattcasey, line 52: thanks for adding this too

mattcasey (changes requested):
left comments

pages/api/v1/rewards/submissions.ts
Devorein: Yeah that sounds good. Responding to the first comment we do have logic to differentiate between applications and submissions in this component `components/rewards/components/RewardApplications/RewardApplications.tsx#L136`

pages/api/v1/rewards/submissions.ts
Devorein: Yeah that sounds good and agree with mapping `inProgress` to `in_progress` and same for review.

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+      text: string;
+      markdown: string;
+    };
+  };
+  application?: {
+    content: {
+      text: string;
+    };
+  };
+  status: Omit<ApplicationStatus, 'inProgress' | 'review'> | 'in_progress' | 'in_review';
```
mattcasey, line 112: please use Pick, not Omit

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+ *                type: array
+ *                items:
+ *                  $ref: '#/components/schemas/PublicApiSubmission'
+ */
+async function getSubmissions(req: NextApiRequest, res: NextApiResponse<PublicApiSubmission[]>) {
+  const spaceId = req.query.spaceId as string;
+  const rewardId = req.query.rewardId as string;
+  const authorizedSpaceId = req.authorizedSpaceId;
+
+  if (!spaceId && !rewardId) {
```
mattcasey, line 150: we don't require a spaceId on the "Space API" if you look at most other endpoints. i think it's only required if you're using a partner key

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+    throw new InvalidInputError('Missing spaceId or rewardId');
+  }
+
+  const submissions = await prisma.application.findMany({
+    where: {
+      ...(spaceId ? { spaceId } : { bountyId: rewardId, spaceId: authorizedSpaceId }),
```
mattcasey, line 156: this is unsafe - we should only let you query by authorizedSpaceId

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+                  text: submission.message
+                }
+              }
+            : undefined,
+          credentials: submission.issuedCredentials
+            .map((credential) => {
+              return credential.ceramicRecord;
+            })
+            .filter(isTruthy)
+            .map((ceramicRecord: any) => {
```
mattcasey, line 230: where is this type actually? this type-casting seems very dangerous especially in our API

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+ *        attester:
+ *          type: string
+ *        recipient:
+ *          type: string
+ *        schemaId:
+ *          type: string
+ *        timeCreated:
+ *          type: number
+ *        chainId:
+ *          type: string
```
mattcasey, line 39: pretty sure chainId is a number type

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+ *        type:
+ *          type: string
+ *          enum: [onchain, charmverse, gitcoin]
+ *        verificationUrl:
+ *          type: string
+ *          nullable: true
+ *        iconUrl:
+ *          type: string
+ *          nullable: true
+ *        issuedCredentialId:
```
mattcasey, line 49: rename to "id"

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+ *          type: string
+ *        recipient:
+ *          type: string
+ *        schemaId:
+ *          type: string
+ *        timeCreated:
+ *          type: number
+ *        chainId:
+ *          type: string
+ *        type:
```
mattcasey, line 40: "type" is confusing here; it could be both charmverse and on-chain, i'm not sure why we went with this enum. rename to "source" ?

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+ *          example: 3fa85f64-5717-4562-b3fc-2c963f66afa6
+ *        content:
+ *          type: object
+ *        attester:
+ *          type: string
+ *        recipient:
+ *          type: string
+ *        schemaId:
+ *          type: string
+ *        timeCreated:
```
mattcasey, line 36: This should be "createdAt": "string"

pages/api/v1/rewards/submissions.ts
```diff
@@ -0,0 +1,240 @@
...
+
+const handler = apiHandler();
+
+handler.get(getSubmissions);
+
+/**
+ * @swagger
+ * components:
+ *  schemas:
+ *    EASAttestationFromApi:
```
mattcasey, line 21: this type wont make sense to non-charmverse people. i'm not even sure what "FromApi" means. whats a better name?

pages/api/v1/rewards/submissions.ts
Devorein: Good point, updated.

pages/api/v1/rewards/submissions.ts
Devorein: yep. Updated.

pages/api/v1/rewards/submissions.ts
Devorein: Updated

pages/api/v1/rewards/submissions.ts
Devorein: Correct

pages/api/v1/rewards/submissions.ts
Devorein: done

pages/api/v1/rewards/submissions.ts
Devorein: This will be removed since for charmverse sourced credentials issuedCredentialId is equal to id

pages/api/v1/rewards/submissions.ts
Devorein: Get go with `EASCredential`

pages/api/v1/rewards/submissions.ts
Devorein: Updated

PR #3801, opened 2024-03-14
Improvements for proposal token voting

Devorein:
### WHAT

Please read commit messages!


lib/proposals/createProposal.ts
```diff
@@ -240,10 +240,13 @@ export async function createProposal({
       ? await upsertProposalFormAnswers({ proposalId, answers: formAnswers })
       : null;
 
-  await createVoteIfNecessary({
-    createdBy: userId,
-    proposalId
-  });
+  // make sure proposal is published before we create a vote
+  if (!isDraft) {
```
Devorein, line 244: This was a bug, if your first step was a vote and you first create a draft then published it, it threw an error

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/EvaluationStepSettings.tsx
```diff
@@ -54,6 +60,15 @@ export function EvaluationStepSettings({
     });
   }
 
+  useEffect(() => {
+    if (evaluation.type === 'vote' && evaluation.voteSettings?.strategy === 'token') {
```
mattcasey, line 64: suggestion: just define this outside:
```
const isTokenVoting = evaluation.type === 'vote' && evaluation.voteSettings?.strategy === 'token'
```
and use isTokenVoting as a dependency. i don't think u need an extra state

lib/votes/getVotingPowerForVotes.ts
```diff
@@ -20,14 +20,29 @@ export async function getVotingPowerForVotes({
     }
   });
 
+  const user = await prisma.user.findUniqueOrThrow({
```
mattcasey, line 23: suggestion: request the user's wallets here too, instead of doing two queries to prisma

mattcasey (approved):

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/EvaluationStepSettings.tsx
Devorein: Updated.

lib/votes/getVotingPowerForVotes.ts
Devorein: Good one. Updated.

PR #3808, opened 2024-03-15
Use BigInt to convert token id to decimal string

Devorein:
### WHY

[Discussion](https://discord.com/channels/894960387743698944/1217815145296957533)

### DIFF
--- a/lib/blockchain/getNFTs.ts
+++ b/lib/blockchain/getNFTs.ts
@@ -38,7 +38,7 @@ export type SupportedChainId = SupportedChainIdByAlchemy | SupportedChainIdByAnk
 export type NFTData = {
   id: string;
   tokenId: string;
-  tokenIdInt: number | null;
+  tokenIdInt: string | null;
   contract: string;
   title: string;
   description: string;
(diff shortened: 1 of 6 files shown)


lib/blockchain/getNFTs.ts
```diff
@@ -38,7 +38,7 @@ export type SupportedChainId = SupportedChainIdByAlchemy | SupportedChainIdByAnk
 export type NFTData = {
   id: string;
   tokenId: string;
```
valentinludu, line 40: I understand that `tokenIdInt` can't be a number anymore.

If `tokenId` is already a string then tokenIdInt is not necessary anymore or can be a BigInt.
In your current situation `tokenIdInt` will be the same as `tokenId`, right?

Devorein: not necessarily since tokenId could be a hex value, `tokenIdInt` is guaranteed to be the decimal representation. But we can probably make use of a single one since hex values are not used.

motechFR (approved):

lib/blockchain/getNFTs.ts
motechFR: Good suggestion, let's see if we can simplify to just 1 key

valentinludu: This is what I was thinking too. Alchemy doesn't return hex values from what I know but if you want to be sure you could just do
`tokenId: BigInt(nft.tokenId).toString() and you're good.`

Devorein: let's do that then. I will update and merge. thnx

valentinludu (approved):
Looks good

PR #3819, opened 2024-03-19
Project profile & member form

Devorein:


components/projects/ProjectMemberFields.tsx
```diff
@@ -0,0 +1,125 @@
...
+  | 'previousProjects'
+  | 'telegram'
+  | 'twitter'
+  | 'walletAddress'
+  | 'warpcast'
+>;
+
+export type ProjectMemberField = keyof ProjectMemberPayload;
+
+export const ProjectMemberFieldProperties: ProjectFieldProperty[] = [
```
mattcasey, line 28: this should be either lowercase camelcase or all-caps snake case. Capital names should still be reserved for types, classes, and components

components/projects/interfaces.ts
mattcasey, line 1: there should not exist a components/projects folder, because projects is not an actual page path. I think it would make sense to put all of this inside settings/projects for now, except for the things specific to proposals, they would go inside proposals

components/projects/interfaces.ts
```diff
@@ -0,0 +1,49 @@
+import type { Project, ProjectMember } from '@charmverse/core/prisma-client';
+
+import type { ProjectPayload } from './ProjectFields';
+import type { ProjectMemberPayload } from './ProjectMemberFields';
+
+export type ProjectFieldConfig = Partial<
```
mattcasey, line 6: this file could probably live in lib/projects. so that lib is not importing from components

components/settings/projects/hooks/useGetDefaultProject.ts
```diff
@@ -0,0 +1,31 @@
+import { projectDefaultValues } from 'components/projects/ProjectFields';
+import { projectMemberDefaultValues } from 'components/projects/ProjectMemberFields';
+import { useMembers } from 'hooks/useMembers';
+import { useUser } from 'hooks/useUser';
+import type { TelegramAccount } from 'pages/api/telegram/connect';
+
+export function useGetDefaultProject() {
```
mattcasey, line 7: this should be a utility method that takes user and members as an input, not a stateful hook

components/projects/interfaces.ts
Devorein: Alright updated.

components/settings/projects/hooks/useGetDefaultProject.ts
Devorein: Good point, updated.

PR #3825, opened 2024-03-20
include page metadata on blocks returned by the API

mattcasey:
### WHAT

Update to backend APIs to return page information

### WHY

Do this first, so we can then update the frontend to use the data without a breaking change

Interface notes:
- "BlockWithDetails" is designed to be a replacement for "Block" from prisma (backend)
- "UIBlockWithDetails" is designed to be a replacement for "Block" from focalboard (frontend)


pages/api/blocks/[id]/subtree.ts
```diff
@@ -26,80 +27,34 @@ export const config = {
 };
 
 async function getBlockSubtree(req: NextApiRequest, res: NextApiResponse<Block[] | { error: string }>) {
-  const blockId = req.query.id as string;
+  const pageId = req.query.id as string;
+
   const page = await prisma.page.findFirstOrThrow({
     where: {
-      OR: [{ boardId: blockId }, { cardId: blockId }]
+      id: pageId
```
Comment (mattcasey) [inline pages/api/blocks/[id]/subtree.ts:34]: turns out we are using pageId on the frontend and even in our tests, so i updated it. but we could maybe put this under pages/[pageId]/blocks.ts.

pages/api/blocks/[id]/subtree.ts
```diff
@@ -26,80 +27,34 @@ export const config = {
...
+    resourceId: pageId,
     userId: req.session.user?.id
   });
 
   if (computed.read !== true) {
     return res.status(404).json({ error: 'page not found' });
   }
-  const blocks = await prisma.block.findMany({
-    where: {
-      OR: [{ id: blockId }, { rootId: blockId }, { parentId: blockId }]
```
Comment (mattcasey) [inline pages/api/blocks/[id]/subtree.ts:49]: im not sure why we need to query by parentId. it doesn't have an index so i removed it until we have a reason

testing/setupDatabase.ts
```diff
@@ -972,7 +972,6 @@ export async function generateProposal({
 export async function generateBoard({
   createdBy,
   spaceId,
-  parentId,
```
mattcasey, line 975: we were only using parentId for testing linked databasese. It does not make sense to me why we would pass a pageId as parentId to a block. From what i could tell, we don't use parentId in the actual code for inline db's, this relationship was only being used by the e2e test

PR #3831, opened 2024-03-22
Don't rely on page context for card metadata

mattcasey:
- removed usePages() inside of DatabaseEditor component, except a few places where we need it for boards or rewards
- gets rid of "CardPage" interface, altho I made a new one "CardWithRelations" when we are passing around reward or proposal info


pages/api/blocks/[id]/subtree.ts
```diff
@@ -51,8 +53,27 @@ async function getBlockSubtree(req: NextApiRequest, res: NextApiResponse<BlockWi
   }
 
   const { blocks } = await getRelatedBlocks(blockId);
-
-  return res.status(200).json(blocks);
+  const block = blocks.find((b) => b.id === blockId);
+  if ((block?.fields as BoardFields).sourceType === 'proposals') {
+    // Filter blocks based on proposal permissions
+    const permissionsById = await permissionsApiClient.proposals.bulkComputeProposalPermissions({
```
Comment (motechFR) [inline pages/api/blocks/[id]/subtree.ts:59]: Looks good, this returns a map of proposalIds, so just got to watch out for using those as 1:1 for pageIds

motechFR (approved):

PR #3835, opened 2024-03-22
update template script to export all space data

mattcasey:


motechFR (approved):
Also updated screenshots

PR #3845, opened 2024-03-26
Copy block link & Scroll to linked block

Devorein:
### WHY

[]

### DIFF
--- a/components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
+++ b/components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
@@ -1,5 +1,5 @@
 import { history } from '@bangle.dev/base-components';
-import { EditorState } from '@bangle.dev/pm';
+import { EditorState, NodeSelection } from '@bangle.dev/pm';
 import type { Plugin } from '@bangle.dev/pm';
 import { objectUid } from '@bangle.dev/utils';
 import { log } from '@charmverse/core/log';
@@ -145,6 +145,26 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
       editor.view.dispatch(editor.view.state.tr.setMeta(threadPluginKey, threadIds));
     }
   }, [(threadIds ?? []).join(','), isLoadingRef.current]);
+
+  useEffect(() => {
+    const hash = window.location.hash.slice(1);
+    if (hash && editor && !isLoadingRef.current) {
+      let nodePos: number | undefined;
+      editor.view.state.doc.descendants((node, pos) => {
+        if (node.attrs.id === hash) {
+          nodePos = pos;
+          return false;
+        }
+      });
+
+      const domNode = document.getElementById(hash);
+      if (domNode && nodePos !== undefined) {
+        editor.view.dispatch(editor.view.state.tr.setSelection(NodeSelection.create(editor.view.state.doc, nodePos)));
+        domNode.scrollIntoView();
+        window.history.replaceState({}, '', window.location.href.split('#')[0]);
+      }
+    }
+  }, [isLoadingRef.current]);
 
   function _onConnectionEvent(_editor: CoreBangleEditor, event: ConnectionEvent) {
     if (onConnectionEvent) {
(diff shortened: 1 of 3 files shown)


components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -146,6 +147,26 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
     }
   }, [(threadIds ?? []).join(','), isLoadingRef.current]);
 
+  useEffect(() => {
+    const hash = window.location.hash.slice(1);
+    if (hash && editor && !isLoadingRef.current) {
```
mattcasey, line 152: please put all this logic into the heading component folder

components/common/CharmEditor/components/@bangle.dev/react/ReactEditor.tsx
```diff
@@ -146,6 +147,26 @@ export const BangleEditor = React.forwardRef<CoreBangleEditor | undefined, Bangl
...
+          nodePos = pos;
+          return false;
+        }
+      });
+
+      const domNode = nodePos ? (editor.view.domAtPos(nodePos)?.node as HTMLElement) : null;
+      if (domNode && nodePos !== undefined) {
+        editor.view.dispatch(editor.view.state.tr.setSelection(NodeSelection.create(editor.view.state.doc, nodePos)));
+        domNode.scrollIntoView();
+        window.history.replaceState({}, '', window.location.href.split('#')[0]);
```
mattcasey, line 165: i dont think we should remove the hash

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
```diff
@@ -66,6 +72,29 @@ function Component({ menuState }: { menuState: PluginState }) {
...
+    }
+
+    // calculate the node at the mouse position. do it on click in case the content has changed
+    const topPos = view.state.doc.resolve(rowPosition);
+    const node = topPos.node();
+
+    if (node && node.type.name === 'heading') {
+      const text = node.textContent;
+
+      copyFn(`${window.location.href.split('#')[0]}#${slugify(text)}`)
```
mattcasey, line 89: should use window.lcoation.hash, right?

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
```diff
@@ -66,6 +72,29 @@ function Component({ menuState }: { menuState: PluginState }) {
...
+        .then(() => {
+          showMessage('Link copied to clipboard', 'success');
+        })
+        .catch(() => {});
```
mattcasey, line 93: if you catch it at least log the error

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
```diff
@@ -66,6 +72,29 @@ function Component({ menuState }: { menuState: PluginState }) {
...
 
+  async function copyLinkToBlock() {
+    const rowPosition = menuState.rowPos;
+    if (rowPosition === undefined) {
+      popupState.close();
+      return null;
+    }
+
+    // calculate the node at the mouse position. do it on click in case the content has changed
+    const topPos = view.state.doc.resolve(rowPosition);
```
mattcasey, line 83: why are you not using getNodeForRowPosition?

mattcasey (changes requested):
Bugs found:
- there is always "link to clipboard", even if a row does not have a heading. It seems to show the last heading row i hovered over
- "👋 Welcome to CharmVerse!" turns into #-welcome-to-charmverse. Ideally it shouldn't start with a hyphen
- the page doesn't actually scroll all the way to the heading when it's far down, it only goes about 400px down. it jumps down as soon as the content loads. I would try adding a setTimeout (of even just 0ms) before scrolling. Tested on Firefox and Chrome, adding a heading to the end of the Getting Started page as well as my own custom page. But let me know if u want a video

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
Devorein: Good catch. Updated.

Devorein: There is an even better approach

```js
const url = new URL(window.location.href);
url.hash = '';
const urlWithoutHash = url.toString();
```

components/common/CharmEditor/components/rowActions/RowActionsMenu.tsx
Devorein: It was unfortunately not getting the currently hovered block.  And a simple resolve seemed to have worked.

PR #3849, opened 2024-03-27
Fix rewards rollup

motechFR:
### WHAT
Rewards were moved to complete status, even though they still had open slots


mattcasey (approved):
thanks

PR #3853, opened 2024-03-27
Fix @ links for proposal templates

motechFR:
### WHAT

This should enable using formal proposals, linking them as pages


mattcasey (changes requested):
this is a ton of hacky logic, i thought you would have just adde dback proposal_templates to the types in getAccessibleIds in permissions? It look slike we're adding another query for all pages. I'd rather wait until we add the search in teh UI

PR #3854, opened 2024-03-27
retrieve pages for mentions like we do for nested pages and linked pages

mattcasey:
This is a cleanup + fix so that mentions query page details just like page links

### DIFF
--- a/components/common/CharmEditor/components/mention/components/Mention.tsx
+++ b/components/common/CharmEditor/components/mention/components/Mention.tsx
@@ -11,7 +11,7 @@ import { usePages } from 'hooks/usePages';
 import { useRoles } from 'hooks/useRoles';
 import { isUUID } from 'lib/utils/strings';
 
-import { useGetPageDetails } from '../../../hooks/useGetPageDetails';
+import { useGetPageMetaFromCache } from '../../../hooks/useGetPageMetaFromCache';
 import type { NodeViewProps } from '../../@bangle.dev/core/node-view';
 import type { MentionSpecSchemaAttrs } from '../mention.specs';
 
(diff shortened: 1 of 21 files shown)


components/common/CharmEditor/components/nestedPage/components/NestedPage.tsx
```diff
@@ -58,25 +56,18 @@ function resetPageNodeDropPluginState(view: EditorView) {
...
-  const { data: sourcePage, isLoading: isPageLoading } = useGetPageMeta(
-    !pageFromPagesContext && isDocumentPath && node.attrs.id
-  );
-
-  const documentPage = sourcePage || pageFromPagesContext;
-  const isLoading = isPageLoading && loadingPages;
+  const { page: documentPage, isLoading } = useGetPageDetails({
+    pageId: isDocumentPath ? node.attrs.id : null
+  });
+  const isProposalTemplate = documentPage?.type === 'proposal_template';
```
motechFR, line 70: Quick UX pointer; Feedback we received was that people were using "mentioned page" instead of selecting a template, because the "Submit Proposal" prefix didn't look good

mattcasey: please share that with chris

components/common/CharmEditor/components/mention/components/Mention.tsx
```diff
@@ -36,14 +37,17 @@ export default function Mention({ node }: NodeViewProps) {
   const { showUserProfile } = useMemberProfileDialog();
   const attrs = node.attrs as MentionSpecSchemaAttrs;
   const { getMemberById } = useMembers();
-  const { pages } = usePages();
   const member = getMemberById(attrs.value);
   const { roles } = useRoles();
   const { space } = useCurrentSpace();
 
+  const isDocumentPath = attrs.type === 'page';
+  const { page, isLoading } = useGetPageDetails({
```
mattcasey, line 45: this allows us to retrieve titles of cards/templates that do not live in usePAges()

PR #3866, opened 2024-03-28
support relation to parent page

mattcasey:


pages/api/spaces/[id]/pages.ts
```diff
@@ -67,17 +70,26 @@ async function getPages(req: NextApiRequest, res: NextApiResponse<PageMeta[]>) {
...
+        }
+      }
     }
   });
 
   pages.forEach((page) => {
     page.galleryImage = replaceS3Domain(page.galleryImage);
     page.headerImage = replaceS3Domain(page.headerImage);
     page.icon = replaceS3Domain(page.icon);
+    // hide children of pages that do not belong in the sidebar
```
Comment (mattcasey) [inline pages/api/spaces/[id]/pages.ts:86]: This is where we will determine if a page should be hidden in the navigation (based on its parent page's type)

PR #3867, opened 2024-03-29
fix relative date util + test

mattcasey:


mattcasey:
accidentally merged before tests passed, but letting it go thru

PR #3871, opened 2024-03-29
Questbook grant proposals

valentinludu:
Create credentials for questbook approved proposals


lib/questbook/createProjectCredentials.ts
```diff
@@ -0,0 +1,83 @@
+import { log } from '@charmverse/core/log';
+import { prisma } from '@charmverse/core/prisma-client';
+import { getAddress } from 'viem';
+import { optimism } from 'viem/chains';
+
+import { signAndPublishCharmverseCredential } from 'lib/credentials/attestOffchain';
+import type { ExternalProjectMetadata } from 'lib/credentials/schemas/external';
+import { getProjectOwners } from 'lib/gitcoin/getProjectDetails';
```
motechFR, line 8: Question; Does getProjectOwners only return safes and are all the projects owned by safes?

Looking at the code, it seems an EOA wallet would fall through and not be returned

lib/questbook/createProjectCredentials.ts
```diff
@@ -0,0 +1,83 @@
...
+                  Source: 'Questbook',
+                  Event: 'Approved',
+                  GrantRound: application.grantTitle,
+                  Date: credentialDate,
+                  GrantURL: '',
+                  URL: application.proposalUrl
+                }
+              },
+              chainId: optimism.id,
+              recipient: owner
```
motechFR, line 70: A unit test here where we mock external dependencies & API response would we good to confirm the call data

In particular, we want to ensure the owner is saved to lowercase

lib/questbook/createProjectCredentials.ts
```diff
@@ -0,0 +1,83 @@
...
+
+          try {
+            const cred = await signAndPublishCharmverseCredential({
+              credential: {
+                type: 'external',
+                data: {
+                  Name: application.projectName,
+                  ProjectId: externalProject.id,
+                  Source: 'Questbook',
+                  Event: 'Approved',
```
motechFR, line 62: Suggestion; Rename to "Grant Approved" for better context

motechFR (approved):

lib/questbook/createProjectCredentials.ts
valentinludu: getSafeOwners returns all addresses in lowercase

lib/questbook/createProjectCredentials.ts
valentinludu: Done

PR #3872, opened 2024-03-29
use a different port for permissions in test mode by default

mattcasey:
You can already do this with env files by setting PERMISSIONS_API_URL in webapp and PORT in the permissions folder, but it's not the default.

This way you can run a dev and test versions of permissions at once. I'm changing the port in CI too just for consistency


mattcasey:
since its doable by env files alerady, i figure not waste time trying to make this work or breaking peopele's envs

PR #3892, opened 2024-04-03
Project multiple wallet addresses

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/project-wallet-per-chain-9395122206771407)

### DIFF
--- a/components/settings/projects/ProjectsSettings.tsx
+++ b/components/settings/projects/ProjectsSettings.tsx
@@ -29,7 +29,7 @@ import type { ProjectWithMembers } from 'lib/projects/interfaces';
 
 import { CreateProjectForm } from './components/CreateProjectForm';
 import { ProjectFormAnswers } from './components/ProjectForm';
-import { useProjectForm } from './hooks/useProjectForm';
+import { convertToProjectValues, useProjectForm } from './hooks/useProjectForm';
 
 function ProjectRow({
   projectWithMembers,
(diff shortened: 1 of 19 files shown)


components/settings/projects/ProjectsSettings.tsx
```diff
@@ -82,48 +82,35 @@ function ProjectRow({
...
         ...projectValues,
-        id: projectWithMembers.id,
         projectMembers: projectValues.projectMembers.map((member, index) => ({
           ...projectWithMembers.projectMembers[index],
           ...member
         }))
-      };
-
-      updateProject(updatedProject);
+      }).then((updatedProjectWithMembers) => {
```
valentinludu, line 91: You can use the official swr API
e.g. updateProject({}, { onSuccess: () => {} })

components/settings/projects/hooks/useProjectForm.ts
```diff
@@ -14,36 +14,67 @@ import type {
...
+            if (isRequired && (!addressChainCombos || addressChainCombos.length === 0)) {
+              return false;
+            }
+
+            return addressChainCombos.every((addressChainCombo) => {
+              const { address } = addressChainCombo;
+              if (!isRequired && address === '') {
+                return true;
+              }
+              return ethers.utils.isAddress(address) || address.endsWith('.eth');
```
valentinludu, line 48: Better to use `isAddress` from `viem`.

In general we want to move from ethers to viem where possible.

components/settings/projects/hooks/useProjectForm.ts
valentinludu, line 1: Not sure if possible, but from what I see you could use `when()` and `context` to have the same behaviour of creating different types of validation based on other fields.

A good example can be found here `useCollectablesForm.tsx`.

I know this could be a very big change, so it's more of a nice to have.

PR #3902, opened 2024-04-04
allow sorting member directory

mattcasey:
### WHAT

New Sort options and clean up for member directory header

### WHY

We want to let a space view all members based on join date


Devorein (approved):
@mattcasey, that was a nice cleanup, but the sort for the member directory doesn't persist. I don't know if that's intentional or not but it seems like it should be stored on localstorage at least.

PR #3905, opened 2024-04-05
External Credentials - remove demo grants and questbook title

valentinludu:


mattcasey (approved):

lib/credentials/queriesAndMutations.ts
```diff
@@ -290,11 +290,17 @@ export async function getExternalCredentialsByWallets({
...
+    .map((credential) => ({
+      ...credential,
+      iconUrl: null,
+      favoriteCredentialId: null,
+      index: -1,
+      issuedCredentialId: undefined
+    }))
+    .filter(
+      (c) =>
+        !(c.content?.Name as string)?.toLowerCase().includes('test') &&
```
mattcasey, line 303: it would be a bit nicer if you defined a blacklist up top, maybe even using regex:
```
const blacklistedNameRegexes = [/test/i, /demo/i];
const filtered = credentials.filter((r) => !blacklistedNameRegexes.test('test'));
```

valentinludu: Done

PR #3907, opened 2024-04-08
Select project member in project form

Devorein:
### WHY

[Discussion](https://app.charmverse.io/charmverse/select-team-members-per-proposal-7074687353259137)

### DIFF
--- a/components/[pageId]/DocumentPage/DocumentPage.tsx
+++ b/components/[pageId]/DocumentPage/DocumentPage.tsx
@@ -7,6 +7,7 @@ import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { FormProvider } from 'react-hook-form';
 import { useResizeObserver } from 'usehooks-ts';
 
+import { useGetProposalFormFieldAnswers } from 'charmClient/hooks/proposals';
 import { useGetReward } from 'charmClient/hooks/rewards';
 import { CharmEditor } from 'components/common/CharmEditor';
 import { CardPropertiesWrapper } from 'components/common/CharmEditor/CardPropertiesWrapper';
@@ -107,7 +108,9 @@ function DocumentPageComponent({
   const { proposal, refreshProposal, onChangeEvaluation, onChangeWorkflow, onChangeRewardSettings } = useProposal({
     proposalId
   });
-
+  const { data: proposalFormFieldAnswers = [], isLoading: isLoadingAnswers } = useGetProposalFormFieldAnswers({
+    proposalId
+  });
   // We can only edit the proposal from the top level
   const readonlyProposalProperties = !page.proposalId || readOnly || !!proposal?.archived;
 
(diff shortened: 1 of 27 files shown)


components/[pageId]/DocumentPage/DocumentPage.tsx
```diff
@@ -107,7 +108,9 @@ function DocumentPageComponent({
   const { proposal, refreshProposal, onChangeEvaluation, onChangeWorkflow, onChangeRewardSettings } = useProposal({
     proposalId
   });
-
+  const { data: proposalFormFieldAnswers = [], isLoading: isLoadingAnswers } = useGetProposalFormFieldAnswers({
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/DocumentPage.tsx:111]: this should just be retrieved as part of useProposal, this is loading a lot of unecessary data

components/settings/projects/components/ProjectMemberFields.tsx
```diff
@@ -29,14 +25,7 @@ export function ProjectMemberFieldAnswers({
       fieldConfig={fieldConfig}
       properties={projectMemberFieldProperties}
       onChange={(updatedProjectMemberPayload) => {
```
mattcasey, line 27: you could just pass onChange now, in instead of making a new function?

components/settings/projects/hooks/useProjectUpdates.ts
```diff
@@ -55,54 +99,63 @@ export function useProjectUpdates({ projectId }: { projectId: MaybeString }) {
...
+          ...project,
+          projectMembers: [...project.projectMembers, createdProjectMember]
+        });
+        mutate(
+          (projects) => {
+            if (!projects) {
+              return projects;
+            }
+
+            return projects.map((_project) => {
```
mattcasey, line 133: you should DRY this with a utility method

lib/projects/patchProject.ts
```diff
@@ -0,0 +1,28 @@
...
+export type UpdateProjectPayload = Partial<ProjectPayload> & {
+  id: string;
+};
+
+export async function patchProject({ projectValues }: { projectValues: UpdateProjectPayload }) {
+  return prisma.project.update({
+    where: {
+      id: projectValues.id
+    },
+    data: {
```
mattcasey, line 14: suggestion: just spread all the properties? sometimes it helps to spell out each field individually, but in this case it just seems like we will end up having to make sure to update this any time a new property gets added

pages/api/projects/[id]/patch.ts
```diff
@@ -0,0 +1,39 @@
+import type { Project } from '@charmverse/core/prisma-client';
```
Comment (mattcasey) [inline pages/api/projects/[id]/patch.ts:1]: this should be a PATCH endpoint inside pages/api/projects/[id].ts

lib/projects/patchProject.ts
```diff
@@ -0,0 +1,28 @@
+import { prisma } from '@charmverse/core/prisma-client';
```
mattcasey, line 1: patchProject and updateProject mean the same thing, can you come up with a better name for this?

pages/api/projects/[id]/patch.ts
Comment (mattcasey) [reply pages/api/projects/[id]/patch.ts:1]: i realized after i left this comment that there already is a 'put' handler in that file...

components/settings/projects/components/ProjectMemberFields.tsx
Devorein: Good catch.

pages/api/projects/[id]/patch.ts
Comment (Devorein) [reply pages/api/projects/[id]/patch.ts:1]: yep that is to update the project with the project members, this one is just for the project

lib/projects/patchProject.ts
Devorein: yep, updated patchProject to updateProject and updateProject to updateProjectAndMembers

components/settings/projects/hooks/useProjectUpdates.ts
Devorein: Its a bit tricky since all three operations are different and update the cache in different way, I've tried a few but those are making the code less dry.

components/[pageId]/DocumentPage/DocumentPage.tsx
Comment (Devorein) [reply components/[pageId]/DocumentPage/DocumentPage.tsx:111]: But why would it be called twice? We are using the useSWR hook, and it's using the same proposalId, so it's calling it once only.

PR #3911, opened 2024-04-08
fix scrolling to page headers

mattcasey:
### WHAT
Fix for scrolling to page headers. I also moved the trigger into the fidus editor code which is responsible for updating the content, so we don't need a timeout.

### WHY

The issue was that `domAtPos` returns an _offest_ to its parent HTML node if it's a text node. So we were receiving the entire dom container. There's a util that handles this and returns the proper dom node instead: https://github.com/atlassian/prosemirror-utils/blob/fbdf888a4c24bae5f124d166bd21692d7b5285be/src/selection.ts#L208


components/common/CharmEditor/components/heading.ts
```diff
@@ -192,27 +193,30 @@ export function insertEmptyParaBelow() {
...
-      setTimeout(() => {
-        // Need to reference the domNode again because the node might have been re-rendered
-        const _domNode = editor.view.domAtPos(nodePos!)?.node as HTMLElement;
-        hideSelectionTooltip(pluginKey)(editor.view.state, editor.view.dispatch, editor.view);
-        document.querySelector('.document-print-container')?.scrollTo(0, _domNode.offsetTop);
-      }, 500);
+  const domNode = nodePos ? (findDomRefAtPos(nodePos, view.domAtPos.bind(view)) as HTMLElement) : false;
+  if (domNode && typeof nodePos !== 'undefined') {
+    view.dispatch(view.state.tr.setSelection(NodeSelection.create(view.state.doc, nodePos)));
+    // hideSelectionTooltip(pluginKey)(view.state, view.dispatch, view);
```
mattcasey, line 216: Not sure why this was needed? It could be that we should be creating a TextSelection instead. Here is code that fiduseditor uses to scroll to an "anchor" page type: https://github.com/fiduswriter/fiduswriter/blob/38d52bf84a976471129eefc46ae5c5667b06409f/fiduswriter/document/static/js/modules/editor/index.js#L664. (which btw would be a nice way to introduce adding anchors anywhere in the page)

Devorein (approved):

PR #3913, opened 2024-04-09
Unlimited rewards

valentinludu:
Unlimited rewards instead of dash.
Disable validation when unlimited.

Card:
https://app.charmverse.io/charmverse/rewards-available-unlimited-leads-to-this-red-box-in-the-rewards-table-9809057999884654


valentinludu:
The "#Available" is a type number input field and it is validating/highlighting the field in red because "Unlimited" is a string.
As discussed in the  discord thread, the validation needs to be turned off for this field.

> What is the component we're using for card proeprties? why does it not need this extra validation rule or copy?

PR #3919, opened 2024-04-11
Onchain credentials ux

motechFR:
### WHAT

Make credentials a step inside the proposal workflow

Fetch offchain and onchain credentials by ID

Refactor credentials selector for better UX


config/constants.ts
```diff
@@ -2,10 +2,11 @@ import env from '@beam-australia/react-env';
 
 // Note: NODE_ENV can only be 'development' or 'production' according to Next.js, but we don't want to mix them with test env
 export const isTestEnv = (env('APP_ENV') ?? process.env.REACT_APP_APP_ENV ?? process.env.NODE_ENV) === 'test';
-export const isStagingEnv = (env('APP_ENV') ?? process.env.REACT_APP_APP_ENV) === 'staging';
+export const isStagingEnv = true;
+//  (env('APP_ENV') ?? process.env.REACT_APP_APP_ENV) === 'staging';
 export const isProdEnv = process.env.NODE_ENV === 'production' && !isTestEnv && !isStagingEnv;
-export const isDevEnv =
-  (!process.env.NODE_ENV || process.env.NODE_ENV === 'development') && !isProdEnv && !isStagingEnv && !isTestEnv;
+export const isDevEnv = true;
```
mattcasey, line 8: remember to reset this plz

mattcasey (changes requested):
The issuedCredentials should be returned to the frontend as part of the populated proposal, and then passed down like other relational data. Calling refreshProposal() will be all that's needed to update them. This ought to clean up a lot of code, we can get rid of useGetIssuedProposalCredentials and its endpoint for instance

components/[pageId]/DocumentPage/hooks/useProposal.ts
```diff
@@ -19,6 +20,7 @@ export function useProposal({ proposalId }: { proposalId?: string | null }) {
   const { trigger: upsertRubricCriteria } = useUpsertRubricCriteria({ proposalId });
   const { trigger: updateProposalWorkflow } = useUpdateWorkflow({ proposalId });
   const { subscribe } = useWebSocketClient();
+  const { refreshIssuableCredentials } = useProposalCredentials({ proposalId });
```
Comment (mattcasey) [inline components/[pageId]/DocumentPage/hooks/useProposal.ts:23]: this is pattern that doesn't work! you're just creating a new instance of useSWR, so refreshing only mutates the instance inside this component

components/proposals/hooks/useProposalCredentials.tsx
```diff
@@ -126,3 +131,24 @@ export function useProposalCredentials() {
...
+    proposalId
+  });
+  const multiCredentials = useMultiProposalCredentials({ proposalIds: proposalId ? [proposalId] : null });
+
+  return {
+    ...multiCredentials,
+    issuedCredentials,
+    isLoadingIssuedCredentials,
+    refreshIssuedCredentials,
+    hasPendingOnchainCredentials:
```
mattcasey, line 151: this value can also be returned from the backend instead of the frontend, as part of the proposal details

PR #3920, opened 2024-04-11
sometimes SafeKit gets imported with an extra level of default, cant …

mattcasey:
…figure out why


esbuild.config.ts
```diff
@@ -5,7 +5,11 @@ import * as esbuild from 'esbuild';
 esbuild
   .build({
     bundle: true,
-    entryPoints: ['./background/initWebsockets.ts', './background/cron.ts'],
+    entryPoints: [
+      './background/initWebsockets.ts',
```
mattcasey, line 9: i deleted the unecessary copy of esbuild for ceramic

mattcasey: only difference is we didnt externalize packages - but i prefer we do it this way

Devorein (approved):

PR #3929, opened 2024-04-12
Fix/project-members-validation

Devorein:

### DIFF
--- a/__e2e__/proposals/structuredProposalWithProject.spec.ts
+++ b/__e2e__/proposals/structuredProposalWithProject.spec.ts
@@ -284,6 +284,9 @@ test.describe.serial('Structured proposal template with project', () => {
     // Add a new project member and update their project fields
     await proposalPage.projectTeamMembersSelect.click();
     await proposalPage.getProjectMemberOption(0).click();
+    await proposalPage.page.waitForResponse((response) => {
+      return response.request().method() === 'PUT' && response.url().endsWith('/form/answers');
+    });
     await proposalFormFieldPage.fillProjectField({ fieldName: 'projectMembers[0].email', content: 'doe@gmail.com' });
     await proposalFormFieldPage.fillProjectField({
       fieldName: 'projectMembers[1].email',


__e2e__/po/formField.po.ts
```diff
@@ -22,6 +22,31 @@ export class FormField extends DocumentPage {
...
+    textArea = false
+  }: {
+    textArea?: boolean;
+    fieldName: string;
+    content: string;
+  }) {
+    let pendingApiCall: Promise<any>;
+
+    if (fieldName.startsWith('projectMember')) {
+      pendingApiCall = this.page.waitForResponse(/\/api\/projects\/[^\\/]+\/members\/[^\\/]+/);
```
mattcasey, line 37: it seems weird that a standard method should be waiting for a request... if you're just trying to fill out a field, why should there be an API call before you reach it?

__e2e__/po/formField.po.ts
```diff
@@ -22,6 +22,31 @@ export class FormField extends DocumentPage {
...
+    } else {
+      pendingApiCall = this.page.waitForResponse(/\/api\/projects\/[^\\/]+\/patch/);
```
mattcasey, line 39: why bother making if/else just for a subpath? why not just waitForResponse(/\/api\/projects\/).

mattcasey, line 39: i would remove the if/else and just check that `/\/api\/projects\/` gets called

__e2e__/po/formField.po.ts
```diff
@@ -22,6 +22,31 @@ export class FormField extends DocumentPage {
...
+    }
+    const result = await this.page
+      .locator(`data-test=project-field-${fieldName} >> ${textArea ? 'textarea' : 'input'}`)
```
mattcasey, line 42: an improvement would be to have data-test props on input and textarea so we don't have to care

components/settings/projects/hooks/useProjectForm.ts
```diff
@@ -162,28 +177,32 @@ export function useProjectForm(options: {
...
     mode: 'onChange'
   });
 
   useEffect(() => {
-    if (options.projectId) {
-      form.reset(projectWithMembers);
-    } else {
-      form.reset(defaultProjectAndMembersPayload);
+    const projectWithMembers =
+      options.defaultProjectValues ?? projectsWithMembers?.find((project) => project.id === options.projectId);
```
mattcasey, line 190: shouldn't we be trying to use the real data before falling back to defaults?

lib/projects/constants.ts
```diff
@@ -101,50 +103,67 @@ export const projectFieldProperties: ProjectFieldProperty[] = [
...
-      warpcast: '',
-      github: '',
-      linkedin: '',
-      telegram: '',
-      otherUrl: '',
-      previousProjects: ''
-    }
-  ]
-};
+export function createDefaultProjectAndMembersPayload() {
```
mattcasey, line 106: we should build the UI so that all these strings except for name can be undefined

lib/projects/constants.ts
```diff
@@ -101,50 +103,67 @@ export const projectFieldProperties: ProjectFieldProperty[] = [
...
+      },
+      walletAddress: {
+        private: true
+      }
     }
+  } as ProjectAndMembersFieldConfig;
+}
+
+export const projectInclude = {
+  projectMembers: {
```
mattcasey, line 159: this is clever but i think this sorting makes more sense to be done in the UI

lib/projects/getProjectById.ts
```diff
@@ -0,0 +1,10 @@
+import { prisma } from '@charmverse/core/prisma-client';
+
+import { projectInclude } from './constants';
+
+export async function getProjectById(projectId: string) {
```
mattcasey, line 5: this fucntion should have a return type: ProjectWithMembers

lib/projects/interfaces.ts
```diff
@@ -1,6 +1,6 @@
 import type { Project, ProjectMember } from '@charmverse/core/prisma-client';
 
-type ProjectPayload = Pick<
+export type ProjectPayload = Pick<
```
mattcasey, line 3: I think this type is only used in updateProposal (and then re-exported as UpdateProjectPayload), we should just move it there. I'm not sure why ProjectField inherits only these fields - what about the other fields we don't pick? but maybe it would make more logic if these were called "UpdatedableProjectFields" or something, and this also goes into updateProposal.ts

lib/projects/interfaces.ts
```diff
@@ -17,7 +17,7 @@ type ProjectPayload = Pick<
 
 export type ProjectField = keyof ProjectPayload;
 
-type ProjectMemberPayload = Pick<
+export type ProjectMemberPayload = Pick<
```
mattcasey, line 20: same comment as above. Don't think of an interface file as a schema.prisma, where you go to find all the models. More ofte nthan not, we are going from function to type, not type to function. So putting everything in one jar just adds another hop to get to the truth.

components/settings/projects/hooks/useProjectForm.ts
```diff
@@ -44,7 +44,17 @@ function addMatchersToSchema({
       }
     });
   } else if (fieldType === 'email') {
-    schemaObject[fieldType] = schemaObject[fieldType]!.email('Invalid email');
+    schemaObject[fieldType] = schemaObject[fieldType]!.test('is-valid-email', 'Invalid email address', (value) => {
```
mattcasey, line 47: suggestion: we should just put all the schema/yup stuff into lib now that this is used in the backend too. It's better if components imports from lib, not the other way around

__e2e__/po/formField.po.ts
Devorein: Its a bit difficult in this instance. I will skip it for now.

__e2e__/po/formField.po.ts
Devorein: Made it work with a bit of modification

components/settings/projects/hooks/useProjectForm.ts
Devorein: This is just a naming issue, the defaultProjectValues is the real data. I renamed it to `initialProjectValues`

components/settings/projects/hooks/useProjectForm.ts
Devorein: Moved to `lib/createProjectYupSchema`

components/settings/projects/ProjectsSettings.tsx
```diff
@@ -46,11 +51,20 @@ function ProjectRow({
 }) {
   const { user } = useUser();
   const isTeamLead = projectWithMembers.projectMembers[0].userId === user?.id;
```
mattcasey, line 53: This component has no guarantee that team lead is first in the list

mattcasey (approved):

lib/projects/constants.ts
Devorein: We are going with both approach just to be safe. Sorting on the client side and also making sure the order is correct when sending data

PR #3939, opened 2024-04-16
Remove goerli

motechFR:
### WHAT

Drop goerli from connectors, and prepare migration

TODO: Find a way to handle existing rewards in UI so they don't crash the app


valentinludu (approved):
Looks good!

connectors/chains.ts
```diff
@@ -158,7 +145,7 @@ export const RPC: Record<string, IChainDetails> = {
     ...EVM_DEFAULT,
     chainId: baseSepolia.id,
     viem: baseSepolia,
-    chainName: 'Base - Goerli Testnet',
+    chainName: 'Base - Sepolia Testnet',
```
mattcasey, line 148: wow nice catch

PR #3941, opened 2024-04-16
Added chain select for wallet address form field

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/extend-wallet-input-to-allow-a-blockchain-requirement-9395122206771407)

### DIFF
--- a/components/common/form/FormField.tsx
+++ b/components/common/form/FormField.tsx
@@ -369,8 +369,8 @@ export function FormField(
                     formField.private ? <Chip sx={{ mx: 1 }} label='Private' size='small' /> : undefined
                   }
                   walletInputConfig={
-                    // Only show if chainId is present in field config
-                    isWalletConfig(formField.fieldConfig)
+                    // Only show if chainId is present in field config and its a valid chainId
+                    isWalletConfig(formField.fieldConfig) && formField.fieldConfig.chainId
                       ? {
                           chainId: formField.fieldConfig.chainId
                         }
--- a/components/common/form/FormField.tsx
+++ b/components/common/form/FormField.tsx
@@ -218,6 +218,15 @@ function ExpandedFormField({
           onCreateOption={onCreateOption}
           onDeleteOption={onDeleteOption}
           onUpdateOption={onUpdateOption}
+          walletInputConfig={{
+            chainId: (formField.fieldConfig as { chainId: number } | undefined)?.chainId,
+            onChangeChainId: (chainId) => {
+              updateFormField({
+                fieldConfig: { chainId },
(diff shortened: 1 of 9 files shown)


components/common/form/FormField.tsx
```diff
@@ -218,6 +218,15 @@ function ExpandedFormField({
           onCreateOption={onCreateOption}
           onDeleteOption={onDeleteOption}
           onUpdateOption={onUpdateOption}
+          walletInputConfig={{
+            chainId: (formField.fieldConfig as { chainId: number } | undefined)?.chainId,
+            onChangeChainId: (chainId) => {
+              updateFormField({
+                fieldConfig: { chainId },
```
mattcasey, line 225: shouldn't this include existing field config if there is one?

components/common/form/FormFieldAnswers.tsx
```diff
@@ -229,6 +229,13 @@ export function FormFieldAnswersControlled({
                     labelEndAdornment={
                       formField.private ? <Chip sx={{ ml: 1 }} label='Private' size='small' /> : undefined
                     }
+                    walletInputConfig={
+                      (formField.fieldConfig as { chainId: number })?.chainId
```
mattcasey, line 233: you should use a type guard, we dont use them a lot but you had to write the same code 3 times

components/common/form/fields/FieldTypeRenderer.tsx
```diff
@@ -31,23 +38,44 @@ type SelectProps = {
 } & FieldProps &
   Required<ControlFieldProps>;
 
-type Props = (Omit<TextProps, 'type'> | Omit<SelectProps, 'type'>) & {
+type WalletProps = {
+  type: 'wallet';
+  walletInputConfig?: WalletInputConfig;
```
mattcasey, line 43: is walletInputConfig really undefined sometimes?

mattcasey (approved):

components/common/form/FormFieldAnswers.tsx
Devorein: Good idea. Updated.

components/common/form/FormField.tsx
Devorein: This is specifically for the wallet input config, so there won't be any existing field config

components/common/form/fields/FieldTypeRenderer.tsx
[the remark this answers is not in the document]
Devorein: We are making it optional in the Props underneath, this type isn't even necessary

PR #3943, opened 2024-04-16
Feat/kyc credentials

valentinludu:
Main changes

add synaps api key and secret in the integrations tab
add persona api key and secret in the integrations tab
add switch for setting the default kyc integration
api routes for handling credentials


__e2e__/po/settings/integrationsSettings.po.ts
```diff
@@ -0,0 +1,45 @@
+import type { Locator, Page } from '@playwright/test';
+
+import { SettingsModal } from './settings.po';
+
+export class IntegrationsSettings extends SettingsModal {
+  readonly saveButton: Locator;
```
mattcasey, line 6: this is not the right way to define locators anymore, because it's so verbose. See this example: https://github.com/charmverse/app.charmverse.io/blob/main/__e2e__/po/formField.po.ts#L9

background/tasks/processPersonaWebhookMessages/task.ts
```diff
@@ -0,0 +1,13 @@
+import { log } from '@charmverse/core/log';
+
+import { processMessages } from 'lib/aws/webhookSqs';
+import { SQS_WEBHOOK_PERSONA_QUEUE_NAME } from 'lib/kyc/config';
+import { processWebhookMessage } from 'lib/kyc/persona/webhook/processWebhookMessage';
+
+log.info('Persona Queue url:', SQS_WEBHOOK_PERSONA_QUEUE_NAME);
```
mattcasey, line 7: suggestion: move this info to the log when we start processing (line 11). i think its kinda noisy when things just log when the app starts up

components/settings/integrations/components/PersonaModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  completed: 'Completed',
+  needs_review: 'Needs review',
+  approved: 'Approved',
+  created: 'Created',
+  failed: 'Failed',
+  expired: 'Expired',
+  declined: 'Declined'
+};
+
+export default function PersonaModal({ spaceId }: { spaceId: string }) {
```
mattcasey, line 23: dont use default exprts

components/settings/integrations/components/SpaceIntegrations.tsx
```diff
@@ -1,89 +1,115 @@
...
-    },
+    defaultValues: getDefaultValues({ kycCredentials, space }),
     resolver: yupResolver(schema),
     mode: 'onSubmit'
   });
 
+  useEffect(() => {
+    reset(getDefaultValues({ kycCredentials, space }));
+  }, [
+    kycCredentials?.synaps?.apiKey,
```
mattcasey, line 58: why not just watch kycCredentials?

components/settings/integrations/components/SpaceIntegrations.tsx
```diff
@@ -97,7 +123,66 @@ export function SpaceIntegrations({ space }: { space: Space }) {
...
+              Save
+            </Button>
+          </Grid>
+        )}
       </Grid>
     </form>
   );
 }
+
+function getDefaultValues({ kycCredentials, space }: { kycCredentials?: KycCredentials; space: Space }) {
```
mattcasey, line 178: this and maybe some of the other state seems like it belongs in a sub-component of SpaceIntegrations, since it's specific to KYC

lib/kyc/config.ts
```diff
@@ -0,0 +1,14 @@
+import { isProdEnv } from 'config/constants';
+
+// Synaps
+export const synapsUrl = 'https://api.synaps.io/v4' as const;
+export const SQS_WEBHOOK_SYNAPS_QUEUE_NAME = isProdEnv
+  ? `https://sqs.us-east-1.amazonaws.com/310849459438/prd-webhook-synaps`
+  : 'https://sqs.us-east-1.amazonaws.com/310849459438/stg-webhook-synaps';
```
mattcasey, line 7: we put these in environment secrets usually but i think i actually prefer it this way if it's not an api secret. it reduces our AWS costs and makes it easier for other devs to work on. just be conscious that this is open-source code and visible to everyone

lib/kyc/getKycCredentials.ts
```diff
@@ -0,0 +1,23 @@
+import type { PersonaCredential, SynapsCredential } from '@charmverse/core/prisma-client';
+import { prisma } from '@charmverse/core/prisma-client';
+
+export type KycCredentials = {
+  synaps: Omit<SynapsCredential, 'id'> | null;
+  persona: Omit<PersonaCredential, 'id'> | null;
+};
+
+export async function getKycCredentials(spaceId: string): Promise<KycCredentials> {
+  const synapsCredential = await prisma.synapsCredential.findUnique({
```
mattcasey, line 10: these could be done in parallel. nbd tho

mattcasey (approved):

__e2e__/po/settings/integrationsSettings.po.ts
valentinludu: I like it! It's cleaner this way.
Changed!

background/tasks/processPersonaWebhookMessages/task.ts
valentinludu: Good point. Moved it for both task files!

components/settings/integrations/components/PersonaModal.tsx
valentinludu: Done!

components/settings/integrations/components/PersonaModal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const { user } = useUser();
+  const { data: personaUserKyc, isLoading: isPersonaUserKycLoading } = useGetPersonaInquiry(spaceId);
+  const {
+    data: personaInquiry,
+    trigger: initPersonaInquiry,
+    isMutating: isLoadingPersonaInquiry
+  } = useInitPersonaInquiry(spaceId);
+  const popupConfirmationState = usePopupState({ variant: 'popover', popupId: 'confirm-kyc' });
```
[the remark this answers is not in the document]
valentinludu: Nice. I added the component!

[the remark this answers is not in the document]
valentinludu: The confirmation modal was necessary for Synaps so I added it for Persona as well just to keep the same UX.

components/settings/integrations/components/SpaceIntegrations.tsx
valentinludu: I want to prevent the useEffect from running if a kycCredentials is updated with the same values.

components/settings/integrations/components/SpaceIntegrations.tsx
valentinludu: Since we decided to move from a general Save to a more localised one, I moved it to it's own integration component

lib/kyc/config.ts
valentinludu: Yeah, I usually add those in .env, but this time I presumed it's fine to copy pasta from the other task files since we are doing the same thing for github and others.

lib/kyc/getKycCredentials.ts
valentinludu: Added it in a prisma.$transaction.

PR #3951, opened 2024-04-17
Share modal UX

mattcasey:
1. add a Copy link icon to the footer

2. Show "No access" instead of "Remove" for default permissions:
image

image


valentinludu (approved):
Code looks good.

Tested on Desktop and Mobile.

PR #3957, opened 2024-04-19
Add project columns to proposal-as-a-source

mattcasey:
Changes:
- we now query project and members when building properties for proposal-as-a-source. I chose to use unique id's instead of type, as we decided a while back to do going forward. This allows us to reuse types like "text" and "multiSelect"
- I moved some things out of lib/projects/<interfaces|constants.ts> to lib/projects/formField. This helps de-couple project-related and form-related logic.
- enabled multiSelect component to show values that don't exist in the template's options. There's no need when the column is a readonly one.


lib/databases/proposalsSource/getBoardProperties.ts
```diff
@@ -147,176 +107,257 @@ export function getBoardProperties({
...
+  boardProperties: IPropertyTemplate[],
+  fieldConfig: ProjectAndMembersFieldConfig
+) {
+  applyToPropertiesById(boardProperties, {
+    id: constants.PROJECT_NAME_ID,
+    name: 'Project',
+    options: [],
+    type: 'text'
+  });
+  if (getFieldConfig(fieldConfig.excerpt).show) {
```
mattcasey, line 126: things i dont love about this function:
- no type-safety over the properties in field config
- not very DRY, maybe i could add the property ids to the configuration in lib/projects/formField? then we could just loop thru the list of fields..

Devorein: +1 to adding property ids to the configuration

PR #3960, opened 2024-04-22
Fix rewards FE errors

valentinludu:
1st error: Rewards applicants returned a null value.
2nd error: Mapped Fragment without keys


components/common/DatabaseEditor/components/properties/UserAndRoleSelect.tsx
```diff
@@ -89,7 +89,7 @@ function SelectedOptions({
     <>
       {value.map((option) => {
         return (
-          <>
+          <Fragment key={option.id}>
```
mattcasey, line 92: thanks for catching

components/common/DatabaseEditor/components/propertyValueElement.tsx
```diff
@@ -482,7 +482,7 @@ function PropertyValueElement(props: Props) {
     Array.isArray(propertyValue) &&
     !card.fields.isAssigned
   ) {
-    propertyValueElement = null;
+    propertyValueElement = <Box />;
```
mattcasey, line 485: this doesn't seem right, null is a valid value to pass in leiu of a JSX element

mattcasey (changes requested):
see question

components/common/DatabaseEditor/components/propertyValueElement.tsx
valentinludu: At line 705 that `null` set at line 485 is logging an error.
I could verify that it's not a `REWARDS_APPLICANTS_BLOCK_ID` before showing the error.

mattcasey (approved):

PR #3962, opened 2024-04-22
Fixed non issue related events from throwing error in github sqs

Devorein:
### WHY

[Datadog](https://app.datadoghq.com/logs?query=%40logger%3Asqs&cols=host%2Cservice&fromUser=true&index=*&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1713788335767&to_ts=1713791935767&live=true)


valentinludu (approved):
Thank you for looking into this.

PR #3964, opened 2024-04-23
Fix titles disappearing in kanban boards

mattcasey:
When I added the listener to page update, i was assuming that title would exist all the time. But if the icon fo a page was changing, we ended up passing in `undefined` for `title`  and removing the page title inside of card state


valentinludu (approved):
Tested. All good!

PR #3966, opened 2024-04-23
Draft reward

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/add-support-for-draft-rewards-33931382913021957)


components/rewards/components/RewardStickyFooter.tsx
```diff
@@ -0,0 +1,73 @@
...
+  reward: RewardWithUsers;
+  page: { title: string; type: PageType };
+  refreshReward?: VoidFunction;
+}) {
+  const { showMessage } = useSnackbar();
+  const { trigger: publishReward, isMutating } = usePublishReward(reward.id);
+
+  async function onClick() {
+    try {
+      await publishReward({
```
mattcasey, line 26: this footer is only used in DocumentPage, we should not need to pass in any attributes from the reward to publish it. publish action should just be switching the status

lib/rewards/publishReward.ts
```diff
@@ -0,0 +1,56 @@
...
+  const errors = getRewardErrors({
+    page,
+    reward: { assignedSubmitters, rewardAmount, rewardToken, chainId, customReward, reviewers },
+    rewardType
+  });
+  if (errors.length > 0) {
+    throw new InvalidInputError(errors.join(', '));
+  }
+
+  await updateRewardSettings({
```
mattcasey, line 45: this seems like way too much overkill just to flip the status column on reward table

pages/api/rewards/[id]/publish.ts
```diff
@@ -0,0 +1,38 @@
...
+    throw new InvalidInputError('Reward is not in draft state.');
+  }
+
+  if (reward.createdBy !== userId) {
+    throw new UnauthorisedActionError('You do not have permission to publish this reward.');
+  }
+
+  const rewardPublishData = req.body as RewardPublishData;
+
+  const publishedReward = await publishReward({
```
Comment (mattcasey) [inline pages/api/rewards/[id]/publish.ts:30]: validation should be done in this endpoint like we do for proposals to be consistent

components/rewards/components/RewardApplications/NewWorkButton.tsx
```diff
@@ -1,35 +1,30 @@
...
   addIcon?: boolean;
   variant?: ButtonProps['variant'];
   buttonSize?: ButtonProps['size'];
   color?: ButtonProps['color'];
 };
 
-export function NewWorkButton({ color, buttonSize, addIcon, rewardId, variant = 'contained' }: Props) {
-  const { rewards } = useRewards();
+export function NewWorkButton({ color, buttonSize, addIcon, reward, variant = 'contained' }: Props) {
+  const rewardId = reward?.id;
```
mattcasey, line 23: The reward prop is requried in the prop so no need for "?"

components/rewards/components/RewardApplications/RewardApplications.tsx
```diff
@@ -3,22 +3,21 @@ import { Box, Divider, Grid, IconButton, Stack, Tooltip } from '@mui/material';
...
-import type { ApplicationMeta } from 'lib/rewards/interfaces';
+import type { ApplicationMeta, RewardWithUsers } from 'lib/rewards/interfaces';
 import { formatDate, formatDateTime } from 'lib/utils/dates';
 import type { LoggedInUser } from 'models';
 
 import { RewardApplicationStatusChip } from '../RewardApplicationStatusChip';
 
 type Props = {
-  rewardId: string;
+  reward: RewardWithUsers;
```
mattcasey, line 20: suggestion: inherit the type for reward from NewWorkButton.

lib/rewards/updateRewardSettings.ts
```diff
@@ -73,6 +78,7 @@ export async function updateRewardSettings({ rewardId, updateContent }: RewardUp
         id: rewardId
       },
       data: {
+        status: isPublished ? 'open' : reward.status,
```
mattcasey, line 81: just because a reward is published doesn't necessarily mean it is "open". maybe it is "complete". but i dont tihnk we should use this function to publish a  reward

components/common/DatabaseEditor/components/table/tableRows.tsx
```diff
@@ -50,6 +51,7 @@ function TableRows(props: Props): JSX.Element {
   } = props;
   const hasSubPages = allCards.some((card) => card.subPages?.length);
   const [pageSize, setPageSize] = useState(DEFAULT_PAGE_SIZE);
+  const { rewards = [] } = useRewards();
```
mattcasey, line 54: Please remove: focalboard components should only rely on Redux data.

components/common/DatabaseEditor/components/table/tableRows.tsx
```diff
@@ -94,51 +96,55 @@ function TableRows(props: Props): JSX.Element {
...
+            emptySubPagesPlaceholder={
+              reward ? (
+                <Box
+                  p={1}
+                  pl={5}
+                  display='flex'
+                  justifyContent='flex-start'
+                  style={{ backgroundColor: 'var(--input-bg)' }}
+                >
+                  <NewWorkButton color='secondary' reward={reward} addIcon variant='text' buttonSize='small' />
```
mattcasey, line 139: it made more sense for NewWorkButton to import reward context locally so that context changes do not re-render the entire table

mattcasey (changes requested):
- publishReward should only update the status field and not change any other fields on the reward - look at publishProposal. These two methods should have the same responsibliities
- seems inappropriate to pull in reward context for all table rows

components/rewards/components/RewardStickyFooter.tsx
Devorein: Right but we need to disable the button if the properties aren't valid, thats why I'm removing passing the properties when publishing but still keeping the reward and page prop for validation.

lib/rewards/publishReward.ts
Devorein: Yeah good point, updated.

components/rewards/components/RewardApplications/NewWorkButton.tsx
Devorein: That was missed from a refactor, good catch.

components/rewards/components/RewardApplications/NewWorkButton.tsx
```diff
@@ -1,35 +1,30 @@
...
-import { useRewards } from 'components/rewards/hooks/useRewards';
 import { useCharmRouter } from 'hooks/useCharmRouter';
 import { useIsCharmverseSpace } from 'hooks/useIsCharmverseSpace';
 import { useUser } from 'hooks/useUser';
+import type { RewardWithUsers } from 'lib/rewards/interfaces';
 import { statusesAcceptingNewWork } from 'lib/rewards/shared';
 
 type Props = {
-  rewardId?: string;
+  reward: RewardWithUsers;
```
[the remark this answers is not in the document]
Devorein: Point. Updated.

components/common/DatabaseEditor/components/table/tableRows.tsx
Devorein: The reward attached to card doesn't have the application information. But I agree with the point that we shouldn't be loading rewards context just because it might have a reward, I will move the context inside NewWorkButton by passing rewardId.

components/common/DatabaseEditor/components/table/tableRows.tsx
Devorein: Removed.

PR #3970, opened 2024-04-23
expose comments to all viewable pages

mattcasey:
### WHAT

- Add a 'sign in to comment' button when you aren't logged in
- Allow users to read comments if they can view the page

### WHY

Comment privleges match Notion. Also requested by OP


lib/pages/comments/listPageComments.ts
```diff
@@ -19,6 +19,14 @@ export async function listPageComments({ pageId, userId }: ListPageCommentsProps
       createdAt: 'desc'
     }
   });
+  const page = await prisma.page.findFirst({
```
valentinludu, line 22: I don't see this being used anywhere. Do you need it?

testing/setupDatabase.ts
```diff
@@ -22,6 +23,7 @@ import type {
 import { Prisma } from '@charmverse/core/prisma';
 import type { Application, PagePermission, PageType } from '@charmverse/core/prisma-client';
 import { prisma } from '@charmverse/core/prisma-client';
+import type { AnyIfEmpty } from 'react-redux';
```
valentinludu, line 26: This is not used anywhere.

valentinludu (approved):

PR #3972, opened 2024-04-23
add proposal link column back to rewards

mattcasey:
re-enabling a feature that got lost during refactors... i also removed a bunch of unecessary logic to check if milestones exist or not, which chris was OK with

### DIFF
--- a/components/rewards/hooks/useRewardsBoardAdapter.ts
+++ b/components/rewards/hooks/useRewardsBoardAdapter.ts
@@ -45,9 +45,8 @@ export function useRewardsBoardAdapter() {
   const { space } = useCurrentSpace();
   const { membersRecord } = useMembers();
   const { rewards, mutateRewards } = useRewards();
-  const { rewardsBoardBlock: board, rewardBlocks } = useRewardBlocks();
   const { getRewardPage } = useRewardPage();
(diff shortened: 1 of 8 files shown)


components/rewards/hooks/useRewardsBoardAdapter.ts
```diff
@@ -193,7 +174,8 @@ export function mapRewardToCard({
         : reward.applications
       : [];
 
-  const proposalLinkValue = rewardPage ? [rewardPage.title, `/${rewardPage.path}`] : '';
+  const sourceProposalPage = (reward as RewardWithUsers).sourceProposalPage;
+  const proposalLinkValue = sourceProposalPage ? [sourceProposalPage.title, `/${sourceProposalPage.id}`] : '';
```
Devorein, line 178: I am getting a link like this `https://pr-3972-proposal-link.charmverse.co/relative-salmon-crow/Proposal 1,/d1d6a94e-e12d-489c-9135-46671bfc6e96`, when it should be `https://pr-3972-proposal-link.charmverse.co/relative-salmon-crow/proposal-1-6715228079665687` why don't we just use the path here or the id atleast

Devorein (changes requested):
Please fix the proposal link on the rewards board. Its currently broken

Devorein (approved):

PR #3979, opened 2024-04-24
Add support for preventing new props if board is locked

motechFR:


components/common/DatabaseEditor/utils/blockUtils.ts
```diff
@@ -19,14 +19,17 @@ export function blockToFBBlock(block: BlockWithDetails): UIBlockWithDetails {
...
   return {
     ...block,
     deletedAt: block.deletedAt ? new Date(block.deletedAt).getTime() : 0,
     createdAt: new Date(block.createdAt).getTime(),
     updatedAt: new Date(block.updatedAt).getTime(),
     title: block.title || '',
     type: block.type as UIBlockWithDetails['type'],
-    fields
+    fields,
+    isLocked: block.isLocked
```
mattcasey, line 32: this seems unecessary, since we spread block above on line 25?

components/common/DatabaseEditor/utils/blockUtils.ts
```diff
@@ -40,6 +43,7 @@ export function fbBlockToBlock(fbBlock: UIBlockWithDetails): Omit<BlockWithDetai
     fields: fbBlock.fields,
     deletedAt: fbBlock.deletedAt === 0 ? null : fbBlock.deletedAt ? new Date(fbBlock.deletedAt) : null,
     createdAt: !fbBlock.createdAt || fbBlock.createdAt === 0 ? new Date() : new Date(fbBlock.createdAt),
-    updatedAt: !fbBlock.updatedAt || fbBlock.updatedAt === 0 ? new Date() : new Date(fbBlock.updatedAt)
+    updatedAt: !fbBlock.updatedAt || fbBlock.updatedAt === 0 ? new Date() : new Date(fbBlock.updatedAt),
+    isLocked: fbBlock.isLocked
```
mattcasey, line 47: i'm not sure this is necessary here.. why do we need to send isLocked to the backend? its' not part of the prisma model

lib/databases/block.ts
```diff
@@ -116,14 +117,14 @@ export function createPatchesFromBlocks(newBlock: Block, oldBlock: Block): Block
...
+  Object.keys(newBlock.fields).forEach((key) => {
+    if (oldBlock.fields[key] !== newBlock.fields[key]) {
+      newUpdatedFields[key] = newBlock.fields[key];
     }
   });
-  Object.keys(newBlock).forEach((val) => {
-    if (val !== 'fields' && (oldBlock as any)[val] !== (newBlock as any)[val]) {
-      newUpdatedData[val] = (newBlock as any)[val];
+  Object.keys(newBlock).forEach((key) => {
+    if (key !== 'fields' && (oldBlock as any)[key] !== (newBlock as any)[key] && key !== 'isLocked') {
```
mattcasey, line 126: i think the reason you are having to filter it out here is because you're passing it through in `fbBlockToBlock`, but that is where we should be filtering out non-standard fields.

components/common/DatabaseEditor/store/boards.ts
```diff
@@ -51,7 +51,8 @@ const boardsSlice = createSlice({
         if (block.type === 'board' && block.fields.isTemplate) {
           state.templates[block.id] = block as Board;
         } else if (block.type === 'board' && !block.fields.isTemplate) {
-          state.boards[block.id] = block as Board;
+          const currentBoard = state.boards[block.id] ?? {};
+          state.boards[block.id] = { ...currentBoard, block } as Board;
```
mattcasey, line 55: i tested this PR without merging props and it seems to work fine.

The payloads being passed here are defined in components/common/DatabaseEditor/store/databaseBlocksLoad.ts, so if it is somehow not evaluating isLocked, we should  be able to fix it there.

pages/api/blocks/index.ts
```diff
@@ -329,6 +332,8 @@ async function updateBlocks(req: NextApiRequest, res: NextApiResponse<BlockWithD
 
   const updatedBlocks = await prisma.$transaction(
     blocks.map((block) => {
+      delete block.isLocked;
```
mattcasey, line 335: this is unecessary if u fix fbBlockToBlock

pages/api/blocks/[id]/index.ts
```diff
@@ -81,6 +82,10 @@ async function getBlock(req: NextApiRequest, res: NextApiResponse<BlockWithDetai
     Object.assign(result, updatedFields);
   }
 
+  if (page?.isLocked) {
+    (block as Pick<Board | Card, 'isLocked'>).isLocked = true;
```
Comment (mattcasey) [inline pages/api/blocks/[id]/index.ts:86]: This line should be unecessary, since we're already calling applyPageToBlock above, which applies isLocked. It's also odd that you reference `block` variable even though we return `result`.

lib/databases/block.ts
```diff
@@ -228,7 +229,8 @@ export function blockToUIBlock(block: BlockWithDetails): UIBlockWithDetails {
     updatedAt: new Date(block.updatedAt).getTime(),
     title: block.title || '',
     type: block.type as UIBlockWithDetails['type'],
-    fields
+    fields,
+    isLocked: block.isLocked
```
mattcasey, line 233: seems unecessary since we spread block above

lib/databases/block.ts
```diff
@@ -246,6 +248,7 @@ export function applyPageToBlock(block: PrismaBlock, page: PageFields): BlockWit
   blockWithDetails.pageType = page.type;
   blockWithDetails.updatedAt = page.updatedAt;
   blockWithDetails.updatedBy = page.updatedBy;
+  blockWithDetails.isLocked = !!page.isLocked;
```
mattcasey, line 251: this is the most important place to apply page properties to blocks. And then any place where isLocked is not showing up means we need to be calling applyPageToBlock() there as well

pages/api/blocks/[id]/subtree.ts
```diff
@@ -60,6 +60,10 @@ async function getBlockSubtree(req: NextApiRequest, res: NextApiResponse<BlockWi
   const { blocks } = await getRelatedBlocks(blockId);
   const block = blocks.find((b) => b.id === blockId);
 
+  if (block?.type === 'board') {
+    (block as any as Board).isLocked = !!page.isLocked;
+  }
```
Comment (mattcasey) [inline pages/api/blocks/[id]/subtree.ts:65]: this shoudl be unecessary since we call applyPageToBlock on all the blocks returned from getRelatedBlocks()

PR #3983, opened 2024-04-25
Refresh proposal data after saving rubric draft scores

valentinludu:
Fix for:
https://www.loom.com/share/43bc0fa965e9481fb9155d7509b319a2?sid=81f9a7be-553d-4b0c-b48b-3aaf17750a23

### DIFF
--- a/components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/RubricEvaluation/RubricAnswersForm.tsx
+++ b/components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/RubricEvaluation/RubricAnswersForm.tsx
@@ -1,7 +1,7 @@
 import type { ProposalRubricCriteria, ProposalRubricCriteriaAnswer } from '@charmverse/core/prisma-client';
 import styled from '@emotion/styled';
 import { Alert, Box, FormGroup, FormLabel, Stack, TextField, Rating, Typography } from '@mui/material';
-import { useEffect, useMemo, useState, forwardRef } from 'react';
+import { useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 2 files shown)


components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/RubricEvaluation/RubricAnswersForm.tsx
```diff
@@ -26,7 +25,8 @@ type Props = {
   answers?: ProposalRubricCriteriaAnswer[];
   draftAnswers?: ProposalRubricCriteriaAnswer[];
   criteriaList: ProposalRubricCriteria[];
-  onSubmit: (props: { isDraft: boolean }) => void;
+  onSubmit: (props: { isDraft: boolean }) => Promise<void>;
+  refreshProposal?: VoidFunction;
```
mattcasey, line 29: this should be handled inside onSubmit method defined inside RubricEvaluation. It's actually already handled there, we just don't refresh the proposal if it's draft, because that used to mean there was no proposal saved to the DB yet. So we need to update the logic and only save if there is a proposal

mattcasey (changes requested):
the fix should go inside RubricEvaluation.tsx

components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/RubricEvaluation/RubricAnswersForm.tsx
valentinludu: Nice. Updated!

mattcasey (approved):

PR #3984, opened 2024-04-25
Simplify creating a Milestone

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/simplify-creating-a-milestone-8540352399347748)


mattcasey:
Token should be the defafult reward type when adding a milestone. In this PR, it is "None"

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/RewardSettings.tsx
```diff
@@ -82,42 +52,14 @@ export function RewardSettings({ value, onChange }: RewardSettingsProps) {
           options={options}
           onChange={async (rewardsTemplateId: string | string[]) => {
             if (rewardsTemplateId === 'add_new') {
-              openNewPage({
-                type: 'bounty_template'
-              });
+              const absolutePath = space ? getAbsolutePath('/rewards/new?type=bounty_template', space.domain) : '';
```
mattcasey, line 55: please don't use the word "bounty" in our APIs, URLs, etc. even though it is the page type. "type=template" would be enough too. it's not like you can create a new reward using proposal template :) this wasn't as much an issue with proposal templates but we could change that URL too to be consistent

mattcasey: Also, we shouldn't even try to call window.open if space is undefined

components/proposals/ProposalPage/components/ProposalProperties/hooks/useProposalRewards.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  function showReward({ reward, page, draftId }: ProposalPendingReward) {
+    setRewardValues(reward);
+    openNewPage(page || undefined);
+    setCurrentPendingId(draftId);
+  }
+
+  return {
+    newRewardErrors,
```
mattcasey, line 132: when you have 15 props being passed back and forth, i am not sure putting it in a separate file is actually helpful. This is really hard to jump back and forth when reading to understand what each value does

mattcasey: another idea: move the NewDocumentPage component and children to a new component. Use it as a container for some of this state - things like the dialog popup state which has nothing to do with creating a milestone

lib/rewards/createDefaultReward.ts
```diff
@@ -80,3 +80,55 @@ export async function createDefaultReward({ spaceId, userId }: { spaceId: string
...
+      }
+    },
+    pageProps: {
+      type: 'bounty_template',
+      title: 'Default milestone template',
+      sourceTemplateId: null,
+      headerImage: null,
+      icon: null,
+      contentText: '',
+      content: {
```
mattcasey, line 127: default content should be null

lib/rewards/getRewardErrors.ts
```diff
@@ -61,14 +65,17 @@ export function getRewardErrors({
...
+  if (!isTemplate && !isProposalTemplate && !isMilestone) {
     // these values are not required for templates
     if (!reward.reviewers?.length) {
       errors.push('Reviewer is required');
-    } else if (reward.assignedSubmitters && reward.assignedSubmitters.length === 0 && !isProposalTemplate) {
+    } else if (reward.assignedSubmitters && reward.assignedSubmitters.length === 0) {
       errors.push('You need to assign at least one submitter');
     }
   }
+  if (isMilestone && !templateId) {
```
mattcasey, line 76: suggestion: rename 'templateId' to "hasTemplate".

lib/rewards/getRewardErrors.ts
```diff
@@ -61,14 +65,17 @@ export function getRewardErrors({
   if (!page?.title && !linkedPageId) {
     errors.push('Page title is required');
   }
-  if (!isTemplate) {
+  if (!isTemplate && !isProposalTemplate && !isMilestone) {
```
mattcasey, line 68: i wonder if we can just reduce this to "isMilestone", since if it is a proposal template then it will always be a milestone?

mattcasey (approved):
Approved but left some requests

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/RewardSettings.tsx
Devorein: Without the domain it doesn't work, it takes me to `http://localhost:3000/rewards/new?type=template`

components/proposals/ProposalPage/components/ProposalProperties/hooks/useProposalRewards.ts
Devorein: dialog related functions like openNewPage, closeDialog and showReward are used by components other than `NewDocumentPage`, so its quite difficult to do the refactoring you suggested

lib/rewards/createDefaultReward.ts
Devorein: Thnx. Updated.

lib/rewards/getRewardErrors.ts
Devorein: templateId is not required anymore

PR #3987, opened 2024-04-25
Reward credentials evaluation step

Devorein:


motechFR:
When I am inside the application, we should only show existing information inside the submit step

The fields are already all inside the modal, and having this empty data is a bit confusing as an applicant looking at the sidebar

Screenshot 2024-04-25 at 21 45 27

motechFR:
Race condition in the reward, when I finished last evaluation, it skipped credentials in UI and went to payment

When I came back to the navigator tab, the credentials step is the active one

Here, me might just want to bundle issuableCredentials inside the reward

PR #3988, opened 2024-04-26
Reward workflow updates

Devorein:


mattcasey (approved):

lib/rewards/getReward.ts
```diff
@@ -25,6 +25,11 @@ export function rewardWithUsersInclude() {
         permissionLevel: true
       }
     },
+    page: {
```
mattcasey, line 28: thanks for making this update

PR #3991, opened 2024-04-26
Reward workflow updates 2

Devorein:


mattcasey (approved):

lib/rewards/publishReward.ts
```diff
@@ -53,14 +55,23 @@ export async function publishReward(rewardId: string) {
...
+  await prisma.$transaction([
+    prisma.bounty.update({
+      where: {
+        id: rewardId
+      },
+      data: {
+        status: 'open'
+      }
+    }),
+    prisma.pagePermission.create({
```
mattcasey, line 67: was this a bug before? was no one able to access the rewards we published except admin?

mattcasey (approved):

PR #3996, opened 2024-04-27
allow application authors to change the reward amount

mattcasey:
This update adds more granular readonly permissions when a template is used on a reward

I deployed an update to allow templates to not have token amount, but you couldn't set the amount because we blanket-mark the entire sidebar as readOnly if there was a template.

image


components/rewards/components/RewardEvaluations/components/Settings/components/PaymentStepSettings/PaymentStepSettings.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+      {rewardType === 'custom' && (
+        <RowStack>
+          <FieldLabel required={!readOnly} style={{ width: 150 }}>
+            Custom {getFeatureTitle('Reward')}
+          </FieldLabel>
+          <TextField
+            data-test='custom-reward-input'
+            onChange={(e) => {
+              onChange({
```
Devorein, line 98: We should ideally add some sort of debouncing here

components/rewards/components/RewardEvaluations/components/Settings/components/PaymentStepSettings/components/RewardTokenForm.tsx
Devorein, line 1: The modal looks quite strange when I try to add a custom token. There is no backdrop and it doesn't close even if I click on the background.
image

pages/api/rewards/templates/[id].ts
```diff
@@ -0,0 +1,31 @@
...
+  const userId = req.session.user?.id;
+  const pageId = req.query.id as string;
+  const template = await getRewardTemplate({ pageId });
+
+  const { isAdmin } = await hasAccessToSpace({
+    spaceId: template.spaceId,
+    userId
+  });
+
+  if (template.status === 'draft' && !isAdmin) {
```
Comment (Devorein) [inline pages/api/rewards/templates/[id].ts:24]: We can make use of the page permission but this works fine

Devorein (approved):
image
1. The "Reward token details" inputs look a bit weird, the labels should be left-aligned and the inputs right-aligned.
2. In the screenshot above I'm creating a new proposal template, the token amount for the milestone should be optional correct?

components/rewards/components/RewardEvaluations/components/Settings/components/PaymentStepSettings/PaymentStepSettings.tsx
mattcasey: I think debouncing is best done at the highest level, for example if we are calling an API then i would put it there. this is kind of low-level and it is just calling a callback

mattcasey: I checked and half the time we're just updating state, half the time we call "updateReward" from useReward. maybe we can add debouncing there?

components/rewards/components/RewardEvaluations/components/Settings/components/PaymentStepSettings/components/RewardTokenForm.tsx
mattcasey: Agree, but i believe this is how it was before. we could fix in a followup

pages/api/rewards/templates/[id].ts
Comment (mattcasey) [reply pages/api/rewards/templates/[id].ts:24]: i wasnt sure if page permissions actually manages draft template status already, since u just added it?

mattcasey:
> image
> 
>     1. The "Reward token details" inputs look a bit weird, the labels should be left-aligned and the inputs right-aligned.
> 
>     2. In the screenshot above I'm creating a new proposal template, the token amount for the milestone should be optional correct?

Agree i didnt' actually c heck this UI. i'll fix it thanks

PR #3998, opened 2024-04-29
Hide project fields that are private

mattcasey:
The original bug was to change validation, but I think this is a better fix, to make it consistent we do not show form fields that are private if you don't have view access

To test, you can fill out a proposal w/project that has private fields then view it as a non-admin. THe fields should not be visible


lib/forms/interfaces.ts
```diff
@@ -0,0 +1,67 @@
+import type { FormFieldType, MemberPropertyType } from '@charmverse/core/prisma';
```
mattcasey, line 1: I moved this to lib. The components/common/form folder is starting to contain a lot of not-common stuff. The form data model should stay separate, since it is opinionated and related to storing forms. Whereas the components form folder sould just be UI-related things that could work for ANY form

lib/proposals/form/getProposalFormFields.ts
```diff
@@ -1,9 +1,28 @@
-import type { FormField } from '@charmverse/core/prisma-client';
+import type { FormFieldInput } from 'components/common/form/interfaces';
+import type { ProjectAndMembersFieldConfig } from 'lib/projects/formField';
 
-export function getProposalFormFields(fields: FormField[] | null | undefined, canViewPrivateFields: boolean) {
+export function getProposalFormFields(fields: FormFieldInput[] | null | undefined, canViewPrivateFields: boolean) {
   if (!fields) {
     return null;
   }
+  if (canViewPrivateFields) {
```
mattcasey, line 8: This is the fix

lib/forms/interfaces.ts
```diff
@@ -0,0 +1,36 @@
+import type { FormFieldType, MemberPropertyType, FormField } from '@charmverse/core/prisma';
```
mattcasey, line 1: i moved form tpyes that are related to our Postgres data model here. We should try to avoid putting stuff in components/common/form that isn't generic

Devorein (approved):

PR #4015, opened 2024-05-02
Standardize Integrations panel in Settings

mattcasey:
Example

image


components/settings/integrations/components/Snapshot/SnapshotSettings.tsx
```diff
@@ -46,20 +52,53 @@ export function SnapshotIntegration({ isAdmin, space }: { isAdmin: boolean; spac
...
   });
 
-  const updateSnapshotDomain = async () => {
+  const isConnected = !!space.snapshotDomain;
+
+  async function updateSnapshotDomain() {
     const values = getValues();
     if (isDirty && isAdmin) {
-      await updateSpace({ snapshotDomain: values.snapshotDomain || null }, { onSuccess: () => refreshCurrentSpace() });
+      try {
```
valentinludu, line 60: No need for try catch here

You can use the `onError` method after the `onSuccess` one.

valentinludu:
One UX thing:
 If I click Connect on the KYC card, select Synaps or Persona and then I change my mind and click Cancel. The kyc option remains and as a user I don't have the ability to reset it to NO option. 
 I think Cancel should reset the form.

valentinludu (approved):
I like the new design! Nice work.

✅ Code looks good.

✅ Tested on mobile and desktop.

components/settings/integrations/components/Snapshot/SnapshotSettings.tsx
mattcasey: try/catch is more recognizable

mattcasey:
> One UX thing:
If I click Connect on the KYC card, select Synaps or Persona and then I change my mind and click Cancel, then click Connect again on the KYC card. The kyc option selection remains and as a user I don't have the ability to reset it to NO option.
I think Cancel should reset the form.

Agree, thanks.

> I like the new design! Nice work.

Thanks

components/settings/integrations/components/Snapshot/SnapshotSettings.tsx
valentinludu: It was more of a suggestion really.

This is a pattern I started using after we updated SWR to 2.X and it's the same thing as try catch. 
 
e.g.:
```
await updateSpace({ snapshotDomain: values.snapshotDomain || null }, 
{
onSuccess: () => refreshCurrentSpace(),
onError: (err) => doSomething()
}
);
```

PR #4027, opened 2024-05-03
allow grouping rewards by milestone

mattcasey:
we already supported grouping on regular db's, but not on rewards. i also had to allow you to sort by 'proposalUrl' field, since it only supported select/multi-select before. i'd like to add other types in the future
image


Devorein (approved):
1. The Hide button doesn't work. We can remove it from the UI for the rewards board
image

2. Rewards dnd doesn't quite work. It works if you are in the same group but not when it's in a different group. Notion doesn't allow inter-group dnd either.

Nothing blocking but those two work fine for regular boards.

PR #4030, opened 2024-05-06
automatically create and save proposals in draft form

mattcasey:


pages/api/proposals/[id]/publish.ts
```diff
@@ -110,40 +121,6 @@ async function publishProposalStatusController(req: NextApiRequest, res: NextApi
...
-              id: true,
-              type: true
-            }
-          }
-        }
-      }
-    }
-  });
-
-  if (proposalForm?.projectId) {
```
Comment (Devorein) [inline pages/api/proposals/[id]/publish.ts:139]: Shouldn't we add a check for the project form values when you are publishing a proposal?

Devorein (approved):

pages/api/proposals/[id]/publish.ts
Comment (mattcasey) [reply pages/api/proposals/[id]/publish.ts:139]: this is handled inside getProposalErrors() now

PR #4032, opened 2024-05-07
Fix user select property

valentinludu:
Card:
https://app.charmverse.io/charmverse/when-adding-a-person-to-a-card-property-it-doesnt-actually-remove-the-original-person-1513873279848057

## WHY
The `value` state was not synced with the `defaultValue` state. Only after the user closes the dropdown the component was rerendered and the value was shown correctly.

## FIX
Remove the `value` state, since we have the `defaultValue` which should be the same thing.

### DIFF
--- a/components/common/form/InputSearchMember.tsx
+++ b/components/common/form/InputSearchMember.tsx
@@ -147,36 +147,25 @@ export function InputSearchMemberMultiple({
   ...props
 }: IInputSearchMemberMultipleProps) {
   const { members, membersRecord } = useMembers();
-  const defaultMembers = (defaultValue ?? []).map((userId) => membersRecord[userId]).filter(Boolean);
-  const [value, setValue] = useState<Member[]>(defaultMembers);
+  const selectedMembers = (defaultValue ?? []).map((userId) => membersRecord[userId]).filter(Boolean);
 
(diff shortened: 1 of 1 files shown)


components/common/form/InputSearchMember.tsx
```diff
@@ -147,36 +147,25 @@ export function InputSearchMemberMultiple({
   ...props
 }: IInputSearchMemberMultipleProps) {
   const { members, membersRecord } = useMembers();
-  const defaultMembers = (defaultValue ?? []).map((userId) => membersRecord[userId]).filter(Boolean);
-  const [value, setValue] = useState<Member[]>(defaultMembers);
+  const selectedMembers = (defaultValue ?? []).map((userId) => membersRecord[userId]).filter(Boolean);
```
mattcasey, line 150: i think this should probably be memo'd since you're passing it in as a value and it would re-create itself every render?

mattcasey (changes requested):
should memoize the value too avoid too many renders

components/common/form/InputSearchMember.tsx
valentinludu: Done

mattcasey (approved):

PR #4034, opened 2024-05-07
Disable token chain and token inputs when using a reward template

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/disable-token-chain-and-token-inputs-when-using-a-reward-template-8641440715535329)

### DIFF
--- a/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
+++ b/components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
@@ -62,7 +62,6 @@ export function ProposalRewardsForm({
   const readOnlyType = !isAdmin && (readOnly || !!template?.rewardType);
   const readOnlyDueDate = !isAdmin && (readOnly || !!template?.dueDate);
   const readOnlyToken = !isAdmin && (readOnly || !!template?.rewardToken);
-  const readOnlyChain = !isAdmin && (readOnly || !!template?.chainId);
   const readOnlyTokenAmount = !isAdmin && (readOnly || !!template?.rewardAmount);
   const readOnlyCustomReward = !isAdmin && (readOnly || !!template?.customReward);
 
@@ -255,7 +254,6 @@ export function ProposalRewardsForm({
                   Token
                 </PropertyLabel>
                 <RewardTokenProperty
-                  readOnlyChain={readOnlyChain}
                   readOnlyTokenAmount={readOnlyTokenAmount}
                   readOnlyToken={readOnlyToken}
                   requireTokenAmount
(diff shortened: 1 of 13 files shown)


components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
```diff
@@ -57,16 +52,18 @@ export function ProposalRewardsForm({
...
+  rewardStatus
 }: Props) {
   const { getFeatureTitle } = useSpaceFeatures();
-  const isAdmin = useIsAdmin();
   const [isExpanded, setIsExpanded] = useState(!!expandedByDefault);
   const { data: template } = useGetRewardTemplate(templateId);
-  const readOnlyProperties = !isAdmin && (readOnly || !!template);
-  const readOnlyDueDate = !isAdmin && (readOnly || !!template?.dueDate);
-  const readOnlyToken = !isAdmin && (readOnly || !!template?.rewardToken);
+  const readOnlyType = readOnly || !!template?.rewardType;
```
mattcasey, line 60: admins are always supposed to be able to override template settings

components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
```diff
@@ -57,16 +52,18 @@ export function ProposalRewardsForm({
...
+  const readOnlyDueDate = readOnly || !!template?.dueDate;
+  const readOnlyToken = readOnly || !!template?.rewardToken;
+  const readOnlyChain = readOnly || !!template?.chainId;
```
mattcasey, line 63: i dont think we actually have a use case where only chain is set but token is optional ,so i dont think we should support it. if the token's read only, then chain is readonly

components/proposals/ProposalPage/components/ProposalProperties/hooks/useProposalRewards.ts
```diff
@@ -68,7 +68,7 @@ export function useProposalRewards({
   }).join(', ');
 
   function getTemplateAppliedReward(template?: RewardTemplate) {
-    const rewardType = template ? getRewardType(template) : 'token';
+    const rewardType = template ? template.rewardType : 'token';
```
mattcasey, line 71: thx for cleanup

mattcasey (changes requested):
left some comments. blocking issue is that admins cannot override templates

components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
Devorein: My bad I thought this restriction will be for every user (including admin). Added admin override.

components/proposals/ProposalPage/components/ProposalProperties/components/ProposalRewards/ProposalRewardsForm.tsx
Devorein: Sure lets go with that.

PR #4037, opened 2024-05-08
Fix proposal labels in list and db source

motechFR:


mattcasey (approved):
feels a little hacky that we ever used "unpublished" for both draft and credentials/reward steps, like rewards_unpublished and credentials_unpublished should really be new values. but so long as this works i think it's good enough for now

PR #4039, opened 2024-05-08
Handle case where proposal blocks are undefined

motechFR:


lib/databases/proposalsSource/getBoardProperties.ts
```diff
@@ -118,6 +118,10 @@ function applyProjectProfileProperties(
   boardProperties: IPropertyTemplate[],
   fieldConfig: ProjectAndMembersFieldConfig
 ) {
+  if (!fieldConfig) {
+    return;
```
motechFR, line 122: Not the fix, we need to find why fieldConfig ends up undefined

mattcasey:
Can you provide a reason for this change?

mattcasey:
Updated answer: reason is that the board seems to be loaded already but not its cards when viewing a public page

PR #4043, opened 2024-05-09
Allow creating Assigned workflow reward templates without applicants

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/allow-creating-reward-templates-with-assigned-workflow-that-do-not-have-any-applicants-8935643864307441)


scripts/migrations/2024_05_09_addRewardFieldsWorkflowId.ts
```diff
@@ -0,0 +1,77 @@
...
+          userId: true
+        }
+      }
+    },
+  });
+
+  const totalBounties = bounties.length;
+  let currentBounty = 0;
+
+  const workflows = getRewardWorkflows('');
```
motechFR, line 46: Since we are not yet using spaceId, we should drop the param as it's confusing

motechFR (approved):

PR #4052, opened 2024-05-10
optimize form answer components

mattcasey:
- save the input of individual form field answers one at a time, instead of the whole set of answers (which could have stale data)
- remove the subscription to formState of the proposal form input. this had a huge impact, since it re-rendered the entire form whenever a single field is updated. This is no longer the case, except in the Project form, which is still subscribing to the formState You can tell because editing a single field inside project form re-renders the entire form


components/common/form/FormFieldAnswers.tsx
```diff
@@ -65,67 +63,35 @@ const StyledStack = styled(Stack)`
...
-      showMessage(e.message, 'error');
-    } finally {
-      setIsFormDirty(false);
-    }
-  }
+  const saveFormFieldAnswers = useCallback(
+    async (answer: { id: string; value: FormFieldValue }) => {
+      if (onSave && !disabled) {
+        try {
+          // await onSave([answer]);
```
Devorein, line 87: This line should be uncommented? Its not saving the regular form field answers otherwise.

components/common/form/fields/FieldTypeRenderer.tsx
```diff
@@ -49,10 +50,43 @@ type Props = (Omit<TextProps, 'type'> | Omit<SelectProps, 'type'>) & {
...
+    const _onChange = fieldProps.onChange;
+    if (_onChange) {
+      fieldProps.onChange = (e) => {
+        _onChange(e);
+        // currently only used by proposal forms
+        // do not save updates if field is invalid
+        if (formFieldId && onChangeDebounced && !fieldProps.error) {
+          onChangeDebounced({
+            id: formFieldId,
+            value: typeof e?.target?.value === 'string' ? e.target.value : e
```
Devorein, line 85: This is storing the number fields in string. We might need to cast the values based on the form field type.

components/common/form/FormFieldAnswers.tsx
valentinludu, line 87: Shouldn't this be available?

valentinludu (approved):

components/common/form/fields/FieldTypeRenderer.tsx
```diff
@@ -49,10 +50,43 @@ type Props = (Omit<TextProps, 'type'> | Omit<SelectProps, 'type'>) & {
...
+      () => (_onChangeDebounced ? debounce(_onChangeDebounced, 300) : null),
+      [_onChangeDebounced]
+    );
+    const _onChange = fieldProps.onChange;
+    if (_onChange) {
+      fieldProps.onChange = (e) => {
+        _onChange(e);
+        // currently only used by proposal forms
+        // do not save updates if field is invalid
+        if (formFieldId && onChangeDebounced && !fieldProps.error) {
```
Devorein, line 82: The error map is not up to date. It's always one step behind. So even with invalid values, we are sending requests, and ironically, when it's valid, it's not sending one (since the last one was invalid.)

Devorein (changes requested):
Please uncomment the `onSave` before merging.

components/common/form/FormFieldAnswers.tsx
mattcasey: yes thanks! was commenting it out for testing

components/common/form/fields/FieldTypeRenderer.tsx
Devorein: nvm currently we are storing it in string too.

mattcasey:
Requested changes were addressed

PR #4055, opened 2024-05-10
Refetch nft metadata again if it was not retrieved on initial request

Devorein:
### WHY

[Discussion](https://discord.com/channels/894960387743698944/1235705281925742753)

### DIFF
--- a/lib/blockchain/provider/alchemy/client.ts
+++ b/lib/blockchain/provider/alchemy/client.ts
@@ -137,7 +137,10 @@ export async function getNFT({
   chainId: SupportedChainId;
 }) {
   const url = `${getAlchemyBaseUrl(chainId, 'nft', 'v3')}/getNFTMetadata`;
-  const res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId });
+  let res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId });
+  if (res.name === 'Failed to load NFT metadata') {
+    res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId, refreshCache: true });
+  }
   const updateRes = await updateNFTResp(res);
   return mapNFTData(updateRes, null, chainId);
 }


lib/blockchain/provider/alchemy/client.ts
```diff
@@ -137,7 +137,10 @@ export async function getNFT({
   chainId: SupportedChainId;
 }) {
   const url = `${getAlchemyBaseUrl(chainId, 'nft', 'v3')}/getNFTMetadata`;
-  const res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId });
+  let res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId });
+  if (res.name === 'Failed to load NFT metadata') {
```
valentinludu, line 141: Names can change.

I think it should be better to check the code number if available.
https://docs.alchemy.com/reference/error-reference
It states here that a request can return status 200 and a JSON-RPC error code.

lib/blockchain/provider/alchemy/client.ts
```diff
@@ -137,7 +137,10 @@ export async function getNFT({
...
+    res = await GET<AlchemyNft>(url, { contractAddress: address, tokenId, refreshCache: true });
```
valentinludu, line 142: Also

Instead of copy and paste the same request line, it would be better to look at how we are doing retries in other places and use that logic if possible.

lib/blockchain/provider/alchemy/client.ts
Devorein: This is the [doc](https://docs.alchemy.com/reference/getnftmetadata-v3) for that specific endpoint. It always returns 200, even if it is not able to load the metadata. That's why we need to look at the `name` field rather than in the catch block.

lib/blockchain/provider/alchemy/client.ts
Devorein: This is more of a refetch with the param `refreshCache: true` rather than a retry since the request isn't technically failing. The second request is bound to return the correct metadata, albeit with a bit more delay.

valentinludu (approved):

PR #4057, opened 2024-05-14
Proposal evaluation final step settings

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/allow-setting-final-step-9200318097829283)


motechFR:
UI Bug
- When the proposal has an intermediate step which is "final" and that step has passed and the next step is credentials, credentials does not show in proposals table or proposals as source table

components/common/WorkflowSidebar/components/StepperResultIcon.tsx
```diff
@@ -38,11 +38,13 @@ export type Result = 'pass' | 'fail';
...
   result,
   isCurrent,
-  position
+  position,
+  showDash
 }: {
   isCurrent?: boolean;
   result: Result | null;
   position: number;
+  showDash?: boolean;
```
motechFR, line 47: Suggestion; Add a comment about why we want to show a dash

components/settings/proposals/components/EvaluationDialog.tsx
```diff
@@ -205,17 +215,46 @@ function StepRequiredReviews({
...
+  isLastEvaluation: boolean;
+  finalStep: WorkflowEvaluationJson['finalStep'];
+  setValue: UseFormSetValue<FormValues>;
+}) {
+  return (
+    <Box>
+      <FieldLabel>Final step</FieldLabel>
+      <Stack flexDirection='row' justifyContent='space-between' alignItems='center'>
+        <Typography color='textSecondary' variant='body2'>
+          If this step passes, the entire workflow passes
```
motechFR, line 219: Suggestion: If this step passes, the entire ${getFeatureTitle('proposal')} passes

motechFR (approved):
Approved, but plz fix the table display bug first, and add some e2e coverage

Suggested flows
- Displaying proposal status correctly in the table

- Test that a flow with a final step stays at that step

PR #4059, opened 2024-05-14
Fix sort databases

valentinludu:
React was not updating the component that was consuming the sorted cards.
The cards array was sorted and the original array was overwritten and that means no rerender.

Card:
https://app.charmverse.io/charmverse/sort-doesnt-work-on-databases-6821939509445178

### DIFF
--- a/components/common/DatabaseEditor/components/table/tableHeader.tsx
+++ b/components/common/DatabaseEditor/components/table/tableHeader.tsx
@@ -185,22 +185,21 @@ function TableHeader(props: Props): JSX.Element {
             }}
           />
         </Stack>
-        {template.id !== Constants.titleColumnId && (
-          <>
-            <MenuItem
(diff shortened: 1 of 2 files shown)


components/common/DatabaseEditor/components/table/tableHeader.tsx
```diff
@@ -185,22 +185,21 @@ function TableHeader(props: Props): JSX.Element {
             }}
           />
         </Stack>
-        {template.id !== Constants.titleColumnId && (
```
valentinludu, line 188: If you are wondering what did I do here:
- Found an FE error that was thrown because you can't add a fragment inside MenuList.
- The fix was to add the 2 components inside an array.
- This was brought to light when testing the actual bug, and it was very annoying to see the red console instead of variables for debugging.

motechFR (approved):

PR #4069, opened 2024-05-20
Fix card reordering

valentinludu:
Issue:
- Drag and drop in the same column is not working properly. The cardOrder is updated, but the cards positioning inside the group is not updated because it is not taking into consideration the cardOrder.

More info:
- before it was working because we were overriding the cards object when sorting. The cards should not be mutated, it should get the cardOrder array and sort it by that.
- this was the PR that fixed sorting but broke the cards order
https://github.com/charmverse/app.charmverse.io/pull/4059/files

Solution:
- Send sortedCards to the group


Devorein (approved):

Devorein:
Tested and it works 🚀

PR #4071, opened 2024-05-20
Proposal form performance

valentinludu:


components/common/form/FormFieldAnswers.tsx
```diff
@@ -158,10 +158,16 @@ export function FormFieldAnswersControlled({
 
   useEffect(() => {
     // auto-save form fields if the form is dirty and there are no errors
-    if (debouncedValues) {
-      saveFormFieldAnswers();
+    const deboiuncedFunc = debounce(saveFormFieldAnswers, 300);
```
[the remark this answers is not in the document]
mattcasey: One other concern here is that saveFormFieldAnswers is grabbing external state, it's not using references which would have the most up-to-date values. Seems like bugs will happen, or e2e tests at least will be flaky, but i havent verified

components/proposals/ProposalPage/components/ProposalFormFieldAnswers.tsx
```diff
@@ -76,7 +74,7 @@ export function ProposalFormFieldAnswers({
     <FormFieldAnswers
       milestoneProps={milestoneProps}
       enableComments={enableComments}
-      formFields={fields}
+      formFields={isLoading ? undefined : fields}
```
mattcasey, line 77: there's no need for this line because we return LoadingComponent above

mattcasey (approved):

components/proposals/ProposalPage/components/ProposalFormFieldAnswers.tsx
valentinludu: Oh, yeah. Resolved!

PR #4072, opened 2024-05-20
fix evaluation button

mattcasey:
We were displaying "Decline" when someone wanted to approve rubric criteria:
https://www.loom.com/share/e990952f059341f3a98c7becce6374c9?sid=caa4defc-cbe2-4fbe-ba12-7d8890ae3a7f

I did a bit of cleanup too. In general we should try not to write components that serve so many use cases, and instead compose them into what we need. for instance the component that rubric uses shouldn't include any appeal logic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -407,7 +407,8 @@
     "react-dom": "^18.2.0",
     "typescript": "^5.4.2",
     "esbuild": "^0.19.5",
-    "viem": "^2.8.13"
+    "viem": "^2.8.13",
+    "next": "^14.1.1"
   },
   "msw": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -407,7 +407,8 @@
     "react-dom": "^18.2.0",
     "typescript": "^5.4.2",
     "esbuild": "^0.19.5",
-    "viem": "^2.8.13"
+    "viem": "^2.8.13",
```
mattcasey, line 410: this also fixes an unrelated production build issue

PR #4076, opened 2024-05-21
Refactored proposal evaluation appeal reviews model

Devorein:
### WHY

1. We were using the `ProposalEvaluationReview` model previously with appeal: true, this PR makes the necessary changes to support a different table just for appeal reviews


mattcasey (approved):
thanks for cleaning up

PR #4080, opened 2024-05-22
Import and create proposal from OP Rounds API

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/op-import-proposal-from-agora-24561164849175032)


mattcasey (approved):

lib/proposals/op/importOpProjects.ts
mattcasey, line 1: this file belongs in scripts

PR #4100, opened 2024-05-27
Fixed pasting text inside code block

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/pasting-code-into-code-block-breaks-the-block-9259967878949904)


mattcasey (changes requested):
please add a test in markdown.plugin.spec.ts

Also, did you consider adding this rule to the code plugin? i'm fine with it here, tho, so long as we have a test someplace. it may almost be better in the future if we just had one plugin to manage copy/pasting, cuz it's hard to make sure they are in the correct order

PR #4103, opened 2024-05-28
Save criteria updated

mattcasey:


components/proposals/hooks/useProposalFormAnswers.ts
```diff
@@ -0,0 +1,67 @@
...
+    | undefined;
+
+  const { control, getFieldState } = useFormFields({
+    fields: formFields
+  });
+
+  const projectForm = useProjectForm({
+    initialProjectValues: proposal?.project,
+    projectId: proposal?.projectId,
+    selectedMemberIds: projectAnswer?.selectedMemberIds,
```
mattcasey, line 46: Please note @Devorein that I think this value is never updated after first load anymore, since I added 'revaliate: false' to the PUT request to our API. But it looks like it is only used to set the initial values inside Project form anyway?

I didn't want to add a watcher on the field to keep it up-to-date, because that would have to watch all form answers. If we need this to be updated I think i'd add a special setter that can be called from the project form.

Devorein: Right it's only used to set the initial values and the validation schema, but I think we refactor the useProjectForm hook and move it inside this new hook.

PR #4106, opened 2024-05-28
Bugfix/adding x regex

airinterface:
### WHAT

Twitter regex  

### WHY

Add support. to accept x as a domain name as well.


mattcasey (approved):
thanks for the fix!

PR #4121, opened 2024-05-31
Select properties for proposal source boards

Devorein:


pages/api/blocks/index.ts
```diff
@@ -27,6 +27,7 @@ import { hasAccessToSpace } from 'lib/users/hasAccessToSpace';
 import { getCustomDomainFromHost } from 'lib/utils/domains/getCustomDomainFromHost';
 import { getSpaceDomainFromHost } from 'lib/utils/domains/getSpaceDomainFromHost';
 import { UnauthorisedActionError } from 'lib/utils/errors';
+import { prettyPrint } from 'lib/utils/strings';
```
mattcasey, line 30: unused?

pages/api/blocks/index.ts
```diff
@@ -288,6 +288,19 @@ async function updateBlocks(req: NextApiRequest, res: NextApiResponse<BlockWithD
     }
   });
 
+  const { isAdmin } = await hasAccessToSpace({
```
mattcasey, line 291: please put this extra db call iinside a conditional, once we figure out a block is for proposals

pages/api/pages/[id]/proposal-source.ts
```diff
@@ -43,11 +46,20 @@ async function createProposalSource(req: NextApiRequest, res: NextApiResponse) {
     userId
   });
 
+  const { isAdmin } = await hasAccessToSpace({
```
Comment (mattcasey) [inline pages/api/pages/[id]/proposal-source.ts:49]: i don't think we should add this extra logic. if an admin does for some reason give other users ability to edit the board, they would expect it to work.

pages/api/pages/[id]/proposal-source.ts
```diff
@@ -43,11 +46,20 @@ async function createProposalSource(req: NextApiRequest, res: NextApiResponse) {
...
+    userId,
+    spaceId: boardPage.spaceId
+  });
+
+  if (!isAdmin) {
+    throw new ActionNotPermittedError('You do not have permission to update this page');
```
Comment (mattcasey) [inline pages/api/pages/[id]/proposal-source.ts:55]: better error message would be sometihnf like "Only admins can use proposals as a source"

pages/api/blocks/index.ts
mattcasey: i actually think this check is unecessary. what if an admin wants to give a non-admin edit access? this would be confusing

mattcasey: this check might be confusing to users if an admin wants to give non-admins ability to make changes to the database. The only requirement was that non-admins cannot create a new database. I think its ok to leave in for now but we should bring it up to Chris

lib/databases/proposalsSource/getBlocks.ts
```diff
@@ -52,7 +55,26 @@ export async function getBlocks(
...
+        id: DEFAULT_BOARD_BLOCK_ID,
+        spaceId: board.spaceId
+      }
+    },
+    select: {
+      fields: true
+    }
+  })) as null | { fields: BoardFields };
+
+  const proposalCustomProperties = (proposalBoardBlock?.fields.cardProperties ?? []) as IPropertyTemplate[];
```
mattcasey, line 70: this logic belongs inside of updateBoardProperties

lib/proposals/getProposalTemplates.ts
```diff
@@ -45,7 +48,31 @@ export async function getProposalTemplates({
           id: true,
           archived: true,
           formId: true,
-          status: true
+          status: true,
+          form: {
```
mattcasey, line 52: this is just supposed to be meta data for templates. it needs tobe small, it was very slow in OP's space to load all the templates with all this data. please use a new endpoint or just query templates individually

__integration-tests__/server/pages/api/blocks/[id]/subtree.spec.ts
```diff
@@ -577,6 +578,17 @@ describe('GET /api/blocks/[id]/subtree - proposal databases', () => {
         }
       ]
     });
+    await updateBoardProperties({
+      boardId: proposalsDatabase.id,
+      selectedProperties: {
```
Comment (mattcasey) [inline __integration-tests__/server/pages/api/blocks/[id]/subtree.spec.ts:583]: this should just be added to the generateBoard interface

components/common/DatabaseEditor/components/viewSidebar/viewSourceOptions/viewSourceOptions.tsx
```diff
@@ -174,21 +178,13 @@ export function ViewSourceOptions(props: ViewSourceOptionsProps) {
             )}
 
             {/** Only admins can create proposals as datasource, to avoid accidentally revealing proposal data */}
-            {allowedSourceOptions.includes('proposals') && isAdmin && (
+            {allowedSourceOptions.includes('proposals') && (
               <SourceType
+                disabled={!isAdmin}
                 data-test='source-proposals'
+                disabledTooltip={!isAdmin ? 'Only admins can create proposals as datasource boards' : ''}
```
mattcasey, line 185: you should just let the caller define disabledTooltip, and it is only used if disabled is true. this avoids having to put logic here

components/common/PageLayout/components/Sidebar/components/SidebarButton.tsx
```diff
@@ -64,7 +64,7 @@ export function SidebarLink({
 }: {
   active: boolean;
   href?: string;
-  icon: any;
+  icon?: any;
```
mattcasey, line 67: making this optional doesnt seem necessary

lib/databases/board.ts
```diff
@@ -90,6 +92,7 @@ export type GoogleFormSourceData = {
 
 export type BoardFields = {
   icon: string;
+  selectedProperties?: SelectedProposalProperties;
```
mattcasey, line 95: i was able to remove this and it still compiles

lib/databases/proposalsSource/getBlocks.ts
Devorein: The reason I'm moving it is because in the other place `proposal-source.ts` its using `selectedProperties` provided by the user, which for the rest of the calls we auto generate them.

lib/databases/board.ts
Devorein: my bad forgot to remove it. Good catch.

components/common/PageLayout/components/Sidebar/components/SidebarButton.tsx
Devorein: Another good catch, initially I used this component without icon but then replaced it with a custom one.

components/common/DatabaseEditor/components/viewSidebar/viewSourceOptions/viewSourceOptions.tsx
Devorein: yeah sure

__integration-tests__/server/pages/api/blocks/[id]/subtree.spec.ts
Comment (Devorein) [reply __integration-tests__/server/pages/api/blocks/[id]/subtree.spec.ts:583]: Good point. Updated.

pages/api/blocks/index.ts
Devorein: Alright I will just remove this restriction for non admins for now.

PR #4122, opened 2024-05-31
Poc monorepo

motechFR:


apps/builder-api/tsconfig.json
```diff
@@ -0,0 +1,104 @@
...
+    // "noUncheckedIndexedAccess": true,                 /* Add 'undefined' to a type when accessed using an index. */
+    // "noImplicitOverride": true,                       /* Ensure overriding members in derived classes are marked with an override modifier. */
+    // "noPropertyAccessFromIndexSignature": true,       /* Enforces using indexed accessors for keys declared using an indexed type. */
+    // "allowUnusedLabels": true,                        /* Disable error reporting for unused labels. */
+    // "allowUnreachableCode": true,                     /* Disable error reporting for unreachable code. */
+
+    /* Completeness */
+    // "skipDefaultLibCheck": true,                      /* Skip type checking .d.ts files that are included with TypeScript. */
+    "skipLibCheck": true                                 /* Skip type checking all .d.ts files. */
+  }
```
mattcasey, line 103: please remove any fields we dont use. its just noisy

apps/builder-network/.gitignore
```diff
@@ -0,0 +1,36 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
```
mattcasey, line 4: seems like we could add these all to our top-level .gitignore?

apps/builder-network/.gitignore
```diff
@@ -0,0 +1,36 @@
...
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# local env files
+.env*.local
+
+# vercel
+.vercel
```
mattcasey, line 32: we dont use vercel

apps/builder-network/app/globals.css
```diff
@@ -0,0 +1,107 @@
+:root {
```
mattcasey, line 1: Should be scss

apps/builder-network/app/globals.css
mattcasey, line 1: please copy existing naming conventions, in this case it should be "cssVariables.scss" - or we can rename them both to "globals.scss", i dont care too much except that we have consistency

apps/builder-network/next.config.mjs
mattcasey, line 1: whats .mjs extension for?

apps/builder-network/tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
+  "compilerOptions": {
```
mattcasey, line 2: can this import from tsconfig.next.json?

package.json
```diff
@@ -72,7 +75,11 @@
     "update-core": "npm install @charmverse/core",
     "build:storybook": "npx storybook build",
     "upgrade:storybook": "npx storybook@latest upgrade",
-    "chromatic:ci": "chromatic --only-changed --exit-zero-on-changes"
+    "chromatic:ci": "chromatic --only-changed --exit-zero-on-changes",
+    "builder-network": "cd apps/builder-network && npm run dev",
+    "builder-api": "npm run start --workspace=apps/builder-api",
+    "builder-all": "npm run builder-api & npm run builder-app",
```
mattcasey, line 81: use concurrently to run tasks in parallel, this way we can see logs from both for example, i think it handles closing the processes better too

mattcasey, line 81: the default command should be the 'local' command. that's because we use it all the time. this could be an alias if you want to run "npm run builder-all:local"

package.json
```diff
@@ -72,7 +75,11 @@
     "update-core": "npm install @charmverse/core",
     "build:storybook": "npx storybook build",
     "upgrade:storybook": "npx storybook@latest upgrade",
-    "chromatic:ci": "chromatic --only-changed --exit-zero-on-changes"
+    "chromatic:ci": "chromatic --only-changed --exit-zero-on-changes",
+    "builder-network": "cd apps/builder-network && npm run dev",
```
mattcasey, line 79: maybe use builder as a prefix, eg "builder:client", "builder:api", "builder:all". I think "client" makes more sense than network, which could be anything, but am open to other ideas?

apps/builder-network/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "builder-network",
```
mattcasey, line 2: suggestion: "builder-client" or even "@charmverse/builder/client?

shared/components/elements/Button.tsx
```diff
@@ -0,0 +1,106 @@
+'use client';
```
mattcasey, line 1: i think for right now we could just import components as needed where they exist today. i don't like the idea of adding yet another place to organize components - see our test folders

theme/AppThemeProvider.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
```
mattcasey, line 1: i'd like to see MUI + SCSS set up properly using app router. That was a complicated bit of work

apps/builder-network/next.config.mjs
```diff
@@ -0,0 +1,13 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  rewrites: async () => {
+    return [
+      {
+        source: '/api/:path*',
```
mattcasey, line 6: nice, was hoping we could do somethin like this

apps/connect/next.config.mjs
```diff
@@ -0,0 +1,16 @@
+/** @type {import('next').NextConfig} */
```
mattcasey, line 1: why is this mjs

apps/connect/public/vercel.svg
mattcasey, line 1: why do we need this file

apps/connect/public/next.svg
mattcasey, line 1: why do we need this file

components/common/Button.tsx
mattcasey, line 1: remove 'use client'

apps/connect/tsconfig.json
```diff
@@ -0,0 +1,42 @@
...
+      "@charmverse/core/pages": ["../../node_modules/@charmverse/core/dist/cjs/pages"],
+      "@charmverse/core/pages/utilities": ["../../node_modules/@charmverse/core/dist/cjs/pages-utilities"],
+      "@charmverse/core/proposals": ["../../node_modules/@charmverse/core/dist/cjs/proposals"],
+      "@charmverse/core/bounties": ["../../node_modules/@charmverse/core/dist/cjs/bounties"],
+      "@charmverse/core/utilities": ["../../node_modules/@charmverse/core/dist/cjs/utilities"],
+      "@charmverse/core/errors": ["../../node_modules/@charmverse/core/dist/cjs/errors"]
+    }
+  },
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "exclude": ["../../node_modules"]
```
mattcasey, line 41: why do we need to exclude a parent folder

apps/connect/tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
+  "compilerOptions": {
```
mattcasey, line 2: can this extend our base tsconfig?

apps/connect/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "connect",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "npx next dev",
```
mattcasey, line 6: Next should be defined as a dependency. npx is good for command line utils but i don't think we should use it to install code that's run in production. for starters, the current command `npx next` would always grab whatever is the latest version of next.js

apps/connect/next.config.mjs
```diff
@@ -0,0 +1,16 @@
...
+
+const CONNECT_API_URL = process.env.CONNECT_API_URL;
+
+const nextConfig = {
+  rewrites: async () => {
+    return [
+      {
+        source: '/api/:path*',
```
mattcasey, line 9: thinking about this more, while it's a neat trick to use a proxy, CORS would work just fine too. i think proxying requests thru the webapp will just add overhead and complexity to our system. are there any benefits i might be missing?

apps/connect-api/jest.config-server.ts
```diff
@@ -0,0 +1,9 @@
+import type { JestConfigWithTsJest } from 'ts-jest';
+
+import defaultJestConfig from './jest.config';
+
+const serverJestConfig: JestConfigWithTsJest = {
+  ...defaultJestConfig,
+  testMatch: ['**/__integration-tests__/**/*.spec.ts']
```
mattcasey, line 7: we don't need integration tests for a single koa server. it's fine to use supertest to test endpoints and use unit tests

mattcasey (approved):

PR #4125, opened 2024-06-03
Fix: Database filters are not responsive on mobile

valentinludu:
Task:
https://app.charmverse.io/charmverse/database-filters-are-not-responsive-on-mobile-48855657842594336

Fix:
Let the user scroll left-right in the popover so he can access all the filters.


mattcasey (approved):
It is not really "responsive" if you still have to scroll to see everything, but it's better than it was before

PR #4126, opened 2024-06-03
Fix: Cursor jumps out of box when adding rubric criteria

valentinludu:
Task:
https://app.charmverse.io/charmverse/each-time-i-start-typing-a-new-criteria-in-the-add-a-label-field-the-cursor-jumps-out-after-the-first-letter-or-two-8551001433787735

Fix:
- Remove useEffect since it was making the whole component to rerender and the input was loosing focus
- Added debounce so there aren't so many updates per second


components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/RubricCriteriaSettings.tsx
```diff
@@ -91,19 +92,18 @@ export function RubricCriteriaSettings({ readOnly, showDeleteConfirmation, value
...
+      if (newCriteriaList.every((rubricCriteria) => isValidCriteria(rubricCriteria, answers))) {
+        onChange(newCriteriaList);
       }
     }
   }
 
-  useEffect(() => {
-    setCriteriaList(value);
-  }, [value]);
+  const debouncedSetCriteriaProperty = debounce(setCriteriaProperty, 300);
```
mattcasey, line 106: debounced methods should be memoized, otherwise i dont think this is actually debouncing anything. but i dont think that was causing the problem

mattcasey (changes requested):
The useEffect is necessary for a template feature where we copy rubric criteria from other templates. Without it, the criteria do not update. i'm open to other ideas to make it work:
image

components/proposals/ProposalPage/components/ProposalEvaluations/components/Settings/components/RubricCriteriaSettings.tsx
valentinludu: Done!

valentinludu:
I see...

I added it back but only when the length of the value changes so it doesn't affect the inner values that are changing.

> The useEffect is necessary for a template feature where we copy rubric criteria from other templates. Without it, the criteria do not update. i'm open to other ideas to make it work

mattcasey (approved):

PR #4131, opened 2024-06-04
Custom email branding for space

Devorein:


components/settings/notifications/components/SpaceEmailBrandingSettings.tsx
```diff
@@ -0,0 +1,65 @@
+import { Box, Grid, Stack, TextField, Typography } from '@mui/material';
```
mattcasey, line 1: NotificationSettings really should be a folder and this should be inside of NotificaitonSettings/components

components/settings/notifications/components/SpaceEmailBrandingSettings.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const watchEmailBrandColor = watch('emailBrandColor');
+
+  return (
+    <>
+      <Grid item>
+        <FieldLabel>Email Brand Image</FieldLabel>
+        <Typography variant='caption' mb={2} component='p'>
+          Customize the image shown in email.
+        </Typography>
+        <Avatar
```
mattcasey, line 29: <Controller would be cleaner, instead of using watch() and setValue

components/settings/notifications/components/SpaceEmailBrandingSettings.tsx
```diff
@@ -0,0 +1,65 @@
...
+            helperText={errors.emailBrandColor?.message}
+            placeholder={blueColor}
+          />
+          <Box
+            sx={{
+              border: (theme) => `1px solid ${theme.palette.divider}`,
+              width: '40px',
+              height: '40px',
+              borderRadius: (theme) => theme.spacing(0.5),
+              backgroundColor: watchEmailBrandColor?.startsWith('#') ? watchEmailBrandColor : `#${watchEmailBrandColor}`
```
mattcasey, line 58: we should not need to be supporting two formats. either # is included or it is not

mattcasey: and please make sure we store it WITH the "#"

Devorein: yep guaranteed to store with `#`. I will enforce providing the # just to be consistent.

components/settings/notifications/components/SpaceNotificationSettings.tsx
```diff
@@ -185,18 +116,18 @@ function ToggleInput({
...
   disabled,
   types: typeColumns
 }: {
   disabled: boolean;
-  control: any;
   enabled?: boolean;
   title: string;
   types: (typeof notificationTypes)['rewards']['types'];
 }) {
+  const { control } = useFormContext<FormValues>();
```
mattcasey, line 129: this isnt used?

components/settings/notifications/components/SpaceNotificationSettings.tsx
```diff
@@ -252,3 +183,63 @@ function NotificationRuleComponent({
     </Box>
   );
 }
+
+export function NotificationTogglesInput({ isAdmin }: { isAdmin: boolean }) {
+  const { watch } = useFormContext<FormValues>();
+  const { getFeatureTitle } = useSpaceFeatures();
+  const formValues = watch();
```
mattcasey, line 190: not sure u are using this

lib/mailer/emails/templates/components/Link.tsx
```diff
@@ -4,14 +4,21 @@ import type { ReactNode } from 'react';
 
 import { blueColor } from 'theme/colors';
 
-export default function Link({ children, ...props }: { children: ReactNode } & LinkProps) {
+export default function Link({
+  primaryColor = blueColor,
```
mattcasey, line 8: this default should really be defined in one place, and then each of thes e components just require a color to be provided. that way its easy to change the deafult. but this is fine for now

mattcasey (approved):

components/settings/notifications/components/SpaceNotificationSettings.tsx
Devorein: Its being used

PR #4139, opened 2024-06-06
Create page comment for email replies

Devorein:


.ebstalk.apps.env/webapp.env
```diff
@@ -24,7 +24,8 @@ GOOGLE_OAUTH_CLIENT_SECRET_SENSITIVE="{{pull:secretsmanager:/io.cv.app/prd/googl
 GOOGLE_OAUTH_ENCRYPTION_SECRET="{{pull:secretsmanager:/io.cv.app/prd/google-auth:SecretString:google_oauth_encryption_secret}}"
 LOOPS_EMAIL_API_KEY="{{pull:secretsmanager:/io.cv.app/prd/loops-email:SecretString:loops_email_api_key}}"
 MAILGUN_DOMAIN="{{pull:secretsmanager:/io.cv.app/shared/mailgun:SecretString:mailgun_domain}}"
-MAILGUN_KEY="{{pull:secretsmanager:/io.cv.app/shared/mailgun:SecretString:mailgun_key}}"
+MAILGUN_API_KEY="{{pull:secretsmanager:/io.cv.app/shared/mailgun:SecretString:mailgun_api_key}}"
+MAILGUN_SIGNING_KEY="{{pull:secretsmanager:/io.cv.app/shared/mailgun:SecretString:mailgun_signing_key}}"
```
mattcasey, line 28: technically you don't need a signing key for webapp right? this is just used to retrieve messages from the webhook. its good to avoid adding envs unless they are used

background/tasks/processMailgunWebhookMessages/task.ts
```diff
@@ -0,0 +1,13 @@
+import { log } from '@charmverse/core/log';
+
+import { processMessages } from 'lib/aws/webhookSqs';
+import { SQS_WEBHOOK_MAILGUN_QUEUE_NAME } from 'lib/mailer/mailgunClient';
+import { processWebhookMessage } from 'lib/mailer/webhook/processWebhookMessage';
+
+log.info('Mailgun Queue url:', SQS_WEBHOOK_MAILGUN_QUEUE_NAME);
+const queueUrl = SQS_WEBHOOK_MAILGUN_QUEUE_NAME || '';
```
mattcasey, line 8: if queue url is not defined, the task shouldn't even run. just log an error instead maybe. and then u don't need a fallback value

lib/mailer/getMailgunMesage.ts
```diff
@@ -0,0 +1,37 @@
+import * as http from 'adapters/http';
+
+import { API_KEY, DOMAIN } from './mailgunClient';
+
+interface EmailMessage {
```
mattcasey, line 5: type?

lib/mailer/getMailgunMesage.ts
```diff
@@ -0,0 +1,37 @@
...
+  'Message-Id': string;
+  'Message-Headers': [string, string][];
+  'stripped-text': string;
+  'Body-HTML': string;
+}
+
+async function getMailgunMessage({ domainName, messageId }: { messageId: string; domainName: string }) {
+  const emailMessage = await http.GET<EmailMessage>(
+    `https://api.mailgun.net/v3/domains/updates.${DOMAIN}/messages/${messageId}`,
+    {},
```
mattcasey, line 28: i believe marek fixed it so that you dont need to include this argument if its empty

lib/mailer/getMailgunMesage.ts
mattcasey, line 1: this filename has a typo

lib/mailer/webhook/processWebhookMessage.ts
```diff
@@ -0,0 +1,142 @@
...
+import { WebhookEventNames } from 'lib/webhookPublisher/interfaces';
+import { publishDocumentEvent } from 'lib/webhookPublisher/publishEvent';
+
+import { SIGNING_KEY } from '../mailgunClient';
+
+type WebhookPayload = {
+  body: string;
+};
+
+interface MailgunWebhookPayload {
```
mattcasey, line 17: type

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -376,7 +377,8 @@ export async function createDocumentNotifications(webhookData: {
               pageCommentId: documentId ? commentId : undefined,
               postCommentId: postId ? commentId : undefined,
               userId: authorId,
-              content: comment.content
+              content: comment.content,
+              messageId
```
mattcasey, line 381: reminder: we dont store incoming message ids

mattcasey (approved):

.ebstalk.apps.env/webapp.env
Devorein: Good point. It should only be on the cron.

background/tasks/processMailgunWebhookMessages/task.ts
Devorein: Updated. But we do it the same in all the other places. Thats why I just copied and pasted this part.

lib/mailer/getMailgunMesage.ts
Devorein: We actually don't even need to use this method anymore.

lib/mailer/webhook/processWebhookMessage.ts
Devorein: nice

lib/notifications/documents/createDocumentNotifications.ts
Devorein: Right. Updated.

PR #4166, opened 2024-06-12
Proposal source board rubric evaluation criteria reviewer comment and score properties

Devorein:

### DIFF
--- a/lib/databases/board.ts
+++ b/lib/databases/board.ts
@@ -21,7 +21,9 @@ export const proposalPropertyTypesList = [
   'proposalEvaluationType',
   'proposalCreatedAt',
   'proposalStep',
-  'proposalReviewerNotes'
+  'proposalReviewerNotes',
+  'proposalRubricCriteriaReviewerScore',
+  'proposalRubricCriteriaReviewerComment'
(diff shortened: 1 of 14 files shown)


lib/databases/board.ts
```diff
@@ -21,7 +21,9 @@ export const proposalPropertyTypesList = [
   'proposalEvaluationType',
   'proposalCreatedAt',
   'proposalStep',
-  'proposalReviewerNotes'
+  'proposalReviewerNotes',
+  'proposalRubricCriteriaReviewerScore',
+  'proposalRubricCriteriaReviewerComment'
 ] as const;
```
mattcasey, line 27: Not for this PR, but i want to find a way that doesn't require maintaining a list like this. 'type' should really just be 'text' or 'select', etc.. maybe a new field that is not 'id' or 'type', maybe we just need more configuration properties

mattcasey (approved):

lib/databases/board.ts
Devorein: I agree, but I believe we decided on using a consistent ID to indicate a special type of column.

PR #4168, opened 2024-06-12
Farcaster login

valentinludu:
Test cases:
- login with Farcaster and create an account automatically
- login with Metamask and create an account. Then login with farcaster. It should fail because you already have an account created with metamask
- login with metamask, go to account settings and connect Farcaster
- Disconnect Farcaster
- Use Farcaster as your identity


mattcasey:
This adds .03MB to our default app size, can we lazy-load the library only when users want to log in w/farcaster?

valentinludu:
Done!

> This adds .03MB to our default app size. It goes from 1.33 -> 1.36MB. can we lazy-load the library only when users want to log in w/farcaster?

mattcasey (approved):

PR #4173, opened 2024-06-13
Collapse evaluation section for creating proposal with template

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/proposals-collapse-the-step-sections-when-you-are-editing-a-template-and-using-a-proposal-template-7553013986177897)


mattcasey (approved):
i tested out a few cases and it seems to work well!

PR #4179, opened 2024-06-14
Revert "Fix user select property"

mattcasey:
This is a fix for cases like AddPermissionsPageForm which does not update the defaultValue when onChange handler is called. This input should maintain its own internal state of what is selected or else require a 'value' prop

Reverts charmverse/app.charmverse.io#4032 and also fixes original issue in a different way

### DIFF
--- a/components/common/DatabaseEditor/components/properties/UserSelect.tsx
+++ b/components/common/DatabaseEditor/components/properties/UserSelect.tsx
@@ -1,11 +1,13 @@
 import styled from '@emotion/styled';
 import CloseIcon from '@mui/icons-material/Close';
+import type { AutocompleteRenderGetTagProps } from '@mui/material';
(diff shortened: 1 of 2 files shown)


valentinludu:
Actually the fix should be here: components/common/PageLayout/components/Header/components/ShareButton/components/PaidPagePermissions/AddPagePermissionsForm.tsx

```
<InputSearchMemberMultiple
              defaultValue={selectedUserIds}
```

The InputSearchMemberMultiple was missing a defaultValue

mattcasey:
found a better solution, instead of the UserSelect component expecting InputSearchMemberMultiple to track defaultValue, it uses an internal method from Autocomplete to remove items. This feels cleaner, now the tags update the internal state of Autocomplete directly and we don't have to require that parent components manage everything.

valentinludu (approved):
Tested and works

PR #4183, opened 2024-06-18
Farcaster member directory property and required login identity

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/allow-space-admins-to-require-members-to-connect-farcaster-identity-4995024008006026)


components/login/components/WarpcastLogin.tsx
```diff
@@ -12,7 +12,7 @@ import type { LoggedInUser } from 'models';
 
 import { FarcasterLoginModal } from './FarcasterModal';
 
-function WarpcastLoginButton({ type }: { type: LoginType }) {
+export function useWarpcastLogin({ type }: { type: LoginType }) {
```
mattcasey, line 15: if this is a hook, it should be in a hook file

mattcasey (changes requested):
this adds .02MB to the payload, looks like it is probably adding the warpcast SDK to our app. please use lazy loading like we are doing with LoginButton

components/_app/components/ConnectedAccounts.tsx
```diff
@@ -1,10 +1,13 @@
+import { AuthKitProvider } from '@farcaster/auth-kit';
 import { Stack, Typography } from '@mui/material';
 import { type ReactNode } from 'react';
 
 import { useAddUserWallets } from 'charmClient/hooks/profile';
 import { Button } from 'components/common/Button';
 import { FieldWrapper } from 'components/common/form/fields/FieldWrapper';
+import { FarcasterLoginModal } from 'components/login/components/FarcasterModal';
 import { WalletSign } from 'components/login/components/WalletSign';
+import { useWarpcastLogin } from 'components/login/components/WarpcastLogin';
```
mattcasey, line 10: this should be lazy-loaded

components/members/components/MemberDirectorySidebar/MemberPropertyItem.tsx
```diff
@@ -14,6 +15,29 @@ import { iconForPropertyType } from 'components/common/DatabaseEditor/widgets/ic
...
 import { MEMBER_PROPERTY_CONFIG, PREMIUM_MEMBER_PROPERTIES } from 'lib/members/constants';
 import DiscordIcon from 'public/images/logos/discord_logo.svg';
+import FarcasterLogo from 'public/images/logos/farcaster.svg';
+
+function FarcasterIcon() {
+  const theme = useTheme();
+
+  if (theme.palette.mode === 'dark') {
+    return (
+      <SvgIcon viewBox='0 0 20 20' fontSize='small'>
```
mattcasey, line 25: we should be using CSS to modify the SVG based on light mode. using JS is very hacky, but not a blocker

lib/members/getSpaceMembers.ts
```diff
@@ -142,7 +143,8 @@ export async function getSpaceMembers({
           properties,
           searchValue: getMemberSearchValue(spaceRole.user, visiblePropertiesMap, memberUsernameRecord[spaceRole.id]),
           roles,
-          isBot: userData.isBot ?? undefined
+          isBot: userData.isBot ?? undefined,
+          farcasterUser: userData.farcasterUser || undefined
```
mattcasey, line 147: please map this to the fields you need. its a bad practice to just send the whole JSON forward, also requires type-casting

lib/members/interfaces.ts
```diff
@@ -54,6 +55,7 @@ export type Member = Pick<User, 'id' | 'createdAt' | 'updatedAt' | 'username'> &
   roles: Pick<Role, 'name' | 'id'>[];
   onboarded: boolean;
   searchValue: string;
+  farcasterUser?: FarcasterUser;
```
mattcasey, line 58: just pick the fields we need here

mattcasey (changes requested):
Required changes requested:
- dont send the entire farcaster blob to the frontend
- lazy-load the auth kit

components/login/components/WarpcastLogin.tsx
Devorein: Updated

components/_app/components/ConnectedAccounts.tsx
Devorein: Updated

components/members/components/MemberDirectorySidebar/MemberPropertyItem.tsx
Devorein: Removed the explicit fill value and it worked for both theme.

lib/members/interfaces.ts
Devorein: Updated

PR #4188, opened 2024-06-19
combine staging workflows

mattcasey:


apps/connect-api/src/server.ts
```diff
@@ -1,23 +1,24 @@
...
-
 // CORS middleware configuration
 app.use(
   cors({
     origin: (ctx) => {
       const origin = ctx.request.headers.origin as string;
-      const isAllowed = allowedOrigins.includes(origin as string);
-      if (isAllowed) {
-        return origin; // Allow the request from this origin
+      if (isDevEnv || isTestEnv) {
```
mattcasey, line 16: i removed this allowed origins setting and replaced it with the logic we have for web sockets, which doesn't require extra CI magic - just check the top-level domain. We'll need a solution to support non-charmverse domains too but that can just be a simple static env var

apps/connect/components/projects/ProjectsList.tsx
```diff
@@ -1,3 +1,5 @@
+import 'server-only';
```
valentinludu, line 1: Niiiice. 

I was just reading about this one. We should discuss with the team about it in our meeting.

apps/connect/package.json
```diff
@@ -9,6 +9,8 @@
     "lint": "npx next lint"
   },
   "dependencies": {
-    "@mui/material-nextjs": "^5.15.11"
+    "@mui/material-nextjs": "^5.15.11",
+    "material-ui-popup-state": "^5.1.2",
```
valentinludu, line 13: I tested, had the same issue.

You can remove it from here and update it in the main package.

valentinludu:
Since you updated @types/react you could update @types/react-dom as well. These 2 are always needed to be kept up to date.

Also, a small detail, @types/koa-router should be in devDependencies.

apps/connect/package.json
mattcasey: done thanks

PR #4190, opened 2024-06-20
Allow users to select Projects from OP

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/allow-users-to-select-projects-from-op-576010101925365)


lib/projects/getOpProjectsByFarcasterId.ts
```diff
@@ -0,0 +1,6 @@
+import { getOpProjects } from 'lib/optimism/getOpProjects';
+
+export async function getOpProjectsByFarcasterId({ farcasterId }: { farcasterId: number }) {
+  const opProjects = await getOpProjects();
+  return opProjects.filter((project) => project.team.find((member) => member.farcasterId !== farcasterId.toString()));
```
mattcasey, line 5: dont u want to do the opposite? only find projects where your farcater matches?

mattcasey (approved):

lib/projects/getOpProjectsByFarcasterId.ts
Devorein: yes but that returns no data since there is a single mock data and it doesn't match with any farcaster account. We can just probably comment out this line but removing and adding an extra ! was easier.

PR #4196, opened 2024-06-21
UI improvements for proposal appeal step

Devorein:


components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/PassFailEvaluationContainer.tsx
```diff
@@ -211,7 +211,13 @@ export function PassFailEvaluationContainer({
                         Appeal reason
                       </Typography>
                     </FormLabel>
-                    <Typography>{evaluation.appealReason}</Typography>
+                    <Typography
+                      sx={{
+                        wordBreak: 'break-word'
```
mattcasey, line 216: won't this be weird? why would we break in the middle of the word

components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/PassFailEvaluationContainer.tsx
```diff
@@ -226,6 +232,12 @@ export function PassFailEvaluationContainer({
                   evaluationResult={evaluation.result}
                   evaluationReviews={evaluation.appealReviews}
                   requiredReviews={evaluation.appealRequiredReviews ?? 1}
+                  reviewerOptions={
+                    evaluation.appealReviewers?.map((reviewer) => ({
```
mattcasey, line 236: for future reference this mapping should be done inside PassFailEvaluation, because it's the one calling UserAndRoleSelect that requires this type

scripts/optimism/updateAppealReviewers.ts
```diff
@@ -0,0 +1,139 @@
+import { getCurrentEvaluation } from '@charmverse/core/dist/cjs/proposals';
+import { prisma } from '@charmverse/core/prisma-client';
+import { isTruthy } from 'lib/utils/types';
+
+const proposalConfigs = [
```
mattcasey, line 5: do not commit this to code. it contains personal information, it shouldnt even be in this PR

scripts/optimism/updateAppealReviewers.ts
```diff
@@ -0,0 +1,139 @@
...
+                  some: {
+                    email: reviewer.toLowerCase()
+                  }
+                }
+              },
+            ]
+          }
+        })))).filter(isTruthy);
+
+        await prisma.$transaction([
```
mattcasey, line 114: please add checks that we actually found a user for all reviewers

scripts/optimism/updateAppealReviewers.ts
```diff
@@ -0,0 +1,139 @@
...
+          id: true,
+          evaluations: {
+            orderBy: {
+              index: "asc"
+            }
+          }
+        }
+      });
+  
+      const currentEvaluation = getCurrentEvaluation(proposal.evaluations);
```
mattcasey, line 81: dont use current evaluation. they explicitly gave us the step to update

scripts/optimism/updateAppealReviewers.ts
```diff
@@ -0,0 +1,139 @@
...
+            data: reviewers.map(reviewer => ({
+              evaluationId: currentEvaluation.id,
+              userId: reviewer.id,
+              proposalId: proposal.id,
+            }))
+          })
+        ])
+      }
+    } catch (error) {
+      console.error(error);
```
mattcasey, line 131: script should fail and not continue

components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/PassFailEvaluationContainer.tsx
Devorein: No actually this won't break in the middle of words only if it has number or urls it would split it at the edge. Here is a before and after comparison.

Before:
image

After:
image

components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/PassFailEvaluationContainer.tsx
Devorein: Its already done in the parent component for regular non appeal view, this just overrides it for the appeal view.

PR #4199, opened 2024-06-24
Fix UI explanation for public proposals

motechFR:


lib/utils/prepareDatadogQuery.ts
mattcasey, line 1: did you mean to include this?

mattcasey: please make it lib/utils/datadog.ts if we do add datadog utils

PR #4207, opened 2024-06-25
Connect project view

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/project-view-03436387489019421)


apps/connect-api/src/main.ts
```diff
@@ -2,7 +2,7 @@ import { log } from '@charmverse/core/log';
 
 import { app } from './server';
 
-const port = process.env.PORT || 4000;
+const port = 4000;
```
mattcasey, line 5: let docker override port

apps/connect-api/lib/aws/uploadToS3Server.ts
mattcasey, line 1: is this a copy from lib? i think u can just import from our top-level lib now

PR #4211, opened 2024-06-26
Fix generating a signer

motechFR:
### WHAT

TODO
- Select final Fid for connecting
- Clean up types for the signer request


pages/api/farcaster/signer.ts
```diff
@@ -22,8 +22,8 @@ export type FarcasterSignerResponse = {
   token: string;
   deeplinkUrl: string;
 };
-
-const appFid = isProdEnv ? 1501 : 318061;
+// CharmVerse Dev account
```
mattcasey, line 25: why are we removing the production account as a fix?

PR #4221, opened 2024-06-27
Connect Project page

Devorein:
### WHY

image


apps/connect/apiClient/apiClient.ts
```diff
@@ -14,9 +15,9 @@ class ConnectApiClient extends HttpClient {
   }
 
   async uploadImage(file: File) {
-    // Using this causes issue during upload
+    // Using 'this.GET' causes issue during upload
```
mattcasey, line 18: woops, thanks for the comment

apps/connect/app/projects/[id]/page.tsx
```diff
@@ -0,0 +1,11 @@
+import { ProjectDetailsSkeleton } from '@connect/components/projects/components/ProjectDetailsSkeleton';
+import { ProjectDetails } from '@connect/components/projects/ProjectDetails';
```
Comment (mattcasey) [inline apps/connect/app/projects/[id]/page.tsx:2]: include "Page" at the end of the component, and it should live at /projects/[id]/ProjectDetailsPage i think

mattcasey (approved):

apps/connect/app/projects/[id]/page.tsx
Comment (motechFR) [reply apps/connect/app/projects/[id]/page.tsx:2]: Counterpoint to this, other code paths such as profile/page have a very minimal page component, and the actual page imported from components

motechFR (approved):
Works well, but I'm getting these errors from React

https://react.dev/errors/418?invariant=418
https://react.dev/errors/422?invariant=422

apps/connect/app/projects/[id]/page.tsx
Comment (mattcasey) [reply apps/connect/app/projects/[id]/page.tsx:2]: that's not a counter to what i said. i was just asking him to rename the component so it is more standard like the other components

apps/connect/app/projects/[id]/page.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { Suspense } from 'react';
+
+export default function ProjectPage({ params }: { params: { id: string } }) {
+  return (
+    <Suspense fallback={<ProjectDetailsSkeleton />}>
```
Comment (valentinludu) [inline apps/connect/app/projects/[id]/page.tsx:7]: Suspense is not going to work here. The loading.ts will run here.

Suspense is good to run inside nested server components. 
e.g. ProjectsList inside Homepage. The ProjectsList can run slower then the Homepage so it has a skeleton and the user can see the Homepage with a skeleton until ProjectsList will be available.

If you want you could add a loading.ts file and add the skeleton here

apps/connect/components/projects/components/ProjectDescription.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <Typography
+      variant='body1'
+      mt={2}
+      onClick={() => {
+        setShowFullDescription(!showFullDescription);
+      }}
+    >
+      {showFullDescription ? description : fancyTrim(description, 500)}
+      <Typography
+        variant='body2'
```
valentinludu, line 21: Add here 
`variant='body2'`

And the hydration error will go away.

It was actually the validation of the DOM

apps/connect/components/projects/ProjectDetails.tsx
```diff
@@ -0,0 +1,154 @@
...
+                  height: 24
+                }}
+              />
+              <Link href={project.twitter} passHref target='_blank'>
+                <Typography variant='body1'>{project.twitter.replace(/https?:\/\/twitter\.com\//, '')}</Typography>
+              </Link>
+            </Stack>
+          )}
+        </Stack>
+        {project.description && <ProjectDescription description={project.description} />}
```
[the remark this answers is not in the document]
valentinludu: Nvm. I found the problem and commented below.

apps/connect/components/projects/ProjectDetails.tsx
```diff
@@ -0,0 +1,154 @@
+import { fetchProject } from '@connect/lib/actions/fetchProject';
```
valentinludu, line 1: Add `import server-only` here so we know it should not have client related code .

Devorein: Nice suggestion. Added.

apps/connect/app/projects/[id]/page.tsx
Comment (Devorein) [reply apps/connect/app/projects/[id]/page.tsx:7]: Thank you I was wondering why the skeleton was not showing up even with an artificial delay. This works perfectly now.

apps/connect/lib/actions/fetchProject.ts
```diff
@@ -0,0 +1,51 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import type { StatusAPIResponse } from '@farcaster/auth-kit';
+
+export async function fetchProject(projectId: string) {
```
valentinludu, line 4: Please add noStore here or add the 'force-dynamic' flag in the page.tsx

https://nextjs.org/docs/app/api-reference/functions/unstable_noStore

PR #4228, opened 2024-06-27
fix cookie config for staging

mattcasey:


motechFR (commented):
Is this desirable? Wouldn't it be better to have a single, cross app isDevEnv // isStagingEnv etc?

mattcasey:
> Is this desirable? Wouldn't it be better to have a single, cross app isDevEnv // isStagingEnv etc?

Yes i made a task for it.

PR #4229, opened 2024-06-28
add gallery image to applications

mattcasey:
User request to show gallery images for rewards. might require a migration for them


Devorein (changes requested):
1. Should it support submissions too? Clicking on the viewsidebar > Source > Submissions doesn't do anything
2. Newly created board pages are throwing 404 error from `api/blocks/[id]`, maybe related to removing the filter in Gallery cards?

mattcasey:
> Should it support submissions too?

No, we only support kanban for submissions right now. I'll look at the error though, thanks!

Devorein (approved):

PR #4232, opened 2024-06-28
Public connect project pages

Devorein:


apps/connect/middleware.ts
```diff
@@ -12,7 +12,7 @@ export async function middleware(request: NextRequest) {
   const user = session.user;
   const path = request.nextUrl.pathname;
 
-  if (!user && path !== '/') {
+  if (!user && path !== '/' && !path.startsWith('/projects/')) {
```
motechFR, line 15: edge case: projects/new would fall into this case

Devorein: Hmm good catch. Forgot about that one. We are going to replace projects path with /p anyways so will make that change

mattcasey (approved):

lib/farcaster/getFarcasterUsers.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (username) {
+    const farcasterUsersResponse = await GET<FarcasterUsersResponse>(
+      `${neynarBaseUrl}/user/search`,
+      {
+        q: username,
+        limit: 5
+      },
+      {
+        headers: {
+          Api_key: process.env.NEYNAR_API_KEY
```
mattcasey, line 61: do u need to add this to the env config?

Devorein: no this was added before

PR #4235, opened 2024-07-01
Connect user profile page

Devorein:


apps/connect/components/common/FarcasterCard.tsx
```diff
@@ -23,40 +23,42 @@ export function FarcasterCard({
...
-            xs: 'column',
-            sm: 'row'
-          }
-        }}
-      >
-        <Stack
-          alignItems={{
-            xs: 'center',
-            sm: 'flex-start'
+      <CardActionArea LinkComponent={Link} href={`/u/${username}`} hrefLang='en'>
```
valentinludu, line 26: Better to use Link from 'next/link'

I usually use Link from Mui when sending the user outside.

Link from next/link has the prefetch prop turned on, so a new page is prefetch before the user clicks it. Super useful for performance.
https://nextjs.org/docs/pages/api-reference/components/link#prefetch

apps/connect/components/common/FarcasterCard.tsx
```diff
@@ -23,40 +23,42 @@ export function FarcasterCard({
...
+            <Stack direction='row' justifyContent='space-between' width='100%' alignItems='center'>
+              <Typography variant='h6'>{name || 'N/A'}</Typography>
+              {onDelete && (
+                <IconButton size='small'>
+                  <DeleteOutline color='error' onClick={onDelete} fontSize='small' />
+                </IconButton>
+              )}
+            </Stack>
+            <Typography variant='subtitle1' color='secondary'>
+              {username || 'N/A'} #{fid || 'N/A'}
```
valentinludu, line 56: In the new design user has "@" in front of the username

apps/connect/components/common/PageWrapper.tsx
```diff
@@ -1,20 +1,51 @@
...
-      p={3}
+      p={backToProfileHeader ? 0 : 3}
       pb={4}
       borderRadius={8}
       bgcolor='background.default'
       border={{ xs: 'none', md: '1px solid var(--charm-palette-divider)' }}
       maxWidth='500px'
       mx='auto'
       my={2}
+      position={backToProfileHeader ? 'relative' : undefined}
```
valentinludu, line 23: In the new design there is no "back to...". In my PR I am deleting the Back to profile. There will be a user profile in the top-right.
https://www.figma.com/design/uGO5TTywk0VE2yvHcbErdQ/Untitled?node-id=0-1&t=pCcTakt8btZZAG4Q-0

apps/connect/components/profile/[username]/ProfileDetailsPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export async function ProfileDetailsPage({ user }: { user: Pick<LoggedInUser, 'farcasterUser' | 'id'> | null }) {
+  const farcasterDetails = user?.farcasterUser?.account as Required<
+    Pick<FarcasterBody, 'bio' | 'username' | 'displayName' | 'pfpUrl'>
+  >;
+
+  return (
+    <PageWrapper backToProfileHeader>
+      <Box p={2} mt={5}>
+        {user ? (
```
Comment (valentinludu) [inline apps/connect/components/profile/[username]/ProfileDetailsPage.tsx:20]: It will be better to have here `{user.farcasterUser ?... }`

since farcasterUser can be undefined

apps/connect/components/profile/components/ProfileDetailsSkeleton.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { ProjectItemSkeleton } from '@connect/components/projects/components/ProjectItemSkeleton';
+import Card from '@mui/material/Card';
+import Skeleton from '@mui/material/Skeleton';
+import Stack from '@mui/material/Stack';
+
+export function ProfileDetailsSkeleton() {
+  return (
+    <PageWrapper>
+      <Card sx={{ display: 'flex', flexDirection: 'column', gap: 2, p: 2 }}>
+        <Card sx={{ p: 2 }}>
```
valentinludu, line 11: In theory you can't have a card inside a card. This can be CardContent maybe or the first card can be just a Box.

apps/connect/lib/actions/fetchUserByFarcasterUsername.ts
valentinludu, line 1: This is not an action. It belongs to a lib/profile folder.

apps/connect/components/common/PageWrapper.tsx
valentinludu: Actually, you can leave it like this. I will resolve conflicts and delete the header in my PR, after you merge.

valentinludu:
If I'm in the /u/[name] page, the profile card should not have a link to the same page I am in.

valentinludu (approved):
Tested mobile/desktop. Looks good.

apps/connect/components/common/FarcasterCard.tsx
Devorein: Wow thank you. I will use it everywhere we route to an internal page

apps/connect/components/common/PageWrapper.tsx
Devorein: Sounds good

apps/connect/components/profile/components/ProfileDetailsSkeleton.tsx
Devorein: Sure thats produces better semantic html

apps/connect/components/profile/[username]/ProfileDetailsPage.tsx
Comment (Devorein) [reply apps/connect/components/profile/[username]/ProfileDetailsPage.tsx:20]: Hmm I can't think of any case where farcasterUser would be undefined. Either you login/signup with your farcaster account or for team members if they are not present in our db we create farcaster users and when they login we attach that farcaster user to the connected user. But I updated the code nonetheless

PR #4243, opened 2024-07-02
Store and show optimism project attestation

Devorein:


lib/optimism/storeOptimismProjectAttestations.ts
```diff
@@ -0,0 +1,164 @@
...
+        },
+        {
+          schemaId: optimismProjectAttestationSchemaId
+        }
+      ]
+    },
+    type: 'onchain'
+  });
+
+  const optimismProjectMetadataAttestationRecord = optimismAttestations
```
motechFR, line 92: Suggestion; we could also just have 2 separate graphql queries made, to avoid additional filtering

lib/optimism/storeOptimismProjectAttestations.ts
```diff
@@ -0,0 +1,164 @@
...
+                    return parseInt(teamMember).toString();
+                  }
+                  return null;
+                })
+                .filter(isTruthy)
+            : [])
+        ])
+      );
+
+      await prisma.optimismProjectAttestation.upsert({
```
motechFR, line 136: Since attestations are write once, readonly after, why not just do a findFirst and then create if not found?

It would be more accurate

lib/optimism/storeOptimismProjectAttestations.ts
```diff
@@ -0,0 +1,164 @@
...
+      (attestation) =>
+        attestation.schemaId === optimismProjectAttestationSchemaId &&
+        optimismProjectMetadataAttestationRecord[attestation.id]
+    )
+    .reduce((acc, attestation) => {
+      acc[attestation.id] = attestation as EASAttestationFromApi<OptimismProjectAttestation>;
+      return acc;
+    }, {} as Record<string, EASAttestationFromApi<OptimismProjectAttestation>>);
+
+  for (const optimismProjectAttestation of Object.values(optimismProjectAttestationRecord)) {
```
motechFR, line 111: It would be simpler to iterate directly through project metadata attestations, and fetch the project from EAS api if it's not available in list of attestations

components/common/form/fields/OptimismProjectSelector.tsx
```diff
@@ -170,18 +198,18 @@ export function OptimismProjectSelector({ value, disabled, ...props }: Props) {
...
             displayEmpty
-            value={selectedProject?.attestationUid ?? ''}
+            value={selectedProject?.attestationId ?? ''}
             disabled={disabled}
             onChange={(e) => {
               const projectId = e.target.value;
               const newProject =
                 projectId !== 'add-new-project'
-                  ? projects.find((project) => project.attestationUid === projectId)
+                  ? projects.find((project) => project.attestationId === projectId)
```
motechFR, line 207: any refs to project.attestationId will need to be projectRefUID

motechFR (approved):

lib/optimism/storeOptimismProjectAttestations.ts
Devorein: Since we will have a single record which keeps the metadata updated ideally we would use upsert

lib/optimism/storeOptimismProjectAttestations.ts
Devorein: That would require a bunch of EAS Api call right rather than batched requests. And since we are running it in cronjob it shouldn't be an issue

PR #4247, opened 2024-07-02
add a new token gate option for contract method

mattcasey:
Allow calling a custom contract method to check a wallet's balance. This enables one space to use their staking contract: https://sepolia.etherscan.io/address/0x55105f426126952ac6d9c2e7e72c7451318617d3#code

Form:
image

Preview:
image

### DIFF
--- a/lib/tokenGates/validateTokenGateCondition.ts
+++ b/lib/tokenGates/validateTokenGateCondition.ts
@@ -1,4 +1,4 @@
-import { formatUnits, getAddress, parseEther } from 'viem';
+import { formatUnits, getAddress, parseEther, parseAbi } from 'viem';
 
 import { getPoapsFromAddress } from 'lib/blockchain/poaps';
 import { getPublicClient } from 'lib/blockchain/publicClient';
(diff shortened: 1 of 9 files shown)


lib/tokenGates/validateTokenGateCondition.ts
```diff
@@ -97,6 +97,19 @@ export async function validateTokenGateCondition(
 
       return userAddress === ownerAddress;
     }
+    case condition.type === 'ContractMethod': {
+      const minimumQuantity = BigInt(condition.quantity || 1);
+      const result = await publicClient.readContract({
+        abi: parseAbi([`function ${condition.method}(address) view returns (uint256)`]),
+        address: contractAddress,
+        functionName: condition.method,
+        args: [walletAddress as `0x${string}`]
```
valentinludu, line 106: use userAddress since it is verified and typed. No need for assertion anymore.

valentinludu (approved):
Tested! All works well.

lib/tokenGates/validateTokenGateCondition.ts
mattcasey: it's actually typed as string

PR #4254, opened 2024-07-05
Connect app UI

valentinludu:
https://www.figma.com/design/uGO5TTywk0VE2yvHcbErdQ/Untitled?node-id=0-1&t=pCcTakt8btZZAG4Q-0


apps/connect/lib/actions/revalidatePath.ts
```diff
@@ -0,0 +1,8 @@
+'use server';
+
+import { actionClient } from '@connect/lib/actions/actionClient';
+import { revalidatePath } from 'next/cache';
+
+export const actionRevalidatePath = actionClient.metadata({ actionName: 'revalidatePath' }).action<void>(async () => {
```
valentinludu, line 6: FYI. This just revalidates the main layout in client components.

.eslintignore
```diff
@@ -2,6 +2,7 @@ scripts/
 dist/
 
 apps/connect-api/dist/
+apps/connect-api/src/routes.ts
```
mattcasey, line 5: Can we not do this? It feels really hacky to have to ignore a file in src, it's also being added twice. i think this particular file is auto-generated, but that's even more reason that i should pass linting.

In the worst case, add eslint ignore to the src file?

apps/connect-api/src/scripts/generateRoutes.ts
```diff
@@ -34,7 +34,8 @@ async function generateRoutesFile() {
     .map((_, index) => `rootRouter.use(router${index}.routes(), router${index}.allowedMethods());`)
     .join('\n');
 
-  const content = `/* eslint-disable import/no-extraneous-dependencies */
+  const content = `
+/* eslint-disable import/no-extraneous-dependencies */
```
mattcasey, line 38: i think the comment needs to be the first line. if you were getting an eslint error, that may be why

apps/connect/apiClient/apiClient.ts
```diff
@@ -15,6 +15,10 @@ class ConnectApiClient extends HttpClient {
     return this.POST('/api/session/login-with-farcaster', req);
   }
 
+  async logOut(): Promise<void> {
```
mattcasey, line 18: all lowercase to be consistent please: `logout()`

apps/connect/app/layout.tsx
```diff
@@ -1,27 +1,29 @@
...
 import '@connect/theme/cssVariables.scss';
 
 export const metadata: Metadata = {
-  title: 'Charm Connect - onchain builder network',
+  title: 'Charm Connect - Onchain builder network',
   description: 'Onchain network for connecting web3 developers, projects, organizations'
 };
 
-export default function RootLayout({
+export default async function RootLayout({
```
mattcasey, line 15: why do you have to use async here?

apps/connect/components/common/FarcasterCard.tsx
```diff
@@ -72,14 +70,14 @@ export function FarcasterCard(props: {
...
-      <Card>
+      <Card sx={{ border: 'none' }}>
         <FarcasterCardContent {...props} />
       </Card>
     );
   }
 
   return (
-    <Card>
+    <Card sx={{ border: 'none' }}>
```
mattcasey, line 80: This is fine, but do we even need a card component if there's no border?

apps/connect/components/layout/Footer.tsx
```diff
@@ -0,0 +1,30 @@
+import { Box, Link, Typography } from '@mui/material';
+import type { BoxProps } from '@mui/material/Box';
+import Image from 'next/image';
+
+export function Footer(props: BoxProps) {
```
mattcasey, line 5: Please remove props if we don't use it

apps/connect/components/layout/Header.tsx
```diff
@@ -0,0 +1,11 @@
+import 'server-only';
+
+import { getCurrentUser } from '@connect/lib/actions/getCurrentUser';
+
+import { NavBar } from './NavBar';
+
+export async function Header() {
+  const user = await getCurrentUser();
```
mattcasey, line 8: This is fine for now, but user should probably be passed into context or something. how do we avoid re-querying the user every time we need their information?

apps/connect/components/layout/NavBar.tsx
```diff
@@ -1,159 +1,87 @@
...
+import { Box, Container, IconButton, Menu, MenuItem, Toolbar, AppBar } from '@mui/material';
+import Image from 'next/image';
+import Link from 'next/link';
+import { usePathname, useRouter } from 'next/navigation';
 import type { MouseEvent } from 'react';
 import { useState } from 'react';
 
-const pages = ['Projects'];
-const settings = ['Profile', 'Account', 'Dashboard', 'Logout'];
+import type { LoggedInUser } from 'models/User';
```
mattcasey, line 14: Let's stop using LoggedInUser for this app

apps/connect/components/welcome/WelcomePage.tsx
```diff
@@ -11,23 +10,20 @@ import { FarcasterCard } from '../common/FarcasterCard';
 import { ExtraDetails } from './ExtraDetails';
 
 export function WelcomePage({ user }: { user: LoggedInUser }) {
-  const farcasterDetails = user.farcasterUser?.account as
-    | Required<Pick<FarcasterBody, 'bio' | 'username' | 'displayName' | 'pfpUrl'>>
-    | undefined;
+  const farcasterDetails = user.farcasterUser?.account as Required<FarcasterBody>;
```
mattcasey, line 13: this is being cast as if it always exists but it's reading from an optional field

apps/connect/lib/actions/revalidatePath.ts
mattcasey: would be worth adding a comment above on why you would use it

mattcasey (changes requested):
- clean up eslint or can we chat about it offline? i would like to unerstand the issue more
- everything else can be taken as a suggestion

apps/connect/lib/actions/revalidatePath.ts
valentinludu: Done

apps/connect-api/src/scripts/generateRoutes.ts
valentinludu: Done

apps/connect/apiClient/apiClient.ts
valentinludu: done

apps/connect/components/layout/Footer.tsx
valentinludu: I used it but no I don't need it anymore so I removed the props.

apps/connect/components/welcome/WelcomePage.tsx
valentinludu: Added the `| undefined` in all the places to ensure we take care of undefined situations

apps/connect/components/layout/NavBar.tsx
valentinludu: I created another LoggedInUser just for this app

apps/connect/components/common/FarcasterCard.tsx
valentinludu: I saw the border in a few places. I need to circle back with Chris to see where it is needed and where not.

apps/connect/components/layout/Header.tsx
valentinludu: I can think of a method to cache the response for server components. Let's discuss more in our daily.

.eslintignore
valentinludu: The error is thrown only when we commit a new route.

I deleted the apps/connect-api/.eslintignore file.

I deleted this line and it will probably resurface on the next route we will create and commit.

PR #4277, opened 2024-07-11
Connect grants page

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/grants-page-on-connect-app-2081675867556796)


apps/connect/components/grants/GrantsDetailsPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+  applyLink?: string;
+  path: string;
+};
+
+export function GrantsDetailsPage({ grants, currentTab }: { grants: Grant[]; currentTab: 'new' | 'upcoming' }) {
+  return (
+    <PageWrapper>
+      <Stack direction='row' gap={1} justifyContent='center'>
+        <Link href='/grants/upcoming'>
+          <Button size='small' variant={currentTab === 'upcoming' ? 'contained' : 'outlined'}>
```
valentinludu, line 24: You can have only one component here, like `<Button LinkComponent={Link}`

A good example can be found here:
apps/connect/components/projects/ProjectShareButton.tsx

apps/connect/app/grants/new/page.tsx
valentinludu, line 1: I am thinking in the future we could have multiple types of grants lists.

Since /new and /upcoming pages are basically the same thing, you could have only one page like: **/grants?type=new**

And inside page you could use **searchParams** to get the sorting.
Also you can set a default value (new) if no params are present so the page would render the default value server side.

apps/connect/components/grants/GrantsDetailsPage.tsx
valentinludu, line 1: use import server-only to be sure we render it on the server. and only children components will be client components

apps/connect/app/g/[path]/page.tsx
Comment (valentinludu) [inline apps/connect/app/g/[path]/page.tsx:1]: Would be nice to have the same experience like the other pages where we have a loading skeleton

apps/connect/components/grants/GrantsDetailsPage.tsx
Devorein: updated

apps/connect/components/grants/GrantsDetailsPage.tsx
Devorein: nice one. Updated.

apps/connect/app/g/[path]/page.tsx
Comment (Devorein) [reply apps/connect/app/g/[path]/page.tsx:1]: Updated

apps/connect/app/grants/new/page.tsx
Devorein: Updated

valentinludu (approved):

apps/connect/components/projects/[id]/ProjectDetailsPage.tsx
```diff
@@ -18,7 +18,9 @@ export async function ProjectDetailsPage({ project }: { project: ConnectProjectD
   );
 
   return (
-    <PageWrapper header={<ProjectHeader name={project.name} avatar={project.avatar} coverImage={project.coverImage} />}>
+    <PageWrapper
+      header={<PageCoverHeader name={project.name} avatar={project.avatar} coverImage={project.coverImage} />}
+    >
```
Comment (mattcasey) [inline apps/connect/components/projects/[id]/ProjectDetailsPage.tsx:23]: Was this ">" intentional in the first place?

mattcasey:
FYI you tagged this as "deploy" and it just deployed. the main webapp.  Remember to use 'deploy-connect' :)

mattcasey (approved):

PR #4294, opened 2024-07-15
OP mixpanel events

Devorein:


motechFR:
Fix:
- Trigger "create proposal" event when a template is selected from proposal templates dropdown

motechFR (approved):
Approved with caveat we need to fix these 2 events

PR #4295, opened 2024-07-15
PWA notifications for Connect app

valentinludu:
How it works:
1. The user needs to install the pwa app and login
    - This is possible with the help of the manifest file (manifest.json)
2. The user needs to accept notifications from the popup screen
     - Each time a user didn't set his permissions we ask permission to send him notifications (layout).
     - Usually this happens only once
2. We register a subscription for a certain user from a certain device in our db 
    - Once the user accepts the notifications we are saving this subscription permission in our db with the help of a server action (actionSaveSubscription)
4. Once a day at 10 AM a cron job runs and sends a notification to users 
    - Made possible with the web-push package (cron task)
6. The service worker file receives the notification from the backend and serves that notification to the user's device
    - sw.js file handles it


apps/connect/app/api/push-notifications/route.ts
```diff
@@ -0,0 +1,42 @@
+import { prisma } from '@charmverse/core/prisma-client';
```
motechFR, line 1: We should delete this file, or make it a dev only route

apps/connect/app/layout.tsx
```diff
@@ -1,28 +1,68 @@
...
   themeColor: '#fff'
 };
 
-export default function RootLayout({
+export default async function RootLayout({
   children
 }: Readonly<{
   children: ReactNode;
 }>) {
+  const session = await getIronSession<SessionData>(cookies(), getIronOptions());
```
motechFR, line 61: small suggestion, could we wrap this in a util method called getSession() to make it easier to consume?

apps/connect/app/api/push-notifications/route.ts
valentinludu: Done

apps/connect/app/layout.tsx
valentinludu: Done

motechFR (approved):

apps/connect/app/layout.tsx
valentinludu: I finally made it to work. Session should have the same cookieName as the one from connect-api

PR #4298, opened 2024-07-16
Bug/filters follow up

motechFR:


scripts/prompts/traverseAndTokenise.ts
```diff
@@ -0,0 +1,133 @@
+import { exec } from 'node:child_process';
```
mattcasey, line 1: is this part of the filters bug?

lib/databases/getRelatedBlocks.ts
```diff
@@ -99,5 +99,5 @@ export async function getRelatedBlocks(blockId: string): Promise<{ blocks: Block
     // remove orphan and deleted blocks
     .filter((block) => !block.deletedAt && (!!block.pageId || block.type === 'view' || block.type === 'board'));
 
-  return { blocks: validBlocks, source };
+  return { blocks: [...validBlocks] as any, source };
```
mattcasey, line 102: please don't add "any" type here, there should be no need to create a new array either

lib/databases/__tests__/generateCsv.spec.ts
```diff
@@ -205,3 +205,7 @@ describe('loadAndGenerateCsv()', () => {
     expect(firstRow).toContain('first guy,second guy');
   });
 });
+
+describe('generateTableArray()', () => {
```
mattcasey, line 209: this is not actually testing anything

lib/databases/subtree.ts
mattcasey, line 1: files should be named after the exported method - rename to getSubtree.ts?

PR #4300, opened 2024-07-17
Optimism project form field

Devorein:


@connect-shared/lib/projects/createOptimismProject.ts
```diff
@@ -1,14 +1,23 @@
...
+import type { ProjectSource } from '@charmverse/core/prisma-client';
 import { prisma } from '@charmverse/core/prisma-client';
 import { getFarcasterUsers } from '@root/lib/farcaster/getFarcasterUsers';
 import { generatePagePathFromPathAndTitle } from '@root/lib/pages/utils';
 import { stringToValidPath, uid } from '@root/lib/utils/strings';
 import { isTruthy } from '@root/lib/utils/types';
 import { v4 } from 'uuid';
 
-import type { FormValues } from './form';
+import type { FormValues } from '../../../apps/connect/lib/projects/form';
```
motechFR, line 10: we should only import from lib towards apps, never in the other direction.

@connect-shared/lib/projects/editOptimismProject.ts
```diff
@@ -5,9 +5,9 @@ import { uid } from '@root/lib/utils/strings';
 import { isTruthy } from '@root/lib/utils/types';
 import { v4 } from 'uuid';
 
-import type { FormValues } from './form';
+import type { FormValues } from '../../../apps/sunnyawards/lib/projects/form';
```
motechFR, line 8: Same as above, please deduplicate this type and put it inside the lib

apps/connect/components/projects/components/SearchFarcasterUser.tsx
```diff
@@ -8,6 +8,7 @@ import debounce from 'lodash/debounce';
 import { useEffect, useMemo, useState } from 'react';
 
 type FarcasterProfile = Pick<StatusAPIResponse, 'fid' | 'pfpUrl' | 'bio' | 'displayName' | 'username'>;
+const connectApiClient = new ConnectApiClient();
```
motechFR, line 11: Is there a reason for moving this outside?

The reason we ended up initialising it inside the calling code, is because sometimes the environment variable was not initialised

apps/connect/lib/actions/createProject.ts
```diff
@@ -15,7 +15,7 @@ export const actionCreateProject = authActionClient
   .action(async ({ parsedInput, ctx }) => {
     const input = parsedInput;
     const currentUserId = ctx.session.user!.id;
-    const newProject = await createConnectProject({
+    const newProject = await createOptimismProject({
```
motechFR, line 18: It's better to pass 'connect' as the source explicitly

We should also make the parameter required to avoid forgetting it

apps/connect/lib/projects/__tests__/createConnectProject.spec.ts
```diff
@@ -2,7 +2,7 @@ import { prisma } from '@charmverse/core/prisma-client';
 import { testUtilsRandom, testUtilsUser } from '@charmverse/core/test';
 import fetchMock from 'fetch-mock-jest';
 
-import { createConnectProject } from '../createConnectProject';
+import { createConnectProject } from '../../../../../@connect-shared/lib/projects/createOptimismProject';
```
motechFR, line 5: Please fix import path

motechFR (approved):

@connect-shared/lib/projects/createOptimismProject.ts
Devorein: updated

@connect-shared/lib/projects/editOptimismProject.ts
Devorein: updated

apps/connect/components/projects/components/SearchFarcasterUser.tsx
Devorein: ahh thats a good point. I thought its unnecessary to instantiate it multiple times. I will move it back for now.

apps/connect/lib/actions/createProject.ts
Devorein: fair point. Update.

PR #4308, opened 2024-07-22
Next safe actions update

valentinludu:
### WHAT

Next safe action bump to latest version

### WHY

The types were not handled correctly. 
- Actions throwing errors. I removed all the `@ts-ignore` and works perfect!
- If an action does not have an input you don't need to add `{}` to it anymore. This was a ts 5.5 issue and it's fixed.


motechFR (approved):
Nice, those ts-ignore were so annoying!

PR #4338, opened 2024-07-29
Connect navigation + Grants page id fix

mattcasey:
I did these together unintentionally, but it's easier to merge as one

- moves PageWrapper into a shared component
- adds env id for grants page so it can be customized in staging (using AWS console)


apps/connect/lib/grants/getGrants.ts
mattcasey, line 1: most of the grants-related changes are in this file

PR #4340, opened 2024-07-29
hide summon token gate

mattcasey:
[### WHAT

### WHY

](https://discord.com/channels/894960387743698944/1266043391347527782/1266043391347527782)


valentinludu (commented):
Don't forget about the cron job that comes from `lib/summon/syncSummonSpaceRoles`

valentinludu (approved):
looks good!

mattcasey:
Thanks for the review. I intentionally left most of it intact in case we change our mind. We don't have any way to know if it's being used right now AFAICT

PR #4344, opened 2024-07-30
Make animations common to both apps

valentinludu:


mattcasey (approved):
thanks for moving to connect-shared

PR #4361, opened 2024-08-02
Setup docker

motechFR:

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -482,7 +482,7 @@ jobs:
     name: Upload Docker image and assets
     runs-on: ubuntu-latest
     # run whether previous jobs were successful or skipped
-    if: github.ref == 'refs/heads/main' && !(failure() || cancelled())
+    if: github.ref != 'refs/heads/main' && !(failure() || cancelled())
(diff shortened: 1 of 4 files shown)


scripts/migrations/2024_08_02_migrateProposalPublicPermissions.ts
```diff
@@ -1,3 +1,4 @@
+import { ProposalWorkflowTyped } from "@charmverse/core/proposals";
```
mattcasey, line 1: you should keep seprate changes in separate PRs so that we can easily revert

.github/workflows/test_and_deploy.yml
```diff
@@ -482,7 +482,7 @@ jobs:
     name: Upload Docker image and assets
     runs-on: ubuntu-latest
     # run whether previous jobs were successful or skipped
-    if: github.ref == 'refs/heads/main' && !(failure() || cancelled())
+    if: github.ref != 'refs/heads/main' && !(failure() || cancelled())
```
mattcasey, line 485: what is this change?

mattcasey: remmeber to remove this

PR #4369, opened 2024-08-06
Pagination for connect grants page

Devorein:


mattcasey (approved):

mattcasey (changes requested):
As we discussed, please move the api code to the Next API. There's just a lot less boilerplate, and at the moment I do not have a reason to keep the Connect API. So our default should be to use the Next API, unelss a good reason to use the Connect API comes up

It doesn't look like you are actually using 'cursor' yet inside getGrants?

PR #4377, opened 2024-08-06
Remove unused fields from projects

mattcasey:
We will run a migration script before deploying this: https://github.com/charmverse/app.charmverse.io/pull/4379


Devorein (approved):

Devorein:
Approved but faced a few bugs
1. Can't create project due to this error `Exactly one team lead is required`. It seems like during project creation the projectMember that we pass is not the teamLead
2. There are also issues rendering the `TextInputField` as the socialUrls is an array now but its not being handled correctly
image

Devorein (changes requested):
Please check comments

PR #4401, opened 2024-08-10
Product updates farcaster composer action

mattcasey:


apps/farcaster/theme/cssVariables.scss
```diff
@@ -0,0 +1,19 @@
+* {
```
mattcasey, line 1: there are no CSS variables in this file

apps/farcaster/theme/index.d.ts
```diff
@@ -0,0 +1,20 @@
+import type {
```
mattcasey, line 1: i don't think this file's necessary for the farcaster frame?

apps/farcaster/theme/colors.ts
```diff
@@ -0,0 +1,38 @@
+import { darken, lighten } from '@mui/material/styles';
```
mattcasey, line 1: these should only be colors we need for farcaster frames

apps/farcaster/lib/product-updates/action.ts
mattcasey, line 1: we never use hyphens in folder names, should be productUpdates

apps/farcaster/lib/product-updates/createFrame.ts
mattcasey, line 1: folders in lib should be camelcase

apps/farcaster/lib/product-updates/action.ts
```diff
@@ -0,0 +1,15 @@
+'use server';
```
mattcasey, line 1: filename should be "createProductUpdatesFrameAction"

apps/farcaster/components/common/Header.tsx
```diff
@@ -0,0 +1,8 @@
+'use client';
+
+import { useDatadogLogger } from '@root/hooks/useDatadogLogger';
+
+export function Header() {
+  useDatadogLogger({ service: 'farcaster-browser' });
```
mattcasey, line 6: we shouldn't call this "Header", since it's not actually a UI element - I know we use Header for datadoglogger in the other apps, but it's not a good place to put it. We need to move it out so it's more clear where these things go.  How bout rename to something like "GlobalComponent". I don't have any preference but can't come up with anything else :)

apps/farcaster/components/product-updates/ProductUpdatesComposerAction.tsx
```diff
@@ -0,0 +1,136 @@
...
+      createdAtLocal: new Date().toDateString()
+    },
+    resolver: yupResolver(schema),
+    mode: 'onChange'
+  });
+
+  const { execute, isExecuting } = useAction(createProductUpdatesFrameAction, {
+    onSuccess: (data) => {
+      reset();
+      window.parent.postMessage(
```
mattcasey, line 42: suggestion: move anything that talks directly to their api into a separate file/folder. For example it would just export a method `createCast`.

apps/farcaster/components/product-updates/ProductUpdatesComposerAction.tsx
```diff
@@ -0,0 +1,136 @@
...
+              text: `Product updates for ${new Date().toLocaleDateString()}`,
+              embeds: data.data ? [data.data.image] : []
+            }
+          }
+        },
+        '*'
+      );
+    },
+    onError: (err) => {
+      log.error(err.error.serverError?.message || 'Something went wrong', err.error.serverError);
```
mattcasey, line 56: we need to display errors to the user. probably needs to be inline in the form?

apps/farcaster/components/product-updates/ProductUpdatesComposerAction.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+        <Stack alignItems='flex-end'>
+          <Button
+            type='submit'
+            size='large'
+            sx={{
+              width: 'fit-content'
+            }}
+            variant='contained'
+            disabled={isExecuting || !isValid}
```
mattcasey, line 128: From now on, all buttons are always clickable. do not disable if it's invalid

apps/farcaster/theme/index.d.ts
Devorein: Good point. Removed

apps/farcaster/lib/product-updates/action.ts
Devorein: I think to keep things consistent we should just rename the file to be equal to the file name and keep each action separate. I don't think we mix actions anywhere

apps/farcaster/components/product-updates/ProductUpdatesComposerAction.tsx
Devorein: Updated, named it postCreateCastMessage as createCast might be mistaken for a database action

apps/farcaster/components/product-updates/ProductUpdatesComposerAction.tsx
Devorein: Updated

PR #4406, opened 2024-08-12
Move logic from connect-api to connect

valentinludu:


@connect-shared/components/common/MultiTextInputField.tsx
```diff
@@ -45,6 +45,7 @@ export function MultiTextInputField<T extends FieldValues>({
               placeholder={placeholder}
               error={!!fieldState.error}
               {..._field}
+              value={_field.value || ''}
```
mattcasey, line 48: I think u can remove this - i believe i fixed it by providing defaultValue

@connect-shared/lib/session/logoutAction.ts
```diff
@@ -4,11 +4,11 @@ import { log } from '@charmverse/core/log';
 
 import { authActionClient } from '../actions/actionClient';
 
-export const logoutAction = authActionClient.metadata({ actionName: 'login' }).action(async ({ ctx }) => {
+export const logoutAction = authActionClient.metadata({ actionName: 'logout' }).action(async ({ ctx }) => {
```
mattcasey, line 7: thanks for fixing!

apps/connect/app/api/aws/upload-token/route.ts
```diff
@@ -0,0 +1,74 @@
+import { GetFederationTokenCommand, STSClient } from '@aws-sdk/client-sts';
```
mattcasey, line 1: Please dont duplicate code - we have this already written inside sunnyawards, so it should be put into connect-shared

apps/connect/components/projects/components/ProjectForm.tsx
```diff
@@ -127,6 +128,7 @@ export function ProjectForm({
                   aria-labelledby='project-twitter'
                   error={!!fieldState.error}
                   {...field}
+                  value={field.value || ''}
```
mattcasey, line 131: this looks like a hack, i am not sure what problem you are solving. maybe we need to be providing default values in the useForm or provide a defaultValue instead?

apps/sunnyawards/components/common/Header/Header.tsx
```diff
@@ -28,7 +28,7 @@ export function Header({ user }: { user: LoggedInUser | null }) {
 
   const { execute: logoutUser } = useAction(logoutAction, {
     onSuccess: async () => {
-      await revalidatePathAction();
+      revalidatePathAction();
```
mattcasey, line 31: why should we not await? this function returns a promise?

mattcasey (changes requested):
Sunny Awards and Connect should share code that is the same. When I created Connect, we said that it was not worth me going back and de-duping at the time, and agreed to do it as-we-go. Please at least remove the new duplicate code for uploading images. And ideally, now that we are only using server actions in both it would be easy to re-use some other components like SearchFarcasterUser.

@connect-shared/components/common/MultiTextInputField.tsx
valentinludu: defaultValue makes a component to be uncontrolled and then, when the state changes it becomes controlled and the browser throws an error. Usually React recommends to create a component controlled or uncontrolled.

Screenshot 2024-08-13 at 10 11 20

valentinludu: Another way would be to set default values to useForm hook.

apps/connect/app/api/aws/upload-token/route.ts
valentinludu: Agree. Moved it in connect-shared folder!

apps/connect/components/projects/components/ProjectForm.tsx
valentinludu: I commented above.

apps/sunnyawards/components/common/Header/Header.tsx
valentinludu: It makes the change of the page faster. We can discuss more in our call.

mattcasey (approved):

PR #4415, opened 2024-08-13
Improve UI for text inputs

mattcasey:
- use grey instead of red for delete action - red is 'scary' we should try and only use it for errors
- hide Add/Delete on disabled forms. These are confusing as it makes it look like u can edit the fields
- add tooltips to Delete so user knows what it does
- add the smart link icon to multi-input text fields

image


components/common/form/fields/TextInputField.tsx
```diff
@@ -51,17 +42,41 @@ export const TextInputField = forwardRef<HTMLDivElement, Props>(
...
+          <IconButton color='secondary' href={props.value as string} target='_blank' size='small' sx={{ p: 0 }}>
+            <LinkIcon />
+          </IconButton>
+        )
+      }
+    : undefined;
+
+  return (
+    <TextField
+      defaultValue=''
```
valentinludu, line 73: I know this transforms the component into an uncontrolled one.
If you want to keep the component controlled and avoid the error `A component is changing an uncontrolled input to be controlled`, you can delete `defaultValue` and add `value={props.value || ''}`

Same comment for all the places where this is used.

valentinludu (approved):
Left just one comment.

PR #4417, opened 2024-08-13
Fix markdown export

motechFR:
### WHAT

Fixed a bug with mention suggest marks


Devorein (approved):

lib/prosemirror/markdown/generateMarkdown.ts
```diff
@@ -19,10 +22,16 @@ export async function generateMarkdown({
   content: any;
   generatorOptions?: CharmMarkdownGeneratorOptions;
 }): Promise<string> {
+  prettyPrint({ title, content, generatorOptions });
```
Devorein, line 25: small artifact

PR #4429, opened 2024-08-14
track more Mixpanel events for sunny awards

mattcasey:


@connect-shared/lib/session/logoutAction.ts
```diff
@@ -6,7 +6,9 @@ import { authActionClient } from '../actions/actionClient';
 
 export const logoutAction = authActionClient.metadata({ actionName: 'logout' }).action(async ({ ctx }) => {
   const userId = ctx.session.user?.id;
-  ctx.session.destroy();
+  // ctx.session.destroy();
```
valentinludu, line 9: I presume this was a test.

It should be session.destroy().

apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx
```diff
@@ -18,53 +19,28 @@ const Icon = (
...
+        gap: 1,
+        backgroundColor: farcasterPurple,
+        '&:hover': {
+          backgroundColor: farcasterPurple
+        }
+      }}
+      onMouseDown={() => trackEvent('click_share_on_warpcast')}
     >
-      <Typography component='span' sx={textStyles}>
+      <Typography component='span' sx={{ fontSize: '22px' }}>
```
Comment (valentinludu) [inline apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx:43]: You can use a variant here like subtitle or body1 or h6 depending on which one has a fontSize similar to the one here.

apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx
```diff
@@ -18,53 +19,28 @@ const Icon = (
...
       href={href}
-      onClick={onClick}
       rel='noopener noreferrer'
       target='_blank'
       startIcon={Icon}
-      sx={style}
-      disabled={disabled}
+      sx={{
+        gap: 1,
+        backgroundColor: farcasterPurple,
```
Comment (valentinludu) [inline apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx:36]: I added the color 'farcaster.main' to be used anywhere and it will be taken from the theme file.

valentinludu (changes requested):
Logout should destroy() cookies.

@connect-shared/lib/session/logoutAction.ts
mattcasey: tahnk u!!!

apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx
Comment (mattcasey) [reply apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx:43]: +1 it would be better, i was trying to just inline the styles from above... we're about to add a Twiter share button so i'd like to do the cleanup afterward

apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx
Comment (mattcasey) [reply apps/sunnyawards/components/projects/[id]/share/components/WarpcastButton.tsx:36]: i wanted to do that but was worried it would take too much time, thanks!

valentinludu (approved):

PR #4430, opened 2024-08-14
Bugs/sunny patches

motechFR:
### WHAT

Change viem ENS RPC url to avoid throwing an error

Add error states in the form

Fix bug with team members where form was incorrectly updated on deleting user, and editing form was blocked


@connect-shared/lib/projects/createOptimismProject.ts
```diff
@@ -37,9 +37,7 @@ export async function createOptimismProject({
 
   if (hasEnsName) {
     const resolvedAddress = await resolveEnsAddress(input.mintingWalletAddress as string);
```
mattcasey, line 39: please delete/replace lib/blockchain resolveEnsAddress with our own: resolveENSName from lib/blockchain/getENSName. u can split out the file if you want, but i want to use the original one  everywhere

connectors/chains.ts
```diff
@@ -113,7 +113,7 @@ export const RPC: Record<string, IChainDetails> = {
   ETHEREUM: {
     ...EVM_DEFAULT,
     chainId: mainnet.id,
-    viem: mainnet,
+    viem: { ...mainnet, rpcUrls: { default: { http: ['https://rpc.ankr.com/eth'] } } },
```
mattcasey, line 116: this will not change what wagmi/core does

mattcasey (changes requested):
see comment

apps/sunnyawards/components/projects/edit/EditProjectPage.tsx
```diff
@@ -26,13 +27,22 @@ export function EditProjectPage({ user, project }: { user: LoggedInUser; project
...
+    },
     onSuccess: () => {
       router.push(`/p/${project.path}`);
     },
     onError(err) {
-      log.error(err.error.serverError?.message || 'Something went wrong', err.error.serverError);
+      const errorMessage = err.error.validationErrors
+        ? concatenateStringValues(err.error.validationErrors.fieldErrors)
+        : err.error.serverError?.message || 'Something went wrong';
+      log.error(errorMessage || 'Something went wrong', err.error.serverError);
```
mattcasey, line 43: 1. we already do logging on the backend (and the info is usually better), so this log is unecessary
2. Our log util expects an object in the second argument, and has special handling for  the "error" prop, so you should pass it like: `log.error(message, { error: })`

apps/sunnyawards/components/projects/edit/EditProjectPage.tsx
```diff
@@ -120,7 +130,7 @@ export function EditProjectPage({ user, project }: { user: LoggedInUser; project
             execute({
               ...input,
               id: project.id,
-              projectMembers: input.projectMembers.slice(1)
+              projectMembers: input.projectMembers.filter((m) => m.farcasterId !== user.farcasterUser?.fid)
```
mattcasey, line 133: the previous code was pulling out the team lead - are you sure we don't want to do that anymore?

apps/sunnyawards/components/projects/new/CreateProjectPage.tsx
```diff
@@ -26,13 +25,21 @@ export function CreateProjectPage({ user }: { user: LoggedInUser }) {
...
+    onExecute: () => {
+      setError(null);
+    },
     onSuccess: (data) => {
       router.push(`/p/${data.data?.projectPath as string}/share`);
     },
     onError(err) {
-      log.error(err.error.serverError?.message || 'Something went wrong', err.error.serverError);
+      const errorMessage = err.error.validationErrors
+        ? concatenateStringValues(err.error.validationErrors.fieldErrors)
```
mattcasey, line 39: if we have multiple errors, they should be split out in paragraph tags. if u just use a comma-separated list, that is hard to read

mattcasey (approved):

apps/sunnyawards/components/projects/edit/EditProjectPage.tsx
```diff
@@ -5,25 +5,39 @@ import { PageWrapper } from '@connect-shared/components/common/PageWrapper';
...
 import { editProjectAction } from 'lib/projects/editProjectAction';
-import { schema } from 'lib/projects/form';
 import type { FormValues, ProjectCategory } from 'lib/projects/form';
+import { schema } from 'lib/projects/form';
 
 import { ProjectForm } from '../components/ProjectForm';
 
 export function EditProjectPage({ user, project }: { user: LoggedInUser; project: ConnectProjectDetails }) {
   const router = useRouter();
+  const [errors, setErrors] = useState<string[] | null>(null);
```
valentinludu, line 24: Instead of creating a new state, you already have one called result.
And use that result in a constant if you want to transform it.

```
  const { execute, isExecuting, result } = useAction(editProjectAction, {
    onSuccess: () => {
      router.push(`/p/${project.path}`);
    },
    onError(err) {
      log.error(err.error.serverError?.message || 'Something went wrong', err.error.serverError);
    }
  });
  const formError = result.validationErrors?.formErrors;
```

apps/sunnyawards/components/projects/new/CreateProjectPage.tsx
```diff
@@ -17,16 +18,26 @@ import { ProjectForm } from '../components/ProjectForm';
...
+      setErrors(null);
       const projectPath = data.data?.projectPath;
       if (projectPath) {
         router.push(`/p/${projectPath}/share`);
       }
     },
     onError(err) {
-      log.error(err.error.serverError?.message || 'Something went wrong', err.error.serverError);
+      const hasValidationErrors = err.error.validationErrors?.fieldErrors;
+      const errorMessage = hasValidationErrors
```
valentinludu, line 36: Same comment as above

valentinludu (approved):
Left only 1 comment for handling errors

apps/sunnyawards/components/projects/edit/EditProjectPage.tsx
motechFR: Awesome thanks, I will cleanup in next PR

PR #4431, opened 2024-08-15
add share to twitter button

mattcasey:
image


@connect-shared/lib/projects/findProject.ts
mattcasey, line 1: i renamed this file becaus "fetch***" sounds like we are using fetch, and we don't actually use 'fetch' as a verb in many places. would also consider 'getProject'

PR #4439, opened 2024-08-15
Remove 2 contract fields from OP projects

mattcasey:
Chris said OP requested we remove two fields: deployer address and transaction hash from proejct fields. I assume this means we don't have to send to Agora either ,so i removed them everywhere

i have not tested w/agora yet


motechFR:
I just ran a test against the Agora API. This is still required currently for submitting metadata

Here's the error

{"issues":[{"code":"invalid_type","expected":"string","received":"undefined","path":["contracts",0,"deploymentTxHash"],"message":"Required"},{"code":"invalid_type","expected":"string","received":"undefined","path":["contracts",0,"deployerAddress"],"message":"Required"}],"name":"ZodError"}

motechFR (approved):

PR #4451, opened 2024-08-18
Feat/sunny import project

motechFR:
### WHAT

Contains remaining changes for importing OP projects


apps/sunnyawards/components/projects/new/CreateProjectPage.tsx
```diff
@@ -101,9 +96,18 @@ export function CreateProjectPage({
       setValue('primaryContractAddress', metadata.contracts[0].address as Address);
     }
 
+    setValue('projectMembers', project.projectMembers);
```
mattcasey, line 99: you're setting projectMembers twice?

mattcasey (changes requested):
Ideally, we never wait for an API response to load the page component (getFarcasterUsers is being called to get the op projects). These can take a long time or fail. Maybe for now to keep things simple and get the bugs out, just make sure we catch any errors?

apps/sunnyawards/components/projects/new/CreateProjectPage.tsx
```diff
@@ -101,9 +96,18 @@ export function CreateProjectPage({
...
+
     setValue(
       'projectMembers',
-      (metadata.team ?? []).map((farcasterId) => ({ farcasterId: parseInt(farcasterId.toString()), name: 'demo' }))
+      project.projectMembers.map(
```
mattcasey, line 103: using `reset` would make more sense here than setting individual values. No one would expect for any previous fields to carry over if they are picking a new project from the dropdown. It also might solve the issue you're having w/websites, but not sure

apps/sunnyawards/lib/projects/fetchUnimportedOptimismProjectsAction.ts
mattcasey, line 1: we shouldn't use "fetch" verb in our function names - there's no need to add a new verb everywhere, "get*" is fine

apps/sunnyawards/lib/projects/fetchUnimportedOptimismProjectsAction.ts
```diff
@@ -25,5 +32,75 @@ export const fetchUnimportedOptimismProjectsAction = authActionClient
...
+          if (!farcasterUserMap[fid]) {
+            fidsWithoutProfile.push(fid);
+          }
+        }
+
+        return fidsWithoutProfile;
+      })
+    );
+
+    const farcasterProfiles = unresolvedUsers.length ? await getFarcasterUsers({ fids: unresolvedUsers }) : [];
```
mattcasey, line 71: i realize this is like a fallback? if we just remove it, does that remove dependency on any API?

@connect-shared/lib/projects/findProject.ts
```diff
@@ -97,6 +101,8 @@ export async function findProject({ id, path }: { id?: string; path?: string }):
         ...member.user,
         userId: member.userId,
         teamLead: member.teamLead,
+        name: farcasterUser?.displayName as string,
```
mattcasey, line 104: this type casting is dangerous

apps/sunnyawards/components/projects/components/AddProjectMembersForm.tsx
```diff
@@ -28,11 +26,13 @@ export function AddProjectMembersForm({
     name: 'projectMembers' as FieldArrayPath<FormValues>,
     control
   });
-  const [selectedFarcasterProfiles, setSelectedFarcasterProfiles] =
-    useState<FarcasterProfile[]>(initialFarcasterProfiles);
 
   const farcasterDetails = user.farcasterUser?.account as Required<FarcasterBody> | undefined;
 
+  const selectedFarcasterProfiles = (control._formValues.projectMembers as ConnectProjectMember[]).filter(
```
mattcasey, line 32: please add a comment why you need to use control._formValues? I couldn't find any documentation about it, although it exists in the types. It seems like an internal field that is not intended to be used

lib/credentials/__tests__/mapProjectToGitcoin.spec.ts
```diff
@@ -10,10 +10,10 @@ describe('mapProjectToGitcoin', () => {
     const input: ConnectProjectDetails = {
       id: '1',
       name: 'Project X',
+      sunnyAwardsNumber: null,
```
mattcasey, line 13: the default for this column is a number

mattcasey (approved):

PR #4454, opened 2024-08-19
Action form hook

valentinludu:
combine useAction with useForm hook into a new hook.


apps/connect/components/projects/hooks/useCreateProject.ts
```diff
@@ -0,0 +1,72 @@
...
+      for (const { name, type, message } of fieldErrorValues) {
+        setError(name as any, { type, message });
+      }
+
+      for (const formError of formErrors) {
+        setError('root.serverError', { type: 'manual', message: formError });
+      }
+
+      if (serverError?.message) {
+        setError('root.serverError', { type: serverError?.code?.toString(), message: serverError.message });
```
mattcasey, line 58: this is not a good design because we have to watch 'errors' from the top level of the form. this causes unecessary re-rending of all the form inputs. I think server error should just be exposed outside of the form

mattcasey: Also we could just have a react component that can render any server error, instead of doing it in some react-hook-form-middleware, which is not as reusable

PR #4458, opened 2024-08-19
Share reviews

mattcasey:
image


motechFR (approved):
Works as intended, but ran into a bug when creating a proposal from the template

PR #4461, opened 2024-08-20
Convert Connect to Scoutgame

valentinludu:


.ebextensions_scoutgame/00_env_vars.config
mattcasey, line 1: replace "connect" with "scout-game" in this file

mattcasey (changes requested):
I noticed at least these things TODO:
- [x] update the _env_vars.config file
- [x] add a new deploy .yml file which uses a new ECR image and deploys to staging when "deploy-scoutsgame :rocket:" tag is applied
- [ ] Create a new staging env to test that everyone was done correctly
- [ ] Update destroy_staging.yml so that it tears down the new staging env once PR's are merged

mattcasey (approved):

PR #4475, opened 2024-08-21
Proposal evaluation due date notification settings

Devorein:


background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+
+import { count } from 'lib/metrics';
+import { sendNotificationEmail } from 'lib/notifications/mailer/sendNotificationEmail';
+import { saveProposalNotification } from 'lib/notifications/saveNotification';
+import { isTruthy } from 'lib/utils/types';
+
+export async function sendProposalEvaluationNotifications() {
+  const proposals = await prisma.proposal.findMany({
+    where: {
+      page: {
```
mattcasey, line 13: also add archived: false

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+        deletedAt: null,
+        isTemplate: false
+      },
+      status: 'published',
+      evaluations: {
+        some: {
+          dueDate: {
```
mattcasey, line 20: This query is just going to grow larger over time. I think you should add a gte for the current time, so that we are only looking for proposals in a certain window, and we don't send notifications for events that are past due already. And possible result: null, since we only need to notify evaluations that don't have a result yet.

background/tasks/sendProposalEvaluationNotifications.ts
Devorein: Good one. Updated

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+            evaluationId: currentEvaluation.id
+          });
+
+          const sent = await sendNotificationEmail({
+            id: proposalNotification.id,
+            type: 'proposals'
+          });
+
+          if (sent) {
+            notificationCount += 1;
```
mattcasey, line 118: would be good to add a log here, maybe one for each user that we sent an email

background/tasks/sendProposalEvaluationNotifications.ts
Devorein: Updated. Thnx

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+          }
+        }
+      }
+    } catch (error: any) {
+      log.error(`Error sending proposal evaluation notification: ${error.stack || error.message || error}`, {
```
mattcasey, line 123: you don't need to put the error in the string here since you're passing it in the metadata correctly

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+        error,
+        proposalId: proposal.id,
+        userId: proposal.createdBy
```
mattcasey, line 126: this userId is confusing, its the original author even tho we may be trying to send to a reviewer. better to leave it out. i would include the spaceId instead

background/tasks/sendProposalEvaluationNotifications.ts
Devorein: Sure lets do that

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+      });
+    }
+  }
+
+  if (notificationCount > 0) {
+    log.info(`Sent ${notificationCount} proposal evaluation notifications`);
```
mattcasey, line 132: "proposal evaluation notifications" is really vague, can u be more specific, eg "proposal due date reminders"

background/tasks/sendProposalEvaluationNotifications.ts
```diff
@@ -0,0 +1,137 @@
...
+    count('cron.user-notifications.sent', notificationCount);
```
mattcasey, line 133: change to cron.proposal-reminders.sent?

mattcasey (approved):

background/tasks/sendProposalEvaluationNotifications.ts
Devorein: For user notifications we always use this metric

PR #4478, opened 2024-08-21
Re-add gitcoin storage

motechFR:


mattcasey (approved):
Thanks. i was told by chris we want to run a script on Monday for all existing projects, i am guessing this PR can be merged afterwards so that we start creating attestations ad hoc?

PR #4479, opened 2024-08-21
Evaluation rubric UX

mattcasey:
- use a red "X" insteda of thumbs down icon for rejected reviews
- add a new tab + separate the final decision review from "your review" tab


components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/components/PassFailEvaluation.tsx
```diff
@@ -54,15 +55,12 @@ export type PassFailEvaluationProps = {
...
+const ResultCopyStack = styled(Stack)`
   flex-direction: row;
   gap: ${({ theme }) => theme.spacing(1)};
   align-items: center;
   justify-content: center;
-  padding-bottom: ${({ theme }) => theme.spacing(2)};
-  padding-left: ${({ theme }) => theme.spacing(2)};
-  padding-right: ${({ theme }) => theme.spacing(2)};
-  padding-top: ${({ addPaddingTop, theme }) => (addPaddingTop ? theme.spacing(2) : 0)};
+  padding: ${({ theme }) => theme.spacing(2)};
```
mattcasey, line 63: i removed this prop, we might end up with extra spacing but right now we dont have enough spacing in the normal case. we have way too many conditionals in this file, it's impossible to know what 'addPaddingTop' means or why reviews have to be 0. we should re-organize so that components are composed instead, instead of having one component take 20 props.

mattcasey: image

PR #4483, opened 2024-08-22
Sunny - Fix card content strings

valentinludu:


apps/sunnyawards/components/common/FarcasterCard.tsx
```diff
@@ -37,7 +37,7 @@ function FarcasterCardContent({
       >
         <Avatar size={avatarSize} name={username || 'N/A'} avatar={avatar} />
       </Stack>
-      <Box width='100%'>
+      <Box width='100%' sx={{ wordBreak: 'break-all' }}>
```
mattcasey, line 40: can we discuss this in our dev meeting? i dislike this rule because it breaks words in half even when there's space underneath to just split between the words

mattcasey (approved):

PR #4494, opened 2024-08-23
Fix redirect error triggered by actions

valentinludu:
Best practices:
- use actions only on user click, interactions, instead of automatically triggering them
- use actions only for POST requests
- use redirects inside successful actions like a successful POST request
- if actions are used inside an RSC and throw an error, the client will redirect to a global error or crash, depending on the error boundary
- track action should consider that userId can be undefined.

Future considerations:
- Redirects should be handled in the middleware
   - less bugs, less debugging, since it's in one place
   - middleware renders before all pages so will not have issues with other conditions
   - less redundant code. For example we have redirect('/welcome') in all the pages


mattcasey (approved):

@connect-shared/lib/profile/getCurrentUser.ts
```diff
@@ -0,0 +1,43 @@
...
+    },
+    include: {
+      farcasterUser: true
+    }
+  });
+
+  // Handle a bad session user id
+  if (!user) {
+    session.destroy();
+    log.warn('User has a session that is not found in the db', { userId });
```
mattcasey, line 34: if we're destryong the session i would definitely want more clari\ty than just call this "getCurrentUser" it should be like "getSessionUser" or something

valentinludu: I got out this logic to run in only in one place, the layout. This func should be more dumb: just get the user by id.

valentinludu: I just found out that session.destroy() can be used only in actions or api routes that are requested by the client not the server.

@connect-shared/lib/mixpanel/trackEventAction.ts
```diff
@@ -27,7 +27,7 @@ export const trackEventAction = actionClient
     const event = { userId, ...eventPayload };
 
     // Make sure to use userId from session
-    event.userId = sessionUserId;
+    event.userId = sessionUserId || '';
```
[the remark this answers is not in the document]
valentinludu: TypeScript complains because now, actionClient user.id can be undefined

PR #4510, opened 2024-08-27
Create cv user in farcaster composer app

Devorein:


lib/farcaster/createOrUpdateFarcasterUser.ts
```diff
@@ -0,0 +1,177 @@
...
+        },
+        data: {
+          claimed: true
+        }
+      });
+    }
+
+    trackUserAction('sign_in', { userId: farcasterUser.user.id, identityType: 'Farcaster' });
+
+    await trackOpSpaceClickSigninEvent({
```
mattcasey, line 69: why would we track OP events here?

lib/farcaster/createOrUpdateFarcasterUser.ts
```diff
@@ -0,0 +1,177 @@
...
+        where: {
+          id: farcasterUser.userId
+        },
+        data: {
+          claimed: true
+        }
+      });
+    }
+
+    trackUserAction('sign_in', { userId: farcasterUser.user.id, identityType: 'Farcaster' });
```
mattcasey, line 67: this is not a sign-in event

lib/farcaster/createOrUpdateFarcasterUser.ts
```diff
@@ -0,0 +1,177 @@
...
+          upsert: {
+            create: {
+              description: bio || '',
+              social: {
+                farcasterUrl: `https://warpcast.com/${username}`
+              }
+            },
+            update: {
+              description: userWithWallet.profile?.description || bio || '',
+              locale: userWithWallet.profile?.locale,
```
mattcasey, line 107: why are we updating the profile with the fields we just queried?

lib/farcaster/createOrUpdateFarcasterUser.ts
mattcasey: please move events like "sign up", etc out of lib methods. this method is just 'createOrUpdateUser' it could be used inside a cron job for all we know.

lib/farcaster/createOrUpdateFarcasterUser.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+      }
+    },
+    include: {
+      profile: true
+    }
+  });
+
+  if (userWithWallet) {
+    const updatedUser = await prisma.user.update({
```
mattcasey, line 92: I think we should actually be making sure this user doesn't already have a connected Farcaster account, and in that case just make a new user still. The database would throw an error because userId is unique

mattcasey (changes requested):
check mixpanel events

mattcasey (approved):

PR #4511, opened 2024-08-27
Fixed my work tab for proposals

Devorein:
### WHAT

1. Hide private evaluation details
2. Show a modal if a proposal is currently unviewable
3. Include appeal reviewers and approvers in actionable group

### DIFF
--- a/components/proposals/components/UserProposalsTables/ProposalsTable.tsx
+++ b/components/proposals/components/UserProposalsTables/ProposalsTable.tsx
@@ -5,8 +5,11 @@ import { Stack } from '@mui/system';
 import type { UserProposal } from '@root/lib/proposals/getUserProposals';
 import { relativeTime } from '@root/lib/utils/dates';
 import { useRouter } from 'next/router';
+import { useState } from 'react';
 
+import Modal from 'components/common/Modal';
 import { evaluationIcons } from 'components/settings/proposals/constants';
+import { useSpaceFeatures } from 'hooks/useSpaceFeatures';
 
 const StyledTableRow = styled(TableRow)`
   cursor: pointer;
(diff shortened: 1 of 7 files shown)


motechFR:
Thank you for adding the unit test. Could you please also add some E2E tests for this path?

I think 2 e2e tests should be enough, one with the state for the reviewer, and one for the author

In the test for the author, we could have a proposal with a workflow without private evaluations, and one proposal with evaluations

lib/proposals/__tests__/getUserProposals.spec.ts
```diff
@@ -259,15 +264,123 @@ describe('getUserProposals()', () => {
...
+      userId: proposalAuthor.id
+    });
+
+    const proposalReviewerProposals = await getUserProposals({
+      spaceId: space.id,
+      userId: proposalReviewer.id
+    });
+
+    expect(proposalAuthorProposals.authored[0].currentEvaluation).toBeUndefined();
+    expect(proposalReviewerProposals.actionable[0].currentEvaluation).toBeDefined();
```
motechFR, line 384: I think we need to rework this method. I'd expect authored proposal to have a current evaluation of type "private"

You should be able to use this lib method

lib/proposals/concealProposalSteps.ts


It's also pre-optimised so you can fetch data externally and pass it. That way, the step concealment will work out of the box without many extra fetches

components/proposals/components/UserProposalsTables/ProposalsTable.tsx
```diff
@@ -102,6 +111,15 @@ export function ProposalsTable({ proposals, title }: { title: string; proposals:
           </Box>
         </Card>
       )}
+      <Modal
```
mattcasey, line 114: suggestion, maybe for next time: use `useConfirmationModal()`, it's meant to be an easy way to show a popup-message to users. it doesn't just have to be a confirmation, we could extend it to hide the cancel button

mattcasey (approved):

PR #4513, opened 2024-08-27
Add a mini charm-editor to Farcaster app (new @packages folder)

mattcasey:
- created a new minimal charm editor in @packages/charmeditor
- started moving some core 'extensions' into this package and out of components/
- started usign NY Times's library for react-prosemirror: https://github.com/nytimes/react-prosemirror. So far i have really not used much, and it replaces Bangle.dev's ReactEditor component

Next steps:
- i didnt need to use node views or move over the more complicated parts of CharmEditor. but what i'd like to do is leave some of the larger features like the collab plugin, or 'nested pages', etc. in the main webapp, while all the shared plugins, code inside lib/prosemirror, and React integration will live inside @packages/charmeditor


.gitignore
```diff
@@ -12,27 +12,13 @@ serverless/schema.prisma
 serverless/*.node
 
 # next.js
-/.next/
-/out/
-/dist/
+.next/
+dist/
```
mattcasey, line 16: these changes make it so we don't have to explicitly call out nested versions of the folders

Devorein:
* Can't seem to delete bullet points, it creates tabs instead. You have to select the previous line and then press delete to remove it.
* Disabling the editor while the action is being executed would be a nice UX

Devorein (approved):

PR #4516, opened 2024-08-28
Table row open button link

Devorein:
### WHAT

[Card](https://app.charmverse.io/charmverse/cmd-click-right-click-open-new-tab-proposals-from-proposal-list-129493023591712)

### DIFF
--- a/components/common/DatabaseEditor/components/table/tableRow.tsx
+++ b/components/common/DatabaseEditor/components/table/tableRow.tsx
@@ -4,6 +4,8 @@ import CollapseIcon from '@mui/icons-material/ArrowDropDown';
 import ExpandIcon from '@mui/icons-material/ArrowRight';
 import DragIndicatorIcon from '@mui/icons-material/DragIndicator';
 import { Box, Checkbox, Stack } from '@mui/material';
+import Link from 'next/link';
+import { useRouter } from 'next/router';
 import type { Dispatch, MouseEvent, ReactElement, ReactNode, SetStateAction } from 'react';
 import React, { memo, useContext, useEffect, useMemo, useRef, useState } from 'react';
 import { FormattedMessage } from 'react-intl';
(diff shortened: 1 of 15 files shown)


components/common/DatabaseEditor/components/table/tableRow.tsx
```diff
@@ -411,6 +429,7 @@ function ExpandableTableRow({ subPages, ...props }: Props & { isNested?: boolean
                 card={subPage}
                 indentTitle={30}
                 isNested
+                href={subPage.customIconType === 'applicationStatus' ? `` : ``}
```
mattcasey, line 432: what's going on here? we're checking your icon top and either way we just use an empty string :)

components/common/DatabaseEditor/components/table/tableRow.tsx
```diff
@@ -122,14 +125,21 @@ function TableRow(props: Props) {
...
   const [title, setTitle] = useState('');
   const isGrouped = Boolean(activeView.fields.groupById);
   const isDragAndDropEnabled = !isTouchScreen() && !props.readOnly && !props.disableDragAndDrop;
 
+  const router = useRouter();
+  const domain = router.query.domain;
+
+  const href =
+    card.customIconType === 'applicationStatus'
+      ? `/${domain}/rewards/applications/${card.id}`
```
mattcasey, line 140: remember we can't just pass in domain at the beginning of the href. this will break on spaces hosted under custom domains. But I think you just need to use Link from the components/common folder, and it will handle the space domain part for you

mattcasey (approved):
Approved - works in staging :) but please address the custom domain issue first

components/common/DatabaseEditor/components/table/tableRow.tsx
Devorein: Good one. Thnx

PR #4536, opened 2024-09-03
Fix modal padding

motechFR:


apps/sunnyawards/components/common/Modal/Modal.tsx
```diff
@@ -22,8 +22,8 @@ const ModalContainer = styled.div<{ padding?: string; size: ModalSize }>`
   width: ${({ size }) =>
     size === 'large' ? '670px' : size === 'fluid' ? 'auto' : size === 'small' ? defaultSize : size};
   background-color: #262626;
-  border-radius: 1;
-  padding: 4;
+  border-radius: 10px;
+  padding: 16px;
```
motechFR, line 26: Hardcoding styling rather than going through the theme for timesaving purposes

PR #4537, opened 2024-09-03
Add domain to environment

motechFR:
### WHAT

TODO: Fix the bad rendering of the image in staging env


mattcasey (approved):

mattcasey:
Note that DOMAIN is set inside the StagingStack.ts file, and you are only setting it for production.

PR #4541, opened 2024-09-04
Add scripts for backfilling Agora

motechFR:


mattcasey:
similar to this https://github.com/charmverse/app.charmverse.io/pull/4574

PR #4544, opened 2024-09-04
Purchase nft using decent sdk

Devorein:


apps/farcaster/tsconfig.json
```diff
@@ -65,7 +65,8 @@
     "next-env.d.ts",
     "**/*.ts",
     "**/*.tsx",
-    ".next/types/**/*.ts"
+    ".next/types/**/*.ts",
+    "../../components/proposals/NFTPurchase.tsx"
```
mattcasey, line 69: this should not be necessary

package.json
```diff
@@ -137,6 +137,9 @@
     "@composedb/devtools-node": "^0.7.1",
     "@datadog/browser-logs": "^4.42.2",
     "@datadog/browser-rum": "^4.42.2",
+    "@decent.xyz/box-common": "^3.1.15",
```
mattcasey, line 140: please install these directly to scoutgame module, not the top-level ones (`npm install XXX -w apps/scoutgame`)

mattcasey:
Make sure we are not calling/loading Decent during builds. looks like it's throwing an error about an API key

package.json
```diff
@@ -140,6 +140,8 @@
...
+    "@decent.xyz/the-box": "^3.1.15",
```
mattcasey, line 144: these should be installed in the scoutgame package.json

PR #4547, opened 2024-09-04
Waitlist pages

valentinludu:
Home, Score, Builder pages


mattcasey:
Please check: when i first 'signed in with warpcast', the page didn't redirect. once i clicked "sign in with warpcast' again, it instantly showed me my profile

valentinludu:
I commented above about a fix I made for login page and it should be good + made everything revalidate on update actions to be sure the user does not remain on the login page.

I will check with Mo about the Github connection.

> * when i tried to verify thru github and hit Back on my browser (cuz it fails), it took me to the login page, not my profile. If I try to log in, it just says loading forever. I can see /score being POST'd to, and it's returning some HTML

PR #4549, opened 2024-09-05
SG - Use scout user

Devorein:


apps/scoutgame/app/welcome/page.tsx
```diff
@@ -20,9 +19,5 @@ export default async function Welcome() {
     return null;
   }
 
-  if (user?.data?.connectOnboarded) {
```
mattcasey, line 23: maybe we need to update the model to include 'onboarded' and keep this redirect

PR #4556, opened 2024-09-07
Convert cron, ceramic and websockets into monorepo apps

mattcasey:
- enables us to remove some dependencies from the global package.json
- make clear and standardized boundaries between apps
- will make future cleanup easier
- standardize deployment

Tested:
- [x] Deploy cron to staging
- [x] Deploy ceramic to staging
- [x] Deploy websockets to staging


.github/actions/build/action.yml
```diff
@@ -19,30 +17,17 @@ runs:
...
           dist
-          background/ceramicServer/generated
         key: webapp-${{ hashFiles('**/package-lock.json') }}-${{ hashFiles('**/*.ts', '**/*.tsx', '**/*.gql', '**/*.scss', '!node_modules', '!.next', '!dist', '!apps') }}
         restore-keys: |
           webapp-${{ hashFiles('**/package-lock.json') }}-
 
     - name: Build
       shell: bash
       if: steps.restore_build.outputs.cache-hit != 'true'
-      run: npm run build
```
Devorein, line 36: Are we not going to save the build cache anymore 😮

.github/actions/build_app/action.yml
Devorein, line 1: Why do we have two separate build actions? Can't we use this one for building the webapp too?

.github/workflows/deploy_cron.yml
```diff
@@ -0,0 +1,259 @@
...
+
+      - name: Build and Push Docker image
+        id: docker_build_push
+        uses: ./.github/actions/build_docker_image
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: us-east-1
+        with:
+          ecr_registry: cron
```
Devorein, line 174: I like that we are using a separate registry for cron. I believe previously we were using the webapp one

.github/workflows/deploy_scoutgame.yml
```diff
@@ -138,7 +136,7 @@ jobs:
           file_extensions: '**/*.{ts,tsx}'
 
       - name: Run tests
-        run: npm run test -w apps/scoutgame
```
Devorein, line 141: When testing cron we are using the -w apps/cron format. But for scoutgame and I assume other non webapp apps we are using `app_name:test` format. Shouldn't we stick with a single format?

apps/websockets/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "description": "",
+  "scripts": {
+    "build": "npx tsx esbuild.config.ts",
+    "start": "tsx watch src/initWebsockets.ts",
+    "start:prod": "node ./dist/initWebsockets.js",
+    "test": "npx jest",
+    "typecheck": "../../node_modules/typescript/bin/tsc --project ./tsconfig.json  --noEmit"
+  },
+  "dependencies": {
+    "@packages/charmeditor": "^1.0.0"
```
Devorein, line 13: Nice I didn't know you could do that 😮

docker-compose.yml
```diff
@@ -184,36 +193,27 @@ services:
...
-    profiles:
-      - prd-ceramic
-
   prd-sockets:
     extends:
       service: defaults
     ports:
       - '80:3000'
-    command: ['sh', '-c', 'node ./dist/initWebsockets.js']
+    command: ['sh', '-c', 'node ./apps/websockets/dist/initWebsockets.js']
```
Devorein, line 201: Shouldn't we use an npm command like `npm run websockets:start:prod` or something?

Devorein (approved):

docker-compose.yml
mattcasey: we're using node directly in this file. there is a difference when you use npm, i think it changes how the process gets run inside of docker

.github/workflows/deploy_scoutgame.yml
mattcasey: updated

.github/actions/install/action.yml
```diff
@@ -39,13 +43,14 @@ runs:
         path: |
           ./node_modules
           ./apps/*/node_modules
```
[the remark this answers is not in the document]
mattcasey: they won't be built so there's nothing to be copied

.github/actions/build_app/action.yml
mattcasey: the webapp should eventually use build_app but first we should make it a workspace app

PR #4564, opened 2024-09-09
Token gate comment

mattcasey:
Adds the option to token gate settings to include a message

image

image

image


packages/charmeditor/src/schema.ts
```diff
@@ -1,23 +1,42 @@
...
-  hardBreak.spec,
-  paragraph.spec,
-  text.spec,
-  bold.spec(),
-  italic.spec(),
-  tabIndent.spec
-]);
+export type ExtensionGroup = 'product_updates' | 'tokengate_message';
+
+export const groups: Record<ExtensionGroup, Schema> = {
```
mattcasey, line 17: an idea to allow us more flexibility when using the charm editor, if you want a special set of schemas or plugins, you can just create a new extension group. This allows us a central place to manage how the plugins and schemas are constructed. 

Another important think for me was being able to import the schemas without the plugins

Devorein (approved):

PR #4574, opened 2024-09-11
Proposals/sync workflow

motechFR:
### WHAT

Add ability to resync the proposal evaluations with the template


lib/proposals/workflows/syncProposalWithTemplateEvaluationsAndWorkflowPermissions.ts
mattcasey, line 1: this name is really long, is the idea that it just updates the proposal with whatever settings exist in its template and workflow? Maybe instead it could just be "syncProposalWithTemplate". Workflow is implied, since it gets that from the template too

PR #4576, opened 2024-09-11
Manifest and favicon for waitlist, sg and comingsoon

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/update-favicon-manifest-icons-14563396670165152)


mattcasey (approved):

apps/comingsoon/public/sw.js
mattcasey, line 1: it's odd there is a sw.js in coming soon at all, hm..

mattcasey (changes requested):
i deployed to staging and it is still showing the old favicon, which i think is at apps/scoutgame/favicon.ico.

Next.js seems to want it at that location: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons#image-files-ico-jpg-png

apps/comingsoon/public/sw.js
Devorein: Good catch. We didn't add it in .gitignore.

PR #4579, opened 2024-09-11
SG Layout

valentinludu:
## Changes
- upgrade to MUI V6
- light theme is the same as dark theme for now. We can think about the light theme in the future if we want it
- introduced 2 layouts since we have 2 different types of layouts in figma. This concept is new in Next and it is called **Route Groups**
- the main layout has only configuration logic for the entire app and the second level route group layouts have UI logic
- footer icons are react-icons since we need to change their color from black to white. We can discuss how to export the figma icons in order to achieve the same thing. Rn I didn't manage to make it work since the fill props are everywhere
- Added a new button variant called "gradient". You can check `Sign In` button on `/home` page to see how it will look. e.g. `<Button variant='gradient'>Sign in</Button>` . The DX is pretty neat and I hope we can do this for all buttons and links.
- move everything SG related from @connect-shared to scoutgame folder
- updated all @mui/system  imports to @mui/material
- mui is aware of the data-.... hydration error in the FE and it suggest to add the flag suppressHydrationWarning on the html to supress it which I did and now I removed all warnings


apps/scoutgame/app/page.tsx
Devorein, line 1: We will have a few more routes that are inside (info) shown as soon as the scout logs in for the first time. I think if you pull from main you should get them. All of them were under the `welcome` route

apps/scoutgame/components/layout/LearnMore.tsx
Devorein, line 1: Do we need this on the scoutgame?

valentinludu: I will delete this.

lib/notifications/mailer/sendNotificationEmail.ts
```diff
@@ -228,7 +228,7 @@ async function sendEmail({
       userId: user.id
     },
     subject: template.subject,
-    html: template.html
+    html: await template.html
```
mattcasey, line 231: unecessary?

valentinludu: yes

PR #4581, opened 2024-09-11
Setup waitlist frame events

motechFR:
### WHAT

Track events in the waitlist

- Button interactions tracked as actions
- Initial frame renders have the referrer ID
- When handing a button interaction and resending a frame, we can track a frame view with the viewer


apps/waitlist/lib/waitlistSlots/findOrCreateScoutGameUser.ts
```diff
@@ -0,0 +1,61 @@
...
+          username: true
+        }
+      })
+    : null;
+
+  const fidUserName = existingConnectWaitlistSlot?.username;
+
+  const displayName: string =
+    username ?? fidUserName ?? fid
+      ? await getFarcasterProfile({
```
mattcasey, line 48: probably should add a catch() on this API call

apps/waitlist/lib/waitlistSlots/findOrCreateScoutGameUser.ts
```diff
@@ -0,0 +1,61 @@
...
+          fid
+        }).then((profile) => profile?.body.username || `fid:${fid}`)
```
mattcasey, line 50: we should use displayName, not username

apps/waitlist/lib/waitlistSlots/findOrCreateScoutGameUser.ts
```diff
@@ -0,0 +1,61 @@
...
+  fid,
+  username,
+  wallet
+}: {
+  fid: number;
+  username?: string;
+  wallet?: string;
+}): Promise<Scout> {
+  if (fid && wallet) {
+    throw new InvalidInputError('Cannot provide both fid and wallet');
```
mattcasey, line 17: i dont think this should be an error, it's like "try either one of these" however really i dont think we should need to be looking up by wallet at all for waitlist

mattcasey (approved):
Looks good, i left some comments baout a method that i think was just copied from scoutgame, so not a big deal since waitlist is not going to be a long-lived app anyway

PR #4593, opened 2024-09-13
update github libs

mattcasey:
Upgrading octokit to its new ESM-only version required a few changes to get it to work, and I added comments below. The configs seem to work now, it's a little fragile in how everything has to be set just right, but i am optimistic that Node.js and dependencies will make things more seamless in the future


apps/cron/esbuild.config.ts
```diff
@@ -9,9 +10,15 @@ esbuild
...
-    packages: 'external',
+    // packages: 'external',
+    external: ['@charmverse/core'],
+    plugins: [ddPlugin],
     platform: 'node',
-    sourcemap: true
+    sourcemap: true,
+    // this makes it so that we can use require in the built file https://github.com/evanw/esbuild/issues/946
+    banner: {
+      js: "import { createRequire as topLevelCreateRequire } from 'module';\n const require = topLevelCreateRequire(import.meta.url);\n import path from 'path';\n import { fileURLToPath } from 'url';\n const __filename = fileURLToPath(import.meta.url);\n const __dirname = path.dirname(__filename);"
```
mattcasey, line 20: - to make esbuilds work, i opted to shim the `require()` method and also `__dirname` to make CJS work.

Another option is to set "format": "esm"  in the esbuild configs but this created other runtime errors

apps/cron/package.json
```diff
@@ -2,17 +2,19 @@
...
   "version": "1.0.0",
   "description": "",
+  "type": "module",
   "scripts": {
     "build": "npx tsx esbuild.config.ts",
     "start": "tsx watch src/cron.ts",
-    "start:prod": "node ./dist/cron.js",
-    "test": "npx jest",
+    "start:prod": "node --experimental-vm-modules ./dist/cron.js",
+    "test": "node --experimental-vm-modules ../../node_modules/.bin/jest",
```
mattcasey, line 10: moved our non-next.js apps to "type": "module" and also added " --experimental-vm-modules" to get jest to work

lib/databases/generateCsv.ts
```diff
@@ -9,7 +9,7 @@ import { permissionsApiClient } from '@root/lib/permissions/api/client';
 import { formatDate, formatDateTime } from '@root/lib/utils/dates';
 import { isTruthy } from '@root/lib/utils/types';
 import { stringify } from 'csv-stringify/sync';
-import { sortBy } from 'lodash';
+import { sortBy } from 'lodash-es';
```
mattcasey, line 12: Replaced lodash with lodash-es in places that are imported by 'module' apps, since the regular lodash doesn't seem compatible. However, this required us to use next.js to transpile it in the main app, and playwright has no way to support ESM right now. Node v22 supports requiring esm modules so it may be fixed by then

lib/utils/objects.ts
```diff
@@ -42,3 +40,52 @@ export function removeFalseyFields<T extends object, U extends keyof T>(obj: T):
   }
   return result;
 }
+
+// the following methods were copied from lodash to avoid dealing with es imports
+
+function pick<T extends object, K extends keyof T>(obj: T, keys: K[]): Pick<T, K> {
```
mattcasey, line 46: not happy about having to do this part, but the type: module apps want "lodash-es" and next.js can transpile it fine. It's just that I could not get playwright to work with lodash-es, so i just inlined the methods where both apps need it (testing lib)

PR #4606, opened 2024-09-16
Sunny/wallet login

motechFR:


mattcasey:
When I tried it, it does not show Metamask (because i have phantom, but we should make sure Metamask is always available then)

Also, clicking Wallet Connect the QR code shows up behind the selection at least in Firefox.
image

mattcasey:
i fixed the layout issue. by applying a z-index.. i dunno why metamask doesn't work on firefox but don't think it has to block

PR #4611, opened 2024-09-17
SG home

valentinludu:


apps/scoutgame/components/common/Header/components/HeaderMessage.tsx
```diff
@@ -1,10 +1,16 @@
 import 'server-only';
 
 import { Box, Typography } from '@mui/material';
+import { delay } from '@root/lib/utils/async';
 
-export function HeaderMessage() {
+export async function HeaderMessage() {
+  const data = await delay(3000);
```
mattcasey, line 7: is this for debugging or just to add a delay for the component to appear (which would be neat)?

apps/scoutgame/middleware.ts
```diff
@@ -1,33 +1,26 @@
 import { getSession } from '@connect-shared/lib/session/getSession';
-import { isTruthy } from '@root/lib/utils/types';
 import type { NextRequest } from 'next/server';
 import { NextResponse } from 'next/server';
 
+const privateLinks = ['/profile', '/scout', '/notifications', '/u'];
```
mattcasey, line 5: should "/u" be private? does that mean viewing someone's profile requires login?

mattcasey (approved):

apps/scoutgame/components/common/Header/components/HeaderMessage.tsx
valentinludu: It's for mocking a request and seeing a neat skeleton before the component appears. I have this everywhere. When the data will come from the backend this is the place to put them.

apps/scoutgame/middleware.ts
valentinludu: Oh, no. /u is not private. I removed it.

PR #4612, opened 2024-09-17
handle new error with TreeView

mattcasey:
The latest TreeView throws an error if any item id is not unique. This can happen when you have a favorate page that is also nested under another favorite page. I tried adding unique id's to all items but it doesn't seem to work right unless item.id is used... which can break a lot of assumptions so for now i am just filtering out duplicates :P

### DIFF
--- a/pages/api/optimism/projects/index.ts
+++ b/pages/api/optimism/projects/index.ts
@@ -30,7 +30,7 @@ async function getProjectsController(req: NextApiRequest, res: NextApiResponse<O
   });
 
   if (!user || !user.farcasterUser) {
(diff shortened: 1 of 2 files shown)


pages/api/optimism/projects/index.ts
```diff
@@ -30,7 +30,7 @@ async function getProjectsController(req: NextApiRequest, res: NextApiResponse<O
   });
 
   if (!user || !user.farcasterUser) {
-    throw new InvalidInputError('User does not have a farcaster account.');
+    return res.status(200).json([]);
```
mattcasey, line 33: this was bugging me in dev, just fixed in this PR altho unrelated

PR #4638, opened 2024-09-22
Gems payout cron

Devorein:

### DIFF
--- a/apps/scoutgamecron/cron.yml
+++ b/apps/scoutgamecron/cron.yml
@@ -4,3 +4,8 @@ cron:
     url: "/process-pull-requests"
     # every 5 minutes
     schedule: "*/5 * * * *"
+
+  - name: "process-gems-payout"
+    url: "/process-gems-payout"
+    # End of every week
+    schedule: "0 0 * * 0"
--- a/apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
+++ b/apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
@@ -1,0 +1,134 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import { calculatePointsForRank, getTopBuilders } from '@packages/scoutgame/calculatePoints';
+import { currentSeason, getCurrentWeek } from '@packages/scoutgame/utils';
+
+export async function processGemsPayout() {
+  const topBuilders = await getTopBuilders(100);
+  const week = getCurrentWeek();
+
+  for (const { builder, gemsCollected, rank } of topBuilders) {
+    const calculatedPoints = calculatePointsForRank(rank);
+    await prisma.gemsPayoutEvent.create({
+      data: {
+        gems: gemsCollected,
+        points: calculatedPoints,
+        week,
+        builderId: builder.id
+      }
+    });
+
+    const nftHolders = await prisma.nFTPurchaseEvent.groupBy({
+      by: ['scoutId'],
+      where: {
+        builderId: builder.id
+      },
+      _count: {
+        scoutId: true
+      }
+    });
+
+    const totalNftsPurchased = nftHolders.reduce((acc, { _count: { scoutId: count } }) => acc + count, 0);
+
+    for (const {
+      scoutId,
+      _count: { scoutId: count }
+    } of nftHolders) {
+      const scoutPoints = 0.8 * calculatedPoints * (count / totalNftsPurchased);
+      await prisma.$transaction([
+        prisma.pointsReceipt.create({
+          data: {
+            value: scoutPoints,
+            recipientId: scoutId,
+            // Why is the eventId a builder event
+            eventId: '',
+            // Who would be the sender here?
+            senderId: '',
+            // Why is claimedAt always required?
+            claimedAt: new Date()
+          }
+        }),
+        prisma.userSeasonStats.upsert({
+          where: {
+            userId_season: {
+              userId: scoutId,
+              season: currentSeason
+            }
+          },
+          update: {
+            pointsEarnedAsScout: {
+              increment: scoutPoints
+            }
+          },
+          create: {
+            pointsEarnedAsScout: scoutPoints,
+            userId: scoutId,
+            season: currentSeason,
+            pointsEarnedAsBuilder: 0
+          }
+        }),
+        prisma.userAllTimeStats.upsert({
+          where: {
+            userId: scoutId
+          },
+          update: {
+            pointsEarnedAsScout: {
+              increment: scoutPoints
+            }
+          },
+          create: {
+            pointsEarnedAsScout: scoutPoints,
+            userId: scoutId,
+            pointsEarnedAsBuilder: 0,
+            currentBalance: scoutPoints
+          }
+        })
+      ]);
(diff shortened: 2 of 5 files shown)


apps/scoutgamecron/cron.yml
```diff
@@ -4,3 +4,8 @@ cron:
     url: "/process-pull-requests"
     # every 5 minutes
     schedule: "*/5 * * * *"
+
+  - name: "process-gems-payout"
+    url: "/process-gems-payout"
+    # End of every week
```
mattcasey, line 10: This is incorrect, because it will run at 12 UTC Sunday. And the specific UTC time will change with DST, so I think it's safest if the cron just runs on the hour every hour and the code determines if it's the right time or not.

apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
```diff
@@ -0,0 +1,134 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import { calculatePointsForRank, getTopBuilders } from '@packages/scoutgame/calculatePoints';
+import { currentSeason, getCurrentWeek } from '@packages/scoutgame/utils';
+
+export async function processGemsPayout() {
+  const topBuilders = await getTopBuilders(100);
+  const week = getCurrentWeek();
```
mattcasey, line 7: please copy the pattern i started w/processPullRequests:

Inside index.ts: the function which gets called by worker. It should handle the retrieval of high-level context and pass into a method(s) (passing in the data makes it easier to be unit-tested). For example:

```
index.ts:
async function  processGemsPayout() {
   const isTimeToRun = /* check if its Sunday at midnight in NY time */
    if (isTimeToRun) { 
      const currentWeek = getCurrentWeek();
      const getTopBuilders(week);
      for (const builder of builders) {
        const points = getGemsPayoutPoints({ builder, index: builders.indexOf(builder) });
        await createGemsPayout(builder, points);
    }
}

```

apps/scoutgamecron/src/worker.ts
```diff
@@ -35,6 +36,8 @@ addTask('/process-pull-requests', processPullRequests);
 
 addTask('/send-push-notifications', sendNotifications);
 
+addTask('/process-gems-payout', processGemsPayout);
```
mattcasey, line 39: please add at least one test on this endpoint like the one we have for send-push-notifications

packages/scoutgame/src/calculatePoints.ts
```diff
@@ -79,3 +88,32 @@ export function getPointsEarnedAsScout(scoutId: string, receipts: (PointsReceipt
     return acc;
   }, 0);
 }
+
+export async function getTopBuilders(quantity: number) {
+  const userWeeklyStats = await prisma.userWeeklyStats.findMany({
+    where: {
+      week: getCurrentWeek()
```
mattcasey, line 95: you should pass in currentWeek so that it is easy to test this method, it's also better if we calculate currentWeek once at the top method and pass it down, so there is one source of truth

packages/scoutgame/src/calculatePoints.ts
```diff
@@ -79,3 +88,32 @@ export function getPointsEarnedAsScout(scoutId: string, receipts: (PointsReceipt
...
+    },
+    orderBy: [
+      {
+        gemsCollected: 'desc'
+      }
+    ],
+    take: quantity,
```
mattcasey, line 102: We should get more than 100 scouts in the case of a tie, in which case we have to sort the builders by the date they received gems first. 

Suggestion: just get all the weekly stats, then sort the list and take the quantity from that

apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
```diff
@@ -0,0 +1,134 @@
...
+      }
+    });
+
+    const totalNftsPurchased = nftHolders.reduce((acc, { _count: { scoutId: count } }) => acc + count, 0);
+
+    for (const {
+      scoutId,
+      _count: { scoutId: count }
+    } of nftHolders) {
+      const scoutPoints = 0.8 * calculatedPoints * (count / totalNftsPurchased);
```
mattcasey, line 36: this should be captured outside in some kind of lib method

apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
```diff
@@ -0,0 +1,134 @@
...
+  const week = getCurrentWeek();
+
+  for (const { builder, gemsCollected, rank } of topBuilders) {
+    const calculatedPoints = calculatePointsForRank(rank);
+    await prisma.gemsPayoutEvent.create({
+      data: {
+        gems: gemsCollected,
+        points: calculatedPoints,
+        week,
+        builderId: builder.id
```
mattcasey, line 16: you should calculate all the point transactions and create them as part of this create call

```
  gemsPayoutEvent.create({
    data: {
       ...
       pointsTransactions: {
          createMany: { data: transactions } 
        }
   }
})
```          

```

File: apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts [hunk 8]
```diff
@@ -0,0 +1,134 @@
...
+      const scoutPoints = 0.8 * calculatedPoints * (count / totalNftsPurchased);
+      await prisma.$transaction([
+        prisma.pointsReceipt.create({
+          data: {
+            value: scoutPoints,
+            recipientId: scoutId,
+            // Why is the eventId a builder event
+            eventId: '',
+            // Who would be the sender here?
+            senderId: '',
```
mattcasey, line 45: senderId should be null (and made optional in core)

apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
```diff
@@ -0,0 +1,134 @@
...
+      scoutId,
+      _count: { scoutId: count }
+    } of nftHolders) {
+      const scoutPoints = 0.8 * calculatedPoints * (count / totalNftsPurchased);
+      await prisma.$transaction([
+        prisma.pointsReceipt.create({
+          data: {
+            value: scoutPoints,
+            recipientId: scoutId,
+            // Why is the eventId a builder event
```
mattcasey, line 42: see note above, it is the eventId of the gem payout

apps/scoutgamecron/src/tasks/processGemsPayout/processGemsPayout.ts
```diff
@@ -0,0 +1,134 @@
...
+            eventId: '',
+            // Who would be the sender here?
+            senderId: '',
+            // Why is claimedAt always required?
+            claimedAt: new Date()
```
mattcasey, line 47: you should not set claimedAt - the scout needs to come claim these points first :)

packages/scoutgame/src/calculatePoints.ts
Devorein: Good idea. Updated

mattcasey (approved):

PR #4645, opened 2024-09-25
SG profile page

Devorein:


package.json
```diff
@@ -140,7 +140,7 @@
     "@bangle.dev/pm-commands": "^0.31.6",
     "@bangle.dev/utils": "^0.31.6",
     "@beam-australia/react-env": "^3.1.1",
-    "@charmverse/core": "^0.83.0",
+    "@charmverse/core": "^0.83.1-rc-scout-register.1",
```
mattcasey, line 143: there's no PRs open on core right now

packages/scoutgame/src/claimPoints.ts
```diff
@@ -0,0 +1,116 @@
+import { prisma } from '@charmverse/core/prisma-client';
+import { currentSeason } from '@packages/scoutgame/utils';
```
mattcasey, line 2: use relative imports when you're using code from the same package

packages/scoutgame/src/claimPoints.ts
```diff
@@ -0,0 +1,116 @@
...
+        claimedAt: null,
+        event: {
+          season: currentSeason
+        }
+      },
+      data: {
+        claimedAt: new Date()
+      }
+    });
+    await prisma.scout.update({
```
mattcasey, line 57: did you mean to use tx here and below?

packages/scoutgame/src/utils.ts
```diff
@@ -77,3 +77,19 @@ export function getCurrentWeekPoints() {
...
   return 100000;
 }
+
+export function getCurrentScoutGameWeek() {
+  const currentWeek = getCurrentWeek();
+  const currentWeekNumber = parseInt(currentWeek.split('W')[1]);
+  return currentWeekNumber;
+}
+
+export function getRelativeTime(date: Date | string) {
```
mattcasey, line 87: I think this is OK here, but i only used "utils" because i didnt have a better name. it's supposed to be scout-game specific utilities. It should notj ust be any utilities we need for scoutgame apps. This method might be better in @packages/utils/dates for instance

apps/scoutgame/tsconfig.json
```diff
@@ -19,40 +24,100 @@
...
+      ]
     }
   },
   "include": [
     "next-env.d.ts",
     "**/*.ts",
     "**/*.tsx",
-    ".next/types/**/*.ts"
+    ".next/types/**/*.ts",
+    "../../packages/scoutgame/src/claimPoints.ts"
```
mattcasey, line 115: do not add links to packages here, we use node modules to import package code

apps/scoutgame/app/(general)/profile/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+  searchParams
+}: {
+  searchParams: {
+    tab: ProfileTab;
+  };
+}) {
+  const user = await getUserFromSession();
+  const tab = searchParams.tab || (user?.githubUser ? 'build' : 'scout');
+  if (!user) {
+    redirect('/');
```
mattcasey, line 29: dont redirect to home if no user exists (it doesnt even work because home is a public page). this redirect should be handled in middleware.ts.

apps/scoutgame/app/(general)/profile/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export default async function Profile({
+  searchParams
+}: {
+  searchParams: {
+    tab: ProfileTab;
+  };
+}) {
+  const user = await getUserFromSession();
+  const tab = searchParams.tab || (user?.githubUser ? 'build' : 'scout');
```
mattcasey, line 27: instead of adding a JOIN on the query for getCurentUser, just check if user.builder is true please

apps/scoutgame/app/(general)/profile/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+import type { Metadata } from 'next';
+import { redirect } from 'next/navigation';
+
+import type { ProfileTab } from 'components/profile/mine/ProfilePage/ProfilePage';
+import { ProfilePage } from 'components/profile/mine/ProfilePage/ProfilePage';
+import { getUserFromSession } from 'lib/session/getUserFromSession';
+
+export const dynamic = 'force-dynamic';
+
+export const metadata: Metadata = {
```
mattcasey, line 12: thanks for adding this.. i have noted to myself we should go and review metadata for all the pages :)

apps/scoutgame/app/(general)/profile/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+  });
+
+  return (
+    <ProfilePage
+      user={{
+        id: user.id,
+        displayName: user.displayName,
+        username: user.username,
+        avatar: user.avatar || '',
```
mattcasey, line 65: you should just let avatar be null, instead of passing an empty string. that doesn't help anything

apps/scoutgame/app/(general)/profile/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+        githubLogin: user.githubUser?.login,
+        currentBalance: user.currentBalance,
+        bio: user.bio,
+        seasonPoints: {
+          builderPoints: seasonPoints?.pointsEarnedAsBuilder || 0,
+          scoutPoints: seasonPoints?.pointsEarnedAsScout || 0
+        },
+        allTimePoints: {
+          builderPoints: allTimePoints?.pointsEarnedAsBuilder || 0,
```
mattcasey, line 74: suggestion: have the 0 defaults inside the component. it makes this a lot simpler: `alltimePoints: allTimePoints`

apps/scoutgame/app/(general)/scout/page.tsx
```diff
@@ -1,13 +1,11 @@
...
 export default async function Scout({
   searchParams
 }: {
   searchParams: { [key: string]: string | string[] | undefined };
 }) {
   const sort = searchParams.sort as string;
-  const user = await getUserFromSession();
 
-  return <ScoutPage user={user || null} sort={sort || 'top'} />;
+  return <BuilderSearchPage sort={sort || 'top'} />;
```
mattcasey, line 10: please rename this back. "ScoutPage" is a logical name for the component that handles the Scout page

apps/scoutgame/components/scout/ScoutCard.tsx
```diff
@@ -0,0 +1,37 @@
+import { Paper, Stack, Typography } from '@mui/material';
```
mattcasey, line 1: this belongs in a components folder

apps/scoutgame/components/scout/BuilderSearchPage.tsx
```diff
@@ -1,22 +1,21 @@
 import 'server-only';
 
-import type { Scout } from '@charmverse/core/prisma-client';
 import { Box } from '@mui/material';
 
-import type { BuildersSort } from 'lib/builders/getSortedBuilders';
+import { BuildersGallery } from 'components/builder/BuildersGallery';
+import { getSortedBuilders, type BuildersSort } from 'lib/builders/getSortedBuilders';
```
mattcasey, line 6: please try to avoid this type of import (where it's inline). its nice if we can be consistent, and types are imported separately from runtime code

apps/scoutgame/components/scout/BuilderSearchPage.tsx
```diff
@@ -1,22 +1,21 @@
...
 
-import { BuildersGridContainer } from './components/BuildersGridContainer';
 import { SearchBuildersInput } from './components/SearchBuildersInput';
 import { SortOptionTabs, sortOptions } from './components/SortOptionTabs';
 
-export async function ScoutPage({ user, sort }: { user: Scout | null; sort: string }) {
+export async function BuilderSearchPage({ sort }: { sort: string }) {
   const currentSort = sortOptions.some((t) => t.value === sort) ? sort : 'top';
-
+  const builders = await getSortedBuilders({ sort: currentSort as BuildersSort, limit: 10 });
```
mattcasey, line 13: this async query should be inside BuildersGallery so it doesnt block the other components

apps/scoutgame/components/scout/ScoutsGallery.tsx
```diff
@@ -0,0 +1,18 @@
+import { Box, Grid2 as Grid } from '@mui/material';
```
mattcasey, line 1: it doesnt look like this component is even used in teh Scouts Page?

apps/scoutgame/components/profile/[username]/PublicBuilderProfile/MobilePublicBuilderProfile.tsx
Comment (mattcasey) [inline apps/scoutgame/components/profile/[username]/PublicBuilderProfile/MobilePublicBuilderProfile.tsx:1]: this must be combined with the desktop view. we determined in our dev chat there's no need to have separate components. Please fix - this is gonna be extremely hard to maintain

apps/scoutgame/components/profile/[username]/PublicProfileTabsMenu.tsx
Comment (mattcasey) [inline apps/scoutgame/components/profile/[username]/PublicProfileTabsMenu.tsx:1]: please group children into components folders so we can see the heirarchy of things

apps/scoutgame/components/profile/[username]/PublicProfileTabsMenu.tsx
```diff
@@ -0,0 +1,15 @@
+import { TabsMenu } from 'components/common/Tabs/TabsMenu';
+
+export function PublicProfileTabsMenu({ tab, username }: { tab: string; username: string }) {
+  const value = ['scout', 'build'].includes(tab) ? tab : 'scout';
```
Comment (mattcasey) [inline apps/scoutgame/components/profile/[username]/PublicProfileTabsMenu.tsx:4]: Please define this value outside of the tabs so that we have a single source of truth that is also used when we display the view below the tabs

apps/scoutgame/components/builder/Carousel/TodaysHotBuildersCarousel.tsx
mattcasey, line 1: this component is specific to the home page, therefore it belongs in components/home/components

apps/scoutgame/components/builder/BuilderActivitiesList.tsx
mattcasey, line 1: please get rid of "builder" as a folder, since that is not the name of any path in our app and we don't group components by type. they should live either inside a path-based folder or the "common" folder. it's also fine IMO if this lives inside of, say, the folder of one page but is imported by two different page components

apps/scoutgame/components/builder/BuildersGallery.tsx
```diff
@@ -0,0 +1,19 @@
+import { Box, Grid2 as Grid } from '@mui/material';
```
mattcasey, line 1: so this one can just be at components/common/BuildersGallery

apps/scoutgame/components/builder/Card/BuilderCard.tsx
mattcasey, line 1: this should jsut be components/common/BuilderCard

apps/scoutgame/components/welcome/WelcomePage.tsx
```diff
@@ -12,7 +12,14 @@ export function WelcomePage({ user }: { user: Scout }) {
     <SinglePageLayout>
       <SinglePageWrapper>
         <Box display='flex' gap={3} flexDirection='column' alignItems='flex-start'>
-          {user.farcasterId && <UserProfile user={user} />}
+          {user.farcasterId && (
+            <UserProfile
+              user={{
+                ...user,
+                avatar: user.avatar ?? ''
```
mattcasey, line 19: noted elsewhere: UserProfile should handle null value for avatar so that the rest of our code is simple and doesn't have to lie

apps/scoutgame/lib/builders/getBuilderWeeklyStats.ts
```diff
@@ -0,0 +1,14 @@
+import { prisma } from '@charmverse/core/prisma-client';
+
+export async function getBuilderWeeklyStats(builderId: string) {
```
mattcasey, line 3: i am almost positive that you want to pass in the week here, and not just grab the first weekly stats :)

apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -24,18 +24,52 @@ export async function getSortedBuilders({
           orderBy: {
             createdAt: 'desc'
           },
+          where: {
+            bannedAt: null,
```
mattcasey, line 28: good idea to filter on bannedAt!

apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -24,18 +24,52 @@ export async function getSortedBuilders({
...
             avatar: true,
-            createdAt: true
+            createdAt: true,
+            builderNfts: {
+              where: {
+                season: currentSeason
+              },
+              select: {
+                currentPrice: true,
+                nftSoldEvents: {
```
mattcasey, line 43: we should not need to query sold events here, it should be stored on the stats someplace

packages/scoutgame/src/utils.ts
```diff
@@ -77,3 +77,19 @@ export function getCurrentWeekPoints() {
   // TODO: Get points allocation for the week
   return 100000;
 }
+
+export function getCurrentScoutGameWeek() {
```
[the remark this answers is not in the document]
Devorein: Good point. Updated.

apps/scoutgame/app/(general)/profile/page.tsx
Devorein: yes thats perfect. thnx

apps/scoutgame/app/(general)/profile/page.tsx
Devorein: Yes good point. I will update it.

apps/scoutgame/components/scout/BuilderSearchPage.tsx
Devorein: Will keep that in mind. It was the editor.

PR #4648, opened 2024-09-25
Feat/notifications

motechFR:


packages/scoutgame/src/script/generateActivities.ts
```diff
@@ -0,0 +1,265 @@
...
+import { getCurrentWeek } from '../utils';
+
+function getRandomDateWithinLast30Days() {
+  const now = new Date();
+  const pastDate = new Date(now);
+  pastDate.setDate(now.getDate() - 30);
+  return new Date(pastDate.getTime() + Math.random() * (now.getTime() - pastDate.getTime()));
+}
+
+export async function generateActivities({ userId }: { userId: string }) {
```
motechFR, line 19: This method has some bugs, need to fix underlying but was good enough to get a few events

mattcasey (approved):

PR #4649, opened 2024-09-26
Scout game user weekly rank cron

Devorein:
### WHY

* Refactored code a bit
* Improved UI to match mockups
* Added extra test cases


packages/scoutgame/src/updateBuildersRank.ts
```diff
@@ -0,0 +1,24 @@
+import { log } from '@charmverse/core/log';
+import { prisma } from '@charmverse/core/prisma-client';
+
+import { getBuildersLeaderboard } from './getBuildersLeaderboard';
+
+export async function updateBuildersRank({ week }: { week: string }) {
```
mattcasey, line 6: this function seems like it could just go inside of the cron app, i can't think of anything else that would be responsible for updating the weekly stats of builders

packages/scoutgame/src/utils.ts
```diff
@@ -87,3 +87,17 @@ export function getCurrentWeekNumber() {
...
+export function getWeekNumber(date: Date) {
+  const currentDate = DateTime.fromJSDate(date, { zone: timezone });
+  const weeksDiff = currentDate.diff(seasonStartDate, 'weeks').weeks;
+  return Math.floor(weeksDiff) + 1;
+}
+
+export function getWeekNumberFromWeek(weekString: string) {
+  const [year, week] = weekString.split('-W');
+  const weekNumber = parseInt(week);
+  const startOfYear = DateTime.fromObject({ year: parseInt(year), month: 1, day: 1 }, { zone: timezone });
```
mattcasey, line 100: this looks like a false assumption. the week number is not based on the start of the year. but i'm not sure what you are trying to do w/the result of this method

packages/scoutgame/src/utils.ts
```diff
@@ -87,3 +87,17 @@ export function getCurrentWeekNumber() {
   const weeksDiff = currentDate.diff(seasonStartDate, 'weeks').weeks;
   return Math.floor(weeksDiff) + 1;
 }
+
+export function getWeekNumber(date: Date) {
```
mattcasey, line 91: This function shouldn't be exported because it's only used internally.

packages/scoutgame/src/utils.ts
```diff
@@ -87,3 +87,17 @@ export function getCurrentWeekNumber() {
...
+  const currentDate = DateTime.fromJSDate(date, { zone: timezone });
+  const weeksDiff = currentDate.diff(seasonStartDate, 'weeks').weeks;
+  return Math.floor(weeksDiff) + 1;
+}
+
+export function getWeekNumberFromWeek(weekString: string) {
```
mattcasey, line 97: new functions like this need unit tests with various edge cases. Remember that "weeks" in terms of date is usually starting on Monday, but ours start on Sunday. What happens if you are diffing a time between Sunday and Saturday? It should take into account that Saturday is the previous week, but the DateTime definition will probably consider them to be the same week.

apps/scoutgame/app/(general)/layout.tsx
```diff
@@ -20,7 +20,14 @@ export default async function Layout({
   return (
     <Box display='grid' gridTemplateRows='auto 1fr auto' minHeight='100vh' bgcolor='background.default'>
       <Header user={user || null} />
-      <Box component='main' minHeight='100%'>
+      <Box
+        component='main'
+        minHeight='100%'
+        px={{
```
mattcasey, line 26: this padding breaks the layout - the banner is no longer full-width, for instance. margins are too big compared to figma on mobile:

image

Here's the figma;
image

apps/scoutgame/app/(general)/home/loading.tsx
```diff
@@ -0,0 +1,22 @@
+import { Skeleton } from '@mui/material';
+import { Stack } from '@mui/system';
+
+import { LoadingBanner } from 'components/layout/Loading/LoadingBanner';
+import { LoadingCards } from 'components/layout/Loading/LoadingCards';
+import { LoadingTable } from 'components/layout/Loading/LoadingTable';
+
+export default function Loading() {
+  return (
+    <Stack gap={2}>
```
mattcasey, line 10: this is honestly just distracting, it doesnt line up exactly with the layout. i dont think we need skeletons for every page. let's just load the data fast

apps/scoutgame/lib/users/getUserStats.ts
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  const githubUser = currentUser?.githubUser[0];
+  const seasonPoints = currentUser?.userSeasonStats[0];
+  const allTimePoints = currentUser?.userAllTimeStats[0];
+
+  return {
+    githubLogin: githubUser?.login,
+    seasonPoints: {
+      builderPoints: seasonPoints?.pointsEarnedAsBuilder,
```
mattcasey, line 52: why bother renaming the fields? can we just pass seasonPoints directly?

packages/scoutgame/src/utils.ts
mattcasey: I wonder if you could simplify and just use this logic:
```
const currentWeekNumber = parseInt(week.split('-W')[1]);
const seasonStart = getWeek(seasonStartDate);
const seasonStartWeekNumber = parseInt(seasonStart.week.split('-W')[1]);
return seasonStartWeekNumber - currentWeekNumber + 1;
```

packages/scoutgame/src/utils.ts
Devorein: Removed this fn

packages/scoutgame/src/utils.ts
Devorein: Added some tests for it. Thnx for the reminder.

apps/scoutgame/app/(general)/home/loading.tsx
Devorein: Thnx. Figured out a solution. It looks much better now

apps/scoutgame/app/(general)/layout.tsx
Devorein: Thnx. Figured out a solution. It looks much better now

apps/scoutgame/lib/users/getUserStats.ts
Devorein: updated

apps/scoutgame/app/(general)/home/loading.tsx
Devorein: For now I've just removed the loading screens. We can add them back later on if required.

PR #4660, opened 2024-09-27
Builder NFT image upload & display

Devorein:


apps/scoutgame/__e2e__/profilePage.spec.ts
```diff
@@ -20,7 +20,7 @@ test.describe('Profile page', () => {
     await profilePage.selectNavigationLink('/profile').click();
 
     // Logged in user should be redirected
-    await page.waitForURL('**/profile');
+    await page.waitForURL('**/profile*');
```
mattcasey, line 23: curious is this actually necessary? what's added to the /profile URL?

apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfile.tsx
```diff
@@ -43,6 +44,7 @@ export async function PublicBuilderProfile({ tab, user }: { tab: string; user: B
       scouts={scouts}
       builder={{
         ...user,
+        avatar: builderNft?.imageUrl,
```
Comment (mattcasey) [inline apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfile.tsx:47]: This is fine, but we should probably use a different name than "avatar" in our component

mattcasey (approved):

PR #4664, opened 2024-09-27
Dont error out nft purchase

motechFR:

### DIFF
--- a/apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfileContainer.tsx
+++ b/apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfileContainer.tsx
@@ -31,6 +31,7 @@ export type BuilderProfileProps = {
   gemsCollected?: number;
   rank?: number | null;
   user?: {
+    id: string;
     username: string;
   } | null;
 };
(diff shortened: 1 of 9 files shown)


apps/scoutgame/components/nft/NFTPurchase.tsx
```diff
@@ -155,6 +155,31 @@ function NFTPurchaseButton({ builderId, user }: NFTPurchaseProps) {
     }
   }, [tokensToBuy, builderTokenId, refreshAsk]);
 
+  /** TODO - Use this payload when we resume calling the contract directly
```
motechFR, line 158: This works in decent console, and in our app for talking to the contract and buying directly

Currently disabled as its us that make the call instead

apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfileContainer.tsx
```diff
@@ -31,6 +31,7 @@ export type BuilderProfileProps = {
   gemsCollected?: number;
   rank?: number | null;
   user?: {
+    id: string;
```
Comment (motechFR) [inline apps/scoutgame/components/[username]/components/PublicBuilderProfile/PublicBuilderProfileContainer.tsx:34]: Passing down user.id as we need to know which scout to reference in the contract call data

PR #4676, opened 2024-09-29
Show spam policy page after github connect

Devorein:


apps/scoutgame/app/api/connect-github/callback/route.ts
```diff
@@ -131,14 +136,22 @@ export async function GET(req: NextRequest) {
       id: scout.id
     },
     data: {
-      onboardedAt: new Date()
+      onboardedAt: new Date(),
+      githubUser: {
```
mattcasey, line 140: was this just an oversight? how were we not creating github user before!

mattcasey (approved):

PR #4687, opened 2024-09-30
Nft purchase UI

valentinludu:


mattcasey:
In dev, when I click on the ScoutButton on scouts page it takes me to the home page for some reason

mattcasey:
Might be WIP, but I tried it out and on desktop i can't see the whole modal:
image

valentinludu:
Fixed this one. In a latest cleanup the user was removed. I added it back.

> In dev, when I click on the ScoutButton on /scouts page it takes me to the home page for some reason. I am just using the seed data, maybe related. The modal opens if i am on the scouts public profile

valentinludu:
I don't see this error anymore. I know it was an issue with the contract not being the correct one but that is just when you have dummy data.

> The form looks really good. Might be WIP, but I tried it out and on desktop i can't see the whole modal. (I think the error is because i'm using seed data, but not sure why it says 'execution reverted', cuz i didnt click anything)

PR #4699, opened 2024-09-30
SG login info modal

valentinludu:
[Card](https://app.charmverse.io/charmverse/better-sign-in-flow-when-clicking-something-that-takes-you-to-log-in-7484419506941695)


apps/scoutgame/components/common/Header/Header.tsx
```diff
@@ -68,7 +68,7 @@ export function Header({ user }: { user: Pick<Scout, 'username' | 'avatar' | 'cu
             </Link>
             <Stack flexDirection='row' gap={2} alignItems='center'>
               <Hidden mdDown>
-                <SiteNavigation topNav />
+                <SiteNavigation topNav isAuthenticated={!!user} />
```
mattcasey, line 71: i like this choice of prop name, it's very clear

apps/scoutgame/components/common/ScoutButton/ScoutButton.tsx
```diff
@@ -36,6 +37,7 @@ export function ScoutButton({
         ${(Number(builder.price) / 10 ** builderTokenDecimals).toFixed(2)}
       </Button>
       <NFTPurchaseDialog open={isPurchasing} onClose={() => setIsPurchasing(false)} builder={builder} />
+      <SignInModalMessage open={authPopup} onClose={() => setAuthPopup(false)} />
```
mattcasey, line 40: Suggestion, maybe for next time: I think it's even better when we just have a hook like "useSignInPrompt", similar to "useSnackbar", and having one global modal and state in globalcomponents. It's much easier to use and u don't have to worry about managing a new modal state every time

mattcasey: Maybe the hook could return a method that wraps a function, and calls it you are logged in:
```
const { requireLogin } = useRequireLogin();

return <>
   <Button onClick={requireLogin(showPurchaseModal)} />
    {/* when used on an anchor link, it just prevents the link from being triggered */}
   <Link href='/profile' onMouseDown={requireLogin} />
</>

```

mattcasey (changes requested):
Look good but one request to change: the redirect after login should actually go to the page you're trying to get to. For example, if you click on Notifications, i should drop you into notifications after login. In the onClick handler on an anchor, it could use the href instead of window.location.pathname

valentinludu:
> Look good but one request to change: the redirect after login should actually go to the page you're trying to get to. For example, if you click on Notifications, i should drop you into notifications after login. Right now it just takes me back to home. In the onClick handler on an anchor, it could use the href instead of window.location.pathname

Just pushed an improvement rn.

apps/scoutgame/components/common/ScoutButton/ScoutButton.tsx
valentinludu: Yes. We can discuss tomorrow improvements for the whole app.

mattcasey (approved):

PR #4704, opened 2024-10-01
Scoutgame notifications

Devorein:


apps/scoutgame/app/(general)/notifications/page.tsx
```diff
@@ -6,8 +6,12 @@ import { getUserFromSession } from 'lib/session/getUserFromSession';
 export default async function Page() {
   const user = await getUserFromSession();
 
+  if (!user) {
```
mattcasey, line 9: i don't think we should add this -  middleware.ts makes sure that user will exist or it redirects. But really, I would rather the page throw an error so we can detect it than the user just see a blank screen with no logs, nothing to help debug.

mattcasey (approved):
Looks good to me.. that was quite a refactor. How did you do this so fast. Cursor?

PR #4708, opened 2024-10-01
Nft metadata

motechFR:


apps/scoutgamecron/src/scripts/fixArtwork.ts
```diff
@@ -12,7 +13,7 @@ async function refreshArtworks() {
   const builderNfts = await prisma.builderNft.findMany({
     where: {
       season: currentSeason,
-      contractAddress: builderContractAddress
+      contractAddress: getBuilderContractAddress().toLowerCase()
```
mattcasey, line 16: Why do we need to lowercase this? the only reason we lowercased walelt addresses in webapp is to normalize them, honestly not sure we should have done that now looking back

.ebstalk.apps.env/scoutgame.env
```diff
@@ -28,4 +28,5 @@ GITHUB_ACCESS_TOKEN="{{pull:secretsmanager:/io.cv.app/prd/github:SecretString:sc
 
 BUILDER_SMART_CONTRACT_OWNER_PRIVKEY="{{pull:secretsmanager:/io.cv.app/prd/credentials:SecretString:builder_smart_contract_owner_privkey}}"
 
-REACT_APP_ENABLE_BUILDER_SMART_CONTRACT="true"
\ No newline at end of file
+NFT_ARTWORK_S3_PATH="prd-artworks"
```
mattcasey, line 31: "prd" should be enough? its not just artworks, we also put metadata in this folder for each token

packages/scoutgame/src/builderNfts/constants.ts
```diff
@@ -27,17 +25,9 @@ const devOptimismMainnetBuildersContract = '0x1d305a06cb9dbdc32e08c3d230889acb9f
...
-  ? realOptimismSepoliaBuildersContract
-  : devOptimismSepoliaBuildersContract;
-const optimismMainnetBuilderContractAddress = useRealContract
-  ? realOptimismMainnetBuildersContract
-  : devOptimismMainnetBuildersContract;
-
-export const builderContractAddress = useTestnets
-  ? optimismSepoliaBuilderContractAddress
-  : optimismMainnetBuilderContractAddress;
+export function getBuilderContractAddress(): `0x${string}` {
```
mattcasey, line 28: why is this a getter? i dont think u need to do that

mattcasey (approved):
left couple comments but code looks ok. thanks for adding the switch wallet to nft purchase

packages/scoutgame/src/builderNfts/constants.ts
motechFR: Make sure env() evaluates to something as react-env may not be loaded yet

.ebstalk.apps.env/scoutgame.env
motechFR: Fixed & also copied files to new destination

PR #4713, opened 2024-10-01
Show loading state while Rainbow kit loads

mattcasey:
This also fixes an issue when closing RainbowKit modal, you wouldn't be able to open it again because the ScoutButton's state was not being updated. (I really wish Rainbow kit offered event callbacks instead, which is why i'm using useEffect to watch for changes).

### DIFF
--- a/apps/scoutgame/components/common/ScoutButton/ScoutButton.tsx
+++ b/apps/scoutgame/components/common/ScoutButton/ScoutButton.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { Button } from '@mui/material';
+import { LoadingButton } from '@mui/lab';
 import { builderTokenDecimals } from '@packages/scoutgame/builderNfts/constants';
 import dynamic from 'next/dynamic';
 import { useRouter } from 'next/navigation';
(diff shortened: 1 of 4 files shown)


apps/scoutgame/components/common/ScoutButton/ScoutButton.tsx
```diff
@@ -21,21 +27,22 @@ export function ScoutButton({
...
-
   return (
-    <>
-      <Button fullWidth onClick={handleClick} variant='buy' data-test='scout-button'>
-        ${(Number(builder.price) / 10 ** builderTokenDecimals).toFixed(2)}
-      </Button>
-      <NFTPurchaseDialog open={isPurchasing} onClose={() => setIsPurchasing(false)} builder={builder} />
-    </>
+    <div>
+      <DynamicLoadingContext.Provider value={setDialogLoadingStatus}>
```
mattcasey, line 40: The amount of extra code is a little crazy for me, but the result is really nice - the button shows loading while it waits to download RainbowKit, which is >1MB of js

apps/scoutgame/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "declarationMap": true,
+    "declaration": true,
```
mattcasey, line 4: this enables us to cmd+click on things inside of Cursor. Also for some reason typechecking in VS Code was not working without this

PR #4724, opened 2024-10-02
Fix payment bugs

motechFR:

### DIFF
--- a/packages/scoutgame/src/builderNfts/mintNFT.ts
+++ b/packages/scoutgame/src/builderNfts/mintNFT.ts
@@ -2,9 +2,10 @@
 
 import { log } from '@charmverse/core/log';
 import { prisma } from '@charmverse/core/prisma-client';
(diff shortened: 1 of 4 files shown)


packages/scoutgame/src/builderNfts/mintNFT.ts
```diff
@@ -2,10 +2,11 @@
 
 import { log } from '@charmverse/core/log';
 import { prisma } from '@charmverse/core/prisma-client';
-import { builderNftChain } from '@packages/scoutgame/builderNfts/constants';
 import { refreshBuilderNftPrice } from '@packages/scoutgame/builderNfts/refreshBuilderNftPrice';
 import { currentSeason, getCurrentWeek } from '@packages/scoutgame/dates';
 
+import { refreshUserStats } from '../refreshUserStats';
```
mattcasey, line 8: please ermove this - and this method is used to generate seed data, do not use it on production

PR #4728, opened 2024-10-02
SG - Improvements - WIP

valentinludu:
Improvements:
- Card sizes on desktop carousel
- Add user context to prevent prop drilling
- Add Link component to navigation so we can prefetch pages on user hover
- revalidate all server data after purchasing an NFT
- SSR issues with useForm and useMediaQuery because the server didn't match the client


apps/scoutgame/components/layout/AppProviders.tsx
```diff
@@ -6,14 +6,19 @@ import { AppRouterCacheProvider } from '@mui/material-nextjs/v14-appRouter';
...
+import { UserProvider } from './UserProvider';
+
+export function AppProviders({ children, user }: { children: ReactNode; user: SessionUser | null }) {
   return (
     <AppRouterCacheProvider options={{ key: 'css' }}>
       <ThemeProvider theme={theme}>
         <CssBaseline enableColorScheme />
-        <WagmiProvider>{children}</WagmiProvider>
+        <WagmiProvider>
+          <UserProvider userSession={user}>{children}</UserProvider>
```
mattcasey, line 20: THANK YOU for adding this, i kept thinking it'd be a good idea

apps/scoutgame/components/layout/UserProvider.tsx
```diff
@@ -0,0 +1,35 @@
...
+  user: SessionUser | null;
+  updateUser: (updatedUser: Partial<SessionUser>) => void;
+};
+
+export const UserContext = createContext<Readonly<UserContext | null>>(null);
+
+export function UserProvider({ children, userSession }: { children: ReactNode; userSession: SessionUser | null }) {
+  const [user, setUser] = useState<SessionUser | null>(userSession);
+
+  const updateUser = useCallback((_updatedUser: Partial<SessionUser>) => {
```
mattcasey, line 18: You could throw an error here, incase someone tries to call it. "FIXME: User update is not implemented!"

apps/scoutgame/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx
```diff
@@ -70,7 +70,7 @@ export function NFTPurchaseForm(props: NFTPurchaseProps) {
 }
 
 export function NFTPurchaseFormContent({ builder }: NFTPurchaseProps) {
-  const { data: user, isLoading: loadingUser } = useGetUser();
+  const { user } = useUser();
```
mattcasey, line 73: thanks for fixing this! ALthough it is nice that we refresh the user points. Right now when you buy an NFT, the points and everything on the page are incorrect

mattcasey (approved):

apps/scoutgame/components/common/Button/IconButton.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
```
mattcasey, line 1: Are these 'use client's necessary? or just attempting to fix the crashes?

valentinludu: 'use client' is necessary when a client component is used inside a server component. Sometimes the app doesn't throw an error because the parent is a client component. But if used inside a server component it will throw an error.

apps/scoutgame/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx
valentinludu: I added the revalidatePath in the action that awaits the transaction

apps/scoutgame/components/layout/UserProvider.tsx
valentinludu: Added!

PR #4733, opened 2024-10-03
Fixed builder nfts sold stats

Devorein:
### WHY

[Post](https://discord.com/channels/894960387743698944/1291101673443426396)

### DIFF
--- a/apps/scoutgame/lib/scouts/getScoutStats.ts
+++ b/apps/scoutgame/lib/scouts/getScoutStats.ts
@@ -12,7 +12,8 @@ export async function getScoutStats(scoutId: string) {
           season: currentSeason
         },
         select: {
-          pointsEarnedAsScout: true
+          pointsEarnedAsScout: true,
+          nftsPurchased: true
         }
       },
       userAllTimeStats: {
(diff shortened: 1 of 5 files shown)


apps/scoutgame/lib/scouts/getScoutStats.ts
```diff
@@ -42,7 +39,7 @@ export async function getScoutStats(scoutId: string) {
   return {
     allTimePoints: scout.userAllTimeStats[0]?.pointsEarnedAsScout,
     seasonPoints: scout.userSeasonStats[0]?.pointsEarnedAsScout,
-    nftsPurchased: scout.nftPurchaseEvents.reduce((acc, curr) => acc + curr.tokensPurchased, 0),
-    buildersScouted: scout.nftPurchaseEvents.map((event) => event.builderNFT.builderId).length
+    nftsPurchased: scout.userSeasonStats[0]?.nftsPurchased,
+    buildersScouted: scout.nftPurchaseEvents.length
```
motechFR, line 43: We should make sure we dedupe by builderID, as you might purchase the same builder multiple times

motechFR (approved):
Looks good, but 1 item to fix

apps/scoutgame/lib/scouts/getScoutStats.ts
Devorein: Good point. Updated.

PR #4743, opened 2024-10-03
Fixed home page crash with back button

Devorein:

### DIFF
--- a/apps/scoutgame/components/home/components/BuildersCarousel/BuildersCarousel.tsx
+++ b/apps/scoutgame/components/home/components/BuildersCarousel/BuildersCarousel.tsx
@@ -1,9 +1,8 @@
 'use client';
 
-import { useEffect, useState } from 'react';
-
 import { Carousel } from 'components/common/Carousel/Carousel';
 import { LoadingCards } from 'components/common/Loading/LoadingCards';
+import { useIsMounted } from 'hooks/useIsMounted';
 import { useLgScreen, useMdScreen } from 'hooks/useMediaScreens';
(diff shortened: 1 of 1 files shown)


apps/scoutgame/components/home/components/BuildersCarousel/BuildersCarousel.tsx
```diff
@@ -1,16 +1,27 @@
...
 import type { BuilderInfo } from 'lib/builders/interfaces';
 
 import { BuilderCard } from '../../../common/Card/BuilderCard/BuilderCard';
 
 export function BuildersCarousel({ builders, userId }: { builders: BuilderInfo[]; userId?: string }) {
   const isMdScreen = useMdScreen();
   const isLgScreen = useLgScreen();
-
   const size = isLgScreen ? 'large' : isMdScreen ? 'small' : 'x-small';
+  const [isClientSide, setIsClientSide] = useState(false);
```
valentinludu, line 16: To make it easy

``` 
const isMounted = useIsMounted();

  if (!isMounted) {
    return null;
  }
```

valentinludu: And inside Carousel.tsx, you can delete this part since it doesn't do anything.

The LoadingCards is in suspense 2 levels up and server side is not rendering anything.

```
const [isClientSide, setIsClientSide] = useState(false);

  useEffect(() => {
    setIsClientSide(true);
  }, []);

  if (!isClientSide) {
    return <LoadingCards />;
  }
```

PR #4745, opened 2024-10-04
organize cron code and fix collabland

mattcasey:


apps/cron/src/tasks/processCollablandWebhookMessages/webhook/processWebhookMessage.ts
```diff
@@ -125,13 +127,15 @@ export async function processWebhookMessage(message: WebhookMessage): Promise<We
...
-  }
+  log.debug('Processing webhook message from Collabland', { message, event: data?.event, payload: data?.payload });
+  // API Tokens are not required for Collab.Land
+  // const hasPermission = await verifyWebhookMessagePermission(message);
+  // if (!hasPermission) {
+  //   return {
+  //     success: true,
+  //     message: 'Webhook message from Collab.Land without permission to be parsed.'
+  //   };
+  // }
```
mattcasey, line 138: i checked and it doesnt seem like we create api tokens when a space is connected via the public api/. but i'm logging the request in prod to see what they are sending in headers

lib/collabland/collablandClient.ts
```diff
@@ -146,13 +146,15 @@ export async function getDiscordUserState({
...
-    };
+  } catch (e) {
+    // 404 means Unknown Member
+    if ((e as any).status === 404) {
+      return {
+        isVerified: false,
+        roles: []
+      };
+    }
+    throw e;
```
mattcasey, line 157: fixes: we were returning "isVerified = false" if the request was 429 (rate limited)

PR #4746, opened 2024-10-04
Gift points during signup for waitlist users

Devorein:


mattcasey (approved):

apps/scoutgame/lib/users/findOrCreateUser.ts
```diff
@@ -67,6 +83,19 @@ export async function findOrCreateUser({
...
   });
+  if (waitlistRecord?.percentile) {
+    const tier = getTier(waitlistRecord.percentile);
+    await prisma.scout.update({
+      where: {
+        id: newScout.id
+      },
+      data: {
+        currentBalance: {
+          increment: waitlistTierPointsRecord[tier]
```
mattcasey, line 94: probabably should add 0 as a fallback just in case

PR #4754, opened 2024-10-07
SG: Added mixpanel events tracking

Devorein:


packages/mixpanel/package.json
mattcasey, line 1: new package - nice :)

packages/scoutgame/src/builderNfts/mintNFT.ts
```diff
@@ -172,6 +173,12 @@ export async function recordNftMint(params: MintNFTParams & { mintTxHash: string
     }
   });
   log.info('Minted NFT', { builderNftId, recipientAddress, tokenId: builderNft.tokenId, amount, userId: scoutId });
-
+  trackUserAction('nft_purchase', {
+    builderId: builderNft.builderId,
```
mattcasey, line 177: you need to pass in a userId here for it to connect to a profile. i'd also like to see what types u think are needed... 'tokenId' is kind of meaningless for Mixpanel. Really we probably just need the builder's name as an attribute of the nft u bought.

apps/scoutgame/lib/users/findOrCreateUser.ts
```diff
@@ -68,5 +74,19 @@ export async function findOrCreateUser({
...
 
+  trackUserAction('user_signup', {
+    userId: newScout.id,
+    fid: farcasterId,
+    username: userProps.username,
+    displayName: userProps.displayName
+  });
+
+  await updateUserProfile(newScout.id, {
+    userId: newScout.id,
```
mattcasey, line 85: why is userId stored on the profile object? isnt it already the distinct id?

mattcasey (changes requested):
- Left some comments about specific fields to include/exclude.

- For updating the user profiles, can we just set up a cron job that runs once a day for now? Then we don't have to worry about calling updateUserProfile in various places. We don't even need to run a script to update existing people.

Less important:
- I think adding types for the user profile and event meta we are tracking would be helpful. right now u could pass anything into 'updateUserProfile' for instance. I am fine if the mixpanel package is opinionated about scoutgame - we can always refactor later when we want to shaer the code wth other apps. but we may actually be putting this all in its own repo eventually anyway

packages/scoutgame/src/builderNfts/mintNFT.ts
Devorein: yeah I think so, we can also track the amount purchased to see if people are buying in bulk also whether its paid in points or not.

apps/scoutgame/lib/users/findOrCreateUser.ts
Devorein: ahh right

apps/scoutgamecron/src/scripts/createScoutMixpanelProfiles.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+  });
+
+  for (const scout of scouts) {
+    await updateUserProfile(scout.id, {
+      builderStatus: scout.builderStatus,
+      username: scout.username,
+      displayName: scout.displayName,
+      farcasterId: scout.farcasterId,
+      userId: scout.id,
```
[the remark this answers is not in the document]
Devorein: I don't see the point of adding onboarded and agreed to TOS since these will always be true. Unless we programmatically create the scouts and they still haven't logged in to scoutgame

mattcasey (approved):

PR #4755, opened 2024-10-07
Paginated scout page

Devorein:


apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -111,7 +114,15 @@ export async function getSortedBuilders({
           orderBy: {
             rank: 'asc'
           },
-          take: limit,
+          take: limit + 1,
```
mattcasey, line 117: why are you adding 1? should you be using skip: 1 instead?

apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -111,7 +114,15 @@ export async function getSortedBuilders({
...
+          cursor: cursor
+            ? {
+                userId_week: {
```
mattcasey, line 120: doesn't cursor need to be based on the sort of the query (rank in this case)?

apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -229,15 +248,15 @@ export async function getSortedBuilders({
...
 
     default:
       throw new Error(`Invalid sort option: ${sort}`);
   }
 
-  return builders;
+  const nextCursor = builders.length > limit ? builders[limit].id : null;
+  const hasMore = builders.length > limit;
+
+  return { builders: builders.slice(0, limit), nextCursor, hasMore };
```
mattcasey, line 261: this look wrong, that we are removing results - maybe you just need to use skip above.

mattcasey: There is also a util to do this called paginatedPrismaTask

apps/scoutgame/lib/builders/getSortedBuildersAction.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { currentSeason, getCurrentWeek } from '@packages/scoutgame/dates';
+import * as yup from 'yup';
+
+import { actionClient } from 'lib/actions/actionClient';
+import { getSortedBuilders } from 'lib/builders/getSortedBuilders';
+
+import type { BuilderInfo } from './interfaces';
+
+export const fetchMoreBuilders = actionClient
```
mattcasey, line 11: this action file should export a function of the same name

apps/scoutgame/components/scout/components/BuildersGalleryContainer.tsx
```diff
@@ -1,28 +1,90 @@
...
+
+    setIsLoading(true);
+    try {
+      const actionResponse = await fetchMoreBuilders({ sort, cursor: nextCursor });
+      if (actionResponse?.data) {
+        const { builders: newBuilders, nextCursor: newCursor } = actionResponse.data;
+        setBuilders((prev) => [...prev, ...newBuilders]);
+        setNextCursor(newCursor);
+      } else if (actionResponse?.serverError) {
+        // TODO: handle server error
```
mattcasey, line 43: at least log the error as a warning

apps/scoutgame/components/scout/components/BuildersGalleryContainer.tsx
```diff
@@ -1,28 +1,90 @@
...
+      }
+    } catch (error) {
+      // TODO: handle error
```
mattcasey, line 46: if it fails and we have builders we should show the user an error message of some kind

apps/scoutgame/components/scout/components/BuildersGalleryContainer.tsx
```diff
@@ -1,28 +1,90 @@
...
+    if (observerTarget.current) {
+      observer.observe(observerTarget.current);
+    }
+
+    return () => {
+      if (observerTarget.current) {
+        observer.unobserve(observerTarget.current);
+      }
+    };
+  }, [loadMoreBuilders]);
```
mattcasey, line 71: don't u need to add observerTarget into the dependencies here

mattcasey (changes requested):
i dont think cursor is being used properly, it should be a date value for 'new' and rank value for top and hot

PR #4766, opened 2024-10-08
Fix waitlist login

mattcasey:


motechFR (approved):
Looks good, although wondering what the edge case for 0 percentile is testing?

PR #4784, opened 2024-10-10
Nft/use hooks

motechFR:


apps/scoutgame/components/common/NFTPurchaseDialog/hooks/useDecentTransaction.tsx
```diff
@@ -0,0 +1,95 @@
...
+  tokensToPurchase: bigint;
+  scoutId: string;
+};
+
+async function prepareDecentTransaction({ txConfig }: { txConfig: BoxActionRequest }): Promise<BoxActionResponse> {
+  const DECENT_API_KEY = getDecentApiKey();
+
+  const response = await GET<BoxActionResponse>(
+    'https://box-v3-2-0.api.decent.xyz/api/getBoxAction',
+    { arguments: txConfig },
```
mattcasey, line 32: does'nt this need to be serialized? and if we use BigInt we need to transform it

packages/utils/src/http.ts
```diff
@@ -88,6 +88,8 @@ function _appendQuery(path: string, data: Params) {
       const value = data[key];
       return Array.isArray(value)
         ? `${value.map((v: string) => `${key}=${v}`).join('&')}`
+        : typeof value === 'object'
```
mattcasey, line 91: this may be too dangerous here, Dates are objects too, but we dont need to call JSON.stringify

apps/scoutgame/components/common/NFTPurchaseDialog/hooks/useDecentTransaction.tsx
motechFR: Moved the serialiser into this function so that it doesn't impact anything else that uses GET

PR #4801, opened 2024-10-14
only process nft transactions from this season

mattcasey:

### DIFF
--- a/apps/scoutgamecron/cron.yml
+++ b/apps/scoutgamecron/cron.yml
@@ -13,7 +13,7 @@ cron:
   - name: "process-nft-mints"
     url: "/process-nft-mints"
     # Every 5 minutes
-    schedule: "*/5 * * * *"
+    schedule: "*/20 * * * *"
 
   - name: "update-mixpanel-user-profiles"
     url: "/update-mixpanel-user-profiles"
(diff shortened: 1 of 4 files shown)


packages/scoutgame/src/builderNfts/handlePendingTransaction.ts
```diff
@@ -26,8 +26,8 @@ export async function handlePendingTransaction({
   // Atomically set the status to 'processing' only if it's currently 'pending'
   const updatedTx = await prisma.pendingNftTransaction.updateMany({
     where: {
-      id: pendingTransactionId
-      // status: 'pending'
+      id: pendingTransactionId,
+      status: 'pending'
```
mattcasey, line 30: i committed this comment outside of a PR and i don't know why. my commit message was "to fix stats" from 2 weeks ago. i dont think it necessarily matters if this is here or not, but seems like it should be here

apps/scoutgamecron/cron.yml
```diff
@@ -13,7 +13,7 @@ cron:
   - name: "process-nft-mints"
     url: "/process-nft-mints"
     # Every 5 minutes
-    schedule: "*/5 * * * *"
+    schedule: "*/20 * * * *"
```
mattcasey, line 16: process was taking at least 8 minutes, so move to 20 to be safe

motechFR (approved):
Looks good 👍

PR #4807, opened 2024-10-15
Add nft count on scouted by user page

valentinludu:
[Card](https://app.charmverse.io/charmverse/fix-nft-counts-for-scout-profile-29919213251814325)

### DIFF
--- a/apps/scoutgame/lib/scouts/getScoutedBuilders.ts
+++ b/apps/scoutgame/lib/scouts/getScoutedBuilders.ts
@@ -88,8 +88,7 @@ export async function getScoutedBuilders({ scoutId }: { scoutId: string }): Prom
       nftsSoldToScout,
       isBanned: builder.builderStatus === 'banned',
       price: builder.builderNfts[0]?.currentPrice ?? 0,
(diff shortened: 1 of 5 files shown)


apps/scoutgame/lib/scouts/getScoutedBuilders.ts
```diff
@@ -88,8 +88,7 @@ export async function getScoutedBuilders({ scoutId }: { scoutId: string }): Prom
       nftsSoldToScout,
       isBanned: builder.builderStatus === 'banned',
       price: builder.builderNfts[0]?.currentPrice ?? 0,
-      boughtNftsCount: undefined
-      // boughtNftsCount: builder.builderNfts[0]?._count?.nftSoldEvents ?? 0
+      boughtNftsCount: builder.builderNfts[0]?._count?.nftSoldEvents ?? 0
```
mattcasey, line 91: i think u could also just use nftsSoldToScout. or replace nftsSoldToScout with this value

valentinludu: I removed it since it is the same thing

PR #4815, opened 2024-10-16
SG info UI

valentinludu:
- SG info sidebar menu  for desktop and sidebar drawer for mobile
- split info in multiple pages
- update meta for username page
- Header pill  with points and avatar is now clickable
- updated mui (for some reason the theme wasn't loading correctly and I think this fixed it)
- updated next (fixes dynamic imports issues)
- center rewards if only 1  in a notification


mattcasey (changes requested):
can you make it so the sidebar doesn't fade in and out while navigating between info pages? Here's a loom: https://www.loom.com/share/52c5a35688bb4de4908f1f0ed157342a

mattcasey (approved):

PR #4820, opened 2024-10-16
Fix top scouts number of nfts held

valentinludu:


apps/scoutgame/lib/scouts/getTopScouts.ts
```diff
@@ -55,7 +63,7 @@ export async function getTopScouts({ limit }: { limit: number }): Promise<TopSco
       const buildersScouted = Array.from(
         new Set(scout.user.nftPurchaseEvents.map((event) => event.builderNFT.builderId))
       ).length;
-      const nftsHeld = scout.user.nftPurchaseEvents.length;
+      const nftsHeld = scout.user.userSeasonStats[0]?.nftsPurchased || 0;
```
motechFR, line 66: Could we please check that the code that updates this value is correct?

motechFR (approved):
Approved, can you double check the user season stats are counted correctly?

One purchase might have multiple tokens

apps/scoutgame/lib/scouts/getTopScouts.ts
mattcasey: we have a test on recordNftMint

PR #4844, opened 2024-10-18
Scout game user optimisations

valentinludu:
- make user provider more practical
- Cache all scout requests just for the url the user is requesting. Every time the url is changed, the cache will be invalid and cached again for the next requests.


Devorein (approved):
Awesome refactor 🚀

PR #4846, opened 2024-10-18
Fix metadata for user page

valentinludu:


mattcasey (approved):

apps/scoutgame/lib/users/getUserByPath.ts
```diff
@@ -24,3 +25,5 @@ export async function getUserByPath(
     githubLogin: user?.githubUser[0]?.login
   };
 }
+
+export const getUserByPath = cache(getUserByUsernamePath);
```
mattcasey, line 29: interesting.. but wont we have to make sure to clear this cache now on various events?

PR #4857, opened 2024-10-21
Catch errors in react server components

valentinludu:
I realised that we didn't have any error handling mechanism in server components and presume all errors are catched by the global-error.ts file which logs the error as it is.

I was inspired by this ECMAScript proposal which handles errors dead simple and build an utility function that resembles this behaviour.
[Insipiration link](https://github.com/arthurfiorette/proposal-safe-assignment-operator?tab=readme-ov-file)

In essence we should catch server side errors while the page is loading and show the user an error or at least some part of the app instead of throwing the lightning bolt at him and maybe logging the error. (As you know if you are using brave, sending analytical data is blocked).


apps/scoutgame/components/common/ErrorSSRMessage.tsx
```diff
@@ -0,0 +1,9 @@
+import { Alert } from '@mui/material';
+
+export function ErrorSSRMessage({ message }: { message?: string }) {
+  return (
```
mattcasey, line 4: suggestion: add a log here that we showed an ssr message

valentinludu: Like [this](https://github.com/charmverse/app.charmverse.io/pull/4857/commits/432a10f08a0f7d0486690d3ac9fbb78bd3827fe0) ?

mattcasey (approved):

PR #4868, opened 2024-10-22
Use a global provider for optimistic purchases

valentinludu:
### WHAT
After purchase, show the success view and send the user a message when the transaction succeeded or failed on the blockchain.
If the user refreshes the page, we get all pending transactions and show a success message when they are done.
Update server components with import 'server-only'
e2e included

### WHY
We needed a way to optimistically update a user purchase


.env.test.local.example
```diff
@@ -19,6 +19,6 @@ CREDENTIAL_WALLET_KEY=d093b8a44723eac658b4888743f945ff2066f1caba8c460f565002f998
 # Make playwright show color output when CI=true - https://github.com/microsoft/playwright/issues/28260
 FORCE_COLOR=1
 
-REACT_APP_BUILDER_NFT_CONTRACT_ADDRESS=0x1d305a06cb9dbdc32e08c3d230889acb9fe8a4dd
+REACT_APP_BUILDER_NFT_CONTRACT_ADDRESS=0x2cba9c6e0c14da826b0ec689cabf02a6f6b9808e
```
mattcasey, line 22: this is not necessary right? we're not calling a real contract?

mattcasey (approved):

apps/scoutgame/playwright.config.ts
```diff
@@ -13,13 +13,13 @@ import { devices } from '@playwright/test';
 const config: PlaywrightTestConfig = {
   testDir: './__e2e__',
   /* Maximum time one test can run for. */
-  timeout: 10 * 1000,
+  timeout: 10 * 10000,
```
mattcasey, line 16: 100 seconds is too long

mattcasey (changes requested):
please clean up the test config

apps/scoutgame/playwright.config.ts
valentinludu: I reverted everything from this file

.env.test.local.example
valentinludu: I reverted!

PR #4876, opened 2024-10-23
filter hidden blocks and properties if you do not have edit access to a proposal source board

mattcasey:
I noticed that for public boards, we are sharing all the blocks even if you filtered them. This PR utilizes the logic for a "locked" board to reduce the data we return if you also just don't have edit permission.


motechFR (approved):
Unit test would be good here

PR #4883, opened 2024-10-24
Shrink scout card username

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/sg-scout-names-go-beyond-the-box-1599089796442068)

image


apps/scoutgame/components/common/Card/ScoutCard.tsx
```diff
@@ -8,12 +9,44 @@ export type ScoutInfo = BasicUserInfo & {
   nfts: number;
 };
 
+function useDynamicFontSize(text: string, maxWidth: number, minFontSize: number, maxFontSize: number) {
```
Devorein, line 12: Was not able to find a better way to achieve this using native CSS :(

PR #4885, opened 2024-10-24
Proposals/selective publishing

motechFR:


components/proposals/ProposalPage/components/ProposalEvaluations/components/Review/EvaluationsReview.tsx
```diff
@@ -41,6 +41,7 @@ export type Props = {
     ProposalWithUsersAndRubric,
     | 'archived'
     | 'id'
+    | 'createdBy'
```
mattcasey, line 44: why do we care who created this? i'd much rather just use 'authors' which is easy to explain and debug

lib/proposals/mapDbProposalToProposal.ts
```diff
@@ -85,6 +90,8 @@ export function mapDbProposalToProposal({
       })
     : null;
 
+  const isAuthor = !!userId && proposal.createdBy === userId && proposal.authors.some((a) => a.userId === userId);
```
mattcasey, line 93: this requires you to not only create the proposal but also be in the list of authors? i think you meant OR? this is exactly my point above: by adding a second place to look up we can introduce bugs like this

lib/proposals/mapDbProposalToProposal.ts
```diff
@@ -102,9 +109,18 @@ export function mapDbProposalToProposal({
...
+      const showReviewsToAuthor = showRubricAnswersToAuthor({
+        evaluationType: evaluation.type as ProposalEvaluationType,
+        isAuthor,
+        proposalFailed: !!proposalFailed,
+        isCurrentEvaluationStep: evaluation.id === currentEvaluation?.id,
+        showAuthorResultsOnRubricFail: !!evaluation.showAuthorResultsOnRubricFail
+      });
       draftRubricAnswers = [];
-      if (!evaluation.shareReviews) {
+      if (!evaluation.shareReviews && !showReviewsToAuthor) {
```
mattcasey, line 123: needs to consider if this is the failed step or not

PR #4929, opened 2024-10-30
SG: Wallet signin support

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/add-wallet-sign-in-10943231655417174)


apps/scoutgame/lib/users/queries.ts
```diff
@@ -11,7 +11,8 @@ export const BasicUserInfoSelect = {
       login: true
     }
   },
-  builderStatus: true
+  builderStatus: true,
+  farcasterUsername: true
```
mattcasey, line 15: This is not a field that should be on a global type. we do not need your farcaster name for the list of hot builders, builder activities or scouts, etc. we only need farcastername in one place: for your profile. so i only think you need to add it to the return type of getUserByPath, and maybe the SessionUser type

apps/scoutgame/lib/users/interfaces.ts
```diff
@@ -5,6 +5,7 @@ export type MinimalUserInfo = {
   displayName: string;
   path: string;
   avatar?: string | null;
+  farcasterUsername?: string | null;
```
mattcasey, line 8: this is not "minimal" info

mattcasey (changes requested):

apps/scoutgame/lib/utils/generateRandomAvatar.ts
mattcasey, line 1: the SVG should be loaded asyncronously really. there's no reason to put this much into the typescript

apps/scoutgame/lib/users/findOrCreateUser.ts
```diff
@@ -39,6 +41,8 @@ export async function findOrCreateUser({
   bio?: string;
   displayName: string;
   path: string;
+  farcasterUsername?: string;
+  farcasterName?: string;
```
mattcasey, line 45: remove farcasterUsername

apps/scoutgame/lib/blockchain/config.ts
```diff
@@ -1,5 +1,8 @@
 import * as yup from 'yup';
 
-export const authSchema = yup.object({ message: yup.string().required(), signature: yup.string().required() });
+export const loginWithWalletSchema = yup.object({
```
mattcasey, line 3: Not a big deal but we should rename this file. its not a general blockchain config, its just a schema that belongs to the verifyWallet method.

apps/scoutgame/lib/users/generateUserPath.ts
```diff
@@ -0,0 +1,12 @@
+import { v4 } from 'uuid';
+
+import { getUserByPath } from './getUserByPath';
+
+export async function generateUserPath(displayName: string) {
```
mattcasey, line 5: this should go into the package for scoutgame as it will be needed for scoutgameadmin

apps/scoutgame/lib/users/generateUserPath.ts
```diff
@@ -0,0 +1,12 @@
...
+  const path = displayName.toLowerCase().replace(/[^a-z0-9]+/g, '-');
```
mattcasey, line 6: we should make sure the path doesn't end or start with a hyphen, and it would be ideal if you checked for 2+ hyphens

apps/scoutgame/lib/users/findOrCreateUser.ts
```diff
@@ -65,8 +69,21 @@ export async function findOrCreateUser({
...
-  if (userProps?.avatar) {
+  if (!userProps?.avatar) {
+    const randomAvatarSvg = generateRandomAvatar();
+    const avatarFile = Buffer.from(randomAvatarSvg);
+    const pathInS3 = getUserS3FilePath({ userId, url: 'avatar.svg' });
+    try {
+      const { fileUrl } = await uploadFileToS3({
+        pathInS3,
+        content: avatarFile,
+        contentType: 'image/svg+xml'
```
mattcasey, line 80: dont use SVG for these. we should convert to PNG

apps/scoutgame/lib/builders/setupBuilderProfile.ts
```diff
@@ -130,7 +130,7 @@ export async function setupBuilderProfile({
     });
   }
 
-  if (inviteCode !== null) {
+  if (inviteCode !== null && isProdEnv) {
```
mattcasey, line 133: this should work for dev and staging env also

apps/scoutgame/lib/users/queries.ts
Devorein: good point. Updated

mattcasey (approved):

PR #4930, opened 2024-10-30
Simplify recordNftMint to with and without purchase

motechFR:


mattcasey (approved):

mattcasey (changes requested):
can it actually be recordNftMint and recordNftMintAndRefresh. Our methods should have less side effects by default and compose more actions

packages/scoutgame/src/builderNfts/recordNftMint.ts
```diff
@@ -155,41 +182,17 @@ export async function recordNftMintWithoutRefresh(params: MintNFTParams & { crea
...
-    where: {
-      txHash: mintTxHash
-    }
-  });
+  log.info('Minted NFT', { builderNftId, recipientAddress, tokenId: builderNft.tokenId, amount, userId: scoutId });
 
-  if (existingTx) {
-    log.warn(`Tried to record duplicate tx ${mintTxHash}`, { params, existingTx });
-    return;
+  if (!skipMixpanel) {
```
mattcasey, line 187: Please pick one pattern... for instance why not make "refresh" another skip option? or why not add a "recordAndSendMixpanel"

mattcasey (approved):

PR #4932, opened 2024-10-31
Implement merkle trees for claiming

motechFR:
### WHAT

This PR introduces connectivity to the new testnet protocol contracts.

The goal is to run the process in parallel so that we can test it based on real data.

Todo
- Issue the merkle roots in scout points
- Add a demo claiming button
- Add an ability to prepare an "upgrade contract" Gnosis Safe transaction

Screenshot 2024-11-01 at 22 05 38


packages/scoutgame/src/dates.ts
```diff
@@ -103,3 +103,19 @@ export function getSeasonWeekFromISOWeek({ season, week }: { season: ISOWeek; we
   const weeksDiff = weekDate.diff(seasonDate, 'weeks').weeks;
   return weeksDiff + 1;
 }
+
+export function getAllISOWeeksFromSeasonStart(): string[] {
+  const seasonOneStart = '2024-W41';
+  const start = getStartOfSeason(seasonOneStart);
+  const end = DateTime.now();
```
mattcasey, line 110: you should always use DateTime.utc() or this may produce different results in people's dev env

packages/scoutgame/src/dates.ts
```diff
@@ -103,3 +103,19 @@ export function getSeasonWeekFromISOWeek({ season, week }: { season: ISOWeek; we
   const weeksDiff = weekDate.diff(seasonDate, 'weeks').weeks;
   return weeksDiff + 1;
 }
+
+export function getAllISOWeeksFromSeasonStart(): string[] {
+  const seasonOneStart = '2024-W41';
```
mattcasey, line 108: This must be fixed, do NOT hardcode seasons anywhere. you should pass the season in. A test on this method would be good too

package.json
```diff
@@ -260,6 +260,7 @@
     "graphql": "^16.8.1",
     "hot-shots": "^9.0.0",
     "html-to-text": "^8.2.0",
+    "i": "^0.3.7",
```
mattcasey, line 263: typo

.ebstalk.apps.env/scoutgameadmin.env
```diff
@@ -12,11 +12,22 @@ ALCHEMY_API_KEY="{{pull:secretsmanager:/io.cv.app/prd/alchemy:SecretString:alche
...
 SCOUTGAME_CONTRACT_NAME="prd"
+SCOUTGAME_S3_BUCKET="scoutgame.public"
 S3_UPLOAD_REGION="{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:region}}"
 S3_UPLOAD_SECRET="{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:access_key_secret}}"
-S3_UPLOAD_KEY="{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:access_key}}"
\ No newline at end of file
+S3_UPLOAD_KEY="{{pull:secretsmanager:/io.cv.app/prd/aws/s3:SecretString:access_key}}"
+
+# Scout protocol season 2
+SCOUTGAME_METADATA_PATH_PREFIX="prd"
```
mattcasey, line 26: i dont see this being used anywhere?

apps/scoutgameadmin/hooks/useProposeContractUpgrade.tsx
```diff
@@ -0,0 +1,259 @@
...
+//       return 'https://safe-transaction-goerli.safe.global/';
+//     case 137:
+//       return 'https://safe-transaction-mainnet.safe.global/';
+//     // Add other chains as needed
+//     default:
+//       throw new Error(`Unsupported chain ID: ${chainId}`);
+//   }
+// };
+
+// export default useProposeSetImplementation;
```
mattcasey, line 259: why is this commented out? please remove and maybe make a second PR if it's a WIP?

mattcasey: also hook files should not export a default

tsconfig.json
```diff
@@ -65,6 +65,7 @@
       "@charmverse/core/proposals": ["node_modules/@charmverse/core/dist/cjs/proposals"],
       "@charmverse/core/bounties": ["node_modules/@charmverse/core/dist/cjs/bounties"],
       "@charmverse/core/utilities": ["node_modules/@charmverse/core/dist/cjs/utilities"],
+      "@charmverse/core/protocol": ["node_modules/@charmverse/core/dist/cjs/protocol"],
```
mattcasey, line 68: is this actually necessary? if we're only importing from apps and pacakges then we should not need it here

packages/scoutgameattestations/src/getAttestationMetadataS3Path.ts
```diff
@@ -0,0 +1,30 @@
+const SCOUTGAME_S3_BUCKET = process.env.SCOUTGAME_S3_BUCKET ?? 'scoutgame.public';
```
mattcasey, line 1: i dont think we should have a default bucket, that seems dangerous?

packages/scoutgameattestations/src/constants.ts
```diff
@@ -0,0 +1,23 @@
+import env from '@beam-australia/react-env';
+import { baseSepolia } from 'viem/chains';
+
+export const scoutGameAttestationChain = baseSepolia;
+
+export const scoutGameAttestationChainId = scoutGameAttestationChain.id;
```
mattcasey, line 6: yuou should export the RPC URL here instead of using getChainById elsewhere to grab the rpcURL. The latter method ends up importing a whole lot of chain configs just to get a single URL

packages/scoutgameattestations/src/attestOnchain.ts
```diff
@@ -0,0 +1,133 @@
...
+    .then((tx) => tx.wait());
+
+  if (onAttestSuccess) {
+    for (let i = 0; i < attestationUids.length; i++) {
+      await onAttestSuccess({ attestationUid: attestationUids[i], data: records[i].data, index: batchStartIndex + i });
+    }
+  }
+
+  attestationLogger.info(
+    `Issued ${attestationUids.length} attestations for schema ${schemaId} on chain ${scoutGameAttestationChainId} with uids: ${attestationUids.join(', ')}`,
```
mattcasey, line 125: would be better to put the uids in metadata so the logs dont get very long

packages/scoutgameattestations/src/attestOnchain.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  return attestationUid;
+}
+
+const maxPerBatch = 30;
+
+export async function multiAttestOnchain(params: {
+  schemaId: string;
+  records: Omit<ScoutGameAttestationInput, 'schemaId'>[];
+  onAttestSuccess?: (input: { attestationUid: string; data: `0x${string}`; index: number }) => Promise<void>;
```
mattcasey, line 75: you're only using this "onAttestSuccess" in one place. it would be a more familiar pattern if u just returned a list of attestations to process afterwards.

packages/scoutgameattestations/src/attestOnchain.ts
```diff
@@ -0,0 +1,133 @@
...
+  batchStartIndex?: number;
+}): Promise<`0x${string}`[]> {
+  if (params.records.length === 0) {
+    return [];
+  }
+
+  if (params.records.length > maxPerBatch) {
+    const allUids: `0x${string}`[] = [];
+    for (let i = 0; i < params.records.length; i += maxPerBatch) {
+      const uids = await multiAttestOnchain({
```
mattcasey, line 85: i would like to see some kind of test that this doesnt end up in a recursive loop

apps/scoutgameadmin/lib/contract/aggregateProtocolData.ts
```diff
@@ -0,0 +1,100 @@
...
+              returnValue.testClaim = {
+                claim: userClaim,
+                proofs
+              };
+            }
+          }
+
+          return returnValue;
+        })
+        .catch(() => {
```
mattcasey, line 79: why is an error ok? we should at least be logging it

packages/scoutgame/src/protocol/generateWeeklyClaims.ts
```diff
@@ -0,0 +1,221 @@
...
+  week: string;
+}): Promise<{ weeklyClaims: WeeklyClaimsTyped; totalBuilders: number; totalPoints: number }> {
+  const existingClaim = await prisma.weeklyClaims.findUnique({
+    where: {
+      week
+    }
+  });
+
+  if (existingClaim) {
+    throw new Error(`Claims for week ${week} already exist`);
```
mattcasey, line 167: this is an expected case becase you're running the job once an hour. it should not log an error

packages/scoutgameattestations/src/constants.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const SCOUTGAME_METADATA_PATH_PREFIX = process.env.SCOUTGAME_METADATA_PATH_PREFIX || 'dev';
```
mattcasey, line 8: i mentioned above, i dont think this is used

packages/scoutgameattestations/src/logger.ts
```diff
@@ -0,0 +1,3 @@
+import { getLogger } from '@charmverse/core/log';
+
+export const attestationLogger = getLogger('scoutgame_attestations');
```
mattcasey, line 3: i like this approach to provide a special logger

packages/scoutgame/src/users/findOrCreateUser.ts
```diff
@@ -9,7 +9,7 @@ import {
   getFilenameWithExtension
 } from '@packages/aws/uploadToS3Server';
 import { getENSName } from '@packages/blockchain/getENSName';
-import { capitalize } from '@root/lib/utils/strings';
+import { capitalize } from '@packages/utils/strings';
```
mattcasey, line 12: thx for the cleanup

apps/scoutgameadmin/components/providers/wagmi/WagmiProvider.tsx
mattcasey, line 1: ALL folders inside components either belong in a path-related folder or "common". There is no need for exceptions here.. and i think putting wallet-related stuff inside the contract component folder would be fine. there's no plan to add wallet-related actions anyplace else at the moment

apps/scoutgamecron/src/tasks/issueGemsOnchain/index.ts
```diff
@@ -0,0 +1,6 @@
+import { attestGemReceipts } from '@packages/scoutgameattestations/attestGemReceipts';
+import type Koa from 'koa';
+
+export async function issueGemsOnchain(ctx: Koa.Context | null) {
+  await attestGemReceipts();
```
mattcasey, line 5: all the code from attestGemReceipts belongs inside of this task folder. IT does not belong in a special package

packages/scoutgameattestations/src/attestGemReceipts.ts
```diff
@@ -0,0 +1,211 @@
...
+import {
+  scoutGameAttestationChainId,
+  scoutGameContributionReceiptSchemaUid,
+  scoutGameUserProfileSchemaUid
+} from './constants';
+import { attestationLogger } from './logger';
+import { uploadContributionReceiptToS3 } from './uploadContributionReceiptToS3';
+import { uploadScoutProfileToS3 } from './uploadScoutProfileToS3';
+
+const minimumGemsDate = new Date('2024-11-04T00:00:00Z');
```
mattcasey, line 15: where is this date from?

PR #4935, opened 2024-11-01
SG: Editable user profile details

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/edit-avatar-and-name-from-profile-1945025184458542)


mattcasey (changes requested):
image

- change the white icon background - it should be grey or something
- the pencil icon should be directly to the right of your name, and so to the left of social icons
- when editing username, there should be an X icon to cancel

PR #4936, opened 2024-11-01
Added banner on cv spaces for scoutgame

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/show-scout-game-marketing-banner-on-charmverse-2774960321764366)


mattcasey (approved):

valentinludu (commented):
Just FYI: This will be visible only for users with a role in that space. That means, if there are users that are just browsing a public space, they will not see it.

PR #4939, opened 2024-11-01
Congrats og image with nft and stats

valentinludu:
[Card](https://app.charmverse.io/charmverse/builder-profile-share-frame-image-4116717470176401)

Info:
- Need to use @vercel/og instead of next/og since we use it outside of a nextjs app
- install @vercel/og globally since we need it in different places


packages/scoutgame/src/builders/getBuilderActivities.ts
```diff
@@ -0,0 +1,123 @@
...
+  limit = 10
+}: {
+  builderId?: string;
+  limit: number;
+}): Promise<BuilderActivity[]> {
+  const builderEvents = await prisma.builderEvent.findMany({
+    where: {
+      builder: {
+        id: builderId,
+        builderStatus: 'approved'
```
Devorein, line 38: this should also consider the banned user since they are also approved

packages/scoutgame/src/builderNfts/artwork/components/BuilderShareImage.tsx
```diff
@@ -0,0 +1,285 @@
...
+  const overlays = [
+    { name: 'red', src: `/images/profile/builder/red-sky.jpg`, color: '#661933' },
+    { name: 'purple', src: '/images/profile/builder/purple-sky.jpg', color: '#190D4D' },
+    { name: 'black', src: '/images/profile/builder/black-sky.jpg', color: '#1B1B1B' },
+    { name: 'blue', src: '/images/profile/builder/blue-sky.jpg', color: '#02143C' }
+  ];
+
+  const domain = 'http://localhost:3000' || process.env.DOMAIN;
+
+  const random = randomInt(0, 4);
```
Devorein, line 36: should we be using a deterministic way to get the overlays? We went with random for the nft image and that caused issues later

Devorein (approved):

packages/scoutgame/src/builders/getBuilderActivities.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  });
+
+  return builderEvents
+    .map((event) => {
+      if (event.type === 'nft_purchase' && event.nftPurchaseEvent) {
+        return {
+          ...event.builder,
+          path: event.builder.path!,
```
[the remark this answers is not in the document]
valentinludu: Done!

packages/scoutgame/src/builders/getBuilderActivities.ts
valentinludu: This is a method I just copied from main app because I need it to be reusable. I don't think it matters for a congrats image if the user is banned or not. All users that are shown in the main app should have a congrats image and stats.

packages/scoutgame/src/builderNfts/artwork/components/BuilderShareImage.tsx
valentinludu: I don't see a way of doing that rn. I think the best solution is to bring it up in our meeting today and see if there is a business requirement for it.

PR #4950, opened 2024-11-04
Make sure we look at the correct season when retrieving builder stats + test

mattcasey:


apps/scoutgame/lib/builders/getSortedBuilders.ts
```diff
@@ -67,7 +67,8 @@ export async function getSortedBuilders({
             },
             userWeeklyStats: {
               where: {
-                week
+                week,
+                season
```
mattcasey, line 71: also make sure we look at the right season!

PR #4955, opened 2024-11-05
Use the standard folder for congrats artwork

mattcasey:
We use the same bucket for prod and dev right now, so any s3 path should include something related to the env. For NFTs, we include a contract name/address, in this case "prd".


valentinludu (approved):
Looks good

PR #4957, opened 2024-11-05
Scoutgame home scrollbar and fixed tabs

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/homepage-scrolling-45966922568953517)


mattcasey (changes requested):
the top nav should be sticky on all pages

PR #4958, opened 2024-11-05
Autoplay hot builder carousel and updated hot builders query

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/hot-builder-automated-scrolling-686778878621773)


mattcasey (approved):
Looks good but i noticed a bug (which is in prod too). If I resize my screen to mobile and then stretch back to desktop, the left/right arrows do not do anything

PR #4961, opened 2024-11-06
Scoutgame telegram quest and claim page

Devorein:


apps/scoutgame/components/login/LaunchDate.tsx
```diff
@@ -2,6 +2,7 @@
 
 import { Box, Stack, Typography } from '@mui/material';
 import { tierDistributionMap } from '@packages/scoutgame/waitlist/scoring/constants';
+import { timeUntilFuture } from '@packages/utils/time';
```
mattcasey, line 5: anything time-related belongs in utils/dates

apps/scoutgame/theme/theme.tsx
```diff
@@ -60,7 +61,8 @@ const themeOptions: Parameters<typeof createTheme>[0] = {
         },
         secondary: {
           main: secondaryText,
-          light: secondaryLightText
+          light: secondaryLightText,
+          dark: secondaryDarkText
```
mattcasey, line 65: suggestion: use darken() from @mui/material/styles, so that we dont need so many hard-coded hex colors

apps/scoutgametelegram/app/quests/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import { QuestsPage } from 'components/quests/QuestsPage';
+import { getDailyClaims } from 'lib/claims/getDailyClaims';
+import { getUserFromSession } from 'lib/session/getUserFromSession';
+
+export default async function Quests() {
+  const user = await getUserFromSession();
+
+  if (!user) {
+    return notFound();
```
mattcasey, line 11: this should not be a notfound error, but an auth error

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
```diff
@@ -0,0 +1,54 @@
...
+export type DailyClaim = {
+  date: Date;
+  day: number;
+  claimed: boolean;
+  isBonus?: boolean;
+};
+
+export async function getDailyClaims(userId: string): Promise<DailyClaim[]> {
+  const currentWeek = getCurrentWeek();
+  const pointsReceipts = await prisma.pointsReceipt.findMany({
```
mattcasey, line 13: you should query on builderEvents, not pointReceipts.  the only data you are looking at is 'createdAt' which also exists in builderEvents.

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
```diff
@@ -0,0 +1,54 @@
...
+    where: {
+      recipientId: userId,
+      event: {
+        type: 'daily_claim',
+        week: currentWeek
+      }
```
mattcasey, line 19: this query should just query the time window of events that it needs to look at

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
```diff
@@ -0,0 +1,54 @@
...
+import { getCurrentWeek, getDateFromISOWeek, getWeekStartEnd, isToday } from '@packages/scoutgame/dates';
+
+export type DailyClaim = {
+  date: Date;
+  day: number;
+  claimed: boolean;
+  isBonus?: boolean;
+};
+
+export async function getDailyClaims(userId: string): Promise<DailyClaim[]> {
```
mattcasey, line 11: this method needs a test

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
```diff
@@ -0,0 +1,54 @@
...
+      const dailyClaimInfo = {
+        date: date.toJSDate(),
+        day: index + 1,
+        claimed: !!receipt,
+        isBonus: false
+      };
+
+      if (index === 6) {
+        const bonusReceipt = pointsReceipts.find(
+          (_receipt) => isToday(_receipt.createdAt, date) && _receipt.value === 3
```
mattcasey, line 46: dont trust the value to assume the bonus. you should add a new type like "daily_claim_streak" to builder event type

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
```diff
@@ -0,0 +1,54 @@
...
+        );
+        return [dailyClaimInfo, { ...dailyClaimInfo, claimed: !!bonusReceipt, isBonus: true }];
```
mattcasey, line 48: An array of arrays is a confusing return type for this. It requires adding "isBonus" to every claim, as well as every claim is an array when only the last day would ever have two results.

Two other possibilities:
```
{ dailyClaims: DailyClaim[], bonusStreak: { claimed: boolean } }
```
or a dictionary:
```
{ 0: DailyClaim, 1: DailyClaim, ..., bonusStreak: { claimed: boolean } }
```

apps/scoutgametelegram/theme/theme.tsx
```diff
@@ -60,7 +61,8 @@ const themeOptions: Parameters<typeof createTheme>[0] = {
         },
         secondary: {
           main: secondaryText,
-          light: secondaryLightText
+          light: secondaryLightText,
+          dark: secondaryDarkText
```
mattcasey, line 65: same comment above.. maybe use darken()?

apps/scoutgametelegram/tsconfig.json
```diff
@@ -21,27 +26,74 @@
...
-      "@root/abis/*": ["../../abis/*"],
-      "@root/lib/*": ["../../lib/*"],
-      "@root/charmClient/hooks/helpers": ["../../charmClient/hooks/helpers"],
-      "@charmverse/core/prisma": ["../../node_modules/@charmverse/core/dist/cjs/prisma"],
-      "@charmverse/core/prisma-client": ["../../node_modules/@charmverse/core/dist/cjs/prisma-client"],
-      "@charmverse/core/log": ["../../node_modules/@charmverse/core/dist/cjs/lib/log"],
-      "@charmverse/core/http": ["../../node_modules/@charmverse/core/dist/cjs/http"],
-      "@charmverse/core/utilities": ["../../node_modules/@charmverse/core/dist/cjs/utilities"],
-      "@charmverse/core/errors": ["../../node_modules/@charmverse/core/dist/cjs/errors"]
+      "@connect-shared/*": [
```
mattcasey, line 29: this is incorrect formatting

packages/utils/package.json
```diff
@@ -16,6 +16,7 @@
     "./sleep": "./src/sleep.ts",
     "./env": "./src/env.ts",
     "./strings": "./src/strings.ts",
-    "./uuid": "./src/uuid.ts"
+    "./uuid": "./src/uuid.ts",
+    "./time": "./src/time.ts"
```
mattcasey, line 20: we already export dates

public/mockServiceWorker.js
```diff
@@ -8,123 +8,123 @@
  * - Please do NOT serve this file on production.
  */
 
-const PACKAGE_VERSION = '2.2.14';
-const INTEGRITY_CHECKSUM = '26357c79639bfa20d64c0efca2a87423';
-const IS_MOCKED_RESPONSE = Symbol('isMockedResponse');
-const activeClientIds = new Set();
+const PACKAGE_VERSION = '2.2.14'
+const INTEGRITY_CHECKSUM = '26357c79639bfa20d64c0efca2a87423'
```
mattcasey, line 12: please bring back semicolons. u have some weird formatting going on!

mattcasey (approved):
approved with suggestions

apps/scoutgametelegram/components/quests/components/DailyClaimGallery/DailyClaimGift.tsx
```diff
@@ -0,0 +1,175 @@
+import { SvgIcon } from '@mui/material';
+
+export function DailyClaimGift({
+  variant = 'primary',
+  size = 64
+}: {
+  variant: 'disabled' | 'primary' | 'secondary';
+  size: number;
+}) {
+  const primaryColor = '#A06CD5';
```
mattcasey, line 10: why are these hard-coded?

mattcasey (changes requested):
I actually realized that you are saving the 'claims' in UTC time, but in the frontend you are using local time. This needs to be reconciled or timing will be off. If I claim after 6pm my time it would think that I claimed for the next day. So you can't rely on pointReceipt/builderEvent.createdAt to know which day of the week someone claimed, unless we use the same timezone on the frontend and backend

Some ideas:
- the frontend can tell the backend what week and day of the week is actually being claimed and we store that information someplace separate. this would probably be best in a new table that builderEvent links to.
- we use the same timezone on frontend and backend like NY time, so that we CAN infer the date from .createdAt.

apps/scoutgame/theme/theme.tsx
Devorein: its a bit difficult to match it precisely like the Figma mockup. Currently, its just two more extra colors

apps/scoutgametelegram/lib/claims/getDailyClaims.ts
Devorein: updated

public/mockServiceWorker.js
Devorein: its not happening anymore. weird. I copied it again from main branch

PR #4977, opened 2024-11-08
Scoutgame claim modal

Devorein:
### WHAT

- [x] Warpcast frame
- [x] Share copies and links

### WHY

[Card](https://app.charmverse.io/charmverse/claim-share-modal-34030863412072265)


apps/scoutgame/lib/points/claimPointsAction.ts
```diff
@@ -1,13 +1,13 @@
...
 
 import { claimPoints } from '@packages/scoutgame/points/claimPoints';
-import { revalidatePath } from 'next/cache';
 
 import { authActionClient } from 'lib/actions/actionClient';
+import { createUserClaimScreen } from 'lib/users/createUserClaimScreen';
 
 export const claimPointsAction = authActionClient.metadata({ actionName: 'claim_points' }).action(async ({ ctx }) => {
-  const result = await claimPoints({ userId: ctx.session.scoutId });
-  revalidatePath('/profile');
```
mattcasey, line 10: are you sure we don't need these?

apps/scoutgame/lib/points/claimPointsAction.ts
```diff
@@ -1,13 +1,13 @@
...
-  revalidatePath('/claim');
-  return { success: true, claimedPoints: result.total };
+  const userId = ctx.session.scoutId;
+  // const result = await claimPoints({ userId });
```
mattcasey, line 10: this is commented out

apps/scoutgame/app/(general)/points-claim/[path]/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+export default async function Claim({ params }: { params: { path: string } }) {
+  const user = await getUserByPath(params.path);
+  if (!user) {
+    return notFound();
+  }
+
+  const region = process.env.S3_UPLOAD_REGION;
+  const bucket = process.env.S3_UPLOAD_BUCKET;
+  const claimScreenUrl = `https://s3.${region}.amazonaws.com/${bucket}/points-claim/${appEnv}/${user.id}/${getCurrentWeek()}.png`;
```
Comment (mattcasey) [inline apps/scoutgame/app/(general)/points-claim/[path]/page.tsx:15]: do getlast week instead of current weeK?

apps/scoutgame/lib/points/claimPointsAction.ts
Devorein: yeah I commented that out to just show the claim modal with the points. I will revert it back before merging

apps/scoutgame/lib/points/claimPointsAction.ts
Devorein: right we do that after you close the modal the reason being the data inside the claim modal is inaccurate (returns to 0) if we revalidate the path immediately

PR #4999, opened 2024-11-13
Refresh nft price by polling

Devorein:


mattcasey (approved):
Thanks.. altho we should probably just be using SWR for this

PR #5010, opened 2024-11-14
Fix mui system imports

valentinludu:
Block suggestions for @mui/system


mattcasey (approved):
thanks for adding the ide settings

PR #5017, opened 2024-11-14
refresh decent price when chasing amount to pay

mattcasey:


packages/scoutgame-ui/src/components/common/NFTPurchaseDialog/components/NFTPurchaseForm.tsx
```diff
@@ -57,7 +57,7 @@ export type NFTPurchaseProps = {
   builder: MinimalUserInfo & { price?: bigint; nftImageUrl?: string | null };
 };
 
-const PRICE_POLLING_INTERVAL = 10000;
+const PRICE_POLLING_INTERVAL = 60000;
```
mattcasey, line 60: also 10 seconds seems aggressive and annoying when you're looking at an NFT

PR #5018, opened 2024-11-15
Send direct message via Warpcast / Scout game admin

mattcasey:


apps/scoutgameadmin/components/contract/Protocol
valentinludu, line 1: Can you delete this file if it remains empty?

apps/scoutgameadmin/components/contract/ProtocolContract.tsx
```diff
@@ -0,0 +1,13 @@
+'use server';
```
valentinludu, line 1: This should be `import 'server-only';`

'use server' will not have any effect since it is used only for action functions and it is more of a flag. import server-only will throw errors if the component is used in a wrong place.

apps/scoutgameadmin/components/contract/ContractDashboard.tsx
```diff
@@ -0,0 +1,51 @@
...
+          href={{
+            query: { tab: 'protocol' }
+          }}
+        />
+      </Tabs>
+      <Box mt={2}>
+        {currentTab === 'seasonOne' && <SeasonOneView {...seasonOne} />}
+        {currentTab === 'protocol' && (
+          <WagmiProvider>
+            <ProtocolContract />
```
valentinludu, line 45: ProtocolContract is a server only component but this file is client only and will make all children client components by default and will try to fetch db data client side. I assume it throws an error, but to be sure you can add `import 'server-only'`;

packages/farcaster/src/warpcast/directCastApi.ts
```diff
@@ -0,0 +1,35 @@
+import { PUT } from '@packages/utils/http';
+import { v4 as uuid } from 'uuid';
+
+// API Key requires special privileges to use this method
+const warpcastApiKey = process.env.WARPCAST_API_KEY;
+
+// Docs: https://warpcast.notion.site/Direct-Cast-API-Reference-Public-1276a6c0c1018089af2bda0d1697a2fd
+
+// Sends a message to a group chat, conversation or recipient depending on which id is provided.
```
valentinludu, line 9: Can't wait to see it in action!

valentinludu (approved):

apps/scoutgameadmin/components/contract/ProtocolContract.tsx
mattcasey: thanks i was trying to remember which one to use

PR #5025, opened 2024-11-17
Scoutgame tg info and profile pages

Devorein:


apps/scoutgametelegram/app/(general)/layout.tsx
```diff
@@ -12,7 +12,9 @@ export default function Layout({
   return (
     <>
       <Box sx={{ minHeight: 'calc(100vh + 120px)', position: 'relative', overflow: 'auto', px: 1 }}>
-        <Header />
+        <Box sx={{ position: 'relative', zIndex: 1 }}>
```
valentinludu, line 15: You can revert it like it was before. I have a fix in my PR to clean app everything. Because of a `position: static` and a `z-index:0` I see there is this need of adding `position: relative; z-index: 1` which is not necessary.

PR #5031, opened 2024-11-18
Add Top new scouts to admin

mattcasey:
Added a method to get top new scouts based on points, and added the export to Admin

image


apps/scoutgame/lib/builders/__tests__/getSortedBuilders.spec.ts
```diff
@@ -7,8 +7,8 @@ import { getSortedBuilders } from '../getSortedBuilders';
 
 describe('getSortedBuilders', () => {
   it('Should sort builders by new', async () => {
-    const mockWeek = v4();
-    const mockSeason = v4();
+    const mockWeek = '2024-W21';
```
mattcasey, line 10: random v4() makes no sense for week, we should always use standard format weeks otherwise we lose the date aspect of it

apps/scoutgame/lib/builders/getLeaderboard.ts
```diff
@@ -16,15 +16,17 @@ export type LeaderBoardRow = {
...
 }: {
   limit?: number;
   week?: string;
+  season?: string;
 }): Promise<LeaderBoardRow[]> {
-  const season = currentSeason;
   const weeklyTopBuilders = await prisma.userWeeklyStats.findMany({
     where: {
       week,
+      season,
```
mattcasey, line 29: this shouldnt matter but i think its always a good practice to include season as well

packages/scoutgame/src/builders/__tests__/approveBuilder.spec.ts
```diff
@@ -8,6 +8,10 @@ jest.unstable_mockModule('../../builderNfts/registerBuilderNFT', () => ({
   registerBuilderNFT: jest.fn()
 }));
 
+jest.unstable_mockModule('../../importReposByUser', () => ({
```
mattcasey, line 11: disables a call to github api

packages/scoutgame/src/builders/getBuildersLeaderboard.ts
```diff
@@ -22,6 +22,9 @@ export async function getBuildersLeaderboard({
       week,
       user: {
         builderStatus: 'approved'
+      },
+      gemsCollected: {
```
mattcasey, line 26: previously the assumption was you didnt have a weekly stats unless you had a builder event. but no builders should appear on the leaderboard without at least one gem

PR #5034, opened 2024-11-19
Serve waitlist frames from scoutgame app

motechFR:
### WHAT

Serving the frame from scoutgame app.

Redirect occurs from the waitlist app.

If we turn off the app, then we'll need a domain-level redirect


.ebstalk.apps.env/waitlist.env
mattcasey, line 1: we can delete this file

apps/waitlist/components/common/JoinWaitlistButton.tsx
mattcasey, line 1: delete this?

mattcasey (approved):
approved just some leftover files we can delete

PR #5036, opened 2024-11-19
Add alert for builder strike during nft purchase

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/show-of-strikes-for-a-builder-when-purchasing-their-nft-7572758038395089)


mattcasey (changes requested):
I don't think we use actions to retrieve data generally, in the past this has caused some issues but i do'nt remember right now.

Please rewrite to use SWR.. or RSC?

PR #5048, opened 2024-11-20
SG - View weeks in leaderboard table

valentinludu:


packages/scoutgame/src/users/__tests__/validateIsoWeek.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { getCurrentWeek, validateISOWeek } from '../../dates';
+
+describe('validateIsoWeek', () => {
```
mattcasey, line 3: thanks for adding a test! this should go inside scoutgame/src/__tests__?

packages/scoutgame/src/dates.ts
```diff
@@ -93,8 +100,8 @@ export function isToday(date: Date, now = DateTime.utc()) {
   return dateDay.equals(now.startOf('day'));
 }
 
-export function getCurrentSeasonWeekNumber(): WeekOfSeason {
-  return getSeasonWeekFromISOWeek({ season: currentSeason, week: getCurrentWeek() });
+export function getCurrentSeasonWeekNumber(week?: ISOWeek): WeekOfSeason {
+  return getSeasonWeekFromISOWeek({ season: currentSeason, week: week || getCurrentWeek() });
```
mattcasey, line 104: suggestion: put the default in the params: 
```
getCurrentSeasonWeekNumber(week = getCurrentWeek(): ISOWeek)

```

apps/scoutgame/lib/builders/getLeaderboard.ts
```diff
@@ -23,9 +23,11 @@ export async function getLeaderboard({
   week?: string;
   season?: string;
 }): Promise<LeaderBoardRow[]> {
+  const isWeekValid = validateISOWeek(week);
+
   const weeklyTopBuilders = await prisma.userWeeklyStats.findMany({
     where: {
-      week,
+      week: isWeekValid ? week : getCurrentWeek(),
```
mattcasey, line 30: i think this method should throw an error instead of hide the invalid input

packages/scoutgame/src/users/__tests__/validateIsoWeek.spec.ts
valentinludu: Moved! Thx for spotting it.

packages/scoutgame/src/dates.ts
valentinludu: Updated!

apps/scoutgame/lib/builders/getLeaderboard.ts
valentinludu: You're right. I have a default value in the page so this will not throw, but in the future we might reuse it so it is definitely better to add an error here.

I updated the func to throw if week is not valid!

mattcasey (approved):

PR #5049, opened 2024-11-20
Data/op reviewer info export

motechFR:


packages/utils/src/jest.config.ts
```diff
@@ -0,0 +1,8 @@
+export default {
+  // Teardown function after all tests run
```
mattcasey, line 2: there's no teardown function

packages/utils/src/dates.ts
```diff
@@ -30,3 +30,29 @@ export function timeUntilFuture(date?: number) {
 
   return { days, timeString, hours, minutes, seconds };
 }
+
+export function getUniqueWeeksCount(events: Date[]): number {
+  // Helper function to calculate the ISO week number
+  function getWeekNumber(date: Date): number {
```
mattcasey, line 36: Strongly prefer we do not write/maintain our own date logic and use libs instead.

You should be able to just do DateTime.fromJSDate(date).toFormat(`kkkk-'W'WW`);
 (thats what we do for scoutgame)

mattcasey (changes requested):
please use luxon instead of writing our own date logic, or maybe just keep it inside the script and don't commit it to the standard package. it's clever but very hard to maintain. looks like it was ai-generated?

PR #5053, opened 2024-11-21
Show talent and moxie in profile

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/add-social-icons-for-moxie-and-talent-protocol-to-builder-profiles-6268789628550171)


packages/scoutgame/src/moxie/airstackRequest.ts
```diff
@@ -0,0 +1,26 @@
+import { log } from '@charmverse/core/log';
+import { RateLimit } from 'async-sema';
+
+// at most, 10 req per second
```
mattcasey, line 4: "50 per second"

packages/scoutgame/src/talent/getTalentProfile.ts
```diff
@@ -0,0 +1,47 @@
...
+    {
+      headers: {
+        'X-API-KEY': process.env.TALENT_PROTOCOL_API_KEY
+      }
+    }
+  );
+  return passportResponse.passport || null;
+}
+
+export async function getTalentProfile({
```
mattcasey, line 25: we should track the wallet that works so that if we find one we dont have to call talent api with all your wallets

packages/scoutgame/src/talent/getTalentProfile.ts
```diff
@@ -0,0 +1,47 @@
...
+}) {
+  if (wallets.length === 0 && farcasterId) {
+    const farcasterProfile = await getFarcasterUserById(farcasterId);
+    // console.log('farcasterProfile', farcasterProfile?.verifications);
+    wallets = farcasterProfile?.verifications.map((address) => address);
+    log.debug('Found wallets from Farcaster', { farcasterId, wallets });
+  }
+  for (const wallet of wallets) {
+    const talentProfile = await getWalletTalentProfile(wallet).catch(() => null);
+    if (talentProfile && talentProfile.score > minimumTalentScore) {
```
mattcasey, line 42: why do we need a minimum score?

apps/scoutgameadmin/app/api/partners/talent/route.ts
```diff
@@ -37,25 +35,26 @@ export async function GET() {
...
-      if (walletScore && walletScore > minimumTalentScore) {
-        const scout = await prisma.scout.findUniqueOrThrow({
-          where: {
-            id: builder.builder.id
-          }
-        });
-        buildersWithTalent.push({ wallet: wallet.address, rank: builder.rank, score: walletScore, builder: scout });
-        foundScore = true;
-      }
+    const talentProfile = await getTalentProfile({
```
mattcasey, line 38: this should just read the DB and not the api anymore

mattcasey, line 38: it'd be better if this just read the

mattcasey (approved):
few suggestions but look sgood

packages/scoutgame/src/users/getUserByPath.ts
```diff
@@ -39,6 +53,8 @@ export async function getUserByPath(path: string): Promise<
     ...user,
     nftImageUrl: user?.builderNfts[0]?.imageUrl,
     congratsImageUrl: user?.builderNfts[0]?.congratsImageUrl,
-    githubLogin: user?.githubUser[0]?.login
+    githubLogin: user?.githubUser[0]?.login,
+    talent: user.talentProfile,
```
mattcasey, line 57: suggestion: just use the relation name "talentProfile" vs "talent" makes it a little easier i think when properties are consistent

mattcasey (approved):

packages/scoutgame/src/talent/getTalentProfile.ts
Devorein: this was present in sg admin

apps/scoutgameadmin/app/api/partners/talent/route.ts
Devorein: thats much faster. updated.

PR #5056, opened 2024-11-21
set week based on github data in the cron job

mattcasey:
Ensure that no matter when a script is run, the week/season for a PR is relative to the PR data.


apps/scoutgamecron/src/tasks/processBuilderActivity/recordMergedPullRequest.ts
```diff
@@ -101,14 +103,14 @@ export async function recordMergedPullRequest({
...
-        gte: startOfWeek.toJSDate()
+      builderEvent: {
+        week
       }
     }
   });
   const hasFirstMergedPullRequestAlreadyThisWeek = recentFirstMergedPullRequests > 0;
 
-  let isFirstMergedPullRequest = totalMergedPullRequests === 0 && !hasFirstMergedPullRequestAlreadyThisWeek;
+  let isFirstMergedPullRequest = totalMergedPullRequests === 0;
```
mattcasey, line 113: i also changed this, so that we can record whether a PR is someone's first or not regardless of whether we are giving them extra gems for it, which is a separate matter

Devorein (approved):

PR #5060, opened 2024-11-22
Proposals tables export

Devorein:
### WHY

[Card](https://app.charmverse.io/charmverse/add-export-option-to-main-proposal-page-7561508747206613)


mattcasey:
When i click "Export" on the All tab nothing happens. I think the button  there is too visually aggressive next to the normal icons, could we just put it in the context menu?

Also, i like the idea to export My proposals. I think we should include a URL column so u can get to them?

pages/api/spaces/[id]/proposals/work/export.ts
Comment (mattcasey) [inline pages/api/spaces/[id]/proposals/work/export.ts:1]: rename to 'myWork'? 'work' is a little ambiguous. or it could be something like proposals/export?type=my_work

Comment (Devorein) [reply pages/api/spaces/[id]/proposals/work/export.ts:1]: sure updated to 'my-work'

PR #5063, opened 2024-11-22
SG Telegram tracking

valentinludu:
## WHAT

Add a platform value to all events triggered by our apps.

Details:
- Take the platform from an env var
- env var is being verified agains our db to see if the value is valid. If the value is not valid we log it and return 'unknown'

### DIFF
--- a/packages/scoutgame-ui/src/hooks/usePageView.ts
+++ b/packages/scoutgame-ui/src/hooks/usePageView.ts
@@ -2,13 +2,18 @@ import { useTrackEvent } from '@packages/scoutgame-ui/hooks/useTrackEvent';
 import { usePathname } from 'next/navigation';
 import { useEffect } from 'react';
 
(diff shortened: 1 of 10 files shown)


packages/scoutgame-ui/src/hooks/usePageView.ts
```diff
@@ -2,13 +2,18 @@ import { useTrackEvent } from '@packages/scoutgame-ui/hooks/useTrackEvent';
...
   const trackEvent = useTrackEvent();
   const pathname = usePathname();
+  const isMounted = useIsMounted();
   // use this in case we want to update on search params
   // const searchParams = useSearchParams();
 
   useEffect(() => {
-    trackEvent('page_view');
-  }, [pathname, trackEvent]);
+    if (isMounted) {
```
mattcasey, line 15: just curious.. why do we have to do this?

mattcasey (approved):

PR #5070, opened 2024-11-26
Mobile footer

mattcasey:
This requires v 8 of the Bot API and i dont know how to enable it..
https://core.telegram.org/bots/webapps#recent-changes

### DIFF
--- a/apps/scoutgametelegram/components/layout/StickyFooter.tsx
+++ b/apps/scoutgametelegram/components/layout/StickyFooter.tsx
@@ -1,9 +1,10 @@
-import { Box } from '@mui/material';
+'use client';
+
 import { styled } from '@mui/material/styles';
 
 import { SiteNavigation } from 'components/common/SiteNavigation';
 
-const StickyContainer = styled('div')`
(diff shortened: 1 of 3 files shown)


apps/scoutgametelegram/components/layout/StickyFooter.tsx
```diff
@@ -1,11 +1,28 @@
 import { Box } from '@mui/material';
+import { styled } from '@mui/material/styles';
 
 import { SiteNavigation } from 'components/common/SiteNavigation';
 
+const StickyContainer = styled('div')`
```
valentinludu, line 6: This needs to be `footer` instead of `div`.

apps/scoutgametelegram/components/layout/StickyFooter.tsx
```diff
@@ -1,11 +1,28 @@
 import { Box } from '@mui/material';
```
valentinludu, line 1: Need `use client` here. I get some errors because styled() is a client function

valentinludu (approved):
Needs to be tested on iPhone!

PR #5071, opened 2024-11-26
Home page for logged out user

Devorein:
### WHY

[Card](https://app.charmverse.io/cvt-scout-game/new-home-screen-752184524260676)


apps/scoutgame/components/home/LandingPage.tsx
```diff
@@ -0,0 +1,286 @@
+'use client';
+
+import type { StackProps } from '@mui/material';
+import { Button, Container, Stack, Typography } from '@mui/material';
+import { useMdScreen } from '@packages/scoutgame-ui/hooks/useMediaScreens';
+import Image from 'next/image';
+import Link from 'next/link';
+
+function ContainerStack({ children, ...props }: { children: React.ReactNode } & StackProps) {
+  const isDesktop = useMdScreen();
```
valentinludu, line 10: This will flicker because of useMdScreen

Better to add inside the `sx` prop the xs and md values:
e.g. 

`p: { xs: 1, md: 4}`

valentinludu: Also after doing this, you can remove use client so we make sure it is prerendered on the server

apps/scoutgame/components/home/LandingPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+  iconSrc,
+  additionalContent
+}: {
+  stepNumber: string;
+  title: string;
+  description: string | React.ReactNode;
+  iconSrc: string;
+  additionalContent?: React.ReactNode;
+}) {
+  const isDesktop = useMdScreen();
```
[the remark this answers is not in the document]
valentinludu: Or use sx prop. Doesn't matter. Both have screen props for different sizes

apps/scoutgame/components/home/LandingPage.tsx
valentinludu: Same comment for all the places you are using sx or mui props for styling different sizes

apps/scoutgame/components/home/LandingPage.tsx
```diff
@@ -0,0 +1,286 @@
...
+        <Button variant='contained' sx={{ width: '50%' }}>
+          <Link href='/login'>Get started</Link>
+        </Button>
+      </Stack>
+    </Stack>
+  );
+}
+
+export function LandingPage() {
+  const isDesktop = useMdScreen();
```
valentinludu, line 263: I know we can't avoid the usage here of the useMdScreen.

My comments above are more for mui props.

apps/scoutgame/app/(general)/home/page.tsx
```diff
@@ -1,12 +1,20 @@
 import { getCurrentWeek, validateISOWeek } from '@packages/scoutgame/dates';
+import { getUserFromSession } from '@packages/scoutgame/session/getUserFromSession';
 
 import { HomePage } from 'components/home/HomePage';
+import { LandingPage } from 'components/home/LandingPage';
```
valentinludu, line 5: This should be loaded dynamic
or
useIsMounted in the `LandingPage` component to ensure it's a client component.

valentinludu: Because the window will flicker when a component is first rendered on the server and hooks return undefined at first, after the client renders the components then you will see the real value of useMdScreen

valentinludu (approved):
Just a small polish on the screen sizes.

Looks good.

PR #5074, opened 2024-11-26
Intercept external link click and navigate using tg api

Devorein:
### WHY

[Card](https://app.charmverse.io/cvt-scout-game/better-handling-of-external-links-in-telegram-23519699011459738)

### DIFF
--- a/packages/scoutgame-ui/src/providers/LinkInterceptor.tsx
+++ b/packages/scoutgame-ui/src/providers/LinkInterceptor.tsx
@@ -1,0 +1,34 @@
+'use client';
+
+import { useEffect } from 'react';
+
+export function LinkInterceptor() {
+  useEffect(() => {
+    const handleLinkClick = (event: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


packages/scoutgame-ui/src/providers/LinkInterceptor.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+export function LinkInterceptor() {
+  useEffect(() => {
+    const handleLinkClick = (event: MouseEvent) => {
+      const anchor = (event.target as Element).closest('a');
+
+      if (anchor && anchor.href) {
+        const url = new URL(anchor.href);
+
+        const isExternal = url.origin !== window.location.origin;
```
mattcasey, line 13: i think we should check target="_blank" instead, seems more robust - we only care if we're trying to send you to a new tab. You're not supposed to use the URL constructor with relative paths. When I try it, i get an uncaught exception

mattcasey (approved):

PR #5088, opened 2024-12-02
Metamask deeplink

mattcasey:
I ran thru the rainbowkit source code and tried to find a solution that would work by just modifying the "rkDetails" configured by Rainbowkit, but it's just not possible. Next best thing was to listen to the click event and override the behavior

https://github.com/rainbow-me/rainbowkit/discussions/2091


motechFR:
Testing on iOS: Sent me to the app store

PR #5099, opened 2024-12-04
Quests for both apps

valentinludu:


Devorein (changes requested):
Great everything seems to be working like before. A few small design changes are required.

image
1. The typography color seems to be using primary color rather than white
2. The width of the Quests section should be narrower according to the mockup
3. Clicking on quests when logged out should show a different modal like claim page
4. Scrollbar even when its not required

apps/scoutgametelegram/lib/session/getSession.ts
```diff
@@ -1,5 +1,5 @@
 import { getSession as getSessionBase } from '@packages/scoutgame/session/getSession';
 
 export const getSession = () => {
-  return getSessionBase({ sameSite: 'none' });
+  return getSessionBase({ sameSite: 'none', secure: true, domain: 'ngrok.app' });
```
Devorein, line 4: oops a leftover from dev

valentinludu: Reverted!

mattcasey (approved):
nice!

Devorein (approved):

PR #5102, opened 2024-12-05
Builder nfts sold stats script

Devorein:
### WHY

[Discussion](https://discord.com/channels/894960387743698944/1313553319989612575)


mattcasey (approved):

mattcasey (approved):
why do we need to do this? is the issue fixed?

PR #5115, opened 2024-12-06
Fix starter pack

mattcasey:
- Fixes builders appearing multiple times in the banner
- Fixes a bunch of place we need to be sure to grab your default NFT. The price from default NFTs was being set on the starter pack


packages/scoutgame/src/builderNfts/refreshBuilderNftPrice.ts
```diff
@@ -30,7 +30,8 @@ export async function refreshBuilderNftPrice({
     const existingNft = await prisma.builderNft.findFirstOrThrow({
       where: {
         builderId,
-        season
+        season,
+        nftType: BuilderNftType.default
```
motechFR, line 34: Nice, this is what caused the starter pack price to be wrong

We should be sure that if this throws it doesn't interrupt other payment processes

mattcasey: I checked and registerBuilderNFT expects a result, so i think it should throw. let the caller handle errors

PR #5121, opened 2024-12-09
Fix starterpack cards commit bullets

valentinludu:


Devorein (approved):
approved but it will be great if we have some sort of util function to get the last 7 days gems. Currently its used in a few places and its not straightforward to use this field.

valentinludu:
> approved but it will be great if we have some sort of util function to get the last 7 days gems. Currently its used in a few places and its not straightforward to use this field.

I pushed n util func for that to be used everywhere

PR #5126, opened 2024-12-09
Common header/footer for both apps

valentinludu:


mattcasey (approved):

mattcasey:
code looks good, i didnt do a visual check

PR #5129, opened 2024-12-09
Mark a card as purchased

motechFR:
### WHAT

Todo: Address case where viewer of a profile sees the card price


packages/scoutgame/src/builders/getStarterPackBuilders.ts
```diff
@@ -76,6 +94,7 @@ export async function getStarterPackBuilders({
     nftImageUrl: builder.builderNfts[0]?.imageUrl || '',
     nftType: BuilderNftType.starter_pack,
     farcasterId: builder.farcasterId,
-    congratsImageUrl: builder.builderNfts[0]?.congratsImageUrl || ''
+    congratsImageUrl: builder.builderNfts[0]?.congratsImageUrl || '',
+    purchased: !!(builder.builderNfts[0] as BuilderNft & { nftSoldEvents?: NftPurchaseEvent[] })?.nftSoldEvents?.length
```
mattcasey, line 98: doesnt seem like u need type casting here

mattcasey (approved):
approved but please see my question about disabled prop

packages/scoutgame/src/builders/getStarterPackBuilders.ts
motechFR: We do need it, as in l. 48 we only perform a join with NftSoldEvents if there is a userId

PR #5133, opened 2024-12-10
Builder  page for both apps

valentinludu:


Devorein (approved):
Looks good. Just need to add all the images from `apps/scoutgame/public/images/promos` to telegram app

PR #5135, opened 2024-12-10
Fix carousels with 100% height on mobile

valentinludu:
Can be tested only on iPhone.

Safari on macOS does not have this issue.


packages/scoutgame-ui/src/components/common/Carousel/Carousel.tsx
```diff
@@ -90,9 +90,9 @@ export function Carousel({
         {...swiperProps}
         style={{
           width: '100%',
-          ...swiperProps.style,
           minHeight: isDesktop ? undefined : mobileMinHeight,
-          zIndex: '0'
```
mattcasey, line 95: i wonder if zIndex being a string was causing issues

mattcasey (approved):

PR #5153, opened 2025-01-15
Optimism proposal reviewer notifications

Devorein:

### DIFF
--- a/lib/notifications/documents/createDocumentNotifications.ts
+++ b/lib/notifications/documents/createDocumentNotifications.ts
@@ -503,6 +503,16 @@ export async function createDocumentNotifications(webhookData: {
               proposal: {
                 select: {
                   id: true,
+                  ProposalAppealReviewer: {
+                    where: {
+                      userId: {
+                        not: null
+                      }
+                    },
+                    select: {
+                      userId: true
+                    }
+                  },
                   reviewers: {
                     where: {
                       userId: {
@@ -524,14 +534,14 @@ export async function createDocumentNotifications(webhookData: {
         new Set(document?.proposal?.reviewers.map((reviewer) => reviewer.userId).filter(isTruthy) ?? [])
       ).filter((userId) => userId !== commentAuthorId && !notificationSentUserIds.has(userId));
 
-      if (documentId && proposalId && proposalReviewerUserIds.length) {
+      if (documentId && proposalId && space.domain === 'op-grants' && proposalReviewerUserIds.length) {
         for (const userId of proposalReviewerUserIds) {
           const proposalPermissions = await permissionsApiClient.proposals.computeProposalPermissions({
             resourceId: proposalId,
             userId
           });
 
-          if (proposalPermissions.evaluate || proposalPermissions.evaluate_appeal || proposalPermissions.view) {
+          if (proposalPermissions.view) {
             const { id } = await saveDocumentNotification({
               type: 'comment.created',
               createdAt: webhookData.createdAt,
--- a/lib/notifications/documents/createDocumentNotifications.ts
+++ b/lib/notifications/documents/createDocumentNotifications.ts
@@ -338,15 +338,6 @@ export async function createDocumentNotifications(webhookData: {
         : (webhookData.event.document?.authors.map(({ id }) => id) ?? []);
       const documentId = webhookData.event.document?.id;
       const postId = webhookData.event.post?.id;
-      const space = await prisma.space.findUniqueOrThrow({
-        where: {
-          id: spaceId
-        },
-        select: {
-          domain: true
-        }
-      });
-
       const comment = webhookData.event.post
         ? await prisma.postComment.findFirstOrThrow({
             where: {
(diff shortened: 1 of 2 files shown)


lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +491,56 @@ export async function createDocumentNotifications(webhookData: {
...
+              id: documentId
+            },
+            select: {
+              type: true,
+              proposalId: true
+            }
+          })
+        : null;
+
+      if (documentId && space.domain === 'op-grants' && document?.type === 'proposal' && document?.proposalId) {
```
motechFR, line 506: Can we check for other places we use **op-grants** for features, and make it a shared constant?

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +491,56 @@ export async function createDocumentNotifications(webhookData: {
...
+              createdBy: commentAuthorId,
+              commentId,
+              pageId: documentId,
+              spaceId,
+              userId: spaceRole.userId,
+              content: comment.content
+            });
+
+            ids.push(id);
+            notificationSentUserIds.add(spaceRole.userId);
```
motechFR, line 539: Could we do the query differently?

This space has alot of members, so you will end up making hundreds of calls to permissions API each time a comment happens.

Alternative approach would be to just load the assigned proposal reviewers (joining on role) who also have a spaceRole, and iterate just through those to check if they already received the notification

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -337,6 +338,14 @@ export async function createDocumentNotifications(webhookData: {
         : (webhookData.event.document?.authors.map(({ id }) => id) ?? []);
       const documentId = webhookData.event.document?.id;
       const postId = webhookData.event.post?.id;
+      const space = await prisma.space.findUniqueOrThrow({
```
mattcasey, line 341: this record is not used?

lib/notifications/documents/createDocumentNotifications.ts
mattcasey: i think he removed it but this feature is for all spaces, not just op

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +492,63 @@ export async function createDocumentNotifications(webhookData: {
...
+            where: {
+              id: documentId
+            },
+            select: {
+              type: true,
+              proposalId: true,
+              proposal: {
+                select: {
+                  id: true,
+                  reviewers: {
```
mattcasey, line 506: we should probably include appealReviewers as well

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +492,63 @@ export async function createDocumentNotifications(webhookData: {
...
+      ).filter((userId) => userId !== commentAuthorId && !notificationSentUserIds.has(userId));
+
+      if (documentId && proposalId && proposalReviewerUserIds.length) {
+        for (const userId of proposalReviewerUserIds) {
+          const proposalPermissions = await permissionsApiClient.proposals.computeProposalPermissions({
+            resourceId: proposalId,
+            userId
+          });
+
+          if (proposalPermissions.evaluate || proposalPermissions.evaluate_appeal || proposalPermissions.view) {
```
mattcasey, line 534: we should only be checking for "view" access

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +492,63 @@ export async function createDocumentNotifications(webhookData: {
...
+            const { id } = await saveDocumentNotification({
+              type: 'comment.created',
```
mattcasey, line 536: this needs to go in the block of code where we send notifications for "comment.created". right now this code appears to be sending notifications for any comment reply or user mention in the document. What i would do:

- conditionally extend the list of "authorIds" with the reviewerIds. then you're done

lib/webhookPublisher/publisher.ts
```diff
@@ -5,6 +5,8 @@ import { isTestEnv } from '@root/config/constants';
 import { addMessageToSQS } from '@root/lib/aws/SQS';
 import type { WebhookEvent, WebhookPayload } from '@root/lib/webhookPublisher/interfaces';
 
+import { createNotificationsFromEvent } from '../notifications/createNotificationsFromEvent';
```
mattcasey, line 8: this import is not used?

mattcasey (changes requested):
see comments

lib/notifications/documents/createDocumentNotifications.ts
Devorein: used below

lib/notifications/documents/createDocumentNotifications.ts
Devorein: Chris notified that we should send notifications for all comments

lib/notifications/documents/createDocumentNotifications.ts
```diff
@@ -483,6 +492,77 @@ export async function createDocumentNotifications(webhookData: {
...
+      const proposalId = document?.type === 'proposal' ? document?.proposal?.id : null;
+      const proposalReviewerUserIds = Array.from(
+        new Set(
+          [...(document?.proposal?.reviewers ?? []), ...(document?.proposal?.ProposalAppealReviewer ?? [])]
+            .map((reviewer) => reviewer.userId)
+            .filter(isTruthy) ?? []
+        )
+      ).filter((userId) => userId !== commentAuthorId && !notificationSentUserIds.has(userId));
+
+      if (documentId && proposalId && space.domain === 'op-grants' && proposalReviewerUserIds.length) {
```
mattcasey, line 541: remove check for op-grants

PR #5168, opened 2025-02-03
Supported hidden (deleted) properties for rewards and proposals

mattcasey:
Allow users to delete a property, but just hide it for new reward/proposals and their templates. This way it will still be available on older docs

...We could potentially support this for regular boards, but i think it's a different use case because rewards and proposal boards are used for all the docs in the system.


valentinludu (approved):
Tested! Works great!

PR #5198, opened 2025-05-13
charmverse dev payment integration

Devorein:


apps/cron/src/cron.ts
```diff
@@ -81,6 +82,9 @@ cron.schedule('0 * * * *', storeOptimismProjectAttestations);
 // Send proposal evaluation notifications every hour
 cron.schedule('0 * * * *', () => sendProposalEvaluationNotifications());
 
+// once every month
+cron.schedule('0 0 1 * *', chargeSpacesSubscriptionTask);
```
mattcasey, line 86: it's better to have this run like once an hour, so that if it fails and needs to retry for any reason, you don't have to go and re-run it offline

lib/subscription/chargeSpaceSubscription.ts
```diff
@@ -0,0 +1,56 @@
...
+export const SubscriptionTierAmountRecord: Record<SpaceSubscriptionTier, bigint> = {
+  free: parseUnits('0', 18),
+  bronze: parseUnits('1000', 18),
+  silver: parseUnits('2500', 18),
+  gold: parseUnits('10000', 18),
+  grant: parseUnits('0', 18),
+  readonly: parseUnits('0', 18)
+};
+
+export async function chargeSpaceSubscription({ spaceId }: { spaceId: string }) {
```
mattcasey, line 17: make sure to make this idempotent, so that if a payment was made already, it gets skipped

lib/subscription/chargeSpaceSubscription.ts
```diff
@@ -0,0 +1,56 @@
...
+  const subscriptionTier = space.subscriptionTier;
+  const spaceTokenBalance = await getSpaceTokenBalance({ spaceId });
+
+  const subscriptionTierAmount = SubscriptionTierAmountRecord[subscriptionTier];
+
+  if (parseUnits(spaceTokenBalance.toString(), 18) < subscriptionTierAmount) {
+    await prisma.space.update({
+      where: { id: spaceId },
+      data: {
+        subscriptionTier: 'free'
```
mattcasey, line 40: i believe this should be 'readonly'

lib/spaces/createSpaceContribution.ts
```diff
@@ -0,0 +1,96 @@
...
+  const tx = await publicClient.getTransaction({ hash: hash.toLowerCase() as `0x${string}` });
+
+  if (tx.from.toLowerCase() !== walletAddress.toLowerCase()) {
+    throw new Error('Transaction sender does not match provided wallet address');
+  }
+
+  // Get the transfer logs from the transaction
+  const logs = await publicClient.getLogs({
+    fromBlock: receipt.blockNumber,
+    toBlock: receipt.blockNumber,
```
mattcasey, line 53: should be OK since we have the right block number :)

lib/spaces/getSpaceReceipts.ts
```diff
@@ -0,0 +1,58 @@
...
+    where: { spaceId },
+    select: {
+      id: true,
+      paidTokenAmount: true,
+      createdAt: true
+    }
+  });
+
+  const spaceContributions = await prisma.spaceSubscriptionContribution.findMany({
+    where: { spaceId, txHash: { not: null }, userId: { not: null } },
```
mattcasey, line 32: maybe we should still show contributions even if we dont have a user? u can use the wallet address as a fallback for user name, etc.

mattcasey (approved):

lib/spaces/getSpaceReceipts.ts
Devorein: sure

PR #5215, opened 2025-05-22
Space subscription feature restriction

Devorein:
…into feat/restrict-features-v1### WHAT


mattcasey (approved):

packages/subscriptions/src/updateSubscription.ts
```diff
@@ -0,0 +1,63 @@
...
+  const isDowngrading = currentTierIndex < previousTierIndex;
+
+  await prisma.$transaction([
+    prisma.space.update({
+      where: { id: spaceId },
+      data: {
+        subscriptionTier: newTier
+      }
+    }),
+    prisma.spaceSubscriptionTierChangeEvent.create({
```
mattcasey, line 27: this seems like a bug.. why is a change event being added here? why is it newTier = 'readonly'? wouldn't that lock this space on the next charge/make it look like it was going to be set to Expired?

packages/subscriptions/src/chargeSpaceSubscription.ts
```diff
@@ -75,19 +62,7 @@ export async function chargeSpaceSubscription({ spaceId }: { spaceId: string })
...
-        });
-        await prisma.spaceSubscriptionTierChangeEvent.create({
-          data: {
-            spaceId,
-            newTier: subscriptionTier,
-            previousTier: space.subscriptionTier!
-          }
-        });
-      }
+      await updateSubscription({ spaceId, newTier: subscriptionTier });
```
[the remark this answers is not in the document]
Devorein: yeah this should only happen if you are getting downgraded. fixed

packages/subscriptions/src/chargeSpaceSubscription.ts
```diff
@@ -75,17 +63,15 @@ export async function chargeSpaceSubscription({ spaceId }: { spaceId: string })
...
-          where: { id: spaceId },
-          data: { subscriptionTier }
-        });
-        await prisma.spaceSubscriptionTierChangeEvent.create({
-          data: {
-            spaceId,
-            newTier: subscriptionTier,
-            previousTier: space.subscriptionTier!
-          }
+      const previousTierIndex = space.subscriptionTier ? downgradeableTiers.indexOf(space.subscriptionTier) : 0;
```
mattcasey, line 66: - just let this be called when the tier is not equivalent.
- handle 'cancelledAt' above where we handle setting to readonly
- add a comment here "most likely at this point, the sub is being downgraded because upgrades happen immediately"

Issue #5190, opened 2025-03-08
Add DAOIP-5 Support


Rashmi-278:
DAOIP-5: The Grants Metadata Standard, aims to standardise grant data reporting to ensure interoperability between different grant systems data for ecosystem data analysis and grant tools 

Charmverse being one of the well adopted tool by grant systems, has agreed to support this vision by DAOstar.org

Rashmi-278:
Todo:
- [ ] Map out Charmverse Metadata Schema with DAOIP-5 Schema 
- [ ] identify categories and filters needed for Grant Data Reporting 
- [ ] Setup a live pipeline from Charmverse API to Datalake / DAOIP-5

PR #5221, opened 2025-05-27
Dependency upgrades - React, MUI, Next, etc.

mattcasey:
Also bump requirements to node v22


mattcasey:
https://github.com/charmverse/app.charmverse.io/issues/5190

PR #5242, opened 2025-06-23
Fixes for proration calculation

mattcasey:


apps/webapp/pages/api/admin/token-balance.ts
```diff
@@ -1,24 +1,26 @@
...
-  res: NextApiResponse<{ value: string; formatted: number }>
-) {
+async function getSpaceTokenBalanceController(req: NextApiRequest, res: NextApiResponse<ClientSpaceTokenBalance>) {
   const { id: spaceId } = req.query as { id: string };
 
-  const { value, formatted } = await getSpaceTokenBalance({ spaceId });
+  const { value, formatted, payments, contributions } = await getSpaceTokenBalance({ spaceId });
 
   res.status(200).json({
+    payments,
```
mattcasey, line 22: i was planning on showing this in the Admin UI but didnt get that far. but its in the network tab at least for now :)

packages/subscriptions/src/calculateSubscriptionCost.ts
```diff
@@ -20,13 +20,14 @@ export function calculateSubscriptionCost({
 
   const now = DateTime.now();
   const monthEnd = now.endOf('month');
-  const daysRemaining = Math.floor(monthEnd.diff(now, 'days').days) + 1;
-  const daysInThisMonth = monthEnd.diff(now.startOf('month'), 'days').days + 1;
+  const daysRemaining = Math.ceil(monthEnd.diff(now, 'days').days);
+  const daysInThisMonth = Math.ceil(monthEnd.diff(now.startOf('month'), 'days').days);
```
mattcasey, line 24: fixed some math so that you won't get any proration when u buy the first day of the month

